Issue #10, opened 2017-03-27
Remove scons temp files from repo


tst-ccamp:
The config.log file and .sconf_temp directory are inadvertently checked into the repo, but they are  generated by the scons build. This may cause the config operation of the scons build to not work correctly. Also, where possible the config step should be forced in case a system setting has changed.

PR #11, opened 2017-03-27
[Issue #10] Removed temporary scons files and added to .gitignore

tst-ccamp:
Fixes: https://github.com/google/earthenterprise/issues/10

* Removes scons temp files and add to .gitignore file
* Adds --config=force to build script (once per build should be good enough)


tst-ccamp:
@asmitaw Is this something that you can set inside the SConscript? I've been meaning to look into that but haven't had the time to go through the docs yet? That would definitely be preferable. Honestly, the caching only adds a minuscule amount of time savings anyway, something like 1 second. For a 20+ minute build, that's definitely not worth the headaches that go along with it :)

asmitaw:
@tst-ccamp - here are some options to do it from within the SconsScript ( I have not tried them myself. )
http://scons.org/doc/2.0.1/HTML/scons-user/c2076.html

PR #39, opened 2017-03-30
I19: Portable Build Windows Fixes

tst-ppenev:
This is no longer a work-in-progress pull request.

* FileUnpacker builds on Linux and Windows now.

* Fixed `fileunpacker` build commands for Windows.
  * Added `sys.prefix` and `sys.exec_prefix` as variables that can be expanded in build commands.
  * Fixed compile commands for MinGW. (Spaces seemed not to work for me.)
* Fixed `util.ExecuteCmd` to show complete standard output and standard error on Windows before returning.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,4 +6,4 @@
 *.glb filter=lfs diff=lfs merge=lfs -text
 *.glc filter=lfs diff=lfs merge=lfs -text
 *.glm filter=lfs diff=lfs merge=lfs -text
-earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/test_data/ binary
+earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/test_data/* binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,4 @@
 *.glb filter=lfs diff=lfs merge=lfs -text
 *.glc filter=lfs diff=lfs merge=lfs -text
 *.glm filter=lfs diff=lfs merge=lfs -text
+earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/test_data/ binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,4 +6,4 @@
 *.glb filter=lfs diff=lfs merge=lfs -text
 *.glc filter=lfs diff=lfs merge=lfs -text
 *.glm filter=lfs diff=lfs merge=lfs -text
-earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/test_data/* binary
+earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/test_data/** binary
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
@@ -62,29 +62,29 @@ def RunTests():
   shutil.copyfile("../test.py", "test.py")
 
   old_path = sys.path
-  sys.path = sys.path + [os.getcwd()]
+  sys.path = [os.getcwd()] + sys.path
   import test
   sys.path = old_path
 
   test.main()
 
-def main():
+def main(argv):
   """Main for build and test."""
-  if (len(sys.argv) != 2 or
-      sys.argv[1].lower() not in ["mac", "windows", "linux"]):
+  if (len(argv) != 2 or
+      argv[1].lower() not in ["mac", "windows", "linux"]):
     print "Usage: build_and_test.py <OS_target>"
     print
     print "<OS_target> can be Mac, Windows, or Linux"
     return
 
-  os.chdir(sys.path[0])
-  os_dir = "%s%s" % (sys.argv[1][0:1].upper(), sys.argv[1][1:].lower())
+  os.chdir(os.path.dirname(os.path.realpath(__file__)))
+  os_dir = "%s%s" % (argv[1][0:1].upper(), argv[1][1:].lower())
   print "Build and test file unpacker library for %s Portable Server." % os_dir
 
-  if BuildLibrary(os_dir, sys.argv[1].lower()=="windows"):
+  if BuildLibrary(os_dir, argv[1].lower()=="windows"):
     print "Library built."
     RunTests()
 
 
 if __name__ == "__main__":
-  main()
+  main(sys.argv)
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
@@ -38,8 +38,12 @@ def BuildLibrary(os_dir, ignore_results):
 
   os.chdir("dist")
   fp = open("../%s/build_lib" % os_dir)
+  build_vars = {
+      'prefix': sys.prefix,
+      'exec_prefix': sys.exec_prefix
+    }
   for line in fp:
-    result = util.ExecuteCmd(line, use_shell=True)
+    result = util.ExecuteCmd(line.format(**build_vars), use_shell=True)
     if result:
       if ignore_results:
         print result
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
@@ -39,9 +39,9 @@ def BuildLibrary(os_dir, ignore_results):
   os.chdir("dist")
   fp = open("../%s/build_lib" % os_dir)
   build_vars = {
-      'prefix': sys.prefix,
-      'exec_prefix': sys.exec_prefix
-    }
+    'prefix': sys.prefix,
+    'exec_prefix': sys.exec_prefix
+  }
   for line in fp:
     result = util.ExecuteCmd(line.format(**build_vars), use_shell=True)
     if result:
--- a/earth_enterprise/src/portableserver/build.py
+++ b/earth_enterprise/src/portableserver/build.py
@@ -28,8 +28,7 @@ import os.path
 import shutil
 import sys
 
-SELF_PATH = os.path.realpath(__file__)
-SELF_DIR = os.path.dirname(SELF_PATH)
+SELF_DIR = os.path.dirname(os.path.realpath(__file__))
 
 
 def ensure_directory(path):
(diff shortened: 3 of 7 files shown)


earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
```diff
@@ -38,8 +38,12 @@ def BuildLibrary(os_dir, ignore_results):
 
   os.chdir("dist")
   fp = open("../%s/build_lib" % os_dir)
+  build_vars = {
+      'prefix': sys.prefix,
+      'exec_prefix': sys.exec_prefix
+    }
```
andreisel: nit: indent for closing brace - 2 more spaces?

tst-ppenev: Done.

tst-ppenev:
@andreisel, @tst-ccamp: This is no longer work-in-progress. I added a Python build script to produce '.tgz', and '.zip' packages for Linux and Windows. I think it's ready for review.

tst-ppenev:
I also added a tentative section on building Portable to the Wiki: https://github.com/google/earthenterprise/wiki/Build-Instructions.

tst-ccamp:
Hey, @tst-ppenev, can you add some prerequisite packages to building on Windows? It's mingw, right? Can you use the one that comes with git for windows?

tst-ppenev:
@tst-ccamp: Lol, yeah. I'll probably be less absent-minded after some food.

tst-ppenev:
@tst-ccamp: I added some prerequisites to the Wiki. Hopefully, I didn't forget something. :octopus:

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,256 @@
...
+def main(argv):
+    """Parses command-line arguments, detects build environment, builds, and
+    packages Portable server."""
+
+    build_dir = os.path.join(SELF_DIR, 'build')
+    source_dir = os.path.join(SELF_DIR, '..')
+
+    platform = None
+    arg_index = 1
+    while arg_index < len(argv):
```
andreisel: Use get_opt.gnu_get_opt() or argparse for parsing argv?

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Copy library to package directory:
+        exclude_files = ['test.py', 'util.py']
+        dist_dir = os.path.join(task_build_dir, 'dist')
+        entries = [
+            f
+            for f in os.listdir(dist_dir)
+            if os.path.splitext(f)[1].lower() in ['.so', '.pyd', '.dll', '.py'] \
+                and f not in exclude_files]
+        copy_from_dir_to_dir(dist_dir, self.server_dir, entries)
```
andreisel: use named argument for entries: entries=entries

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,256 @@
...
+
+        self.create_version_txt()
+
+        # Copy configuration files:
+        copy_from_dir_to_dir(self.resources_dir, self.server_dir)
+
+    def create_version_txt(self):
+        """Creates <server/local/version.txt>."""
+
+        self.get_version()
```
andreisel: Remove get_version()-call, since you set it in init?

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,256 @@
...
+"""
+
+import datetime
+import distutils.dir_util
+import os
+import os.path
+import shutil
+import sys
+
+SELF_PATH = os.path.realpath(__file__)
```
tst-ppenev: It seems to me that the longer a line becomes, the less readable it is. I prefer having it in multiple parts, but having it on one line is readable enough for me, as well.

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Copy from <fusion/portableglobe/servers/>:
+        servers_dir = os.path.join(
+            self.source_dir, 'fusion', 'portableglobe', 'servers')
+
+        exclude_entries = ['linux', 'mac', 'windows', 'fileunpacker']
+        entries = [f
+                   for f in os.listdir(servers_dir)
+                   if f.lower() not in exclude_entries]
+        copy_from_dir_to_dir(servers_dir, self.server_dir, entries)
```
tst-ppenev: I thought `entries=entries` looked a bit redundant, but I can change it.

earth_enterprise/src/portableserver/build.py
tst-ppenev: `get_version()` only updates values, if they're `None`, so it can be called multiple times without harm.  There really isn't a use case for calling it again right now, but in an imaginary world where someone did something like:

```Python
builder = Builder(build_dir, source_dir, platform)
builder.build_date = datetime.date(1945, 5, 8) # Travel back in time.
builder.version_string = None # Recalculate version_string
builder.build() # Make a build for Doc Brown's trip.
```

It would build with a valid <version.txt>.

earth_enterprise/src/portableserver/build.py
tst-ppenev: Done.

.gitattributes
```diff
@@ -6,3 +6,4 @@
 *.glb filter=lfs diff=lfs merge=lfs -text
 *.glc filter=lfs diff=lfs merge=lfs -text
 *.glm filter=lfs diff=lfs merge=lfs -text
+earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/test_data/ binary
```
tst-ccamp: I think you need to use .../test_data/* binary here

earth_enterprise/src/portableserver/build.py
tst-ppenev: Done.

earth_enterprise/src/portableserver/build.py
tst-ppenev: Done.

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def get_version(self):
+        """Parses version information, and sets member variables."""
+
+        if self.base_version is None:
+            path = os.path.join(self.source_dir, 'fusion_version.txt')
+            with open(path, 'r') as input_file:
+                while True:
+                    base_version = input_file.readline().strip()
+                    if not base_version.startswith('#'):
```
andreisel: Add check for empty string:

if base_version and not base_version.startswith('#'):

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,252 @@
...
+                self.base_version, self.build_date.strftime('%Y%m%d'))
+
+    def build_fileunpacker(self):
+        """Builds `fileunpacker` shared libraries, and copies them to
+        <server/>."""
+
+        # Keep the source directory clean:
+        task_build_dir = os.path.join(self.build_dir, 'fileunpacker-build')
+        shutil.rmtree(task_build_dir, ignore_errors=True)
+        distutils.dir_util.copy_tree(
```
andreisel: Maybe use copy_from_dir_to_dir()?

earth_enterprise/src/portableserver/build.py
```diff
@@ -0,0 +1,252 @@
...
+        build_and_test.main(['build_and_test.py', self.platform])
+
+        # Copy library to package directory:
+        exclude_entries = ['test.py', 'util.py']
+        dist_dir = os.path.join(task_build_dir, 'dist')
+        entries = [
+            f
+            for f in os.listdir(dist_dir)
+            if os.path.splitext(f)[1].lower() in
+            ['.so', '.pyd', '.dll', '.py']
```
andreisel: add '.pyc', '.pyo' to the list?

earth_enterprise/src/portableserver/build.py
andreisel: LGTM

earth_enterprise/src/portableserver/build.py
tst-ppenev: Done.

earth_enterprise/src/portableserver/build.py
tst-ppenev: Done.

earth_enterprise/src/portableserver/build.py
tst-ppenev: It makes no difference in this case.  If we wanted to eliminate dependence on `distutils`, we could do that, but I don't think we'd gain anything from it.

earth_enterprise/src/portableserver/build.py
tst-ppenev: I don't see real benefit from including <.pyc>, and <.pyo>. They'll be generated the first time someone executes the Python scripts anyway.  I'd rather keep the archive clean.

.gitattributes
tst-ppenev: I think it worked on Windows with two stars.

earth_enterprise/src/portableserver/build.py
andreisel: Ops, never mind, I interpreted them as the exclude_entries.

.gitattributes
tst-ccamp: This was really difficult to test because after resetting my global setting back to autocrlf to true (the default), every time I would clone the repo fresh it would, of course checkout master. That means it would have the version with dos format files. But, switching branches to the one with the updated .gitattributes alone will not convert the line endings (very frustating). It turns out you have to do this in order to convert the line endings:

1. git checkout I19-Portable-build-Windows-fixes
1. rm .git/index from top level of the repo
1. git reset (rebuilds the index, but takes a few minutes)
1. git reset --hard HEAD

If it converts the line endings of the files it will show the files as being modified during git reset and if you do a git status before running git reset -- hard HEAD.

Anyway, long story short, I was able to verify this works, but I worry that users who want to fix this issue if it's encountered need some instructions on how to fix it even with the addition of the .gitattributes. It should work fine for new clones, but won't fix the issue for existing clones without some sort of manual update. Should we add a FAQ entry, maybe?

tst-ppenev:
@tst-ccamp: I think this is an issue that will probably go away soon (if this works), so adding it to the FAQ may just make people look through more questions until they find what they need.

tst-ccamp:
@tst-ppenev So, you are pro-adding to the FAQ?

tst-ppenev:
@tst-ccamp: No.  I agree with your assessment.  Unless, a lot of people suddenly start having this problem, it wouldn't be a FAQ, might be counterproductive.

Issue #40, opened 2017-03-30
Move tutorial files out of repo


tst-ccamp:
Many contributors have difficulty with git lfs. Accordingly, the tutorial files need to be removed from the main earthenterprise repo and will be hosted elsewhere. The goal is to get the size of the repo down small enough that we no longer have to rely on git-lfs.

AC:
1. Tutorial files are removed from the main repo
2. Write follow-on tickets to further reduce the size of the repo and remove git-lfs requirement

tst-ccamp:
Added tutorial to opengee.org website, but need to remove data files from repo

PR #45, opened 2017-03-31
Adds links for downloading fusion source files

tst-ccamp:
Fixes https://github.com/google/earthenterprise/issues/40
Added links and instructions for downloading and installing fusion source files

A preview of the changes is available here: https://tst-ccamp.github.io/earthenterprise/geedocs/answer/4412441.html


avnishb:
Looks great!  Think these links should be exposed on elsewhere too?

PR #49, opened 2017-04-03
[#40] Removes Fusion tutorial files from the repo

tst-ccamp:
Fixes: https://github.com/google/earthenterprise/issues/40

Removes tutorial files. They are now available for download at data.opengee.org. See http://www.opengee.org/geedocs/answer/6028272.html for more details.


earth_enterprise/tutorial/GoogleEarthFusionTutorial.spec
```diff
@@ -1,49 +0,0 @@
-%define _prefix /opt/google/share/tutorials/fusion
```
andreisel: Chris let's leave .spec, SConstruct and Makefile in repo? So that the only tutorial data files have been moved.
Then developers will have a chance to run scons install for tutorial, and they will know what to do in case of getting error.

earth_enterprise/tutorial/Makefile
```diff
@@ -1,24 +0,0 @@
-#
```
andreisel: keep in repo?

earth_enterprise/tutorial/SConstruct
```diff
@@ -1,63 +0,0 @@
-#-*- Python -*-
```
andreisel: keep in repo?

earth_enterprise/tutorial/GoogleEarthFusionTutorial.spec
tst-ccamp: I suppose I can add a README.md file with a link to download the files if you want to create the RPM.

andreisel: I would suggest to keep files related to the build system and rpm in repo (.rpmrc, GoogleEarthFusionTutorial.spec, SConstruct, Makefile), so they can be easily edited/updated, and only put FusionTutorial folder (data files) into the tarball and on geeoss-site.

Agreed, would be nice to add README.md with the link to download FusionTutorial.
Thanks.

tst-ccamp:
@andreisel I add the files back and added a README.md file with a link. I tried building with scons and make but couldn't get either to work right, so I left any additional instructions out of the README. If you know what I need to use to successfully run the build, I'll add it. Otherwise I plan to just leave that out until someone else has a reason to figure it out.

PR #51, opened 2017-04-03
Fix for search schema parser to handle empty values for search fields in

andreisel:
old POIfile format (xml). Fix for #50


earth_enterprise/src/server/wsgi/serve/push/search/util/search_schema_parser.py
```diff
@@ -275,26 +279,31 @@ def __EndElementData(self, elem):
     Args:
       elem: current element.
     """
+    if self._current_tag == SearchSchemaParser.R_TAG:
+      self._within_record = False
+      # b19268986 Calculate point data structure as hex and write to file
```
andreisel: It is a reference on the bug, cleaned up.

PR #61, opened 2017-04-10
Ee 44 moving dev docs

tst-ahernandez:


tst-ahernandez:
Did not mean to create two pull requests. Removing this one.

Issue #15, opened 2017-03-28
Several fusion gtest unit tests fail on RHEL 7.3


tst-ccamp:
Running the unit tests on RHEL 7.3 results in several failed tests. Investigate failures and correct as needed or create follow-up issues.

Test failures:
```
Running dbroot_generator_tests ...         FAILED
----------
[==========] Running 6 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 5 tests from Test_dbrootgen
[ RUN      ] Test_dbrootgen.Vector1
Fusion Warning:	XML file does not exist: fusion/testdata/dbroot/vector1/config.xml
Fusion Warning:	Unable to read fusion/testdata/dbroot/vector1/config.xml
unknown file: Failure
C++ exception with description "Unable to load config file fusion/testdata/dbroot/vector1/config.xml" thrown in the test body.
[  FAILED  ] Test_dbrootgen.Vector1 (4 ms)
[ RUN      ] Test_dbrootgen.ImageryTest1
Fusion Warning:	XML file does not exist: fusion/testdata/dbroot/imagery1/config.xml
Fusion Warning:	Unable to read fusion/testdata/dbroot/imagery1/config.xml
unknown file: Failure
C++ exception with description "Unable to load config file fusion/testdata/dbroot/imagery1/config.xml" thrown in the test body.
[  FAILED  ] Test_dbrootgen.ImageryTest1 (0 ms)
[ RUN      ] Test_dbrootgen.TerrainTest1
Fusion Warning:	XML file does not exist: fusion/testdata/dbroot/terrain1/config.xml
Fusion Warning:	Unable to read fusion/testdata/dbroot/terrain1/config.xml
unknown file: Failure
C++ exception with description "Unable to load config file fusion/testdata/dbroot/terrain1/config.xml" thrown in the test body.
[  FAILED  ] Test_dbrootgen.TerrainTest1 (1 ms)
[ RUN      ] Test_dbrootgen.ProtoFormatTest1
Fusion Warning:	XML file does not exist: fusion/testdata/dbroot/vector1/config.xml
Fusion Warning:	Unable to read fusion/testdata/dbroot/vector1/config.xml
unknown file: Failure
C++ exception with description "Unable to load config file fusion/testdata/dbroot/vector1/config.xml" thrown in the test body.
[  FAILED  ] Test_dbrootgen.ProtoFormatTest1 (0 ms)
[ RUN      ] Test_dbrootgen.EncodedFormatTest1
[...]

PR #70, opened 2017-04-16
Ensure unit tests can find test data (#15).

bradh:
I considered adding a symlink, but since the tests create files in the test data directory, a copy seemed safer.

### DIFF
--- a/earth_enterprise/src/common/tests/SConscript
+++ b/earth_enterprise/src/common/tests/SConscript
@@ -16,6 +16,7 @@
 
 Import('env')
 import os
+import shutil
 
 env.Append(LIBPATH = ['/usr/lib/', '/usr/lib64'])
 
@@ -50,8 +51,12 @@ map(lambda test: env.test(test, [test + '.cc'],
     ['etencoder_tests',
      ])
 
+testdir = os.path.join(env.exportdirs['bin'],'tests')
+env.copyfile(testdir, 'RunAllTests.pl')
 
-env.copyfile(os.path.join(env.exportdirs['bin'],'tests'),
-             'RunAllTests.pl')
+# Drop in the test data
+fusiontestdir = os.path.join(testdir, 'fusion')
+shutil.rmtree(fusiontestdir, ignore_errors=True)
+shutil.copytree('../../fusion', fusiontestdir)
 
 env.sharedLib('getests', ['qtpathgen.cpp'])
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/common/tests/SConscript
```diff
@@ -50,8 +51,12 @@ map(lambda test: env.test(test, [test + '.cc'],
...
 
+testdir = os.path.join(env.exportdirs['bin'],'tests')
+env.copyfile(testdir, 'RunAllTests.pl')
 
-env.copyfile(os.path.join(env.exportdirs['bin'],'tests'),
-             'RunAllTests.pl')
+# Drop in the test data
+fusiontestdir = os.path.join(testdir, 'fusion')
+shutil.rmtree(fusiontestdir, ignore_errors=True)
+shutil.copytree('../../fusion', fusiontestdir)
```
andreisel: This seems will copy src/fusion to ${bin}/tests/fusion? Why do we need this?
I couldn't find an issue for this fix,
Can you please add an issue, so it will be easier to figure out what this fix is for?
Thanks.

bradh: It is issue #15

andreisel: Thank you. 
I believe you need to copy only src/fusion/testdata.

Also, please see comment at #15. We may skip copying of testdata if we make a message more informative, e.g. suggesting to run tests from src directory.
e.g.  for this tests:
Running idl_tests ...                      FAILED
----------
[==========] Running 10 tests from 1 test case.
[----------] Global test environment set-up.
common/khxml/idl_tests.cc:39: Failure
we can just add message
https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/common/khxml/idl_tests.cc#L39

ASSERT_TRUE(khDirExists(source_dir_)) << "Couldn't find testdata dir: " << source_dir_ << ". Try to run test from src-directory";

And Assert like that can be added to other tests.

bradh: If you run from the source root, you'll end up with some test products in the source dir (rather than the build dir). That is messy (pollutes git, which could be overcome with .gitignore) and might be an issue for building on multiple architectures when we get to that stage.

tst-ccamp: I'm a little torn on this issue. On the one hand, it seems more natural to run the tests from the build directory instead of src, and I certainly don't want to keep track of a bunch of random temp files in .gitignore. But, on the other, copying all the test data can end up taking up a lot of space. I think ideally we would read the data out of the source data but keep the working directory under <builddir>/bin/tests. I'll leave this up to you, @bradh, how much more effort you'd like to put into it. But, I do agree with @andreisel that we should minimize the data that is copied by limiting to the testdata directory if possible. It's 113MB for testdata vs. 317M for all of fusion.

bradh:
As discussed on slack, a better solution would be to do the mkdir() for `fusion/testdata` then just link in the data. While unlikely, builddir and srcdir may be on different mounts, so a symlink would probably be safest.

PR #77, opened 2017-04-19
[#15] Fixes unit test issues

tst-ccamp:
Fixes https://github.com/google/earthenterprise/issues/15

This is a follow up to this PR: https://github.com/google/earthenterprise/pull/70

### DIFF
--- a/earth_enterprise/src/common/tests/RunAllTests.pl
+++ b/earth_enterprise/src/common/tests/RunAllTests.pl
@@ -22,7 +22,7 @@ use File::Basename;
 # gdal_rasterize,...) used in tests.
 $ENV{'PATH'} = join ":", "$Bin/..", $ENV{'PATH'};
 
-my @tests = glob("$Bin/*");
+my @tests = glob("$Bin/*test*");
 
 my $longest = 0;
 foreach my $test (@tests) {
--- a/earth_enterprise/src/common/tests/RunAllTests.pl
+++ b/earth_enterprise/src/common/tests/RunAllTests.pl
@@ -17,12 +17,22 @@
 
 use FindBin qw($Bin $Script);
 use File::Basename;
+use Cwd 'abs_path';
+
+# This script should be run from NATIVE-*-x86_64/bin/tests
+my $scriptDir = dirname(abs_path($0));
+if ($scriptDir !~ /NATIVE-[a-zA-Z0-9_-]*\/bin\/tests$/) {
+    print "This script should be run from <build_dir>/bin/tests.\n";
+    print "<build_dir> is usually something like NATIVE-REL-x86_64.\n";
+    exit;
+}
+chdir($scriptDir);
 
 # Add path to build binary directory for searching executables (ogr2ogr,
 # gdal_rasterize,...) used in tests.
-$ENV{'PATH'} = join ":", "$Bin/..", $ENV{'PATH'};
+$ENV{'PATH'} = join ":", "..", $ENV{'PATH'};
 
-my @tests = glob("$Bin/*test*");
+my @tests = glob("./*test*");
 
 my $longest = 0;
 foreach my $test (@tests) {
--- a/earth_enterprise/src/common/tests/SConscript
+++ b/earth_enterprise/src/common/tests/SConscript
@@ -16,7 +16,6 @@
 
 Import('env')
 import os
-import shutil
 
 env.Append(LIBPATH = ['/usr/lib/', '/usr/lib64'])
 
@@ -55,8 +54,12 @@ testdir = os.path.join(env.exportdirs['bin'],'tests')
 env.copyfile(testdir, 'RunAllTests.pl')
 
 # Drop in the test data
-fusiontestdir = os.path.join(testdir, 'fusion')
-shutil.rmtree(fusiontestdir, ignore_errors=True)
-shutil.copytree('../../fusion', fusiontestdir)
+env.Command(
+    '%s/tests/fusion/' % Dir(env.exportdirs['bin']).abspath, # TARGET
+    Dir('#fusion/testdata').abspath, # SOURCE
+    [env.rsync_cmd % (
+        '$SOURCE',
+        '$TARGET'),
+     Touch('$TARGET')])
 
 env.sharedLib('getests', ['qtpathgen.cpp'])
--- a/earth_enterprise/src/common/tests/SConscript
+++ b/earth_enterprise/src/common/tests/SConscript
@@ -16,6 +16,7 @@
 
 Import('env')
 import os
+import shutil
 
 env.Append(LIBPATH = ['/usr/lib/', '/usr/lib64'])
 
@@ -50,8 +51,12 @@ map(lambda test: env.test(test, [test + '.cc'],
     ['etencoder_tests',
      ])
 
+testdir = os.path.join(env.exportdirs['bin'],'tests')
+env.copyfile(testdir, 'RunAllTests.pl')
 
-env.copyfile(os.path.join(env.exportdirs['bin'],'tests'),
-             'RunAllTests.pl')
+# Drop in the test data
+fusiontestdir = os.path.join(testdir, 'fusion')
+shutil.rmtree(fusiontestdir, ignore_errors=True)
+shutil.copytree('../../fusion', fusiontestdir)
 
 env.sharedLib('getests', ['qtpathgen.cpp'])


bradh:
Confirm submission IAW CLA.

earth_enterprise/src/common/tests/SConscript
```diff
@@ -50,8 +51,13 @@ map(lambda test: env.test(test, [test + '.cc'],
...
 
+testdir = os.path.join(env.exportdirs['bin'],'tests')
+env.copyfile(testdir, 'RunAllTests.pl')
 
-env.copyfile(os.path.join(env.exportdirs['bin'],'tests'),
-             'RunAllTests.pl')
+# Drop in the test data
+fusiontestdir = os.path.join(testdir, 'fusion/testdata')
+shutil.rmtree(fusiontestdir, ignore_errors=True)
+testdatasrc = Dir('../../../fusion/testdata').abspath
```
andreisel: Maybe use src-base dir:
testdatasrc = Dir('#fusion/testdata').abspath

earth_enterprise/src/common/tests/SConscript
```diff
@@ -50,8 +51,13 @@ map(lambda test: env.test(test, [test + '.cc'],
...
+shutil.copytree(testdatasrc, fusiontestdir)
```
andreisel: Can use env.rsync_cmd % (testdatasrc, fusiontestdir). It is being used in other SConscript, but shutil is LGTM.

earth_enterprise/src/common/tests/SConscript
tst-ccamp: Done

earth_enterprise/src/common/tests/SConscript
tst-ccamp: Done. I actually prefer this method because it makes sure that it's executed at the correct build stage, unlike shutil which is executed immediately.

tst-ccamp:
@andreisel I've implemented your suggested changes

earth_enterprise/src/common/tests/SConscript
```diff
@@ -16,6 +16,7 @@
 
 Import('env')
 import os
+import shutil
```
tst-lsavoie: This import is no longer needed.

tst-ccamp: Done

tst-ppenev:
It worked for me, if I ran `./RunAllTests.pl` from <src/NATIVE-OPT-x86_64/bin/tests>, but not if I ran it from a different working directory.

I thought the point of using `FindBin` was to make it independent of the working directory.  @tst-ccamp , could we update `RunAllTests.pl` to change to the directory the script is located in, before running tests?

Optionally, it might be nice to get a warning if you try running <src/common/tests/RunAllTests.pl>, rather than from a build directory.

earth_enterprise/src/common/tests/SConscript
```diff
@@ -50,8 +50,16 @@ map(lambda test: env.test(test, [test + '.cc'],
     ['etencoder_tests',
      ])
 
+testdir = os.path.join(env.exportdirs['bin'],'tests')
+env.copyfile(testdir, 'RunAllTests.pl')
 
-env.copyfile(os.path.join(env.exportdirs['bin'],'tests'),
-             'RunAllTests.pl')
+# Drop in the test data
+env.Command(
```
andreisel: Chris,
maybe put copying of script into the command as well.
So, we will have the command that manages everything we need to run tests.

could be something like this:

\# put this somewhere on the top of file. I am expecting it should be a path in bin../.. 
current_dir = Dir('.').abspath    
...
testdir = os.path.join(env.exportdirs['bin'],'tests')

\# Copy the test data and the script that run all tests.
test_target = '%s/.install_for_test' % current_dir
env.Command(
    test_target, testdir,
    [env.copyfile(testdir, 'RunAllTests.pl'), 
     env.rsync_cmd % (Dir('#fusion/testdata').abspath, '%s/tests/fusion/' % Dir(env.exportdirs['bin']).abspath),
     Touch('$TARGET')])

tst-ccamp: @andreisel I don't really understand why this is necessary. The command to copy the RunAllTests.pl file works fine the way it is. And I think moving it underneath the env.Command call just makes it a little more confusing in my opinion.

tst-ccamp: This may be because I don't really like the syntax for the env.Command in the first place. The problem is made somewhat worse because scons doesn't have thorough documentation for it. Also, I don't know if env.copyfile will return a "Command" that the env.Command function can use. I think it just does the copy instead, but I could be wrong. Scons is a bit complex still in my mind. :)

andreisel: Yeah, scons syntax is confusing.
My suggestion is combining all test setup actions into one command. So if anything else is required for test, it will be added to the list in the env.Command().

Why do we need to use/return anything from env.copyfile() in env.Command().

Source in env.Command() seems allow to set some dependencies:
Suggested above:
- specifies to run set of commands in the list [..] for target "install_for_test";
- specifies dependency on testdir: $SOURCE=testdir;
- creates .install_for_test file (TARGET) showing that the target "install_for_test" has been successful.

tst-lsavoie: @andreisel I tried implementing your suggestion.  It turns out that the Command builder expects a list of strings, so it tries to execute the return value from env.copyfile as a string.  This ends up running the tests as a part of the build, which is not the intended behavior.  I think leaving it as two separate steps is the most scons-ish way of doing this.

tst-lsavoie:
@tst-ppenev I added a check to detect when the script is run outside of the expected directory and print an error.  Take another look.

earth_enterprise/src/common/tests/SConscript
andreisel: SG. Thanks.

tst-ppenev:
@tst-lsavoie: I'm rebuilding. I was thinking the script could `cd` into `$scriptDir` before running the tests, if necessary, but just a warning could be OK. At least, a new user wouldn't be confused about why the tests fail.

earth_enterprise/src/common/tests/RunAllTests.pl
```diff
@@ -17,12 +17,24 @@
 
 use FindBin qw($Bin $Script);
 use File::Basename;
+use Cwd;
+use Cwd 'abs_path';
+
+# This script should be run from NATIVE-*-x86_64/bin/tests
+my $workingDir = abs_path(cwd());
+my $scriptDir = dirname(abs_path($0));
+my $notInBuildDir = (index($scriptDir, "NATIVE") == -1);
```
tst-ppenev: Minor: Since Perl is famous for regular expressions, or regular expressions are famous because of Perl, you could use something like:

```Perl
my $notInBuildDir = $scriptDir !~ /NATIVE-[a-zA-Z0-9_-]*\/bin\/tests$/;
```

This should work in more cases than `(index($scriptDir, "NATIVE") == -1)`.

earth_enterprise/src/common/tests/RunAllTests.pl
```diff
@@ -17,12 +17,24 @@
...
+if ($workingDir ne $scriptDir || $notInBuildDir) {
```
tst-ppenev: Minor: Instead of testing `$workingDir ne $scriptDir` you could also do:

```Perl
chdir($scriptDir);
```

That way the script would work from anywhere, as long as it's called from a build directory.

In total it would be something like:

```Perl
# This script should be run from NATIVE-*-x86_64/bin/tests
my $scriptDir = dirname(abs_path($0));
chdir($scriptDir);
if ($scriptDir !~ /NATIVE-[a-zA-Z0-9_-]*\/bin\/tests$/) {
# . . .
}
```

tst-ccamp: I agree with this. It would be more user friendly in my opinion to just cd into the proper directory.

tst-lsavoie:
@tst-ppenev I tried it with the chdir command and I still got test failures.  Rather than try to debug it, I'm just leaving it as a warning.  I did add the regex, though.

tst-ppenev:
@tst-lsavoie: You also need to replace the use of the `$Bin` variable with `"."` after the `chdir`. It seems to work for me.

tst-lsavoie:
@tst-ppenev I'm still getting failures.  If I run from src it works, but if I run from earth_enterprise (one directory above src) it fails.

tst-lsavoie:
@tst-ppenev figured it out and made the suggested changes.  Take another look.

PR #85, opened 2017-04-21
Remove jpg from git lfs

### LINKED ISSUE #21: Untrack *.jpg files from git-lfs
JPG files are currently tracked in the git-lfs configuration. This not really necessary as the vast majority of them are less that 100KB and all are under 1MB. 

Removing them from LFS will remove over half of the files tracked in LFS. This will greatly speed up the cloning process and will make it less likely for LFS to receive an error while cloning due to timeouts,  network instability, or excessive API calls.

tst-lsavoie:
Fixes #21.

This untracks jpg files from git lfs and tracks them as part of the regular git repo.  The result is that many fewer files are downloaded from git lfs, resulting in a faster clone operation.


tst-ccamp:
Tested checking out the branch and opened jpg file. Reviewed changes. Looks good. Once this is merged, I'll test cloning, but don't anticipate an issue.

PR #86, opened 2017-04-22
Rename old installer directory to avoid confusion.

### LINKED ISSUE #82: Rename the old installer directory to avoid confusion
People might think the earth_enterprise/installer directory is the one used for Open GEE. It's only provided as a reference for future installer development. Renaming it something like "legacy_installer" and adding a README.md file in the directory will clear up some confusion.

bradh:
Resolves #82


bradh:
Facepalm.

PR #89, opened 2017-04-22
Rename old installer directory to avoid confusion

### LINKED ISSUE #82: Rename the old installer directory to avoid confusion

bradh:
Resolves #82


earth_enterprise/legacy_installer/README.txt
```diff
@@ -0,0 +1 @@
+This is the obsolete installer. See ../src/installer for the scripts you want.
```
tst-lsavoie: Nit: might want to reword this a bit - perhaps explain what the legacy installer is and why it's still around in addition to pointing to the new installer.  @tst-ccamp you might have a better idea of what to say here.

andreisel: src/installer is also old installer (rpm/deb packaging), though it can be reused.

andreisel: Some scripts from earth_enterprise/installer are still being used. For example, when you run
scons -j8 release=1 install install_dir... on the level of earth_enterprise, it runs installer/Scripts/verify_installerdir.sh
Please look for installer/Scripts/verify_installerdir.sh in the earth_enterprise/SConstruct.

So, maybe do not move files do far, but put in readme of earth_enterprise/installer, that \
"This is legacy installer used for continuous build with pulse and InstallAnywhere packager. 
Some of the script (e.g. installer/Scripts/verify_installerdir.sh) is still being used in the new install-scripts.
TODO: clean up ...

bradh: @andreisel would it be better to move that script (any anything else that is still being used) into `src/installer` ?

andreisel: @bradh Currently, the `src/installer` is also obsolete stuff, though it can be cleaned up and updated to use for generating rpm/deb packages.
Maybe move the `earth_enterprise/installer` to the `earth_enterprise/legacy_installer`, and copy scripts we need (scripts that are being used in Aaron-installer) into the `earth_enterprise/installer`?

bradh: I don't want to have `earth_enterprise/installer` and `earth_enterprise/src/installer` because its so easy to confuse them (results in developer friction).

Can you explain "Aaron-installer" - that doesn't mean anything to me.

tst-ccamp: @andreisel If you take a look at the fusion_installer branch, you'll see that the src/installer directory is actively in use. I guess it doesn't matter whether the directories are under src or not, but I like the idea of separating legacy scripts into legacy_installer and in-use scripts into installer. Is your suggestion to move both directories into the earth_enterprise directory? @arjntc1 do you have an opinion? We'll need to coordinate all this because fusion_installer is in active development.

@bradh I believe Aaron-installer refers to the changes Aaron has made on fusion_installer.

andreisel:
@tst-ccamp OK, if `src/installer` is actively in use, then sure you can copy anything we need from the `earth_enterprise/installer` to that directory. No, I didn't suggest to move all installer dir into earth_enterprise, but it may make sense to have installer out of src, though I would not make any recommendation here. You should see it better.

@bradh the Aaron-installer - new scripts implemented by @arjntc1 that allow to install/deploy Fusion/Server. The script I pointed above (`installer/Scripts/verify_installerdir.sh`) was being used by the old installer when running `scons install ...` from `earth_enterprise/`. So, I just pointed, that it still can be used by the new installer.

Sure, I agree with moving the `earth_enterprise/installer` to the `earth_enterprise/legacy_installer`.

tst-ccamp:
Hey, @bradh When you get a chance can you update it to move the verify_installdir.sh script to src/installer/scripts and update the top-level SConscript to point to the new location? I think that's all we're waiting on at this point.

bradh:
@tst-ccamp Will do.

andreisel:
@bradh, @tst-ccamp Also, need to update path to that file in the main [SConstruct](https://github.com/google/earthenterprise/blob/master/earth_enterprise/SConstruct#L151)

bradh:
@andreisel Just to confirm - you are saying that the helper script **could** be used, or **is** being used? I'm currently doing a full rebuild cycle, but I don't see where that script is currently called from. Can you give me a pointer?

bradh:
I've fixed the `verify_installerdir.sh` issue but now have a stack of unrelated commits in the PR (from the LFS thing, mainly). Will rebase and try again.

PR #95, opened 2017-04-25
Fusion Installer/Uninstaller

arjntc1:
Contains updated fusion installer/uninstaller scripts -- tested against Ubuntu 14.04, 16.04 and RHEL 7.  Instructions for installing Fusion can be found on the wiki:

https://github.com/google/earthenterprise/wiki/Installing-Fusion

### DIFF
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -703,7 +703,7 @@ backup_fusion()
 #-----------------------------------------------------------------
 copy_files_to_target()
 {
-	printf "\nCopying files from source to target directories..."
+	printf "\nCopying files from source to target directories... "
 
 	mkdir -p $BASEINSTALLDIR_OPT/bin
 	mkdir -p $BASEINSTALLDIR_OPT/share/doc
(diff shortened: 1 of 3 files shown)


tst-ccamp:
@asmitaw Can you post that you are OK with your commit being included to satisfy googlebot?

bradh:
We are still missing the default textures (/opt/google/share/datasets/default.gedb) for fusion.

bradh:
The server installer is assumed to still be in work. A quick test showed that the postgresql component of the install fails, and its creating gefusion init scripts instead of geserver init scripts.

arjntc1:
Brad, I'll take a look at the missing texture/file.  Thanks for pointing out!

asmitaw:
Yes, I am ok with this commit

bradh:
@arjntc1 It looks like the installer script is fine, its just that the `scons ... -installdir=/tmp/fusion_os_install install` doesn't copy the dataset directories down. I manually added them to the `/tmp/fusion_os_install` tree (under `/tmp/fusion_os_install/fusion/opt/google/share`) and the install script copied them properly.

arjntc1:
@bradh Yes, that is correct.  That is by design.  The data files are huge and it currently has to be an active choice.  Maybe a better choice would be to make it a command line argument and if the user chooses to copy the tutorial files, it could download it automatically.  Let me run that idea by a few people.  Thanks for the input!

bradh:
@arjntc1 These aren't the tutorial files (which I agree should be optional, or at least only installed if the user has already download them). These are the background textures (maps) for fusion's main window - it won't run right without them.

arjntc1:
@bradh Got it -- sorry for the misunderstanding.  I have updated the installer to include the copying of the datasets directory to the appropriate location.  Thanks again for pointing out!

wangjunqin:
built earthenterprise on ubuntu 14.04 as step by build instruction in wiki  ,and then run  sudo ./install_fusion.sh,it show "df  ‘/gevol/assets ’No such file or directory,the  installer could not determine you machine's operating system"

bradh:
@wangjunqin You could try removing the change to enable RedHat 7. If that works, try debugging what broke.

arjntc1:
@wangjunqin Can you confirm you have the latest install_fusion.sh file?  The issue you spoke of is a bug in an older version of the install script, but you shouldn't see it now.

tst-ccamp:
I've done some testing with this locally and gone through most of the code. So, I'm going to go ahead and approve/merge.

PR #105, opened 2017-04-27
Upgraded Fusion Installer/Uninstaller to work on CentOS 7

arjntc1:
* Upgraded the fusion installer and uninstaller scripts to work on CentOS 7

### DIFF
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -68,6 +68,7 @@ ROOT_USERNAME="root"
 SUPPORTED_OS_LIST=("Ubuntu", "Red Hat Enterprise Linux (RHEL)")
 UBUNTUKEY="ubuntu"
 REDHATKEY="rhel"
+CENTOSKEY="centos"
 MACHINE_OS=""
 MACHINE_OS_VERSION=""
 MACHINE_OS_FRIENDLY=""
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -297,21 +298,23 @@ determine_os()
...
     if [ -f "$OS_RELEASE1" ] || [ -f "$OS_RELEASE2" ]; then
-		if [ -f "$OS_RELEASE1" ]; then
+	if [ -f "$OS_RELEASE1" ]; then
 	        test_os="$(cat $OS_RELEASE1 | sed -e 's:\"::g' | grep ^NAME= | sed 's:name=::gI')"
-    	    test_versionid="$(cat $OS_RELEASE1 | sed -e 's:\"::g' | grep ^VERSION_ID= | sed 's:version_id=::gI')"
-		else
-			test_os="$(cat $OS_RELEASE2 | sed 's:[0-9\.] *::g')"
-			test_versionid="$(cat $OS_RELEASE2 | sed 's:[^0-9\.]*::g')"
-		fi
+    		test_versionid="$(cat $OS_RELEASE1 | sed -e 's:\"::g' | grep ^VERSION_ID= | sed 's:version_id=::gI')"
```
tst-ccamp: We should probably get rid of the tabs altogether

earth_enterprise/src/installer/uninstall_fusion.sh
```diff
@@ -159,21 +160,23 @@ determine_os()
...
 
     if [ -f "$OS_RELEASE1" ] || [ -f "$OS_RELEASE2" ]; then
-		if [ -f "$OS_RELEASE1" ]; then
-	        test_os="$(cat $OS_RELEASE1 | sed -e 's:\"::g' | grep ^NAME= | sed 's:name=::gI')"
-    	    test_versionid="$(cat $OS_RELEASE1 | sed -e 's:\"::g' | grep ^VERSION_ID= | sed 's:version_id=::gI')"
-		else
-			test_os="$(cat $OS_RELEASE2 | sed 's:[0-9\.] *::g')"
-			test_versionid="$(cat $OS_RELEASE2 | sed 's:[^0-9\.]*::g')"
-		fi
+	if [ -f "$OS_RELEASE1" ]; then
```
tst-ccamp: This might make things a little tricky, but I moved this function to a common.sh script when doing install_server.sh so they can be shared between the two install scripts. It's not merged yet, so no harm, but we'll need to remember to copy these changes into there. Git isn't smart enough to figure out that text has been moved from one file to another.

earth_enterprise/src/installer/install_fusion.sh
arjntc1: Let's talk about that -- I'll admit nano is not giving me consistent results at the moment on the tabs, but no tabs?  That's an age old debate -- spaces vs. tabs!  :)

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -68,6 +68,7 @@ ROOT_USERNAME="root"
 SUPPORTED_OS_LIST=("Ubuntu", "Red Hat Enterprise Linux (RHEL)")
 UBUNTUKEY="ubuntu"
 REDHATKEY="rhel"
+CENTOSKEY="centos"
```
tst-lsavoie: I apparently can't comment on the relevant line, but on line 268, "Mame" should be "Name".  Might as well fix that here.

earth_enterprise/src/installer/install_fusion.sh
tst-ccamp: I prefer spaces, but I don't mind tabs. What I don't like are a mixture of the two, especially on the same line :) Technically, we don't have a standard yet, so I won't require any whitespace changes before approving it, but I think we should probably follow up soon to determine our standard.

bradh:
I think this is going to conflict with the #106. Suggest merging that, then rebasing this.

earth_enterprise/src/installer/install_fusion.sh
tst-lsavoie: I'll throw my hat into the ring: I like the idea of tabs, but they never seem to work well in practice.  Spaces seems to be the most consistent and the least likely to get messed up.

earth_enterprise/src/installer/uninstall_fusion.sh
arjntc1: Hey Chris -- it does look a little crazy, right?  The issue here is that I wrote the scripts in one editor and then all of my recent work has been done in nano (for the most part).  The tabbing has been inconsistent between the two.  I'll need to mod that process.  :)

tst-lsavoie:
We should update the wiki with build instructions for CentOS 7 (that can be done separately from this ticket).

tst-ccamp:
We have an OK from @tst-lsavoie, so I'll go ahead and merge this one

PR #108, opened 2017-04-28
Rename old installer directory to avoid confusion.

bradh:
This is just a cleaner (mergable) form of #89.


tst-ppenev:
I think you may have lost the reviews on this change.

bradh:
@tst-ppenev The reviews are still on #89, but that shows all the changes being dragged in (maybe because of the git lfs changes), so its hard to review and would require everyone to agree.

tst-ppenev:
@bradh: Yeah, this one basically needs a re-review.  It may have been worth getting the `git rebase` working, so we can get approvals from the people who were keeping up with it.

PR #115, opened 2017-04-29
Fix python path for pywms server scripts.

bradh:
Completely untested.


bradh:
Raised a new ticket for the removal: #118

tst-ppenev:
Is it worth merging this, if we're going to delete the whole directory?

bradh:
@tst-ppenev I'd like to, because deleting directories is likely to be more contentious, and it does keep showing up in my search.

tst-ccamp:
Since @tst-lsavoie approved this already, that's good enough for me. I'll rebase and merge.

PR #128, opened 2017-05-04
Add a gitter badge and link the earth gitter chat.

iparanza:
Adding a Gitter badge so people can join chat using their GitHub accounts.


tst-ccamp:
I've added the badge to  the wiki, though admittedly it isn't very obvious. 

I don't really want to promote gitter over slack since currently most of the team uses slack. Of course that could change in the future depending on community interest. The two main chatrooms are linked. So, if we modify the README, I'd like to add a link to both and users can choose which they prefer.

We should also create an issue for this.

iparanza:
There is a slack badge if you'd like that instead of the gitter one. 

```
[![Slack Status](https://slack.yourdomain.com/badge.svg)](https://yourdomain.com)
```

I just wanted to make it obvious that there is a real time chat if people needed help.

iparanza:
Closing in favor of #133

PR #149, opened 2017-05-07
Avoid create a "login" user.

bradh:
The real change is to add `--system`.

There is minor command option expansion and some reformatting (I'd prefer spaces, but most of the function had tabs, so went with that) to make it clearer what is happening.

### DIFF
--- a/earth_enterprise/src/installer/common.sh
+++ b/earth_enterprise/src/installer/common.sh
@@ -179,7 +179,7 @@ check_username() {
     # add user if it does not exist
     if [ -z "$USERNAME_EXISTS" ]; then
         mkdir -p $BASEINSTALLDIR_OPT/.users/$1
-        useradd -d $BASEINSTALLDIR_OPT/.users/$1 -g gegroup $1
+        useradd --home $BASEINSTALLDIR_OPT/.users/$1 --system --gid gegroup $1
         NEW_GEFUSIONUSER=true
     else
         echo "User $1 exists"
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -177,17 +177,17 @@ main_install()
 	# add group if it does not exist
 	if [ -z "$GROUP_EXISTS" ]; then
 		groupadd -r $GROUPNAME &> /dev/null 
-        NEW_GEGROUP=true 
+		NEW_GEGROUP=true 
 	fi
 
 	# add user if it does not exist
 	if [ -z "$USERNAME_EXISTS" ]; then
-        mkdir -p $BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER_NAME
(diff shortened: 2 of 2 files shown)


earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -177,17 +177,17 @@ main_install()
...
 	# add user if it does not exist
 	if [ -z "$USERNAME_EXISTS" ]; then
-        mkdir -p $BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER_NAME
-		useradd -d $BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER_NAME -g gegroup $GEFUSIONUSER_NAME
-        NEW_GEFUSIONUSER=true
-    else
-        # user already exists -- update primary group
-        usermod -g $GROUPNAME $GEFUSIONUSER_NAME
+		mkdir -p $BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER_NAME
+		useradd --home $BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER_NAME --system --gid gegroup $GEFUSIONUSER_NAME
```
asmitaw: use $GROUPNAME  instead of gegroup
I don't remember needing a home dir for geuser, can you please confirm from previous installations?

earth_enterprise/src/installer/common.sh
```diff
@@ -179,7 +179,7 @@ check_username() {
     # add user if it does not exist
     if [ -z "$USERNAME_EXISTS" ]; then
         mkdir -p $BASEINSTALLDIR_OPT/.users/$1
-        useradd -d $BASEINSTALLDIR_OPT/.users/$1 -g gegroup $1
+        useradd --home $BASEINSTALLDIR_OPT/.users/$1 --system --gid gegroup $1
```
asmitaw: please use $GROUPNAME instead of gegroup

earth_enterprise/src/installer/install_fusion.sh
bradh: @asmitaw The groupname is a different issue, so created a tracking ticket #150. I'll look to add that, but it complicates the testing of this change a bit.

I don't know about the user directories - they don't seem to be used, but this is how it was from the initial commit of that installer script: https://github.com/google/earthenterprise/blob/6b20145854c41daa25cd365698554594ab5941d0/earth_enterprise/src/installer/install_fusion.sh#L203 
Perhaps @arjntc1  or @tst-ccamp  can advise?

earth_enterprise/src/installer/common.sh
bradh: Will fix per #150

bradh:
Added fix for group name. Just need to work through the user directory part.

bradh:
@asmitaw After discussion on slack, would like to keep the home directories for now. Can you check this meets your requirements?

tst-ccamp:
This has an approval, so it's good to go. I'm squashing and merging

PR #154, opened 2017-05-08
Server uninstaller

### LINKED ISSUE #48: Write scripts for geserver uninstall.

tst-lsavoie:
Uninstaller for GEE server.  Tested on Ubuntu 14.04 and RHEL 7.

Fixes #48.

### DIFF
--- a/earth_enterprise/src/installer/uninstall_server.sh
+++ b/earth_enterprise/src/installer/uninstall_server.sh
@@ -1,0 +1,197 @@
+#!/bin/bash
+#
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set +x
+
+# get script directory
+SCRIPTDIR=`dirname $0`
+
+. $SCRIPTDIR/common.sh
+
+# versions and user names
+GEES="Google Earth Enterprise Server"
+GEE="Google Earth Enterprise"
+LONG_VERSION="5.1.3"
+SHORT_VERSION="5.1"
+
+# script arguments
+BACKUPSERVER=true
+
+# user names
+ROOT_USERNAME="root"
+GEAPACHEUSER_NAME=""
+GEPGUSER_NAME=""
+GEFUSIONUSER_NAME=""
+GEGROUP_NAME=""
+
+# base directories
+BININSTALLROOTDIR="/etc/init.d"
+BININSTALLPROFILEDIR="/etc/profile.d"
+BASEINSTALLLOGROTATEDIR="/etc/logrotate.d"
+BASEINSTALLDIR_OPT="/opt/google"
+BASEINSTALLDIR_ETC="/etc/opt/google"
+BASEINSTALLDIR_VAR="/var/opt/google"
+
+# derived directories
+BACKUP_DIR="$BASEINSTALLDIR_VAR/server-backups/$(date +%Y_%m_%d.%H%M%S)"
+
+# additional variables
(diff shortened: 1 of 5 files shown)


tst-lsavoie:
This is now tested on Ubuntu 14.04 and RHEL 7, so it's ready for review.

earth_enterprise/src/installer/uninstall_server.sh
```diff
@@ -0,0 +1,352 @@
...
+check_user_group_delete()
+{
+  local retval=0
+    
+  if [ $DELETE_USERS_GROUPS == true ] && [ $HAS_FUSION == true ]; then
+    echo -e "\nYou cannot delete the GEE server users [$GEAPACHEUSER_NAME, $GEPGUSER_NAME, $GEFUSIONUSER_NAME]\n"
+    echo -e "and group [$GEGROUP_NAME] because $GEEF is installed on this server."
+    echo -e "$GEEF uses these accounts too."        
+    echo -e "\nExiting the uninstaller.\n"
+    retval=1
```
tst-lsavoie: I think we need another ticket to look at how the fusion and server uninstallers can work together when both are installed (there is also the problem that some files aren't deleted if fusion is installed).  I can make these changes, though, to make it easier.

tst-lsavoie:
@tst-ccamp I think I've addressed your comments.  Take another look.

earth_enterprise/src/installer/uninstall_server.sh
```diff
@@ -0,0 +1,350 @@
...
+  GEPGUSER_EXISTS=$(getent passwd $GEPGUSER_NAME)
+  GEGROUP_EXISTS=$(getent group $GEGROUP_NAME)
+}
+
+check_user_group_delete()
+{
+  local retval=0
+    
+  if [ $DELETE_USERS_GROUPS == true ] && [ $HAS_FUSION == true ]; then
+    echo -e "\nYou cannot delete the GEE server users [$GEAPACHEUSER_NAME, $GEPGUSER_NAME]"
```
tst-ccamp: I think it would be better if you always remove the users. I think that is actually what the commit message says it will do. I actually don't think this check needs to be here at all except maybe as an explanation on why gegroup will not be removed. The if statement added to remove_users_groups() should work well enough.

tst-ccamp: Doh! If you don't remove the geapacheuser and gepguser then it breaks the fusion uninstaller. It gets this message when trying to remove gegroup:
```
Delete group gegroup
groupdel: cannot remove the primary group of user 'geapacheuser'
```
Basically all the users will need to be removed before gegroup can be removed. I think this can be fixed just by removing this check. We'll also want to make sure that this will work if the uninstall is done in the reverse order.

tst-lsavoie: I'm going to argue that this is good enough for now.  It's true that the Fusion and Server uninstallers don't play well together, but fixing that should be it's own issue.  Here are the options I can see -- none of them are good:

1. Leave it as is - breaks the fusion uninstaller if you don't remove the users.
2. Provide separate arguments for deleting the users and group.  This results in the same problem listed above - if you try to delete the group but not the users, you get an error.  We'd have to add logic to ensure that the user never selected this case.
3. Always delete the users (don't give the user a choice) - would work but this is inconsistent because it's different from what the fusion installer offers.  Plus, we'd have the same problem in reverse (if you don't delete the users in the fusion uninstaller it breaks the server uninstaller).

I'd recommend going with this as an initial step and then working the conflicts between the two uninstallers separately.  This provides a working uninstaller to users and there are sufficient options to allow someone to uninstall fusion and server without errors.  We can then come back and rework both installers and uninstallers together to make them play nicely in all situations.

tst-lsavoie:
@tst-ccamp I've made a change that's similar to what you suggested.  I'm testing it right now, but go ahead and take a look and let me know what you think.

tst-lsavoie:
Tests pass.  There is still one potential problem that you encounter in the following situation:

1. Uninstall fusion but specify -ndgu so that you don't delete the users.
1. Uninstall server.

The server uninstall will try to delete the group because fusion is not installed.  The group delete will fail because the fusion users still exist.  The same thing happens if you switch fusion and server in the repro steps.

The failure is silent and the rest of the uninstaller works fine, so I think leaving it as is will work fine for now.

tst-ccamp:
Tested and reviewed. LGTM

PR #158, opened 2017-05-09
Verify that git lfs is installed when building

### LINKED ISSUE #68: Add check to scons build to make sure git-lfs is installed
Many people have attempted to clone the repo without first installing git-lfs. The error that they receive upon building is confusing ("gzip: stdin: not in gzip format"). It would be better if scons would perform the check to make sure the repo was successfully cloned and print out a helpful error message and direct the developer on how to fix the problem.

tst-lsavoie:
Adds a check to the parent SConscript that ensures that at least one tar.gz file in the repo is valid.  This implies that git lfs is installed and is set up correctly.

Fixes #68

### DIFF
--- a/earth_enterprise/src/SConscript
+++ b/earth_enterprise/src/SConscript
@@ -24,6 +24,30 @@ import os
 import time
 Import('env')
 
+# Verify that git lfs is installed by checking that a tar
+# file is valid.  This runs on every build before any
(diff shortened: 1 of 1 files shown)


tst-ccamp:
Worked like a charm once I figured out how to uninstall git-lfs :)

PR #169, opened 2017-05-15
Updated references to .jp2 files to .tif

tst-ahernandez:
Replaced the file extensions of all jp2 files in the documentation to 'tif'.


bradh:
Looks good.

Can we also strip the JPEG2000 stuff out of the fusion tutorial download to make it smaller? Perhaps a separate ticket.

PR #175, opened 2017-05-19
Create build and install targets in top level SConstruct (#131, #123)

### LINKED ISSUE #123: Update installdir value for "scons install" to match the default for the install scripts
Currently you are required to provide the installdir parameter when running the scons install target. However, since by default the install_fusion.sh and install_server.sh scripts always look for it in `installdir=/tmp/fusion_os_install`. To simplify this process it would be easier to use that as a default location and print a message to the user (preferably at the end of the scons build) which location is used.

### LINKED ISSUE #131: Build install files in a single step
Creating the install files for the fusion or server installer requires two steps, which can be confusing:

```
cd src
scons -j3 release=1 installdir=/tmp/fusion_os_install install
cd ..
scons -j3 installdir=/tmp/fusion_os_install install
```

These steps should be combined to make it easier to create the files for install.

tst-lsavoie:
This change allows the user to run a complete build and to stage install files using a single command for each. The old method of building and installing is still available, but this method is simpler because it requires fewer commands and doesn't require changing directories.  This also sets the default install directory to /tmp/fusion_os_installer.

The new instructions for building and creating install files are as follows:
```
cd earthenterprise/earth_enterprise
scons -j8 optimize=1 build
scons -j8 optimize=1 stage_install
```
Then go to `src/installer` and run the installers.  I will update the wiki with these instructions when this PR is merged.

Tested on Ubuntu 14.04 and RHEL 7.

Fixes #131 
Fixes #123

### DIFF
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -154,6 +154,7 @@ stage_tutorial_cmds = {'stage_tutorial': [
 stage_tutorial = env.PhonyTargets(**stage_tutorial_cmds)
 env.Depends(stage_tutorial, stage_data)
 env.Alias('stage_install', 'stage_tutorial')
+env.Alias('install', 'stage_install')
 
 Export('env')
 Export('installdir')
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -24,23 +24,8 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
-# only accept known arguments
-ValidOptions = { 'installdir' : 1,
-               }
-for argkey in ARGUMENTS.keys():
-   if argkey not in ValidOptions:
-      print "Unrecognized option: '" + argkey + "'"
-      Exit(1)
-
 # process commandline arguments
-installdir = ARGUMENTS.get('installdir', 0)
-
-# this fallback really shouldn't be used, whenever you specify the install
-# target you should also specify an install dir.
-# But having a default makes the rest of the SCons code easier
-if (not installdir) and ('install' in COMMAND_LINE_TARGETS):
-   print "installdir=... must be specified with install target"
-   Exit(1)
+installdir = ARGUMENTS.get('installdir', '/tmp/fusion_os_install')
 
 installdir = Dir(installdir)
 def dir_cat(installdir, subdir1, subdir2):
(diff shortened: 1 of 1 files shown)


earth_enterprise/SConstruct
```diff
@@ -134,21 +119,54 @@ env = khEnvironment(ENV={'PATH' : '/bin:/bin:/usr/bin'},
...
+# Build Fusion and Server
+build_cmds = {'build': [
+    'cd src; scons -j%d %s third_party' % (num_jobs, arg_vars),
+    'cd src; scons -j%d %s' % (num_jobs, arg_vars)
+]}
+run_build = env.PhonyTargets(**build_cmds)
+
+# Copy the binaries to the install staging directory
+stage_bin_cmds = {'stage_bin': [
+    'cd src; scons -j%d installdir=%s %s install' % (num_jobs, installdir, arg_vars)
```
tst-ppenev: Is this going to work, if `installdir` contains spaces, or other special characters?

earth_enterprise/SConstruct
```diff
@@ -134,21 +119,54 @@ env = khEnvironment(ENV={'PATH' : '/bin:/bin:/usr/bin'},
...
+]}
+stage_bin = env.PhonyTargets(**stage_bin_cmds)
+
+# Copy the data and docs to the install staging directory
+stage_data = env.Alias('stage_data', installdirs['root'])
+env.Depends(stage_data, stage_bin)
+
+# Copy the tutorial to the install staging directory if it exists.
+stage_tutorial_cmds = {'stage_tutorial': [
+    'cd tutorial; if [ -d FusionTutorial ]; then scons -j%d installdir=%s; else echo "Skipping tutorial; no tutorial files found."; fi' % (num_jobs, installdir)
```
tst-ppenev: Same as above for `installdir`.

earth_enterprise/SConstruct
```diff
@@ -134,21 +119,54 @@ env = khEnvironment(ENV={'PATH' : '/bin:/bin:/usr/bin'},
...
 SConscript(dirs=['datasets', 'geplaces', 'searchexample', 'docs'],
            exports=['env', 'installdir'])
 
-# The following target is added to phony_target(or alias_target) 'install'
-# so that it will be executed last while trying to create target install
-this_dict = {'install': ['/bin/bash ./src/installer/scripts/verify_installerdir.sh %s' % installdir]}
+# The following target is added to phony_target(or alias_target) 'stage_install'
+# so that it will be executed last while trying to create target stage_install
+this_dict = {'stage_install': [
+    '/bin/bash ./src/installer/scripts/verify_installerdir.sh %s' % installdir,
```
tst-ppenev: Looks like you should escape `installdir` here, as well.

earth_enterprise/SConstruct
```diff
@@ -134,21 +119,54 @@ env = khEnvironment(ENV={'PATH' : '/bin:/bin:/usr/bin'},
...
+
+arg_vars = getArgumentVarsAsString()
+num_jobs = GetOption('num_jobs')
+
+# Build Fusion and Server
+build_cmds = {'build': [
+    'cd src; scons -j%d %s third_party' % (num_jobs, arg_vars),
+    'cd src; scons -j%d %s' % (num_jobs, arg_vars)
+]}
+run_build = env.PhonyTargets(**build_cmds)
```
tst-lsavoie: The thinking here is that if third_party is already built, it will build pretty quickly anyway.  Doing this will only save us ~30 seconds per build, which, admittedly, might be worth it.

tst-ccamp: I think it's worth it. If the third_party scan was faster (< 5 seconds or so) then it wouldn't be so bad. There should be plenty of examples on how to do this with scons.

tst-lsavoie: I timed this and it is actually just under 5 seconds on my machine.  Given that the only reason we need to call the third_party build anyway is because of the skia library (see #176), I'm going to suggest waiting until the skia build is fixed and then removing the extra call as part of that ticket.  What do you think?

tst-ccamp: That makes sense. I can live with it since it's just a temporary work-around.

tst-lsavoie:
@tst-ppenev @tst-ccamp I believe I've addressed your comments.

tst-ccamp:
@tst-ppenev We can always bounce the ticket back if you think something should be fixed

tst-ppenev:
Well, it's a rather simple fix, and it shouldn't create a security issue, since anyone building a project should be careful. It may just cause strange failures under rather rare circumstances.

I think it would work just fine in Python, if you got rid of the extra double quotes, and used backticks, or `repr(installdir)`.

I also prefer the newer `string.format()` style in Python, which also allows you to `repr` a value before expanding it.

Example:

```Python
print "String: {!r}".format("a'b\"")
```

PR #179, opened 2017-05-20
Update date parsing to handle default date string.

### LINKED ISSUE #92: Default sourcedate for terrain / imagery import command doesn't work
Shown by:
```
genewimageryresource --nomask -o Resources/Imagery/BlueMarbleBathy /gevol/src/prebuilt/BlueMarbleBathy/BlueMarbleBathy.kip
Fusion Warning: Can't parse date/time: 0000-00-00
```

Basically the --sourcedate option isn't optional at the moment.

bradh:
Resolves #92, does not change definition of default (which I had considered, but decided was risky in terms of existing resource / project definition, or against a mismatched server version).

### DIFF
--- a/earth_enterprise/src/common/khConstants.cpp
+++ b/earth_enterprise/src/common/khConstants.cpp
@@ -91,6 +91,7 @@ const std::string kEarthClientSearchAdapter = "ECV4Adapter";
 const std::string kMapsSearchAdapter = "MapsAdapter";
 
 const std::string kUnknownDate = "0000-00-00";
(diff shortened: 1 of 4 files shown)


tst-lsavoie:
I'm seeing one oddity here: if you use the command from #92 to create an imagery resource and then open it in the asset manager, the date shows as 1900-01-00.  However, if you create an asset without a date from within the asset manager, the date remains 0000-00-00.  It would be nice to have the same behavior.

defaultdate

bradh:
@tst-lsavoie Should now be consistent.

PR #180, opened 2017-05-20
Do not remove staging tmpdir during fusion install.

### LINKED ISSUE #160: Don't remove the /tmp/fusion_os_install directory by default
Most users are likely to install both Fusion and Server. Deleting the /tmp/fusion_os_install directory by default will be confusing because they will then need to go back and run the scons install commands again.

bradh:
Resolves #160.

This is one way forward. The alternative would be to keep an option to delete the staging directory and invert the default, but it would still require renaming to make sense. Overall, automatically deleting the source doesn't seem like an option that would get used much, so I think its better to remove it all together.

It also makes fusion and server installers more consistent with each other.

### DIFF
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -902,6 +902,8 @@ show_final_success_message()
     echo -e "\n-------------------"
     echo -e "Congratulations! $GEEF $LONG_VERSION has been successfully $INSTALLED_OR_UPGRADED"
     echo -e "in [$USER_INSTALL_DIRECTORY]."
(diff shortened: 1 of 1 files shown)


tst-lsavoie:
It might be a good idea to print something at the end saying you can delete the temp files after running this script if you're not also installing GEE server.

bradh:
@tst-lsavoie should be fixed. Result now looks like:

> -------------------
> Congratulations! Google Earth Enterprise Fusion 5.1.3 has been successfully upgraded
> in [/opt/google].
> You can remove the staged install files from /tmp/fusion_os_install if you do not
> need to install Google Earth Enterprise Server.
>

PR #184, opened 2017-05-22
Sync docs with gh-pages branch

### LINKED ISSUE #191: Synchronize GEE manual content between opengee.org and master branches
Documentation exists both in the master branch and the gh_pages branch. Changes have been made on both sides, so all changes should be synchronized so that the two versions of the documentation are in sync.

AC
Documentation on gh_pages and master is identical

tst-lsavoie:
Fixes #191

Pulls in documentation changes that were made on the gh-pages branch but not on the master branch.


tst-ccamp:
@tst-lsavoie Do you think the changes to remove the special characters should be done on the gh-pages branch as well?

tst-lsavoie:
@tst-ccamp yes, I'm planning to make the same changes on both branches.  I went ahead and marked the two PRs WIP since I'm in the middle of some fairly significant changes.

tst-lsavoie:
@tst-ccamp I configured the apache server to use UTF-8 by default which fixes the special characters.  Also, that change doesn't need to be synced with the gh-pages branch.  Take another look.

tst-ccamp:
Tested locally and reviewed change, LGTM

Issue #159, opened 2017-05-09
Installers should fail if vital files are not found


tst-ccamp:
Fusion and Server installers should fail with an appropriate error message if they cannot find directories they are looking for. Instead they currently just enter an error message in the installer log, such as:
```cp: cannot stat "/tmp/fusion_os_install/common/opt/google/bin": No such file or directory```

It should display an error message that reminds users of the scons install commands that they should run withing the earthenterprise repo.

One way to implement this is to check the return value from the 'cp' command and end the install if it fails.    The tutorial files should still be optional.

asmitaw:
@eclamar is taking a look. we are not able to assign this to them directly, so this note.

tst-ccamp:
Solved with PR #232

PR #232, opened 2017-06-09
Add logic to detect corrupted source install directory and report to …

eclamar:
…user (Bug #159)

### DIFF
--- a/earth_enterprise/src/installer/common.sh
+++ b/earth_enterprise/src/installer/common.sh
@@ -150,14 +150,15 @@ show_no_tmp_dir_message()
 {
     echo -e "\nThe temp install directory specified [$1] does not exist."
     echo -e "Please specify the path of the extracted install files or first run"
-    echo -e "scons release=1 installdir=$1 install\n"
+    echo -e "scons [-j8] release=1 [installdir=$1] stage_install\n"
 }
 
 show_corrupt_tmp_dir_message()
 {
     echo -e "\nThe temp install directory specified [$1] is corrupt (failed on copy)."
+    echo -e "Please see $2 for details"
     echo -e "Please specify the path of the extracted install files or first run"
-    echo -e "scons release=1 installdir=$1 install\n"
+    echo -e "scons [-j8] release=1 [installdir=$1] stage_install\n"
 }
 
 check_bad_hostname() {
--- a/earth_enterprise/src/installer/common.sh
+++ b/earth_enterprise/src/installer/common.sh
@@ -153,6 +153,13 @@ show_no_tmp_dir_message()
     echo -e "scons release=1 installdir=$1 install\n"
 }
 
+show_corrupt_tmp_dir_message()
+{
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -585,44 +585,76 @@ copy_files_to_target()
...
+	if [ $? -ne 0 ]; then error_on_copy=1; fi
 	cp -f $TMPOPENLDAPPATH/ldap.conf.default $BASEINSTALLDIR_ETC/openldap
+	if [ $? -ne 0 ]; then error_on_copy=1; fi
 
 	# TODO: final step: copy uninstall script
 	# cp -f $TMPOPENLDAPPATH/<........> $INSTALL_LOG_DIR
 
+	if [ $error_on_copy -ne 0 ]
+	then
+		show_corrupt_tmp_dir_message $TMPINSTALLDIR
```
bradh: Not sure I follow. Why is this always `$TMPINSTALLDIR` when sometimes that isn't part of the source path?

earth_enterprise/src/installer/common.sh
```diff
@@ -153,6 +153,13 @@ show_no_tmp_dir_message()
     echo -e "scons release=1 installdir=$1 install\n"
 }
 
+show_corrupt_tmp_dir_message()
+{
+    echo -e "\nThe temp install directory specified [$1] is corrupt (failed on copy)."
+    echo -e "Please specify the path of the extracted install files or first run"
+    echo -e "scons release=1 installdir=$1 install\n"
```
tst-ccamp: We should specify that installdir is optional, usually that's done using [installdir=$1]. 

We should probably also update this and the function above to use the stage_install target instead of install to match the instructions on the wiki.

tst-ccamp: Should we also print a message that advises them that they can check the log to see which files failed?

earth_enterprise/src/installer/install_fusion.sh
eclamar: All of the 'cp's are prefaced with $TMPINSTALLDIR;  it is also consistent with the language for a failure if the $TMPINSTALLDIR doesn't exist.

tst-ccamp: @bradh $TMPINSTALLDIR should always be the top level of the install source. The other variables point to subdirectories and are just there to make maintenance of the code a bit easier. 

I suggested in another comment that we might want to point them to the log file for the specific file(s) that failed to copy.

earth_enterprise/src/installer/common.sh
eclamar: Should we add a "-j8" to the 'scons' line to make it consistent with the wiki, also?

tst-ccamp: Sure, but we should probably update both to be in brackets. It's also optional.

eclamar:
I signed it!

tst-ccamp:
Reviewed and tested. Looks good...merging

Issue #199, opened 2017-05-31
Not able to publish the pushed database from server


gautamp86:
After trying to publish the pushed database from server it gives an error (StatusCode:-1)
PFA for more information
![publish_error](https://cloud.githubusercontent.com/assets/28753606/26615998/ff112ab2-45e8-11e7-96e7-818d2127630b.PNG)
![gee_check](https://cloud.githubusercontent.com/assets/28753606/26615999/ff17144a-45e8-11e7-8888-cfbd71bb5e7a.PNG)
Kindly guide me to resolve this issue

asmitaw:
I believe that is because of the gevol is running out of space.
The part of the message points to the transfer (i.e. copy ) failing and 
the 1st error in the geecheck points to the same thing.

gautamp86:
Thanks for the reply. I have checked that also. But the disk/gevol has enough space. PFA for more details.
Is gee server allocate default size? If yes, then how to increase the limit?
![gevol_space](https://cloud.githubusercontent.com/assets/28753606/26619149/d3e386b6-45fa-11e7-9763-2164457459fd.PNG)

bradh:
Is that everything? Do you have a small `/tmp` partition?

gautamp86:
Yes. Everything is under root.

asmitaw:
https://github.com/google/earthenterprise/blob/7dbd0ded23620ffdbd73f21f3a4d3cf9ed8d807e/earth_enterprise/src/server/wsgi/serve/serve_utils.py#L416  - this code gets called when transfer happens. This is copy/ transfer  with some overrides like force_copy

you can try copying the manifest manually/commandline to see there are enough "read" permissions.

Also look at the error logs and access log for more info : /opt/google/gehttpd/logs/ ( match the time when you saw error or reproduce the error and see the latest log info)

asmitaw:
@gautamp86  Can we resolve this as answered, can you please update if there is still an issue with the error details from the access/ error log ?

asmitaw:
this is happening for me for the release as well as internal build ( clean one ), I tried changing the permissions once to geapacheuser:gegroup and it worked , whereas second time I deployed it did not fix the issue.

I believe it could be a regression or something is left behind from the old install that is not getting cleaned by the uninstall_* scripts.

Anyways this is pretty annoying.

tst-ccamp:
Fixed in #240, so closing this issue

PR #240, opened 2017-06-15
fix permissions for stream_space and search_space

asmitaw:
Not able to publish the pushed database from server #199


tst-ccamp:
Looks good. Merging

Issue #241, opened 2017-06-15
Not able to cut the globe


gautamp86:
I have installed fusion and earth server on ubuntu machine. Portable is installed on another machine (windows).
Getting following error while cutting the globe from windows machine.
![error](https://user-images.githubusercontent.com/28753606/27168690-7272e46e-51c4-11e7-9be0-5d7935ebf4cd.PNG)

asmitaw:
can you expand that error  (left arrow) and paste the message here. its hard to see for me.

gautamp86:
![snapshot_error](https://user-images.githubusercontent.com/28753606/27172726-d0b98ff0-51d3-11e7-90ef-83a0b9be9f29.PNG)

asmitaw:
so that is 3D globe you are trying to cut?

gautamp86:
yes, but not having world imagery but having specific area image

asmitaw:
I am not a portable expert, your browser is failing to load the 3D plugin so unless you have the plugin , try 2D globe first is my suggestion ( i.e.  .glm	Portable 2D map )

gautamp86:
while cutting the map also i am getting same error
![maperror](https://user-images.githubusercontent.com/28753606/27173455-67faa802-51d6-11e7-80bf-240f4a489d50.PNG)

asmitaw:
can you inspect the logs at /opt/google/gehttpd/logs/access_log and 
/opt/google/gehttpd/logs/error_log ?

asmitaw:
I mean the geserver logs that are serving the globe

asmitaw:
just see if any error reported in the logs 
tail -10  /opt/google/gehttpd/logs/error_log 
tail -10  /opt/google/gehttpd/logs/access_log

gautamp86:
![error_logs](https://user-images.githubusercontent.com/28753606/27175617-552daace-51dd-11e7-873d-1c7f30502619.PNG)

protodonuts:
Apperantly something wrong here

https://github.com/google/earthenterprise/blob/762190bd1a88dcf8e94959d201801b672bc7ee1d/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py#L308

pointed out by github's syntax highlighting.

bradh:
Certainly that newline in the middle of the string isn't a good sign...

asmitaw:
yeah, take out that copyright text and rebuild

bradh:
[...]

PR #243, opened 2017-06-15
Fix broken line in cutter script.

bradh:
Solution from @protodonuts and @asmitaw.

Note that there are additional issues in #241, this is a partial fix. Please test before merging.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
@@ -305,7 +305,7 @@ class GlobeBuilder(object):
     """Create globe info file."""
     fp = open(self.info_file, "w")
     fp.write("Portable Globe\n")
-    fp.write("Copyright 2017 Google Inc.\nLicensed under the Apache License, Version 2.0.\n");
+    fp.write("Copyright 2017 Google Inc.\nLicensed under the Apache License, Version 2.0.\n")
     fp.write("portable %s:%s\n" % (self.portable_server, self.portable_port))
     fp.write("search %s\n" % (self.search_service))
     fp.write("kml %s:%s\n" % (self.kml_server, self.kml_port))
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
```diff
@@ -305,8 +305,7 @@ def CreateInfoFile(self):
     """Create globe info file."""
     fp = open(self.info_file, "w")
     fp.write("Portable Globe\n")
-    fp.write("Copyright 2017 Google Inc.
-              Licensed under the Apache License, Version 2.0.\n")
+    fp.write("Copyright 2017 Google Inc.\nLicensed under the Apache License, Version 2.0.\n");
```
tst-ccamp: Does this work with the semicolon? You don't typically see it in python code. I say we should remove it for consistency

tst-lsavoie: A semicolon on the end of a line won't do anything in python, but I agree that it should be removed.

tst-ppenev: I've seen plenty of code where the Python interpreter seems to ignore people's semicolons fine, but it's not Python.

You could also use single multi-line string instead of a series of `write`s if you prefer.  You just need to triple the quotes.

bradh:
Fixed stray semicolon.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
bradh: Fixed.

tst-ccamp:
Looks good, but I'll test it out before doing the final merge

tst-ccamp:
Looks like @tst-awankhede verified the changes, so merging this

PR #245, opened 2017-06-15
(WIP) Move gh-pages contents to master

tst-lsavoie:
This PR is not ready to merge - it is a prototype of a possible solution.

This PR pulls the gh-pages content into master, which eliminates duplicate documentation while allowing Github to produce the opengee.org website from the repo.  Content specific to gh-pages was moved to the docs directory, and the geedocs directory (which was identical to earth_enterprise/docs/landing_page/manual/) was replaced with a symlink to earth_enterprise/docs/landing_page/manual/.  I've verified that the website operates exactly the same with this change.  You can see the new website here: https://tst-lsavoie.github.io/earthenterprise/ (spoiler: it's exactly the same as the old one).

Remaining questions:
- Is this the right way to eliminate documentation duplication?
- Do we want to merge the gh-pages branch into master in git so that the git history reflects what actually happened?

Once this change is merged, we will have to update the projects setting to pull the website source from /docs instead of the gh-pages branch.


tst-ccamp:
@tst-lsavoie I don't think you can use a custom directory name for gh-pages, at least that was the case when I was looking at setting it up. It gave you 3 options: Root directory on master, gh-pages directory on master, or root directory on gh-pages branch. I ruled out root on master because it seemed to want to go in and parse all of the earth_enterprise directory upon publishing, which seemed to bog down github. Instead of checking to see if I could exclude the directory, I decided to go with the gh-pages branch. So, if we're going to keep it on master, I would suggest moving it to the gh-pages directory.

If you want to keep the history, you should be able to first move the files on gh-pages branch into the gh-pages directory, then merge the branch with a merge commit into master.

tst-ccamp:
Well, never mind, upon further inspection it looks like gh-pages wants to use the docs directory like you have it :-)

tst-ccamp:
Along with this change, we will want to update the scons files to pull the html documentation for /opt/google/share out of the /docs folder instead of their current location.

tst-lsavoie:
@tst-ccamp I took care of that - the documentation is still in the same location.  Instead of duplicating it under docs, I just added a symlink.  So docs/geedocs is a symlink to earth_enterprise/docs/landing_page/manual/ and scons doesn't need to change because the docs are still in the same place.

tst-lsavoie:
This may not be the right way to handle this change depending on what release we want the website to be on.  There are two options I can see:

1. **Have the website display the latest version of the documentation, including changes since the last release.**  That's what this change will cause - any time someone checks in a doc change, it's immediately reflected on opengee.org.  If we do this, we'll want to copy the docs into a "previous releases" part of the website every time we do a release so that users can find the docs corresponding to their software.
2. **Have the website display the documentation from the last release.** If this is what we want, we're better off leaving the docs on the gh-pages branch and only updating them each time we do a release.  The latest docs would only be available by downloading and building the source.  If we're going with this option, we're currently in an odd spot because we've never had a release, so we update gh-pages every time we update the docs.  Once we finish our first release, though, that can stop.

tst-ccamp:
Those are both good options, but I think I like #1 best. I'd just amend it to have the website point to the latest release by default. We'd just need to update the link every time. But, I actually like the idea of keeping an archive copy of the docs for each release. They are pretty small (38M including pictures), so it's not something we would need to worry about taking up all our space. Plus, this way I think it would be less confusing to explain how to update the webpage.

tst-ccamp:
@tst-lsavoie Is this still a WIP or is it ready for review?

tst-lsavoie:
Closing this.  It will be reworked on a different branch.

Issue #148, opened 2017-05-07
Documentation still points back to old support.google.com path.


bradh:
Most of the answer pages have a footer that points back to the old site.

eclamar:
All of the links to support.google.com seem to be legitimate;  some get redirected at support.google.com to more generic introduction pages.  However, none of them need to be changed.

bradh:
I'm not sure why we're referring to support.google.com at all. Everything relevant should be on opengee.org.

tst-ccamp:
@bradh In some cases the docs refer to Google Earth client documentation which is still actively maintained by Google. I haven't reviewed them all myself, but I don't believe @eclamar noticed any external links that we had a local copy of. Are there any that you're specifically concerned about?

tst-ccamp:
Reopening this as we've found a few links that need to be updated after all.

tst-lsavoie:
Closed by PR #249.

PR #249, opened 2017-06-19
Remove 'support.google.com' references from documentation.

tst-eclamar:
Only half of the pages with references to 'support.google.com' are updated in this push.  The rest will be written up in a new bug.


tst-lsavoie:
Fixes #148.

earth_enterprise/docs/landing_page/manual/answer/6004748.html
```diff
@@ -286,7 +286,7 @@
 </td>
 <td>Localize international URL for support answers
 </td>
-<td><code>https://support.google.com/earth/#topic=4363013</code>
```
tst-lsavoie: There are other configuration values on this page that point to google.com domains (like www.google.com and earth.google.com).  Can we change those to reference opengee.org or something on the github wiki?

tst-eclamar:
Sure, but I don't know the opengee.org or github wiki well enough yet to
know what these values should be.

There are 5 references to google on this webpage(6004748.html), three of
which are the same.

Any recommendations will be appreciated.

Thanks,
Eric.

tst-lsavoie:
Since there's not a good way to fix 6004748.html and it's a fairly minor part of the documentation, we decided to leave it as is and fix it on a future issue.

Issue #188, opened 2017-05-23
GE Server Installer fails to create link to manual under htdocs


tst-ccamp:
On a new install the install_server.sh script does not create the required directory:
/opt/google/gehttpd/htdocs/shared_assets/docs

Or the required symbolic link from:
/opt/google/gehttpd/htdocs/shared_assets/docs/manual
to:
/opt/google/share/doc/manual

Reproduction steps:
1. Start with a clean /opt/google directory
1. Install OpenGEE server
1. Open admin console
1. Select settings (gear button)
1. Select Documentation

Expected:
Manual loads

Actual:
Server returns a 404 error

tst-ccamp:
Hey @tst-awankhede I just noticed another easy to fix issue with the GE Server installer. It should set the owner of the /opt/google/gehttpd/conf.d/virtual_servers directory to geapacheuser:gegroup. Do you think you can add that that to this ticket as you work it? We could make the title of the ticket more generic since it would be fixing more than one problem.

tst-awankhede:
Sounds good

tst-lsavoie:
Fixed in PR# 263.

PR #263, opened 2017-06-20
GE Server Installer fails to create link to manual under htdocs

tst-awankhede:
fix for  #188
permission changes from other requests. 
duplicate code lines removed
cutter permission issue seen in the log, so modified it to the parent.

### DIFF
--- a/earth_enterprise/src/installer/common.sh
+++ b/earth_enterprise/src/installer/common.sh
@@ -250,7 +250,7 @@ create_links()
         ln -s $BASEINSTALLDIR_VAR/run $BASEINSTALLDIR_OPT/run
     fi
 
-    if [ ! -L "$BASEINSTALLDIR_OPT/share/doc" ]; then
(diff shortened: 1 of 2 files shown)


tst-awankhede:
I signed it!

earth_enterprise/src/installer/common.sh
```diff
@@ -250,6 +250,10 @@ create_links()
         ln -s $BASEINSTALLDIR_VAR/run $BASEINSTALLDIR_OPT/run
     fi
 
+    if [ ! -L "$BASEINSTALLDIR_OPT/share/doc" ]; then
```
tst-lsavoie: I'm a little confused by this - should we be checking if $BASEINSTALLDIR_OPT/gehttpd/htdocs/shared_assets/docs exists?

tst-awankhede: yeah. that's the thing I was talking about this am. the wrong check I left in there . sending one that will fix it.

tst-awankhede: please review , its tested and the check is corrected to verify "if the link does not exists, then"

PR #284, opened 2017-06-27
Fusion Pro support for JPEG 2000 (#53)

### LINKED ISSUE #53: Fusion Pro support for JPEG 2000
In the commercial versions of Fusion Pro support for JPEG 2000 came from proprietary libraries compiled with a custom version of GDAL. These libraries will not be used in the GEE OSS version. This task is to find an open source replacement library that can restore this functionality.

tst-awankhede:
resolves #53 
This implementation uses gdal's implementation of openjpeg.

ref:
https://wiki.orfeo-toolbox.org/index.php/JPEG2000_with_GDAL_OpenJpeg_plugin

openjpeg downloaded from:
https://github.com/uclouvain/openjpeg/releases/tag/v2.1.2

### DIFF
--- a/earth_enterprise/src/third_party/gdal/SConscript
+++ b/earth_enterprise/src/third_party/gdal/SConscript
@@ -28,6 +28,7 @@
 Import('third_party_env')
 
 gdal_version = 'gdal-2.1.2'
+openjpeg_version = 'openjpeg-v2.1.2-linux-x86_64'
 # kakadu_version = 'v5_2_6-00418C'
 # geoexpress_version = 'MrSID_DSDK-8.5.0.3422'
 make_cmd_cpu = 'make -j%d' % GetOption('num_jobs')
@@ -38,6 +39,11 @@ current_dir = Dir('.').abspath
 build_root = '%s/%s' % (current_dir, gdal_version)
 gdal_source = File('#/../../earth_enterprise/third_party/gdal/%s.tar.gz'
                    % gdal_version).abspath
+
+
+## Specify path to openjpeg library here:
+openjpeg_source = File('#/../../earth_enterprise/third_party/openjpeg/%s.tar.gz'
+                      % openjpeg_version).abspath
 
 # Specify path to kakadu library here:
 # kakadu_source = File('#/../../earth_enterprise/third_party/kakadu/%s.zip'
@@ -74,7 +80,18 @@ gdal_extract = gdal_env.Command(
         'tar xzf %s\n'
         'touch %s'% (current_dir, current_dir, gdal_source, gdal_target))])
 
-# [2] Extract Kakadu
+
+# [2] Extract openjpeg
+gdal_target = '%s/.extract_openjpeg' % current_dir
+openjpeg_extract = gdal_env.Command(
+    gdal_target, [openjpeg_source, gdal_extract],
+    [gdal_env.MultiCommand(
+        'cd %s\n'
+        'tar xzf %s\n'
+        'touch %s' % (build_root, openjpeg_source, gdal_target))])
+
+
+# [3] Extract Kakadu
 # gdal_target = '%s/.extract_kakadu' % current_dir
 # kakadu_extract = gdal_env.Command(
 #     gdal_target, [kakadu_source, gdal_extract],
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -74,7 +80,18 @@ gdal_extract = gdal_env.Command(
...
+gdal_target = '%s/.extract_openjpeg' % current_dir
+openjpeg_extract = gdal_env.Command(
+    gdal_target, [openjpeg_source, gdal_extract],
+    [gdal_env.MultiCommand(
+        'cd %s\n'
+        'tar xzf %s\n'
+        'touch %s' % (build_root, openjpeg_source, gdal_target))])
+
+
+# [3] Extract Kakadu
```
bradh: Should this be kept (commented out) or just removed? Do we ever plan to support it properly?

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -128,7 +145,7 @@ gdal_extract = gdal_env.Command(
 #             kakadu_version, kakadu_version, kakadu_version, gdal_target))])
 
 
-# [6] Configure Gdal
+# [7] Configure Gdal
```
bradh: Nit: GDAL.

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -171,16 +192,17 @@ else:
...
 
 
 gdal_configure = gdal_env.Command(
 #    gdal_target, [geoexpress_extract, gdal_kakadu_build],
-    gdal_target, gdal_extract,
+    gdal_target, [gdal_extract, openjpeg_extract],
     [gdal_env.MultiCommand(configure_cmd)])
 
-# [7] Build Gdal
+# [8] Build Gdal
```
bradh: Nit: GDAL.

earth_enterprise/src/third_party/gdal/SConscript
tst-awankhede: If your comment is about kakakdu, then I believe we intentionally left it there, so whoever wants to support in their environment, it should be easy to un-comment and use it. 
It serves like config settings in apache or postgress sort of, where not everyone uses everything so its commented out and they can uncomment as required.

On the other hand , I would say we can possibly cleanup and make different scons constructs for kakadu and Mr SID and just have few lines here to include them, but that is not scope of openjpeg work.

tst-awankhede:
@tst-ccamp -  per my understanding for GDAL2.1.2 which we are using today there is a certain version of openjpeg that suppose to work with it not every version of openjpeg would work. 

here is DataSet implementation from GDAL which tracks the version:
#if defined(OPENJPEG_VERSION) && OPENJPEG_VERSION >= 20100  //****version check***
#include <openjpeg-2.1/openjpeg.h>  //********Notice specific version*****
#else
#include <stdio.h> /* openjpeg.h needs FILE* */
#include <openjpeg-2.0/openjpeg.h>
#endif

I looked up my default openjpeg install on ubuntu  16.06 and its lower version
ii  libopenjpeg5:a 1:1.5.2-3.1  amd64        JPEG 2000 image compression/decom

I hope that answers, so either we have a choice to include it in prerequisite and keep updating that list, or just have it as part of install and have people not worry about maintaining different versions.

tst-ccamp:
@tst-awankhede The package that I specified in my comment should be the version that GDAL is looking for. Here's the info I get for that package on Ubuntu 16.04:
```
libopenjp2-7-dev/xenial-updates,xenial-security,now 2.1.0-2.1ubuntu0.1 amd64 [installed,automatic]
  development files for OpenJPEG, a JPEG 2000 image library
```

It's desirable for us to use as many system libraries as possible because when we implement OS packages we wan't to bundle as few packages as possible. At least that was my understanding.

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -224,11 +246,23 @@ gdal_install = gdal_env.Command(
...
-# [10] Install these into various directories as required for build
+
+# [10] Copy Openjpeg libraries (Gdal master installer)
+gdal_target = '%s/.openjpeg_install' % current_dir
+openjpeg_install = gdal_env.Command(
+    gdal_target, gdal_install,
+    [gdal_env.MultiCommand(
+         'cd %s\n'
+         'cp -d  %s/lib/*.so* %s/lib\n'
+        'touch %s' % (
```
tst-lsavoie: Minor: inconsistent formatting

tst-lsavoie:
I've built and run this on Ubuntu 14.04 and RHEL 7.  Everything works fine.  I can build a database using jp2 images on both platforms.  Once the comments are taken care of, I think this is good to go.

earth_enterprise/src/third_party/gdal/SConscript
tst-lsavoie: Edit: I mean inconsistent indenting.

PR #285, opened 2017-06-27
Move documentation files to geedocs

tst-lsavoie:
Originally, geedocs was a symlink to earth_enterprise/docs/landing_page/manual, which contained the actual documentation.  However, Jekyll (the tool that builds opengee.org) has trouble with the symlink and throws an error. To avoid this error, this commit moves the documentation to geedocs and makes landing_page/manual a link to that, which works fine for both the website and the help files installed with fusion and server.


tst-ccamp:
Tested build locally as well as jekyll serving, reviewed code. LGTM

PR #289, opened 2017-06-30
cachedreadaccessor_unittest fails on free() call on Ubuntu 16.04 #239

### LINKED ISSUE #239: cachedreadaccessor_unittest fails on free() call on Ubuntu 16.04
The unit test receives the following error when running in Ubuntu 16.04:
```
*** Error in `./cachedreadaccessor_unittest': free(): invalid next size (normal): 0x000000000209a8c0 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7fb8a97d47e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x7fe0a)[0x7fb8a97dce0a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7fb8a97e098c]
./cachedreadaccessor_unittest[0x4041dd]
./cachedreadaccessor_unittest[0x407675]
./cachedreadaccessor_unittest[0x402bc3]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7fb8a977d830]
./cachedreadaccessor_unittest[0x402d19]
======= Memory map: ========
00400000-0040b000 r-xp 00000000 08:11 127411738                          /opt/dev/tst/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/bin/tests/cachedreadaccessor_unittest
0060b000-0060c000 r--p 0000b000 08:11 127411738                          /opt/dev/tst/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/bin/tests/cachedreadaccessor_unittest
0060c000-0060d000 rw-p 0000c000 08:11 127411738                          /opt/dev/tst/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/bin/tests/cachedreadaccessor_unittest
02038000-039bb000 rw-p 00000000 00:00 0                                  [heap]
7fb894000000-7fb894021000 rw-p 00000000 00:00 0 
7fb894021000-7fb898000000 ---p 00000000 00:00 0 
7fb89b670000-7fb89d7c6000 rw-p 00000000 00:00 0 
...
```

tst-awankhede:
resolves #239 ,
root cause:
The issue is that sometimes request_count is too big and by the time end of function reaches that memory is already taken by some other culprit or overwritten, so at the end when is comes to free() up the string buffer it runs in SEGFAULT.
TestCacheBlock and Pread agravates the problem with multiple long reads

Resolution:
Made string a global member in tests.

### DIFF
--- a/earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
+++ b/earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
@@ -65,7 +65,7 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
     if (!keep_temp_bundle_) {
       khPruneDir(path_base_);
     }
-    delete test_buffer_;
+    delete [] test_buffer_;
   }
 
   // Maintain Test bundle info for all the tests.
--- a/earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
+++ b/earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
@@ -64,6 +64,7 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
   ~CachedReadAccessorUnitTest() {
     if (!keep_temp_bundle_) {
       khPruneDir(path_base_);
+      delete test_buffer_;
     }
   }
 
@@ -72,7 +73,6 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
   uint32 segment_size_;
   std::string path_base_;
   std::string bundle_name_;
-  std::string buffer_;
   char* test_buffer_;
   std::vector<std::string> segment_names_;
   std::string header_path_;
@@ -158,8 +158,10 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
                    uint32 max_blocks,
                    uint32 block_size,
                    std::string prefix) {
-    buffer_.resize(request_count);
-    char* out_buffer = const_cast<char*>(buffer_.data());
+    std::string buffer;
+    buffer.reserve(block_size);
+    buffer.resize(request_count);
+    char* out_buffer = const_cast<char*>(buffer.data());
 
     // Test Read uncached of entire filebundle
     CachedReadAccessor accessor(max_blocks < 2 ? 2 : max_blocks, block_size);
@@ -223,9 +225,10 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
                                     segment_names_[segment_id],
                                     segment_names_[segment_id],
                                     segment_id);
-
-    buffer_.resize(request_count);
-    char* out_buffer = const_cast<char*>(buffer_.data());
+    std::string buffer;
+    buffer.reserve(block_size);
+    buffer.resize(request_count);
+    char* out_buffer = const_cast<char*>(buffer.data());
 
     // Test read within a single block
     // uncached block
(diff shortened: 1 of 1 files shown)


tst-eclamar:
How does making 'buffer' a global solve the problem?  Is this a threading issue?  What am I missing?

Also, where does the 'free()' occur?  I see 'test_buffer_' being allocated, but never deallocated.

tst-awankhede:
@tst-eclamar the memory allocated for std:string would be freed ( made available for other parts of program to use ) when it goes out of scope. There is no guarantee what gets written in that location later or after the resize() when resized to smaller length. Hence keeping it global so does not need to go out of scope.  The "free" call, where it was failing is implicit free() in memory management , not something we explicitly call.

the heap allocation by "new" gets taken care by destructor , so there are no free() calls here in this code (my understanding).

tst-awankhede:
Also,it would be more clear by using debugger.

earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
```diff
@@ -71,8 +71,9 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
   uint32 file_bundle_size_;
   uint32 segment_size_;
   std::string path_base_;
-  char* test_buffer_;
   std::string bundle_name_;
+  std::string buffer_;
+  char* test_buffer_;
```
tst-lsavoie: Looks like this is unused and can be removed.

tst-awankhede: test_buffer_ is used in TestCachedReadAccessorPread

tst-lsavoie:
The problem seems to be on line 250 of the original version of the file.  We allocate a buffer of less than `block_size` bytes and then try to read `block_size` bytes into it.  We can fix that by increasing the size of the buffer or resizing the buffer around line 250.

tst-awankhede:
@tst-eclamar  and @tst-lsavoie  - please see revised fix. I have tested it with my install.

earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
```diff
@@ -238,6 +241,8 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
     // same block cached read
     offset *= 2;
     request_count -= offset;
+    buffer.resize(request_count);
```
tst-lsavoie: There's a potential problem here that I didn't think about when we talked the other day.  Every call to resize could potentially change the location of the underlying character array, which would invalidate out_buffer.  That shouldn't happen since we've already reserved enough capacity in the string, but it's technically possible.  It would be safer to just run buffer.resize(block_size) at the top of the function and remove the extra resize calls.

tst-awankhede: resize would truncate if reducing and push back if adding more space keeping the start of buffer at same address - the reserve would guarantee that a continuous block of memory is reserved hence no overwrite by other parts of program.

earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
```diff
@@ -64,15 +64,16 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
   ~CachedReadAccessorUnitTest() {
     if (!keep_temp_bundle_) {
       khPruneDir(path_base_);
+      delete test_buffer_;
```
tst-awankhede: yes , fixing

earth_enterprise/src/common/filebundle/cachedreadaccessor_unittest.cpp
```diff
@@ -65,14 +65,15 @@ class CachedReadAccessorUnitTest : public UnitTest<CachedReadAccessorUnitTest> {
     if (!keep_temp_bundle_) {
       khPruneDir(path_base_);
     }
+    delete test_buffer_;
```
tst-lsavoie: This should be `delete [] test_buffer_;`

Issue #288, opened 2017-06-30
Add "--listvolumes" flag to geconfigureassetroot


tst-ccamp:
The only way that geconfigureassetroot will list the configured volumes is by using the "--editvolumes" flag, which also requires gefusion services to be stopped. This task will add a flag to list the volumes without going into edit mode. It is also preferable if it would list the volumes without requiring the fusion services to be stopped.

# AC
1. Add "--listvolumes" option to geconfigureassetroot command to list the volumes
1. If possible report the volumes without requiring the gefusion service to be stopped

# Note
Use the output from the "--editvolumes" option as a guideline for formatting the output:
```
Current Volume Definitions:
--- name: netpath,localpath,isTmp ---
1)  assetroot: /gevol/assets,/gevol/assets,N
2)  data: /opt/Data,/opt/Data,N
3)  opt: /opt/google/share/tutorials,/opt/google/share/tutorials,N
4)  src: /gevol/src,/gevol/src,N
```

bradh:
@tst-ccamp Can you re-open. I still owe the documentation part of this change.

tst-ppenev:
@bradh: Do you mean updating documentation under `earthenterprise/docs/geedocs/answer/`?

This ticket doesn't explicitly mention updating the documentation, and has been approved, so, if you'd like to update documentation, we can create a new issue.

bradh:
@tst-ppenev Yes - documentation update to the affected page, identified in the PR  (https://github.com/google/earthenterprise/pull/299): geedocs/answer/3481558.html

tst-ppenev:
@bradh: Feel free to create an issue for that. Or, if you prefer, I can write one.

PR #299, opened 2017-07-05
Add "--listvolumes" flag to geconfigureassetroot utility (#288)

bradh:
Fixes: #288 
Does require running as root, but does not require fusion service to be stopped.

Will follow up with documentation PR if this meets needs: expect that only geedocs/answer/3481558.html would need to be updated.

### DIFF
--- a/earth_enterprise/src/common/khstl.h
+++ b/earth_enterprise/src/common/khstl.h
@@ -358,6 +358,19 @@ makeset(const T &e1, const T &e2, const T &e3, const T &e4, const T &e5)
   tmp.insert(e5);
   return tmp;
 }
+template <class T>
+std::set<T>
+makeset(const T &e1, const T &e2, const T &e3, const T &e4, const T &e5, const T &e6)
+{
+  std::set<T> tmp;
+  tmp.insert(e1);
+  tmp.insert(e2);
+  tmp.insert(e3);
+  tmp.insert(e4);
+  tmp.insert(e5);
+  tmp.insert(e6);
+  return tmp;
+}
 
 
 template <class Key, class Value>
--- a/earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
+++ b/earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
@@ -219,8 +219,9 @@ int main(int argc, char *argv[]) {
       }
       AddVolume(status, volume_name, volume_directory);
     }
-
-    printf("Configured %s.\n", status.assetroot_.c_str());
+    if (!listvolumes) {
+      printf("Configured %s.\n", status.assetroot_.c_str());
+    }
   } catch (const std::exception &e) {
     notify(NFY_FATAL, "\n%s", e.what());
   } catch (...) {
--- a/earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
+++ b/earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
@@ -25,6 +25,7 @@
 #include <config/FixAssetRoot.h>
 #include <config/RecoverIds.h>
 #include <config/gePrompt.h>
+#include <config/geConfigUtil.h>
 #include <DottedVersion.h>
 #include <autoingest/geAssetRoot.h>
 #include <geUsers.h>
@@ -50,7 +51,7 @@ void usage(const char *prog, const char *msg = 0, ...) {
     (stderr,
      "\n"
      "usage:\n"
-     "  %s {--new|--repair|--editvolumes|--fixmasterhost|--addvolume [options]}\n"
+     "  %s {--new|--repair|--editvolumes|--fixmasterhost|--addvolume|--listvolumes [options]}\n"
      "Configure this machine to run Google Earth Fusion.\n"
      "  --help, -?                  Display this usage message\n"
      "\n"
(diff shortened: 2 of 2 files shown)


earth_enterprise/src/common/khstl.h
```diff
@@ -358,6 +358,19 @@ makeset(const T &e1, const T &e2, const T &e3, const T &e4, const T &e5)
   tmp.insert(e5);
   return tmp;
 }
+template <class T>
+std::set<T>
+makeset(const T &e1, const T &e2, const T &e3, const T &e4, const T &e5, const T &e6)
```
tst-ppenev: Are we running a GCC that doesn't support initializer lists?

earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
```diff
@@ -148,7 +155,13 @@ int main(int argc, char *argv[]) {
...
     // and switch to fusion user
-    std::string thishost = ValidateHostReadyForConfig();
+    std::string thishost;
+    if (listvolumes) {
+      AssertRunningAsRoot();
+      thishost = GetAndValidateHostname();
+    } else {
+      thishost = ValidateHostReadyForConfig();
+    }
     printf("Switching to (%s, %s)\n", username.c_str(), groupname.c_str());
```
tst-ppenev: Looks like we need a ticket to convert this I/O, and, possibly, other operations to C++.

earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
```diff
@@ -388,6 +404,42 @@ void RepairExistingAssetRoot(const AssetRootStatus &status, bool noprompt) {
 }
 
 // ****************************************************************************
+// ***  ListVolumes
+// ****************************************************************************
+void listCurrentVolumes(VolumeDefList *const voldefs, std::vector<std::string> volnames) {
+    printf("\nCurrent Volume Definitions:\n");
```
tst-ppenev: Minor: I could see the use of a flag, such as `-q, --quiet` which would disable these prefix and postfix lines. It may be slightly easier to parse the output of this option without them.

earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
```diff
@@ -77,6 +78,8 @@ void usage(const char *prog, const char *msg = 0, ...) {
      "    [--assetroot <dir>]       selected asset root or, modify the local\n"
      "                              path definition for an existing volume,\n"
      "                              or to add a volume definition.\n"
+     "  --listvolumes               List available volumes in the asset root.\n"
```
tst-ppenev: Q: Should this be called, e.g., `--listhostvolumes` to differentiate listing the volumes on the current host from listing all defined volumes? Or, perhaps, this option can take a parameter indicating what host to list volumes for, with some parameter keywords for the current host and all hosts.

Or, perhaps, this can be left for the future, if the future default has the same behavior as this one.

bradh: @tst-ppenev: the task (#288) is pretty clear on what the option is to be called, and what the output format is.

earth_enterprise/src/common/khstl.h
bradh: Just being consistent with the existing code. This seems like too small a change to refactor the STL implementation.

earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
bradh: @tst-ppenev Unchanged code - out of scope for this ticket. Feel free to raise another ticket.

earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
bradh: @tst-ppenev Unchanged code (just refactored to be shared method, avoiding duplication). Out of scope for this ticket.

earth_enterprise/src/common/khstl.h
tst-ppenev: I'm not sure what you mean by refactoring the STL implementation. GCC supports initializer lists since 4.4, and the STL supports them, so you can create, e.g., a set initialized with whatever elements you want in a single line of code.

I see that you're being consistent with the code, I wouldn't hold it up for this particular question.

However, I think we should be getting away from this code, rather than expanding its use.  I would also like to make the version and type of compiler we require for building GEE explicit, so it's clear what code we can write.  I think all the OS distributions that we've listed as supported (RHEL 6, Ubuntu 14) have, at least, GCC 4.4, and GCC is now at version 7.

bradh: This is in the KH STL header (`common/khtstl.h`) - I don't want to refactor (and retest every use) of that code.

earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
```diff
@@ -182,6 +195,9 @@ int main(int argc, char *argv[]) {
     if (fixmasterhost) {
       FixMasterHost(status);
     }
+    if (listvolumes) {
```
tst-ppenev: Minor: There's a `printf` on line 223 below where I can't add a comment from Git Hub. The `Configured /some/assetroot.` message doesn't make much sense when just listing volumes. I don't think it should be shown when `--listvolumes` is specified.

bradh: Will fix to say:

```
Switching to (fusionuser, fusiongroup)
Validating assetroot

Current Volume Definitions:
--- name: netpath,localpath,isTmp ---
1)  assetroot: /gevol/assets,/gevol/assets,N
2)  src: /gevol/src,/gevol/src,N
-------------------------------------
```

earth_enterprise/src/fusion/config/geconfigureassetroot.cpp
tst-ccamp: I like adding host as a separate parameter rather than changing this one. @tst-ppenev can you write a follow-up ticket for that change?

tst-ccamp:
@tst-ppenev has approved this so, I'll go ahead and merge it. Thanks, @bradh!

PR #300, opened 2017-07-05
Bug 168 known volume for tutorial data

tst-eclamar:
Write a script to download tutorial data.


tst-lsavoie:
Replaced by PR #301.

PR #301, opened 2017-07-05
Bug#168, part 2 - add script to download tutorial data

tst-eclamar:
Add script to download data.

### DIFF
--- a/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
+++ b/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+if [ ! -d Imagery ]; then
+	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O /tmp/FusionTutorial-Full.tar.gz
+	tar -xvzf /tmp/FusionTutorial-Full.tar.gz -C .
+	rm -f /tmp/FusionTutorial-Full.tar.gz
+fi
--- a/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
+++ b/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 
 TMPFILE=/tmp/FusionTutorial-Full.tar.gz
+SCRIPTDIR=`dirname $0`
 
 if [ ! -d Imagery ]; then
 	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O $TMPFILE
-	tar -xvzf $TMPFILE -C .
+	tar -xvzf $TMPFILE -C $SCRIPTDIR
 	rm -f $TMPFILE
 fi
--- a/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
+++ b/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-TMPFILE=/tmp/FusionTutorial-Full.tar.gz
-SCRIPTDIR=`dirname $0`
+TMPFILE="/tmp/FusionTutorial-Full.tar.gz"
+SCRIPTDIR=`dirname "$0"`
 
-if [ ! -d Imagery ]; then
-	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O $TMPFILE
-	tar -xvzf $TMPFILE -C $SCRIPTDIR
-	rm -f $TMPFILE
+if [ ! -d "$SCRIPTDIR/Imagery" ]; then
+	wget "http://data.opengee.org/FusionTutorial-Full.tar.gz" -O "$TMPFILE"
+	tar -xvzf "$TMPFILE" -C "$SCRIPTDIR"
+	rm -f "$TMPFILE"
 fi
--- a/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
+++ b/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
+
+TMPFILE=/tmp/FusionTutorial-Full.tar.gz
+
 if [ ! -d Imagery ]; then
-	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O /tmp/FusionTutorial-Full.tar.gz
-	tar -xvzf /tmp/FusionTutorial-Full.tar.gz -C .
-	rm -f /tmp/FusionTutorial-Full.tar.gz
+	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O $TMPFILE
+	tar -xvzf $TMPFILE -C .
+	rm -f $TMPFILE
 fi
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -879,6 +879,7 @@ fix_postinstall_filepermissions()
     find $BASEINSTALLDIR_OPT/share -type f -exec chmod 644 {} \;
     chown -R $ROOT_USERNAME:$ROOT_USERNAME $BASEINSTALLDIR_OPT/share
     chmod ugo+x $BASEINSTALLDIR_OPT/share/support/geecheck/geecheck.pl
+    chmod ugo+x $BASEINSTALLDIR_OPT/share/tutorials/fusion/download_tutorial.sh
 
     # Restrict permissions to uninstaller and installer logs
     chmod -R go-rwx $INSTALL_LOG_DIR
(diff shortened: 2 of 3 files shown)


earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+if [ ! -d Imagery ]; then
+	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O /tmp/FusionTutorial-Full.tar.gz
```
tst-lsavoie: Minor: it might be a good idea to make `/tmp/FusionTutorial-Full.tar.gz` a variable.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
tst-lsavoie: It occurs to me that we should have some instructions about this somewhere.  Maybe we could add a note to the wiki where it talks about installing the tutorial files: https://github.com/google/earthenterprise/wiki/Install-Fusion-or-Earth-Server.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-awankhede: check for the presence of the tar.gz before wget ?

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-eclamar: Ok.  How do I edit it such that it can be reviewed?

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -0,0 +1,6 @@
...
+if [ ! -d Imagery ]; then
+	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O /tmp/FusionTutorial-Full.tar.gz
+	tar -xvzf /tmp/FusionTutorial-Full.tar.gz -C .
```
tst-lsavoie: It would be nice if this script could run from any location but still download the files to the directory that the script is in.  For example, you could call it from a parent directory (`tutorial/fusion/download_tutorial.sh`) and it would download the files to the directory that the script is in.  You can get the directory the script is in by running something like SCRIPTDIR=\`dirname $0\`.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-lsavoie: When you push commits up to the Bug-168_download-tutorial-file-script branch they are automatically added to the PR.

tst-eclamar: Sorry, I meant the webpage 'Install-Fusion-or-Earth-Server' - does that have a different review cycle?

tst-lsavoie: No, you can just edit that directly.  Maybe post a link to it in a comment on this PR so we can look at it.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
+
+TMPFILE=/tmp/FusionTutorial-Full.tar.gz
+
 if [ ! -d Imagery ]; then
```
tst-awankhede: its still not clear to me why checking for Imagery and not  the tar file ?
Also - what part of the code would actually copy the /download_tutorial.sh  into the tmp_install ?

tst-eclamar:
Because the AC has the FusionTutorial-Full.tar.gz to be deleted at the end
of the processing.  Thus, only Imagery (and the other three directories)
will exist.  I can change it to test for all four directories...

The 'download_tutorial.sh' script is installed by 'install_fusion.sh' in
 /opt/google/share/tutorials/fusion, which is where the Imagery, etc.,
directories will exist.

tst-awankhede:
1) I read the ticket again, so its clear about the location /opt/google/share/tutorials/fusion/download_tutorial.sh 

2) in Part1  - sudo  /opt/google/bin/geconfigureassetroot --addvolume opt:/opt/google/share/tutorials --noprompt --nochown
- when I ran this command and, it fails if opt already exists, so can you please add a check not to run this command if opt volume is defined ? 
3) In tar -xvzf $TMPFILE -C .

instead of -C . , it should be -C   /opt/google/share/tutorials/fusion/ otherwise the unzip will put the tutorial resources in whatever dir its run from.

e.g. in the below case 
~/geeinstalldir_copy/common/opt/google$ sudo  /opt/google/share/tutorials/fusion/download_tutorial.sh 
the resources would be put in ~/geeinstalldir_copy/common/opt/google

tst-eclamar:
@tst-awankhede  For comment #3, Lee had a similar comment, so it has been corrected - the output will always go to /opt/google/share/tutorials/fusion.  For comment #2, shall I re-open Part 1, GP-178?

tst-awankhede:
For comment #2, shall I re-open Part 1, GP-178?  - I think we can fix it here in this pull request , just edit that file and "git add" git commit , git push in this PR is okay.

tst-eclamar:
How do I check if 'opt' is defined? - geconfigureassetroot has no query mode (yet).

It would seem to only be an issue if 'opt' is already defined and set to somewhere else.

Do you have an example command line that shows how to do this query?

bradh:
@tst-eclamar Do you need some query mode other than the #288 requirement? If not, we could merge that :-)

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+TMPFILE=/tmp/FusionTutorial-Full.tar.gz
+SCRIPTDIR=`dirname $0`
```
tst-ppenev: Please, use quotes in your shell script. I know we expect to put this script under a path that doesn't include any shell special characters, but that's not a complete guarantee, since this could be symlinked from any location.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -0,0 +1,10 @@
...
+
+if [ ! -d Imagery ]; then
```
tst-ppenev: Are you expecting to be in a specific directory when you run this script? What is `Imagery` relative to?

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -0,0 +1,10 @@
...
+	wget http://data.opengee.org/FusionTutorial-Full.tar.gz -O $TMPFILE
+	tar -xvzf $TMPFILE -C $SCRIPTDIR
```
tst-ppenev: Same comment: Add your quotes.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-eclamar: @tst-ppenev I'm sorry, Pavel, but I'm not sure what you mean by quoting here - please illustrate so I get it right.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-eclamar: You're right, good catch.  Corrected.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-ppenev: @tst-eclamar: Sorry. E.g.:
```bash
SCRIPTDIR=$(dirname "$0")
```

or 

```bash
SCRIPTDIR=`dirname "$0"`
```

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-ppenev: E.g.:
```bash
tar -xvfz "$TMPFILE" -C "SCRIPTDIR"
```

tst-ppenev: I'd, personally, quote all variables that need to expand to one command argument, even for `wget` and `rm`, although the `$TMPFILE` virable doesn't contain any special characters.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
tst-ccamp: I think this demonstrates what @tst-ppenev means:
```
ccamp@xtree-ubuntu:/$ cat /tmp/test\ me/testbash.sh 
#!/bin/bash
SCRIPTDIR=`dirname "$0"`
SCRIPTDIR2=`dirname $0`
echo $SCRIPTDIR
echo $SCRIPTDIR2

ccamp@xtree-ubuntu:/$ /tmp/test\ me/testbash.sh 
/tmp/test me
/tmp me
```

The first line is the correct way to use quotes so that it will pick up weird directories, such as ones that contain spaces. Anywhere you use $0 or $SCRIPTDIR you'll surround it with double quotes. FYI, you don't want to use single quotes because that makes the contents literal. i.e. it will not resolve variables and give you the variable name instead.

tst-lsavoie:
Tested the latest version - looks good.

Issue #311, opened 2017-07-08
Allow `geconfigureassetroot --listvolumes` to List Known Volumes for a Given Host, and for All Hosts


tst-ppenev:
The `--listvolumes` option to the `geconfigureassetroot` command lists the known volumes on the current host. To make it fully-featured, we'd like to allow `geconfigureassetroot` to list the known volumes for any given host, or for all known hosts.

## Notes:

This could be implemented by expanding the `geconfigureassetroot --listvolumes` option as follows.

* Add an optional `host` argument to the `--listvolumes` option. If the argument is a valid host name, list the known volumes for that host.
* If the `host` argument has a special value (such as `@all`, `:allhosts:`, or some other value distinguishable from a host name), list the known volumes on all known hosts.
* If the `host` argument has a different special value (such as `@localhost`, `:localhost:`), or, perhaps, a standard IPv4 or IPv6 localhost value (such as `127.0.0.1`, or `::1`) without special treatment, list the known volumes for the host the command is executed on.
* If the optional `host` argument is not specified, default to listing the known volumes for the host the command is executed on.

# AC:

* `geconfigureassetroot` should support listing the known volumes on a given network host.
* Preferably, a host should be able to be specified by any of its DNS names, or by an IP address. (Unless, that requires too much effort, in which case it could be split into a new ticket.)
* `geconfigureassetroot` should support listing the known volumes on all known network hosts.
* The new command syntax should be backward-compatible with `geconfigureassetroot --listvolumes` which lists the known volumes on the current host.

PR #313, opened 2017-07-08
Update documentation to reflect #299 changes (#312)

### LINKED ISSUE #312: Update documentation to reflect new geconfigureassetroot --listvolumes option
The #288 task (as implemented by #299) requires documentation updates.

bradh:
Fixes #312

### DIFF
--- a/docs/geedocs/answer/3481558.html
+++ b/docs/geedocs/answer/3481558.html
@@ -90,10 +90,10 @@
   <p><i>Default</i>. Uses normal imagery for the project.</p>
   <h2><a name="63542"></a>geconfigureassetroot</h2>
   <pre>
-<code><b>geconfigureassetroot</b> {<b>--new</b> <b>--assetroot </b><i>path </i> [<b>--srcvol </b><i>path</i>] | <b>--repair</b> | <b>--editvolumes</b> | <b>--addvolume</b> | <b>--fixmasterhost</b> | <strong>--noprompt</strong>}  [--nochown]</code></pre>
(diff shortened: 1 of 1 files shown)


docs/geedocs/answer/3481558.html
```diff
@@ -127,6 +127,9 @@
 <code><b>--editvolumes</b> [--assetroot <i>path</i>]</code></pre>
 <p><i>Optional</i>. Follow the prompts to add a volume to the selected asset root or, modify the <code>localpath</code> definition for an existing volume, or to add a volume definition.</p>
 <pre>
+<code><b>--listvolumes</b> [--assetroot <i>path</i>]</code></pre>
+<p><i>Optional</i>. List the available (configured) volumes for the selected asset root.</p>
```
tst-ppenev: Perhaps, we should mention that only the volumes on the host the command is executed from will be shown.

tst-ccamp: I don't think there are any host options for any of the commands for geconfigureassetroot, not just for "--listvolumes". So, if we wanted to add that information it should be a general comment for the whole command, not specifically for list volumes.

bradh: Overall I think this would be better incorporated when #311 goes in.

Right now there is nothing that suggests it would "magically" reach out to another host to list volumes or do any other operation, and command line tools usually work on the host they're run on.

tst-ppenev: You're right. I'm fine with the way it is now.

tst-ccamp:
We've got an approval, so merging

Issue #309, opened 2017-07-07
Check for the FusionConnection before new asset is populated


tst-awankhede:
1. if gefusion assetmanager is stopped
2. and create new resource and save it 
3. save fails as fusion server not up
4.  but by this time saved_asset_request_ is formed
5.  Now even if you start the the server, saving does not have any effect as request is stale and there is no indication of the fact that save had failed before, also because the listener listens to change in values of the asset request and there is no change in value but only server state has changed, the save practically does not proceed. 
6. "Save As" may work but that internally calls save so I believe it may work only with new file name not if user proceeds with same file name because of the condition that checks for request.name != saved_request_Asset_.name -> then only save as new id

for all these reasons and for the main reason of preventing user from loosing all the edited/ populated contents in asset window, we should check the gefusion status before new asset is allowed to be created.

Issue #316, opened 2017-07-10
Add a state to request.config


tst-awankhede:
we would need to have a way to identify that even if the request is saved in the request.config model, in case of error, its not saved on the disk. Hence following scenario is failing 

1. specify a resource path from unknown volume to create fusion resource  from
2. it fails  because of volume issue - expected
3. now we correct the path
4. Save the resource and save succeeds - expected
5. now change anything on resource UI
6. Close the resource - Close does not identify that the resource is modified after Saved in step 4. Hence the IsModified is not detecting that a resource is modified.
proposed is either add a state or a create saved_edit_request_ again after the successful save.

PR #317, opened 2017-07-10
Once failed to save a new resource, the resource cannot be saved again

### LINKED ISSUE #167: Once failed to save a new resource, the resource cannot be saved again correctly
1. In Fusion, open "Asset Manager", click "New Imagery Resource", add some invalid file(s) (e.g. unsupported file, or Tiff file without SRS, or no file at all) to the "Source File(s)" list.
2. Select File menu -> Save. Specify a resource name, then click "Save".

After an error message shows up as expected, there are problems:

- "Save" in File menu is disabled.
- "Save As..." looks OK, but any attempt to correct the "Source File(s)" list would be futile. The resource cannot be saved because the same error is raised again.

To proceed, I have to close the window and open a new one.

(Built on Ubuntu 14)

tst-awankhede:
resolves #167

### DIFF
--- a/earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
+++ b/earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
@@ -660,8 +660,8 @@ void RasterAssetWidget::AssembleEditRequest(
                           khErrnoException::errorString(errno));
       case SourceConfig::NonVolume:
         throw khException(filename + tr(" doesn't reside on a known volume.\n") +
-                         " You can move your asset files to a known volume, or create a new volume that contains their current location." +
-                         " You can list known volumes and create new ones from the command line with geconfigureassetroot.");
+                         " You can move your asset files to a known volume, or create a new volume\n" +
+			 " that contains their current location using geconfigureassetroot command.");
     }
   }
 
--- a/earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
+++ b/earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
@@ -644,6 +644,10 @@ void RasterAssetWidget::AssembleEditRequest(
   request->meta.SetValue("sourcedate", acquisition_date_wrapper_->GetDate());
 
   request->sources.clear();
+  if (source_list->numRows() == 0){
+    throw khException(tr("No source files specified \n") +
+                      khErrnoException::errorString(errno));
+  }
   for (int row = 0; row < source_list->numRows(); ++row) {
     std::string filename = source_list->text(row).latin1();
     SourceConfig::AddResult result = request->sources.AddFile(filename);
--- a/earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
+++ b/earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
@@ -655,7 +655,7 @@ void RasterAssetWidget::AssembleEditRequest(
                           + filename + "\n" +
                           khErrnoException::errorString(errno));
       case SourceConfig::NonVolume:
-        throw khException(filename + tr(" doesn't reside on a known volume."));
+        throw khException(filename + tr(" doesn't reside on a known volume. \n Make sure the volume is configured correctly using geconfigureassetroot and try again."));
     }
   }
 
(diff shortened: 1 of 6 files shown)


tst-eclamar:
I was able to reproduce the error.

However, the code fix does not solve the problem; or, more specifically,  changes the problem.  Now, after selecting 'Save', and getting an error about no resource selected, one can select 'Save' again (note at this point: the 'Build' option is not grayed out either) and it will 'work'.  The 'Save' button is now grayed out, but there is no image specified.

tst-awankhede:
That is happening because the Sources list has not checked in IsModified, hence even if we take out the sources, they don't generate any save error.

However, that was not the case before. Before , any error would block the path of correction and save again so this path was not even tested.

I am looking to implement it, in general enabling the "save->fail->save " as part of same request has many variations to test and go bad.

tst-eclamar:
I'm not sure that I understand your comments.  Do you plan on dealing with the bugs that were exposed by your change in different ticket?  Or are you continuing under the current ticket?

tst-awankhede:
I will try today and update if possible in next couple hours.

tst-awankhede:
I could not come up with a reasonable solution for these existing bugs which would eventually address your observations refer #321.  ( other issues in this part of code are #309 #316 )

tst-awankhede:
updated code it now solves #309 and also checks that there should be at least one src file.

tst-lsavoie:
Just to be clear, this fixes the problem where you try to save with no files and then add a file and save again.  It doesn't fix the problem where you try to save with an invalid file, then change the file and try again.  Do I have that right?  I think we can merge these changes but we're going to need to leave #167 open to make sure we handle the other cases.

tst-awankhede:
yes, fixes #309 and part of #167 , so we can keep the #167 open as discussed.

tst-eclamar:
@tst-lsavoie I haven't looked at it recently - I really should look at it
again, just to be sure.  However, I leave it up to you.

tst-lsavoie:
@tst-eclamar if you have time to look at it tomorrow, go ahead and do that.  Otherwise we'll just merge it in tomorrow.

tst-eclamar:
@tst-lsavoie I've tested it out and it seems good.  Thanks.

earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
```diff
@@ -655,7 +659,7 @@ void RasterAssetWidget::AssembleEditRequest(
                           + filename + "\n" +
                           khErrnoException::errorString(errno));
       case SourceConfig::NonVolume:
-        throw khException(filename + tr(" doesn't reside on a known volume."));
+        throw khException(filename + tr(" doesn't reside on a known volume. \n Make sure the volume is configured correctly using geconfigureassetroot and try again."));
```
tst-ppenev: I think most users want to move their files to a 'known volume', rather than create a new one.

tst-lsavoie: @tst-ppenev would you like this change made?  Other than this, the PR is ready to merge.

tst-awankhede: @tst-ppenev : Hint to use geconfigureassetroot is on purpose, it should help determine the volume ( using --listvolume) and edit or add volume if needed. Move option( as you proposed) is pretty much specific action resolution that might not work for all errors.

If this does not seem reasonable, please propose one the error message.

tst-ppenev: Something like:

> You can move your asset files to a known volume, or create a new one that contains their current location.  You can list known volumes and create new ones from the command line with `geconfigureassetroot`.

tst-ppenev: @tst-lsavoie: It seems like a quick change. Otherwise, if it is taking up more time, I don't think it's important enough to hold the ticket.

tst-awankhede: sure, I will send the change

tst-awankhede:
New error message:     screenshot from 2017-07-14 15-07-51

tst-ppenev:
@tst-awankhede: Lol. I hope that dialog box does word wrapping.

tst-awankhede:
No word wrapping there. Haha

tst-ppenev:
Maybe a newline between the two sentences can help, then. The rule of thumb in typography is 66 characters per line. This looks 'a little' above that.

tst-awankhede:
screenshot from 2017-07-17 14-16-05

changes as per screenshot above

Issue #258, opened 2017-06-19
Create release process


tst-lsavoie:
We need to determine the process to follow when creating a release. It should include at least:

- Building and testing on all supported platforms.
- Updating the OpenGEE website with the documentation for the latest release (we plan to make the documentation for several releases available, but the website should go to the docs for the latest release by default).
- Creating release notes for the new release and redirecting the link in index.html to the new release notes.

Also publish 5.2.0 documentation on opengee.org

tst-eclamar:
The website for this is available at https://tst-eclamar.github.io/earthenterprise/

tst-lsavoie:
The release process is documented here: https://github.com/google/earthenterprise/wiki/Development:-Release-Process.

PR #324, opened 2017-07-12
DO NOT MERGE: Issue #258 - add multiple versions of on-line documentation.

tst-eclamar:


tst-eclamar:
Forgot to mention: The (extremely brief) instructions on how to do this are on the TST google shared drive as 'GEE GP-226 Updating the website for new revisions'.

tst-awankhede:
It would be good to mention the instructions here as well.

docs/geedocs/5.1.3/answer/1066449.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html> <html> <head> <meta charset="UTF-8"/>
 <title>Custom folders</title>
-<link rel="stylesheet" href="../css/styles.css" type="text/css" media="all" /> <link rel="stylesheet" href="../css/containers.css" type="text/css" media="all" /> </head> <body>
+<link rel="stylesheet" href="../../css/styles.css" type="text/css" media="all" /> <link rel="stylesheet" href="../../css/containers.css" type="text/css" media="all" /> </head> <body>
```
tst-lsavoie: I'm thinking we may want to duplicate the art and css directories as well so that we can avoid all of these changes every time we do a release.

docs/geedocs/index.html
```diff
@@ -13,21 +13,8 @@
...
-    <h4><a href="answer/6153619.html">Fusion resources and projects</a></h4>
-    <h4><a href="answer/3468004.html">GEE Server administration</a></h4>
-    <h4><a href="answer/173056.html">GEE Server configuration and security</a></h4>
-    <h4><a href="answer/6006981.html">Google Earth Enterprise client</a></h4>
-    <h4><a href="answer/4580164.html">Portable</a></h4>
-    <h4><a href="answer/6014171.html">Developers</a></h4>
-    <h4><a href="answer/6121524.html">Troubleshooting and error messages</a></h4>
-    <h4><a href="answer/7159245.html">Release notes</a></h4>
+    <h3><a href="./5.2.0/index.html">5.2.0</a></h3>
+    <h3><a href="./5.1.3/index.html">5.1.3</a></h3>
```
tst-lsavoie: Some thoughts about this:

- I don't think we need 5.1.3 documents.  Starting with 5.2.0 should be enough.
- We also need a folder for the next release (the one under development).  I suppose we could make that a part of the release process - copy the existing release docs into a folder for the new release.  Maybe that's the intention.  When we do that, we should modify the link for the release under development to indicate that it's under development.
- The same docs are installed when you install fusion and server.  We need to test that it still works in that case.  It would be great if it only installs the latest version of the docs.  Perhaps updating the link at earthenterprise/earth_enterprise/landing_page/manual to point to the docs for the latest release would do that.  We'd just need to update the release instructions.
- As Asmita noted, we need to get the release instructions into the wiki at some point.
- Asmita also pointed out that it would be great to have a way to avoid duplication and only store the differences between releases.  That doesn't have to be done right now.

docs/geedocs/5.1.3/answer/1066449.html
tst-lsavoie: Never mind on the never mind.  The css directory could perhaps be shared between all the releases, but the art directory should probably be specific for the release because the images could change between releases.  I'm open to discussion on that, though.

tst-eclamar:
@tst-lsavoie:  Ok, will do.  A minor point: I merged my working branch back
into my main branch so that the documentation directories could be viewed
on-line.  Now, where do I do my edits?  Checkout a branch again, then merge
it back into main so that the documentation directories can be viewed, or
just make my edits in my main branch?

docs/geedocs/5.1.3/answer/1066449.html
tst-ppenev: Well, CSS can change, too.

tst-lsavoie:
@tst-eclamar you can just keep making changes on the same branch and then merge it into your master branch again when you're done.  So:
```
git checkout Issue-258_multipe-versions-of-documentation
# Make your changes and commit them
git checkout master
git merge Issue-258_multipe-versions-of-documentation
git push origin master
```
After this has been merged, you'll want to update your master branch to match the master branch on the main repo.  I do it by forcing my master to be the same as the upstream master, which throws away the merges to master you made on your local repo.  That's done like this:
```
git fetch upstream
git checkout master
git reset --hard upstream/master # Force your version of master to be the same as the upstream version
git push --force # Force push your changes to your personal github repo
```
`git reset` and `git push --force` are both dangerous commands that could cause you to lose work, so you should be careful about using them.  The only times I recommend using them are when you are resetting your master branch to the upstream master branch (like above) or when you've done some work and decided it's wrong and want to get rid of it.

tst-lsavoie:
Added a "Do not merge" designation to the title since this is not meant to be merged until the release is complete.

tst-eclamar:
Lee:

Sorry, but I'm not clear as to why I need to do this second set of git
commands.   Why do I want to reset my repo to be the same as the master?
Won't I lose my work?

Thanks,
Eric.

tst-lsavoie:
@tst-eclamar you don't necessarily need to do the second set of git commands.  That's just how I like to do it.  The only commits you will lose are any commits on your master branch that aren't in the upstream master branch.  That should just be the times that you merged your branch into master on your local repo, and those commits are easy to reproduce, so you're not losing anything.  When I do this, I like to wait until my branch is officially merged into the upstream master branch and then do the second set of git commands.  That replaces my merges into master with the official merges into master.

Hopefully that makes some sense.  It's hard to explain git.  The bottom line is that you can safely ignore those commands.

docs/geedocs/5.2.0/css/containers.css
```diff
@@ -0,0 +1,98 @@
+/*@charset "UTF-8";
```
tst-lsavoie: Just double checking - this is duplicated because it is also used by the main index.html?

docs/geedocs/5.2.0/index.html
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+<title>Google Earth Enterprise Documentation</title>
+<link rel="stylesheet" href="../css/styles.css" type="text/css" media="all" />
```
tst-lsavoie: Sorry, found one more problem.  This page doesn't load the CSS correctly when I install it locally.  I think you need to take out the '..' in front of the two css paths.

tst-lsavoie: Also, the Google logo doesn't load when you access it through fusion or GEE server, probably because the common directory isn't copied over.  It might be easiest to copy that logo into the art directory under 5.2.0 as well.

tst-ccamp:
@tst-lsavoie Should we redirect this PR to the release branch or do you want to start the process from scratch?

tst-lsavoie:
I'm pretty sure there have been documentation changes since this was done, so we probably need to at least copy the documents again.  It might be easier to redo it from scratch and pull in the new files from this PR.

tst-ccamp:
The process for creating the new documentation version is documented in our release process. This will need to be redone on release_5.2.0 branch after updates to the docs are complete. So, I think we can close this PR.

PR #338, opened 2017-07-13
Ignore Eclipse project files in entire repo

tst-lsavoie:
This moves the ignore statements for Eclipse project files higher in the directory structure so that they will be ignored anywhere in the repo.


tst-eclamar:
Where did the ".settings" come from?

Other than that, it looks good.

tst-lsavoie:
The .settings file is apparently only created in some scenarios.

tst-eclamar:
Ok, looks good to me.

Issue #270, opened 2017-06-26
Update documentation referring to deprecated Google Earth Plugin


tst-lsavoie:
Several pages refer to the Google Earth API and/or plugin.  While these pages are not necessarily wrong, the plugin is deprecated, and they include many dead links to documentation for the plugin.  We should decide how to handle the Google Earth API and plugin in our documentation and make updates accordingly.

The following pages are affected and may need to be updated:
- 2760566.html
- 3230746.html
- 3230777.html
- 3230807.html
- 3253752.html
- 3424202.html
- 3477797.html
- 4485225.html
- 4556704.html
- 4594620.html
- 6034724.html
- 6051534.html
- 6051568.html
- 6051584.html
- 6053170.html
- 6078762.html
- 6159443.html
- 6259403.html
- 7159245.html

tst-lsavoie:
Closed via PR #346.

PR #346, opened 2017-07-17
Update docs related to GEE plugin

tst-lsavoie:
This PR updates docs that refer to the GEE plugin.  The plugin is discontinued, is no longer publicly available, and doesn't work on most modern browsers, so the docs have been updated to reflect this.


tst-lsavoie:
Fixes #270.

tst-lsavoie:
The updated docs can be seen here: https://tst-lsavoie.github.io/earthenterprise.

Issue #389, opened 2017-07-27
Automate Docker Image Creation for All Supported OS Distributions, and Building Development Code


tst-ppenev:
It would be helpful if we can use Docker images for building and testing GEE Server and Fusion an all platforms supported by GEE.  To distribute release images, we would like to build from a clean clone of the GEE repository.  However, for testing during development, we would like to build from the local clone of the repository that includes any developer changes.

## AC:
* Support building Docker images from a clean clone of the GEE repository, as well as using a local developer clone.
* Support building Docker images running Ubuntu 14, Ubuntu 16, Cento OS 7, and RHEL 7.
* Update documentation on how to build a desired image.
* Update documentation on how to publish a release image, as well as any supporting images to Docker Cloud.

Issue #411, opened 2017-08-01
Update docker images to persist published databases


tst-lsavoie:
The docker images created for #329 do not persist published databases even if the publish root is persisted.  This is probably because the SQL database is not persisted.  We need to update the docker images and instructions to persist the database if the user desires.

tst-lsavoie:
This is low priority work, so I'm removing the issue for now.

PR #383, opened 2017-07-25
#329: Docker Image Build Scripts and Documentation.

tst-ppenev:
* Added scripts for building an Open GEE Docker image, and running GEE Server and Fusion from the container.

* Added a `README.md` describing how to install a Docker image from Docker Cloud, run GEE Server and Fusion in a container, persist globe and map databases, and publish a built image to Docker Cloud.

PR #412, opened 2017-08-01
#389: Support OS Distributions in Docker Image

tst-ppenev:
* Splits the Docker image build into building an image with a build environment for each supported OS distribution, and a derived image that is built in two stages. The first stage obtains the code to be built (either cloning from the upstream repository, or copying the local clone), and builds it. The second stage installs the built binaries, and deletes resources not needed for running GEE to free up space.

* The convenience scripts are updated to allow selection of the build environment OS distribution, first stage (i.e., repository origin), and other options (such as flattening the final image).

* A fix has been added to `earth_enterprise/docker/image-definition/scripts/create_x_user.sh` to allow accessing `/dev/dri` devices owned by groups that don't exist in the Docker container. (This happens, e.g., when running a Cent OS 7 docker container on an Ubuntu host.)

* Sanitizing the Git branch name when generating the Docker image name has been added, so any Git branch names can be used.

* Setting up the build environment on each OS distribution has been extracted into a separate standalone script `get-gee-build-environment.sh`, which can be used to obtain all build prerequisites, and clone the mainline Open GEE GitHub repository on any of the platforms.

* Added a `TST-RUN-STAGE` instruction to Dockerfile templates interpreted by the shell script builder.  That command allows supplying command-line arguments to `docker run` command, so a build second stage can be run with mounting volumes, exposing ports, or anything else required for the build.

* Add an environment variable server running over a named pipe, and a client with a file lock.  The transport is used to send secret values to the build script in the Docker container without reaching logs, or the hard disk.


tst-ccamp:
"Branched from the preliminary implementation for #383. It needs to be rebased with fixes from #383 merged in." Does this still need to be done?

tst-ppenev:
@tst-ccamp: I think I already merged `master` in, so that shouldn't be necessary. I'll remove that point.

tst-ccamp:
@tst-ppenev are you planning on uploading any of the new images to the Docker Store?

tst-ppenev:
@tst-ccamp: Yeah, when this gets approved, I can use it to build 5.2.0.  Should we publish Docker images for all supported platforms?

Did we add publishing a Docker image to the release steps?

earth_enterprise/docker/README.md
```diff
@@ -169,7 +181,62 @@ image:
...
+OS_DISTRIBUTION=ubuntu-16 STAGE_1_NAME=current-clone ./bin/build-docker-image.sh
+```
+
+
+Build a clean clone of branch `release_5.2.0` on Cent OS 7:
+
+```BASH
+OS_DISTRIBUTION=centos-7 CLEAN_CLONE_BRANCH=release_5.2.0 ./bin/build-docker-image.sh
+```
+
```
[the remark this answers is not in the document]
tst-ppenev: No, they're for being able to tell apart the examples from each other easier when editing the source, rather than when looking at the rendered document.

earth_enterprise/docker/README.md
```diff
@@ -169,7 +181,62 @@ image:
...
+### Building a Build-environment Image
+
+The Dockerfiles used by the `build-docker-image.sh` script require Docker
+images with build environments for each platform you are trying to build on.
+
+Normally, a build environment image for each supported platform should be
+published to Docker cloud, so one would be downloaded when you try to build,
+if you don't have it locally.  However, if you want to build a new image to
+publish to Docker cloud, or want to use a custom build environment, you can
+have a look at the `build-build-env-image.sh` script.  It's used as:
```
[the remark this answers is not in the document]
tst-ppenev: Well, it builds an image with a build environment.

The script is not going to customize the build environment.  You'll have to modify the image afterward, or build one by hand.

I doubt anyone would use it as a development environment.  It doesn't contain a GEE clone anyway.

earth_enterprise/docker/bin/build-docker-image.sh
```diff
@@ -1,33 +1,57 @@
...
-        "$STAGE_COMPLETE_MESSAGE")
-            echo "$SELF_NAME: Flattening Docker image."
-            docker export "$STAGE_1_CONTAINER_NAME" | docker import - "$STAGE_2_IMAGE_NAME"
-            STAGE_2_CONTAINER_ID=$(docker run -d "$STAGE_2_IMAGE_NAME" /root/opengee/bin/spin.sh)
-            docker commit --change="CMD /root/opengee/bin/start-servers.sh" "$STAGE_2_CONTAINER_ID" "$OUTPUT_IMAGE_NAME"
-            docker stop "$STAGE_2_CONTAINER_ID"
+function is_string_false()
+{
+    case $(echo "$1" | tr '[:upper:]' '[:lower:]') in
+        false|no|0)
```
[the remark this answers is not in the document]
tst-ppenev: It's more of a default to `True`, unless the value is explicitly `False` function.

It could treat an empty string as `False`, though the way it is used in this script, the tested value couldn't be empty, since it would default to `true` if it was empty.

earth_enterprise/docker/README.md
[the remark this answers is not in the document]
tst-ccamp: I'm open to other suggestions, but still not a fan of build-build-env-image

earth_enterprise/docker/image-definition/Dockerfile.build-env.rhel-6.template
```diff
@@ -0,0 +1,28 @@
...
+RUN mkdir -p /root/opengee/build
+
+
+# Install the build environment.
+#   We register the Docker image with the Red Hat repository provider (you
+# need to have a username and password) so we can install Yum packages.  We
+# unregister when we finish, so we're not taking up a machine license.
+WORKDIR /root/opengee/build
+RUN subscription-manager register \
+        --username=${REDHAT_USERNAME_ESCAPED} --password=${REDHAT_PASSWORD_ESCAPED} \
```
tst-ccamp: If there's any way possible, we should really try to avoid printing out the password for this line

earth_enterprise/docker/image-definition/Dockerfile.build-env.rhel-7.template
```diff
@@ -0,0 +1,26 @@
...
+# Make a build directory:
+RUN mkdir -p /root/opengee/build
+
+# Install the build environment.
+#   We register the Docker image with the Red Hat repository provider (you
+# need to have a username and password) so we can install Yum packages.  We
+# unregister when we finish, so we're not taking up a machine license.
+WORKDIR /root/opengee/build
+RUN subscription-manager register \
+        --username=${REDHAT_USERNAME_ESCAPED} --password=${REDHAT_PASSWORD_ESCAPED} \
```
tst-ccamp: See if printing out the password in plain text can be avoided

tst-ccamp:
When trying to build RHEL 6 build env I got this error:

Initialized empty Git repository in /tmp/opengee-pkgbuild.2bQ56X/gtest-devtoolset-rpm/.git/
./bin/build.sh: line 11: type: docker: not found
--2017-09-27 19:27:47--  https://github.com/google/googletest/archive/release-1.8.0.tar.gz
Resolving github.com... 192.30.253.113, 192.30.253.112
Connecting to github.com|192.30.253.113|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://codeload.github.com/google/googletest/tar.gz/release-1.8.0 [following]
--2017-09-27 19:27:47--  https://codeload.github.com/google/googletest/tar.gz/release-1.8.0
Resolving codeload.github.com... 192.30.253.121, 192.30.253.120
Connecting to codeload.github.com|192.30.253.121|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1281617 (1.2M) [application/x-gzip]
Saving to: `release-1.8.0.tar.gz'

     0K .......... .......... .......... .......... ..........  3%  435K 3s
    50K .......... .......... .......... .......... ..........  7%  897K 2s
   100K .......... .......... .......... .......... .......... 11% 26.6M 1s
   150K .......... .......... .......... .......... .......... 15%  914K 1s
   200K .......... .......... .......... .......... .......... 19% 22.4M 1s
   250K .......... .......... .......... .......... .......... 23% 33.3M 1s
   300K .......... .......... .......... .......... .......... 27% 19.3M 1s
   350K .......... .......... .......... .......... .......... 31%  885K 1s
   400K .......... .......... .......... .......... .......... 35% 6.41M 1s
   450K .......... .......... .......... .......... .......... 39% 43.8M 0s
   500K .......... .......... .......... .......... .......... 43% 38.5M 0s
   550K .......... .......... .......... .......... .......... 47% 13.9M 0s
   600K .......... .......... .......... .......... .......... 51% 13.9M 0s
   650K .......... .......... .......... .......... .......... 55% 14.5M 0s
   700K .......... .......... .......... .......... .......... 59% 2.46M 0s
   750K .......... .......... .......... .......... .......... 63% 2.75M 0s
   800K .......... .......... .......... .......... .......... 67% 4.00M 0s
   850K .......... .......... .......... .......... .......... 71% 3.38M 0s
   900K .......... .......... .......... .......... .......... 75% 28.1M 0s
   950K .......... .......... .......... .......... .......... 79% 31.6M 0s
  1000K .......... .......... .......... .......... .......... 83% 17.6M 0s
  1050K .......... .......... .......... .......... .......... 87% 14.8M 0s
  1100K .......... .......... .......... .......... .......... 91% 14.4M 0s
  1150K .......... .......... .......... .......... .......... 95% 13.6M 0s
  1200K .......... .......... .......... .......... .......... 99% 13.9M 0s
  1250K .                                                     100% 76.1M=0.4s

2017-09-27 19:27:48 (3.10 MB/s) - `release-1.8.0.tar.gz' saved [1281617/1281617]

error: Failed build dependencies:
	tar is needed by gtest-devtoolset2-1.8.0-1.x86_64
The command '/bin/sh -c subscription-manager register         --username=ccamp@t-sciences.com --password=anpw4C4R.C         --auto-attach &&     /root/opengee/bin/get-gee-build-environment.sh -y --clone-repository=no &&     subscription-manager unregister' returned a non-zero code: 1

earth_enterprise/docker/image-definition/Dockerfile.build-env.rhel-7.template
```diff
@@ -0,0 +1,26 @@
+# We only need a few context files, rather than the whole repository:
+# --[ TST-Dockerfile: context-dir=earth_enterprise/docker/image-definition ]--
+# Prompt the user to enter variables required to build this Dockerfile:
+# --[ TST-Dockerfile: required-vars=REDHAT_USERNAME ]--
+# Avoid displaying the values of some variables:
+# --[ TST-Dockerfile: required-password-vars=REDHAT_PASSWORD ]--
+
+FROM registry.access.redhat.com/rhel:7.3
```
tst-ccamp: 7.4 is out now. Maybe we should consider just using :latest

tst-ccamp: Or at least updating to 7.4

tst-ccamp:
Every time I finish building an image I also get a thermopylae version of it, which seems to be smaller.

```
opengee-experimental-centos-7                         latest                      53c0c1d03a2a        5 hours ago          2.44GB
thermopylae/opengee-experimental-build-env-centos-7   latest                      7edf7f2f7700        5 hours ago          1.19GB
```

So, a couple things. Is this behavior documented in the READMEs? Does it make sense to have it specifically say thermopylae in the name? I had thought that meant that it was coming from the thermopylae docker store.

earth_enterprise/docker/README.md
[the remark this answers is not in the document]
tst-ppenev: `make-build-env-image.sh`?

The other script could be renamed to `make-docker-image.sh`.

earth_enterprise/docker/README.md
```diff
@@ -169,7 +181,62 @@ image:
...
+
+
+Build a clean clone of branch `release_5.2.0` from user `unameit`'s GitHub
+repository on Cent OS 7:
+
+```BASH
+OS_DISTRIBUTION=centos-7 CLEAN_CLONE_URL=https://github.com/unameit/earthenterprise.git CLEAN_CLONE_BRANCH=release_5.2.0 ./bin/build-docker-image.sh
+```
+
+### Building a Build-environment Image
```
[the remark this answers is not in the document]
tst-ppenev: Yeah, I was thinking that, normally, we'd have the build environment images in Docker Cloud, so that you wouldn't need to build them, unless you want to modify the build environment. (Assuming, we don't have to change it often.)  So, I put this section afterwards.

earth_enterprise/docker/image-definition/Dockerfile.build-env.rhel-6.template
tst-ppenev: Yes.  What kinds of scenarios should we protect against?

I was thinking of using image variables, at first, but the Docker that comes with Ubuntu 14 is too old, and doesn't support them, so I'm expanding variables in Dockerfile templates instead.  The variables a container is built with a recorded, so, if we did that, we'd have to delete them from the final container, and delete all intermediate container layers, and terminated images.

Currently, the expanded template is stored in a temporary Docker file that is deleted after the build.  Also, a docker image may have the Dockerfile it was built from left behind.  (I need to check for this on RHEL.)

Here are some options:

1. Put the password in a file inside the container.  Delete the file after the password is consumed.

This may be a simple way to avoid having the password in the Docker command logs (which appear on the screen).  There are a lot of scenarios that doesn't protect a user from, but, at least, it won't be shown on the screen.

2. Put the password in an encrypted file in the container.  Delete the file after the password is consumed.

2.1. We could use a temporary 1-time encryption cipher that is automatically generated for each build.

This way we'd passing around the cipher, which wouldn't be useful after the password is consumed and the password-containing file deleted.  The advantage is that we won't be storing the password on disk, so, even if the deleted disk blocks become exposed later, the original password would still not be exposed.  (This may be a threat scenario that requires an attacker stealing someone's disk and running an analysis on it.)

2.2. If the user has multiple passwords in a vault. We could copy the vault, and pass the vault password instead of the RHEL subscription password.

That doesn't improve security from the approach above, but might more convenient in some cases.

3. We could obtain passwords in the container over a(n encrypted) network connection to the host OS.

Passwords would not appear in command logs, and won't be stored on disk.

A tool like [Vault](https://github.com/hashicorp/vault) can encrypt passwords in a file, or provide them over a network connection.

In all of these cases the password would pass in clear text through memory.

tst-ppenev:
@tst-ccamp: It looks like `tar` wasn't installed when you were trying to build on RHEL 6. I'll try to reproduce that.

earth_enterprise/docker/image-definition/Dockerfile.build-env.rhel-7.template
tst-ppenev: I could update it.  The problem with using `latest` is that it'll become RHEL 8 at some time.

tst-ppenev:
@tst-ccamp: I could add something to the `README`.  I was thinking that we'd want the build environment images in the Docker store, so users don't have to build them, unless they need a modified environment.

tst-ppenev:
The RHEL 6 build should be fixed now.

earth_enterprise/docker/README.md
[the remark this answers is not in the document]
tst-ccamp: Yeah, you could say that it's optional if it's available on the Docker Store at the beginning of the section.

earth_enterprise/docker/image-definition/Dockerfile.build-env.rhel-6.template
tst-ccamp: Sure, it may not be the most secure ever, but it's gotta be better than displaying the password in plain text on the screen :) I was very surprised to see that.

earth_enterprise/docker/README.md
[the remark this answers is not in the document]
tst-ppenev: :P That's what I did.

earth_enterprise/docker/image-definition/Dockerfile.build-env.rhel-6.template
tst-ppenev: It seems that using a named pipe works fine.  That way it doesn't reach the disk.  I just need to finish the scripting.

earth_enterprise/docker/README.md
```diff
@@ -156,10 +168,15 @@ If you want to build the Docker image yourself (perhaps, after customization),
 you can use the commands that follow.
```
tst-ccamp: I think this is still confusing. We should put the instructions for Build Build Environment first, then Build Image, then Run Image

tst-ccamp:
When I try actually building a resource in fusion, I always get an error trying to build it. The state for the subprocess changes to Failed with the Resource marked as Blocked. I couldn't find anything relevant in the logs.

This is the command I used to run the container and map the tutorial directory:
```
DOCKER_RUN_FLAGS="-v /opt/google/share/tutorials/fusion:/opt/google/share/tutorials/fusion" OS_DISTRIBUTION=ubuntu-16 ./bin/start-docker-image.sh
```

This is the error that I see in the build output:
```
Completed  78% (137/175) - tiles/sec: 9.44 - time left: 00:00:04 [CPU: 97 Mem: 395 MB PF: 0]
Completed  83% (146/175) - tiles/sec: 9.25 - time left: 00:00:03 [CPU: 97 Mem: 395 MB PF: 0]
Completed  89% (155/175) - tiles/sec: 9.73 - time left: 00:00:02 [CPU: 102 Mem: 395 MB PF: 0]
Completed  94% (164/175) - tiles/sec: 9.55 - time left: 00:00:01 [CPU: 102 Mem: 395 MB PF: 0]
Completed  99% (173/175) - tiles/sec: 10.08 - time left: 00:00:00 [CPU: 100 Mem: 395 MB PF: 0]
Completed 100% (175/175) - tiles/sec: 10.28 - time left: 00:00:00 [CPU: 100 Mem: 372 MB PF: 0]
Processed 175 tiles
Total time to process: 00:00:18
Average tiles per second: 9.60
Output File Sizes
  48.7 MB : /gevol/assets/Test.kiasset/product.kia/ver001/raster.kip
  48.7 MB : TOTAL Output File Sizes
---------- End Command Output ----------
ENDTIME: 2017-10-07 00:22:00
ELAPSEDTIME: 19 seconds
FAILED
```

earth_enterprise/docker/bin/build-docker-image.sh
```diff
@@ -1,33 +1,46 @@
 #! /bin/bash
 
-# Set these variables before running this script to override the
-# default values below:
-: ${STAGE_1_IMAGE_NAME:="opengee-experimental-stage-1"}
-: ${STAGE_2_IMAGE_NAME:="opengee-experimental-stage-2"}
-: ${OUTPUT_IMAGE_NAME:="opengee-experimental"}
+# This script builds a Docker image containing Open GEE Server and Fusion.
+
+source "$(dirname "$0")/lib/input-variables.sh"
```
tst-eclamar: Doesn't this double quoting, in fact, leave $0 unquoted?

tst-ppenev: No, it doesn't.  You could write a short script to test it for yourself, or dig through documentation on command substitution.

This specific expression would fail to produce the correct output, if the current directory name ends in one or more newline characters.  I've considered that an acceptable limitation.

tst-ppenev:
@tst-ccamp: There's a ticket to fix persisting published databases outside of the Docker container running GEE Server and Fusion: https://github.com/google/earthenterprise/issues/411.

The current instructions in the README.md for that are half-baked.  We could consider what storage options we may want to support (e.g., from https://docs.docker.com/engine/admin/volumes/).

I am able to build, publish and open a minimalist globe in the Docker container without using persistent storage.

I could add more fixes to this PR, or do that on the new ticket.

earth_enterprise/docker/README.md
```diff
@@ -29,166 +29,217 @@ service docker start
...
+The build environment image is reused every time you rebuild Open GEE.  This
+should save you time during normal development for targetting another
+platform, since only the steps of transferring the modified source code, and
+running the Open GEE build would need to be repeated.
+
+You can build a build-environment image with a command like this:
 
 ```BASH
-HOST_GEVOL_PATH=/persistent/gevol/ ./bin/start-docker-image.sh
+( . config/rhel-7.sh && ./bin/build-build-env-image.sh )
```
tst-ccamp: Minor: I believe I have it figured out, but can you add a brief explanation on what the config/* files do for you?

tst-ppenev: They allow to avoid indecently long command lines.

There's a subsection called _Image Configuration_ above that tries to explain that.

earth_enterprise/docker/config/rhel-7.sh
```diff
@@ -0,0 +1,24 @@
...
+# Use this file in a command like:
+#
+# ```BASH
+# ( . config/rhel7.sh && ./bin/build-gee-image.sh )
+# ```
+
+export OS_DISTRIBUTION=rhel-7
+export STAGE_1_NAME=clean-clone
+#export REDHAT_USERNAME=put-your-username-here
+#export REDHAT_PASSWORD="your password here"
```
tst-ccamp: Even if I set my password here, it still prints it on the screen:

```
ccamp@Mainframe:/opt/dev/tst/earthenterprise/earth_enterprise/docker$ ( . config/rhel-7.sh && ./bin/build-build-env-image.sh )
REDHAT_PASSWORD: xxxxxxxx
Creating named pipe: /opt/dev/tst/earthenterprise/earth_enterprise/docker/image-definition/build/tmp.eC91n/secrets/socket
Env server listening on: /opt/dev/tst/earthenterprise/earth_enterprise/docker/image-definition/build/tmp.eC91n/secrets/socket
===================
```

tst-ppenev: :stuck_out_tongue_winking_eye: Good catch!  It's not good to leave debugging code behind.

tst-ccamp:
Wrote #552 to document the resource build failures

tst-ccamp:
Wrote #555 to address the second bug I've found

PR #417, opened 2017-08-01
I 386 create run historical script

tst-eclamar:
Create a 'run_historical.sh' script to test out Historical Imagery functionality.


tst-ccamp:
I updated the previous PR to point to master. This stuff will eventually get back into master when we merge release_5.2.0 back into it. So, this PR is unnecessary

PR #418, opened 2017-08-01
Blog: Previous Entries Support ( DO NOT MERGE )

tst-ahernandez:
This change will add a "Previous Posts" section to the Open GEE Blog. Each post's title will be displayed with a link to that post. The UI of the blog should remain the same when viewing each post.

NOTE:

Before merging this PR we should remove the Old Post so it no longer appears. So please notify me when we are ready to merge.

To test this change: https://tst-ahernandez.github.io/earthenterprise/blog.html


docs/_layouts/blogpost.html
```diff
@@ -0,0 +1,73 @@
...
+
+      </div>
+      <!-- Only display the Previous Posts section if there are >1 posts -->
+      {% if num_posts > 1 %}
+        <!-- The previous posts navigator sits in the right 25% -->
+        <div class='large-3 columns large-uncentered opposite darkBorder'>
+          <div>
+            <h3 class='regularText'> Previous Posts </h3>
+            <ul>
+              {% for post in site.posts %}
```
tst-lsavoie: Is this guaranteed to list the posts in reverse chronological order?

docs/css/style.css
```diff
@@ -16,6 +16,13 @@ h1, h2, h3, h4 {
   letter-spacing: 2px;
 }
 
+.darkBorder {
```
tst-lsavoie: This is a matter of opinion, but I'm not a huge fan of the black border around the list of previous posts.  It might look better with no border or perhaps a slightly different color background.  I can live with it as is if everyone else prefers it that way.

docs/_layouts/blogpost.html
tst-ahernandez: Yes

docs/_layouts/blogpost.html
```diff
@@ -0,0 +1,73 @@
...
+          {{ content }}
+        </div>
+
+      </div>
+      <!-- Only display the Previous Posts section if there are >1 posts -->
+      {% if num_posts > 1 %}
+        <!-- The previous posts navigator sits in the right 25% -->
+        <div class='large-3 columns large-uncentered opposite darkBorder'>
+          <div>
+            <h3 class='regularText'> Previous Posts </h3>
```
tst-ccamp: These aren't all going to be Previous Posts. So we should come up with a better name, but I don't really know what would be a good name. "All Posts"? "More Posts"?

docs/_layouts/blogpost.html
```diff
@@ -0,0 +1,73 @@
...
+            <ul>
+              {% for post in site.posts %}
+                <li>
+                  <!-- If the current post is being displayed, do not link it -->
+                  {% if post.title == page.title %}
+                    {{post.title}}
+                  {% else %}
+                    <a href="{{site.url}}{{site.baseurl}}{{ post.url }}">{{ post.title }}</a>
```
tst-ccamp: Minor: inconsistent spacing

docs/_layouts/blogpost.html
```diff
@@ -0,0 +1,73 @@
...
+        <!-- This is the actual content of the post -->
+        <div class="content">
+          {{ content }}
+        </div>
+
+      </div>
+      <!-- Only display the Previous Posts section if there are >1 posts -->
+      {% if num_posts > 1 %}
+        <!-- The previous posts navigator sits in the right 25% -->
+        <div class='large-3 columns large-uncentered opposite darkBorder'>
```
tst-ccamp: If you make the window a bit more narrow (you don't have to go very small) it will move the posts box to the bottom of the page. I'm not sure this is what we wan't. Can you see if you can either make it always on the page or make the threshold a bit bigger before it basically disappears.

docs/css/style.css
tst-ccamp: For what it's worth, I like the border the way it is

tst-ahernandez:
Updated some of the minor comments. The css styling one is fighting me. I kind of like that it goes to the bottom but then the border looks funny. I'll keep poking.

tst-ahernandez:
Updated for smaller resolutions.

PR #416, opened 2017-08-01
I-386 create run historical script

tst-eclamar:
Create a 'run_historical.sh' script to test out Historical Imagery functionality.

PR #425, opened 2017-08-03
I 386 create run historical script

tst-eclamar:
I386 - split off historical scripts, update README


tst-ccamp:
This is a duplicate of #416. Closing in favor of that one.

PR #446, opened 2017-08-14
I413 correct cache clearing documentation

tst-eclamar:
Issue 413
Correct documentation that says where to clear the EC cache.


tst-eclamar:
This PR was closed due to an error on which initial base to use.

PR #447, opened 2017-08-14
I413 Correct cache clearing documentation

tst-eclamar:
Issue 413/GP-287
Correct cache clearing documentation


tst-eclamar:
What problem/issue does deleting the cache solve that clearing the cache doesn't?  That is, was there a specific reason that deleting the cache was recommended over clearing?  It would make more sense to clear than delete..

tst-lsavoie:
@tst-eclamar I'm not sure if there's even a difference between clearing the cache and deleting the cache.  It seems they would solve the same problems.  I think it's fine to leave it as is since that's the way it was before and it does get the job done.

PR #449, opened 2017-08-15
 #414: Reduced the use of 'search tab' in answer document.

### LINKED ISSUE #414: Errors in tutorial on search tabs
The tutorial page on search tabs (4412684.html) seems to be wrong because it has you create a search tab, but then it publishes the database with POI search (not a search tab).  This should be cleaned up.

tst-ppenev:
* Reduced the use of the term 'search tab' in the answer document on Point of Interest search.

Fixes #414.


tst-lsavoie:
@tst-ppenev you may have to personally sign the CLA.  It seems the company one is no longer working.

Issue #436, opened 2017-08-14
Seg fault in fusion when opening vector resource created in old version


tst-lsavoie:
If you create a vector resource in 5.1.3, then uninstall 5.1.3 and install 5.2.0, then try to open the vector resource in fusion 5.2.0, fusion seg faults.

tst-lsavoie:
Closed via PR #451.

PR #451, opened 2017-08-15
Eliminate NULL pointer error in date/time function

tst-lsavoie:
If a QDateWrapper object is initialized without a time (just a date), the hours_edit_, minutes_edit_ and seconds_edit_ objects will be NULL.  I've fixed this by ignoring the hours, minutes, and seconds when the object was initialized, which matches what is done in other places in the file, but I'm interested in other's opinions about whether this is the right fix.  There's also the option of creating the hours_edit_, minutes_edit_ and seconds_edit_ when they are set, but I'm not sure that's the desired behavior.

As a side note, this error actually happens whenever you open a vector resource created by run_tutorial.sh in 5.1.3 and above.  It doesn't just happen when you open a resource created in 5.1.3 in 5.2.0.

### DIFF
--- a/earth_enterprise/src/fusion/fusionui/QDateWrapper.cpp
+++ b/earth_enterprise/src/fusion/fusionui/QDateWrapper.cpp
@@ -27,6 +27,10 @@ static const uint32 kDateStringBufferLength = kDateStringLength + 1;
 static const uint32 kDateTimeStringLength = 20;
 static const uint32 kDateTimeStringBufferLength = kDateTimeStringLength + 1;
 
+bool QDateWrapper::TimeIncluded() const {
+  return hours_edit_ && minutes_edit_ && seconds_edit_;
+}
+
 void QDateWrapper::SetDateValidators(QObject* parent) {
   QIntValidator* year_validator = new QIntValidator(0, 9999, parent);
   QIntValidator* month_validator = new QIntValidator(0, 12, parent);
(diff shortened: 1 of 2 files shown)


tst-eclamar:
I have reviewed this and it looks good.  Eric LaMar.

tst-lsavoie:
@nfarah hours_edit_, minutes_edit_, and seconds_edit_ are always set and modified together.  The only time that minutes_edit_ could be NULL if hours_edit_ was not is if the user passes a valid pointer for hours to the constructor but passes NULL for minutes.  Not technically impossible, but it would be a user error.

I don't think it makes sense to store hours without also storing minutes and seconds.  Perhaps the best fix would be to add a check to the constructor and throw and exception or fail an assertion if any of the three is NULL.  What are your thoughts?

tst-lsavoie:
@tst-nfarah @tst-mswartz I made the suggested changes to check minutes_edit_ and seconds_edit_ as well.  There's still a potential problem in that the user could pass a null pointer for, say, month_edit_ which would still cause a seg fault.  However, if that happens the object has been constructed improperly, so fixing it is going to require more surgery than I'm comfortable performing right before a release.  I suggest we write a ticket to fix that in the future.

tst-nfarah:
It looks good. Agreed, let's write a new ticket to cover other possible case(s). Its bigger than the scope of this specific 'seg fault' ticket.

tst-lsavoie:
I wrote #460 to cover the other possible seg faults in these files.

tst-lsavoie:
The run_tutorial.sh script sets the date for the vector projects to May 10, 2003.  I've verified that the date can be changed and the change is persisted, so everything is working as expected.

tst-lsavoie:
Fixes #436.

PR #454, opened 2017-08-15
#435:   change build_lib compile flags to include -std=gnu++98

### LINKED ISSUE #435: Specify C++ version for portable build
MinGW in Windows now defaults to installing g++ 6.3.0. We discovered that starting with 6.0 the default standard is set to gnu++14 rather than gnu++98. We need to update the build_lib scripts to include the flag "-std=gnu++98". We should probably go ahead and set this for linux and mac as well.

tst-mswartz:
Fixes #435 

Portable wasn't compiling with newer g++ versions , so addition of flag -std=gnu++98 fixed compiler compatibility issues

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/Linux/build_lib
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/Linux/build_lib
@@ -1,5 +1,5 @@
 cp ../shared/* .
 swig -python -c++ glc_unpacker.i
-g++ -O2 -fPIC -I/usr/include/python2.7 -c glc_unpacker_wrap.cxx
-g++ -O2 -c -fPIC glc_unpacker.cpp portable_glc_reader.cpp file_unpacker.cpp file_package.cpp packetbundle.cpp packetbundle_finder.cpp
-g++ -shared glc_unpacker_wrap.o glc_unpacker.o portable_glc_reader.o file_unpacker.o file_package.o packetbundle.o packetbundle_finder.o -o _glc_unpacker.so
+g++ -std=gnu++98 -O2 -fPIC -I/usr/include/python2.7 -c glc_unpacker_wrap.cxx
+g++ -std=gnu++98 -O2 -c -fPIC glc_unpacker.cpp portable_glc_reader.cpp file_unpacker.cpp file_package.cpp packetbundle.cpp packetbundle_finder.cpp
+g++ -std=gnu++98 -shared glc_unpacker_wrap.o glc_unpacker.o portable_glc_reader.o file_unpacker.o file_package.o packetbundle.o packetbundle_finder.o -o _glc_unpacker.so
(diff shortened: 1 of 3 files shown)


tst-mswartz:
I signed it!

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/Linux/build_lib
```diff
@@ -1,5 +1,5 @@
 cp ../shared/* .
 swig -python -c++ glc_unpacker.i
-g++ -O2 -fPIC -I/usr/include/python2.7 -c glc_unpacker_wrap.cxx
-g++ -O2 -c -fPIC glc_unpacker.cpp portable_glc_reader.cpp file_unpacker.cpp file_package.cpp packetbundle.cpp packetbundle_finder.cpp
-g++ -shared glc_unpacker_wrap.o glc_unpacker.o portable_glc_reader.o file_unpacker.o file_package.o packetbundle.o packetbundle_finder.o -o _glc_unpacker.so
+g++ -std=gnu++98 -O2 -fPIC -I/usr/include/python2.7 -c glc_unpacker_wrap.cxx
```
tst-ppenev: I think we should consider using `-std=c++11`.  It'll allow us to use more convenient language constructs, and is supported by a good number of compilers.

tst-lsavoie: @tst-mswartz correct me if I'm wrong, but I think compiling failed with --std=c++11.  We can discuss what standard we ultimately want to use, but I think that should go in a different ticket.

tst-ppenev: It compiles fine with C++11 for me.

tst-mswartz: If I remember correctly, the problem was that newer versions of g++ have a default standard of C++14, not C++11

tst-ccamp: Exactly! https://gcc.gnu.org/gcc-6/changes.html

They just skipped right over C++11 as a default, which seems pretty strange to me, but what are you gonna do?

tst-ccamp: I think if anyone wants to use a C++11 construct in an update, they are welcome to go ahead and update the compile flag at that time.

tst-ccamp:
I see two approvals, so I'm going to merge

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/Linux/build_lib
tst-ppenev: OK. I've approved this for now.

I've seen some code that looked a bit painful, since it wasn't make use of things like initializer lists. I think, as long as we don't get stuck on an old language standard, and make it known what we're supporting, we would save ourselves some pain.

tst-lsavoie:
FYI, I went ahead and filed #467 to consider switching to C++ 11.

PR #457, opened 2017-08-16
I382 - Update documentation to reflect 5.2.0

tst-eclamar:
Update the online documentation to reflect changes to 5.2.0


docs/geedocs/answer/4412441.html
```diff
@@ -186,6 +189,32 @@
...
+<li>Click the first folder icon on the right-hand side for Vector.<br/>
+<br/>
+<img alt="Folder Icon" src="../art/fusion/tutorial/iconProjBrowser.png"/><br/>
+<br/>
+The <strong>Select Folder</strong> dialog appears.<br/>
+<br/>
+<img alt="Select Folder Dialog" src="../art/fusion/tutorial/assetFolderSelection.png"/>
+</li>
+<li>
+Select <strong>Vector</strong>, then click <strong>OK</strong>.
```
tst-lsavoie: I think we should add a step before this step to navigate to /opt/google/share/tutorials/fusion in case the user doesn't start fusion in the tutorial directory.

docs/geedocs/answer/4412446.html
```diff
@@ -63,7 +63,7 @@
 <p><img alt="Push Database Dialog" id="publishDatabase" name="publishDatabase" src="../art/fusion/tutorial/dbasePush-full.png"/></p>
 </li>
 <li>Click <b>Push</b>. Google Earth Enterprise Fusion pushes the database to Google Earth Enterprise Server, and displays a success message when it is done.
-<p class="note"><strong>Note:</strong> If you get an error message, contact your Google Earth Enterprise Server administrator for help, or check the <i>Google Earth Enterprise Administration Guide</i> for more information.
+<p class="note"><strong>Note:</strong> If you get an error message, contact your Google Earth Enterprise Server administrator for help, or check the <i><a href="../answer/3481464.html">Google Earth Enterprise Administration Guide</a></i> for more information.
```
tst-lsavoie: I think it would be good to change the text of the link here since I suspect the "Fusion Administration Guide" is a specific document that is no longer available.  Can we change the link text to something like "documentation on Fusion administraction"?

tst-lsavoie:
There are a couple of things in the ticket that I didn't see:

- 4412446 - at the bottom where it talks about viewing the database in the enterprise client it would be good to have it say something like "Google Earth EC can be downloaded from ..." and provide a link to the relevant wiki page.
- 4412433 - Line 488 says "CA _POIs" with a space between "A" and "_".  We should take the space out.

PR #459, opened 2017-08-16
I362 fix download tutorial execute permission

tst-eclamar:
The execute permission for download_tutorial.sh is set in install_fusion.sh, but install_server.sh clears it again.  So, it is necessary that install_server.sh also sets the execute permission.


earth_enterprise/src/installer/install_server.sh
```diff
@@ -545,6 +545,8 @@ fix_postinstall_filepermissions()
   chmod ugo+x /opt/google/share/support/geecheck/convert_to_kml.pl
   chmod ugo+x /opt/google/share/support/geecheck/find_terrain_pixel.pl
   chmod ugo+x /opt/google/share/support/geecheck/pg_check.pl
+  # Note: this is installed in install_fusion.sh, but needs setting here too.
+  chmod ugo+x $BASEINSTALLDIR_OPT/share/tutorials/fusion/download_tutorial.sh
```
tst-lsavoie: This looks good.  Just for my own edification I did some searching and found that the problem is on line 541 where executable permissions are turned off for everything under /opt/google/share.  This is a good solution.

PR #461, opened 2017-08-16
I362 run download_tutorial.sh requiring root

tst-eclamar:
download_tutorials.sh needs to run as root, so add a check that the current executing context is root.

### DIFF
--- a/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
+++ b/earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
@@ -2,9 +2,10 @@
 
 TMPFILE="/tmp/FusionTutorial-Full.tar.gz"
 SCRIPTDIR=`dirname "$0"`
+SELFNAME=`basename "$0"`
 
 if [ $(id -u) != 0 ] ; then
-    echo "`basename $0`" "must be run as root"
+    echo "$SELFNAME must be run as root"
     exit 1;
 fi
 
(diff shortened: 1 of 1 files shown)


earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -3,6 +3,11 @@
 TMPFILE="/tmp/FusionTutorial-Full.tar.gz"
 SCRIPTDIR=`dirname "$0"`
 
+if [ "$EUID" != "0" ] ; then
```
tst-ppenev: Minor: I think this only works in BASH, and not in other POSIX shell implementations, like `dash` on Ubuntu. You could use `id -u` to be more portable.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -3,6 +3,11 @@
...
+    echo `basename $0` "must be run as root"
```
tst-ppenev: I think you should put everything in quotes here. We're expecting the script name not to have any special characters, but we can be sure it works by quoting.

earth_enterprise/tutorial/FusionTutorial/download_tutorial.sh
```diff
@@ -3,6 +3,11 @@
 TMPFILE="/tmp/FusionTutorial-Full.tar.gz"
 SCRIPTDIR=`dirname "$0"`
 
+if [ $(id -u) != 0 ] ; then
+    echo "`basename $0`" "must be run as root"
```
tst-ppenev: Sorry, I should've given an example. `basename $0` won't work for a script whose name contains any special character.  You can, e.g.:

```BASH
echo "$(basename "$0") must be run as root"
```

or, I usually define `SELF_NAME` near the top of the file, and the use that variable:

```BASH
SELF_NAME=$(basename "$0")

# . . .

echo "$SELF_NAME must be run as root"
```

PR #462, opened 2017-08-16
I382 update reference to nationalatlas.gov to similar on nationalmap.gov

tst-eclamar:
Update reference from nationalatlas.gov to nationalmap.gov


docs/geedocs/answer/4412421.html
```diff
@@ -252,7 +252,7 @@
...
     <p>In this exercise, you create and modify a number of display rules for the CAHighways layer, so Google Earth EC displays the road information at the desired display levels with appropriate labels and coloring to distinguish between major freeways and other roads. </p>
     <p>A key part of knowing how to configure display rules for vector data requires familiarity with the data fields in the source data. For the CAHighways layer of your vector project in this tutorial, you can refer to the description of the data fields at this web site: </p>
-    <code>http://www.nationalatlas.gov/mld/roadtrl.html </code>
+    <code>https://nationalmap.gov/small_scale/mld/roadtrl.html</code>
```
tst-lsavoie: This is more of a question than a comment.  The linked website doesn't describe the data fields directly, though you can get that information if you download the file and look at the .txt file inside the archive.  I think the easiest way to fix it would be to make the sentence more generic: "You can find information about the CAHighways layer of your vector project at this web site:"

I'm open to other solutions, though.

Issue #424, opened 2017-08-02
Fix version tests in geecheck.html


tst-awankhede:
http://localhost/admin/geecheck.html

tst-awankhede:
Following two tests are failing:

Test class:TestVersion
Description:Check Fusion and server versions are aligned.
Test module:fusion_version_test
Test status:ERROR
Test name:testFusionVersionsMatch
Error message:local variable 'version' referenced before assignment
Error type:UnboundLocalError
Error traceback: File "/usr/lib/python2.7/unittest/case.py", line 329, in run testMethod() File "/opt/google/gehttpd/cgi-bin/geecheck_tests/user_tests/fusion_version_test.py", line 59, in testFusionVersionsMatch gee_server_version = common.GetGeeServerVersion() File "/opt/google/gehttpd/cgi-bin/geecheck_tests/common.py", line 89, in GetGeeServerVersion return version
Test class:TestVersion
Description:Check if GEE Server release is the latest available.
Test module:fusion_version_test
Test status:ERROR


Test name:testGeeServerVersion
Error message:local variable 'version' referenced before assignment
Error type:UnboundLocalError
Error traceback: File "/usr/lib/python2.7/unittest/case.py", line 329, in run testMethod() File "/opt/google/gehttpd/cgi-bin/geecheck_tests/user_tests/fusion_version_test.py", line 44, in testGeeServerVersion gee_server_version = common.GetGeeServerVersion() File "/opt/google/gehttpd/cgi-bin/geecheck_tests/common.py", line 89, in GetGeeServerVersion return version

tst-ccamp:
@tst-awankhede This is something we'll need to do for every version, right? If so, we need to update the release process as well.

asmitaw:
yeah, geecheck tests should be run and verified for all releases. However, this test was failing even before we had a version change in our sources.

tst-ccamp:
I mean about the version number change. We will be updating the version number in every release. So, should the tests be updated for every release too?

tst-awankhede:
In release process, 
the step that updates all the version numbers in all applicable files  takes care of this.

The test is failing because of file name fusion_Server_version ( real) vs fusion_server_version( in test)

tst-ccamp:
The process currently lists these files:
` The following files need to be updated with the latest version number:
   - earth_enterprise/src/support/geecheck/geecheck.pl (2 places)
   - earth_enterprise/src/third_party/apache2/apache-ge-index.shtml
[...]

PR #463, opened 2017-08-17
#424: Fixed failing geecheck.html version tests

tst-ppenev:
* Fixed <`fusion_server_version`> file name in `install_server.sh`.
* Fixed Python code to throw exceptions, rather than just create exception objects.
* Removed `testPostgresMaxStackDepth`. (It didn't seem to do a useful check.)
* Add a check and message in the Web client for test suites that were disabled on the command line, or not found on disk. (Fixes Java Script null object exception.)


tst-awankhede:
tested and looks good.
screenshot from 2017-08-22 08-54-18

PR #476, opened 2017-09-01
 #354 - Support building on CentOS6 with Python2.6

### LINKED ISSUE #354: Python version under CentOS 6
When getting Open GEE to compile and execute under CentOS6, it was necessary to upgrade the installed version of python from 2.6 to 2.7.  This, however, broke other system functions that require python2.6, eg, 'yum'.  Python2.7 is required because it is included in Open GEE code (gepublishmanagerhelper.h, line 24 "#include <python2.7/Python.h>").

Can we remove the requirement of python2.7 from the code?  I didn't try 'alias'ing it.

tst-awankhede:
resolves  #354 
regular build and stage_install commands should work

In the http://localhost/admin there is a bug that some icons are not loading hence not showing the db name, but it is not related to this change , hence use geserveradmin commands to publish the db once pushed successfully with fusion.

I will track the icons bug separately.

### DIFF
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -382,22 +382,25 @@ def get_cc_version(cc_path):
 
   return version
 
+def is_version_ge(version_components, comparand_components):
+  """Tests if a version number is greater than, or equal to another version
+  number.  Both version numbers are specified as lists of version components.
+  """
 
-def get_cc_version(cc_path):
-  """Extracts the version number of a C/C++ compiler (like GCC) as a list of
-  strings."""
+  if version_components is None:
+    return False
 
-  process = subprocess.Popen(
-    [cc_path, '--version'], stdout=subprocess.PIPE)
-  (stdout, stderr) = process.communicate()
-  match = re.search('[0-9][0-9.]*', stdout)
-  # If match is None we don't know the version.
-  if match is None:
-    return None
-  version = match.group().split('.')
+  for i in xrange(len(comparand_components)):
+    if i >= len(version_components):
+      version_component = 0
+    else:
+      version_component = int(version_components[i])
+    if int(comparand_components[i]) > version_component:
+      return False
+    if int(comparand_components[i]) < version_component:
+      return True
 
-  return version
-
+  return True
 
 def FusionVersionNumber(source):
   """Read the Fusion Version number from source file.
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -395,6 +395,8 @@ def is_version_ge(version_components, comparand_components):
       version_component = int(version_components[i])
     if int(comparand_components[i]) > version_component:
       return False
+    if int(comparand_components[i]) < version_component:
+      return True
 
   return True
 
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -67,7 +67,7 @@ ValidOptions = set(
      'jobstats',  # Use to enable jobstats profile output.
      'internal',  # Use to enable internal use only debug classes
      'cross_cc',  # Use the cross compiler.
-     'cc_path',
+     'cc_dir',
      'coverage',
      'bundle_3rd_libs',  # Build with bundling 3rd party libs.
      ]);
@@ -92,7 +92,7 @@ internal = int(ARGUMENTS.get('internal', 0))
 native_cc = not int(ARGUMENTS.get('cross_cc', 0))
 coverage = ARGUMENTS.get('coverage', 0)
 bundle_3rd_libs = int(ARGUMENTS.get('bundle_3rd_libs', 0))
-cc_path = ARGUMENTS.get('cc_path', 0)
+cc_dir = ARGUMENTS.get('cc_dir', 0)
 
 # make some directory variables used by other parts of the build system
 top_dir = Dir('#').abspath
@@ -367,6 +367,37 @@ installdirs = {
     'htdocs': InstServerDir('gehttpd/htdocs'),
 }
 
+def get_cc_version(cc_path):
+  """Extracts the version number of a C/C++ compiler (like GCC) as a list of
+  strings."""
+
+  process = subprocess.Popen(
+    [cc_path, '--version'], stdout=subprocess.PIPE)
+  (stdout, stderr) = process.communicate()
+  match = re.search('[0-9][0-9.]*', stdout)
+  # If match is None we don't know the version.
+  if match is None:
+    return None
+  version = match.group().split('.')
+
+  return version
+
+
+def get_cc_version(cc_path):
+  """Extracts the version number of a C/C++ compiler (like GCC) as a list of
+  strings."""
+
+  process = subprocess.Popen(
+    [cc_path, '--version'], stdout=subprocess.PIPE)
+  (stdout, stderr) = process.communicate()
+  match = re.search('[0-9][0-9.]*', stdout)
+  # If match is None we don't know the version.
+  if match is None:
+    return None
+  version = match.group().split('.')
+
+  return version
+
 
 def FusionVersionNumber(source):
   """Read the Fusion Version number from source file.
@@ -407,7 +438,7 @@ env = khEnvironment(
     jobstats=jobstats,
     internal=internal,
     native_cc=native_cc,
-    cc_path=cc_path,
+    cc_dir=cc_dir,
     bundle_3rd_libs=bundle_3rd_libs,
     is_hardy=is_hardy,
     exportdirs=exportdirs,
@@ -440,9 +471,18 @@ env = khEnvironment(
     python_major_version=python_major_version
 )
 
-if cc_path:
-    env['CXX'] = cc_path + '/g++'
-    env['CC'] = cc_path + '/gcc'
+# Check for GCC 4.8 from an alternative toolchain installation on RHEL 6:
+if cc_dir is None and os.path.isfile('/opt/rh/devtoolset-2/root/usr/bin/g++'):
+    # Check if the version of the default compiler is, at least 4.8:
+    if not is_version_ge(get_cc_version(env['CXX']), [4, 8]):
+        cc_dir = '/opt/rh/devtoolset-2/root/usr/bin'
+        rpathlink_dirs += [
+          '/opt/rh/devtoolset-2/root/usr/lib64',
+          '/opt/rh/devtoolset-2/root/usr/lib']
+
+if cc_dir:
+    env['CXX'] = cc_dir + '/g++'
+    env['CC'] = cc_dir + '/gcc'
 
 # The following is to make Scons 1.xx behave like it used to with scons .96
 env['SHCXXFLAGS'] = ['$CXXFLAGS', '-fPIC']
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -367,12 +367,12 @@ installdirs = {
     'htdocs': InstServerDir('gehttpd/htdocs'),
 }
 
-def get_cc_version(cc_path):
+def get_cc_version(cc_dir):
   """Extracts the version number of a C/C++ compiler (like GCC) as a list of
   strings."""
 
   process = subprocess.Popen(
-    [cc_path, '--version'], stdout=subprocess.PIPE)
+    [cc_dir, '--version'], stdout=subprocess.PIPE)
   (stdout, stderr) = process.communicate()
   match = re.search('[0-9][0-9.]*', stdout)
   # If match is None we don't know the version.
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -92,7 +92,7 @@ internal = int(ARGUMENTS.get('internal', 0))
 native_cc = not int(ARGUMENTS.get('cross_cc', 0))
 coverage = ARGUMENTS.get('coverage', 0)
 bundle_3rd_libs = int(ARGUMENTS.get('bundle_3rd_libs', 0))
-cc_dir = ARGUMENTS.get('cc_dir', 0)
+cc_dir = ARGUMENTS.get('cc_dir', None)
 
 # make some directory variables used by other parts of the build system
 top_dir = Dir('#').abspath
@@ -367,19 +367,17 @@ installdirs = {
     'htdocs': InstServerDir('gehttpd/htdocs'),
 }
 
-def get_cc_version(cc_dir):
+def get_cc_version(cc_path):
   """Extracts the version number of a C/C++ compiler (like GCC) as a list of
   strings."""
-
   process = subprocess.Popen(
-    [cc_dir, '--version'], stdout=subprocess.PIPE)
+    [cc_path, '--version'], stdout=subprocess.PIPE)
   (stdout, stderr) = process.communicate()
   match = re.search('[0-9][0-9.]*', stdout)
   # If match is None we don't know the version.
   if match is None:
     return None
   version = match.group().split('.')
-
   return version
 
 def is_version_ge(version_components, comparand_components):
@@ -387,7 +385,7 @@ def is_version_ge(version_components, comparand_components):
   number.  Both version numbers are specified as lists of version components.
   """
 
-  if version_components is None:
+  if version_components is None or comparand_components is None:
     return False
 
   for i in xrange(len(comparand_components)):
@@ -397,8 +395,6 @@ def is_version_ge(version_components, comparand_components):
       version_component = int(version_components[i])
     if int(comparand_components[i]) > version_component:
       return False
-    if int(comparand_components[i]) < version_component:
-      return True
 
   return True
 
@@ -475,6 +471,7 @@ env = khEnvironment(
 )
 
 # Check for GCC 4.8 from an alternative toolchain installation on RHEL 6:
+
 if cc_dir is None and os.path.isfile('/opt/rh/devtoolset-2/root/usr/bin/g++'):
     # Check if the version of the default compiler is, at least 4.8:
     if not is_version_ge(get_cc_version(env['CXX']), [4, 8]):
--- a/earth_enterprise/src/fusion/gepublish/gepublishmanagerhelper.h
+++ b/earth_enterprise/src/fusion/gepublish/gepublishmanagerhelper.h
@@ -23,8 +23,8 @@
 #include <Python.h>
 #include <cstddef>
 
-#include "string"
-#include "vector"
+#include <string>
+#include <vector>
 
 #include "common/ManifestEntry.h"
 
--- a/earth_enterprise/src/fusion/gepublish/gepublishmanagerhelper.h
+++ b/earth_enterprise/src/fusion/gepublish/gepublishmanagerhelper.h
@@ -20,12 +20,11 @@
 #ifndef GEO_EARTH_ENTERPRISE_SRC_FUSION_GEPUBLISH_GEPUBLISHMANAGERHELPER_H_
 #define GEO_EARTH_ENTERPRISE_SRC_FUSION_GEPUBLISH_GEPUBLISHMANAGERHELPER_H_
 
-#include <stddef.h>
+#include <Python.h>
+#include <cstddef>
 
-#include <python2.7/Python.h>  // Note: should be first in include list.
-
-#include <string>
-#include <vector>
+#include "string"
+#include "vector"
 
 #include "common/ManifestEntry.h"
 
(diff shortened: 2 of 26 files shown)


tst-awankhede:
updated CLA.

earth_enterprise/src/SConstruct
```diff
@@ -367,6 +367,37 @@ installdirs = {
...
+  match = re.search('[0-9][0-9.]*', stdout)
+  # If match is None we don't know the version.
+  if match is None:
+    return None
+  version = match.group().split('.')
+
+  return version
+
+
+def get_cc_version(cc_path):
```
tst-ppenev: This looks like a duplicate of the function above, rather than a definition of the `is_version_ge` function.

earth_enterprise/src/server/wsgi/SConscript
```diff
@@ -31,7 +31,7 @@ wsgi_env = env.Clone()
 
 root_dir = Dir(wsgi_env.exportdirs['root']).abspath
 python_bin = wsgi_env['python_bin']
-compiler_cmd = '%s -m compileall -f' % python_bin
+compiler_cmd = '%s -m py_compile ' % python_bin
```
tst-ppenev: Very minor: Looks like there's an extra space at the end of this command.

earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
```diff
@@ -1417,9 +1417,8 @@ def _GetVhNameToBaseUrlDict(self):
       (scheme://host[:port]) -  {vh_name: vh_base_url}
     """
     vh_list = self.QueryVhList()
-    vhname_to_baseurl_dct = {
-        vh_name: self.GetVhBaseUrl(vh_url, vh_ssl) for (
-            vh_name, vh_url, vh_ssl) in vh_list}
+    for (vh_name, vh_url, vh_ssl) in vh_list:
```
tst-ppenev: Q: Have you tested this function? Do you not need to define `vhname_to_base_url_dict = {}` before you can assign its elements?

tst-lsavoie:
@tst-awankhede it looks like your email address wasn't set up correctly when you committed the changes.  It's showing up as invalid-email-address on the commit.  I think that's why it's complaining about CLAs.

earth_enterprise/src/SConstruct
```diff
@@ -367,6 +367,37 @@ installdirs = {
     'htdocs': InstServerDir('gehttpd/htdocs'),
 }
 
+def get_cc_version(cc_path):
+  """Extracts the version number of a C/C++ compiler (like GCC) as a list of
+  strings."""
+
+  process = subprocess.Popen(
+    [cc_path, '--version'], stdout=subprocess.PIPE)
```
tst-lsavoie: Looks like this will need to be changed to `cc_dir`.

earth_enterprise/src/SConstruct
```diff
@@ -367,6 +367,37 @@ installdirs = {
     'htdocs': InstServerDir('gehttpd/htdocs'),
 }
 
+def get_cc_version(cc_path):
+  """Extracts the version number of a C/C++ compiler (like GCC) as a list of
+  strings."""
+
+  process = subprocess.Popen(
```
tst-lsavoie: Minor: I know this is @tst-ppenev's code, but you could simplify this by using the check_output function.  It returns the output directly so you don't have to deal with redirecting the output and using process.communicate().

earth_enterprise/src/SConstruct
```diff
@@ -367,6 +367,40 @@ installdirs = {
...
+  version = match.group().split('.')
+
+  return version
+
+def is_version_ge(version_components, comparand_components):
+  """Tests if a version number is greater than, or equal to another version
+  number.  Both version numbers are specified as lists of version components.
+  """
+
+  if version_components is None:
```
tst-lsavoie: This will fail if comparand_components is none.

tst-awankhede:
@tst-lsavoie where is the email not setup ? It correctly resolves the author to tst-awankhede  profile. Also in git client, this is what is setup
screenshot from 2017-09-01 11-47-31

earth_enterprise/src/SConstruct
tst-ppenev: Yes.  It also doesn't make much sense to try comparing a version to something that you don't specify.

tst-ppenev: I think that's fine.  A caller shouldn't be asking to compare a version to something the caller is not specified.

earth_enterprise/src/SConstruct
```diff
@@ -367,6 +367,40 @@ installdirs = {
     'htdocs': InstServerDir('gehttpd/htdocs'),
 }
 
+def get_cc_version(cc_dir):
```
tst-ppenev: This should be changed back to `cc_path`. This is the path to the compiler whose version is being tested; it's not a path to a directory.

earth_enterprise/src/fusion/gepublish/gepublishmanagerhelper.h
```diff
@@ -20,12 +20,11 @@
...
 
-#include <stddef.h>
+#include <Python.h>
+#include <cstddef>
 
-#include <python2.7/Python.h>  // Note: should be first in include list.
-
-#include <string>
-#include <vector>
+#include "string"
```
tst-ppenev: Q: Why is this changed to `"string"` and `"vector"`? Aren't we including the standard library headers here?

earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
```diff
@@ -1417,9 +1417,9 @@ def _GetVhNameToBaseUrlDict(self):
       (scheme://host[:port]) -  {vh_name: vh_base_url}
     """
     vh_list = self.QueryVhList()
-    vhname_to_baseurl_dct = {
-        vh_name: self.GetVhBaseUrl(vh_url, vh_ssl) for (
-            vh_name, vh_url, vh_ssl) in vh_list}
+    vhname_to_baseurl_dct = []
```
tst-ppenev: Q: I thought this was supposed to be a dictionary, rather than a list. Are the `vh_name`s consecutive integers?

earth_enterprise/src/SConstruct
tst-ppenev: I don't think that's available in Python 2.6 on RHEL 6.

tst-lsavoie:
@tst-awankhede my only guess is that maybe you were working on a different machine or a VM or something like that.  Your other two commits came through fine.  In any case, we know the commit came from you, so I don't know that we need to worry about it much.

tst-ccamp:
@asmitaw If you click on the commits tab it should show you that one of the commits has an incorrect email. 

image

When I look at the commit locally, I see:
```
commit d0d9fe138e4e4affc279e103abc073d1a157b4ca
Author: root <root@localhost.localdomain>
Date:   Wed Aug 30 10:36:44 2017 -0700

     #354 - Support building on CentOS6 with Python2.6
```

earth_enterprise/src/SConstruct
tst-awankhede: its the compiler path, so leaving it as cc_path

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -250,7 +250,7 @@ check_prereq_software()
 		check_prereq_software_retval=1
 	fi
 
-	if ! software_check "$script_name" "python2.7" "python-2.7.*" "python27-*"; then
```
tst-ppenev: I think instead of separating Cent OS from RHEL, and using it as a separation between Cent OS / RHEL 7 and Cent OS / RHEL 6 we should accept both Python 2.6 and Python 2.7 here, since GEE should now run with both.

We just need to change the `grep` expression to match both '2.6' and '2.7'.

earth_enterprise/src/installer/install_server.sh
```diff
@@ -143,7 +143,7 @@ check_prereq_software()
     check_prereq_software_retval=1
   fi
 
-  if ! software_check "$script_name" "python2.7" "python-2.7.*" "python27-*"; then
```
tst-ppenev: Same comment about accepting both Python 2.6 and 2.7.

tst-lsavoie:
When I try to build with these changes I get failures like the ones below:

```
cc1: error: -Werror=unused-result: No option -Wunused-result
cc1: error: -Werror=unused-local-typedefs: No option -Wunused-local-typedefs
```

I'm using the default compiler (gcc 4.4.7).  Do I need to update it to something newer?

tst-ppenev:
@tst-lsavoie: That's what you get when you build with an old compiler. It doesn't support some of the warning command-line options.

tst-lsavoie:
What's the best way to update the compiler on CentOS 6?  We're going to need instructions for this at some point - might as well get started now.

tst-ccamp:
This has some instructions: https://gist.github.com/giwa/b1fb1e44dc0a7d270881

There's also a "cc_dir" flag where you can pass in the path to the compiler.

tst-dsheremata:
@tst-lsavoie  - I've got some draft markdown as a start on how to setup the new compiler for RH.  It's not checked in yet, but coming in an MR soon. Here's a quick snippet: 

# GEE 5.2.1 Build Prerequisites (All platforms) 


* git 1.7.1+
* gcc 4.8.x
* scons 2.3.x+
* python 2.6.x+ 


# RHEL6 Build Notes

## GCC 4.8

To obtain GCC 4.8 on RHEL6, you must enable the DevToolSet repositories in the RHSM: 

```
sudo subscription-manager repos --enable=rhel-x86_64-workstation-dts-2

sudo yum install devtoolset-2-gcc devtoolset-2-binutils  devtoolset-2-gcc-c++ devtoolset-2-gcc-gfortran
```

The gcc 4.8 installation will be located in the `/opt/rh/devtoolset-2/root/usr/bin/` directory   To switch the environment to use gcc 4.8, run: 

```source /opt/rh/devtoolset-2/enable``` 

NOTE:  There is still an issue that we’re working where scons does NOT properly use the gcc 4.8 - we’re actively working to fix this. 


## SCONS 2.3.x+
Scons 2.0 provided by RHSM is not compatible, and it must be upgraded. A “monarch” package is provided from [scons.org](scons-2.5.1-1.noarch.rpm):

```
wget scons-2.5.1-1.noarch.rpm
sudo yum install scons-2.5.1-1.noarch.rpm
```

**Comment by tst-lsavoie** [comment]
@tst-dsheremata @tst-ccamp thanks.

**Comment by tst-lsavoie** [comment]
Is this supposed to support building portable as well or are we handling that separately?  I get python errors when building portable.

**Comment by tst-ppenev** [comment]
@tst-lsavoie: What errors are you getting for Portable? I was able to build and install, but I'm not using exactly this branch.

**Comment by tst-lsavoie** [comment]
```
  File "earthenterprise/earth_enterprise/src/portableserver/build.py", line 251, in <module>
    main(sys.argv)
  File "earthenterprise/earth_enterprise/src/portableserver/build.py", line 248, in main
    builder.build()
  File "earthenterprise/earth_enterprise/src/portableserver/build.py", line 93, in build
    self.build_fileunpacker()
  File "earthenterprise/earth_enterprise/src/portableserver/build.py", line 177, in build_fileunpacker
    build_and_test.main(['build_and_test.py', self.platform])
  File "/home/admin/earthenterprise/earth_enterprise/src/portableserver/build/fileunpacker-build/build_and_test.py", line 84, in main
    if BuildLibrary(os_dir, argv[1].lower()=="windows"):
  File "/home/admin/earthenterprise/earth_enterprise/src/portableserver/build/fileunpacker-build/build_and_test.py", line 46, in BuildLibrary
    result = util.ExecuteCmd(line.format(**build_vars), use_shell=True)
  File "/home/admin/earthenterprise/earth_enterprise/src/portableserver/build/fileunpacker-build/util.py", line 48, in ExecuteCmd
    print "Executing: {}".format(os_cmd if use_shell else ' '.join(os_cmd))
ValueError: zero length field name in format
```

It looks like the fix is similar to some of the other changes already made on this PR but this particular file wasn't modified.

**Comment by tst-ppenev** [comment]
Yeah. I thought that was fixed in this PR, but it may have been missed.

This is just the error message for not recognizing your OS that's failing to be displayed.

We can fix it in this pull request, or leave it for another.  I build and install Portable in my working branch.

**Comment by tst-ppenev** [comment]
@tst-dsheremata: I haven't had to upgrade `scons` on RHEL 6.

**Comment by tst-lsavoie** [comment]
I'm happy to leave it for another PR.  We just need to make sure we get everything working for both CentOS 6 and RHEL 6.

File: earth_enterprise/src/SConstruct (hunk 5 above)
Comment (tst-awankhede) [reply earth_enterprise/src/SConstruct]: sorry, that looks like out of scope for this ticket.

File: earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py (hunk 3 above)
Comment (tst-awankhede) [reply earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py]: those are virtual host names, I have not tested, but I corrected it as per your comments

File: earth_enterprise/src/SConstruct [hunk 12]
```diff
@@ -397,8 +395,6 @@ def is_version_ge(version_components, comparand_components):
       version_component = int(version_components[i])
     if int(comparand_components[i]) > version_component:
       return False
-    if int(comparand_components[i]) < version_component:
```
tst-ppenev: Q: Why was this removed? The comparison won't work correctly, if we don't make this comparison.

earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
tst-ppenev: Q: I don't think the original code created a dictionary mapping keys to lists of `self.GetVhBaseUrl(. . .)` results.  It seemed just a map: key -> `self.GetVhBaseUrl(. . .)`.

earth_enterprise/src/SConstruct
tst-awankhede: why it would not work? I tested and it works. Basically, the line above was redundant, everything else than the check that is retained is and was suppose to return TRUE and its doing so.

earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
tst-awankhede: @tst-ppenev  sorry I am not sure I follow your comment.

earth_enterprise/src/SConstruct
tst-ppenev: It will work for the specific case of comparing 4.4.7 with 4.8, but it won't work for comparing two versions in general.

The line is not redundant.

This is comparing two numbers digit by digit starting from the most significant digit.  At each step, you either know which number is bigger, or the two digits must be equal.  The following digits have no way of telling you which number is bigger, unless all the more significant ones are equal.

You can try this for yourself:

```Python
# p.is_version_ge() is the function with the less-than check removed.
# q.is_version_ge() is the function with the less-than check.

>>> p.is_version_ge([4, 4, 7], [4, 8])
False
>>> q.is_version_ge([4, 4, 7], [4, 8])
False
# Comparing 4.4.7 and 4.8 works with the buggy version.

>>> p.is_version_ge([5, 2, 1], [4, 3])
False
>>> q.is_version_ge([5, 2, 1], [4, 3])
True
# Comparing 5.2.1 and 4.3 doesn't.
```

earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
tst-ppenev: Try this for yourself.

File `p.py`:
```Python
class PublishManagerHelper(object):
  """Mock class with the _GetVhNameToBaseUrlDict() method on the
  'centos-6-support-new' branch."""

  def __init__(self, query_vh_list):
    self.query_vh_list = query_vh_list

  def QueryVhList(self):
    return self.query_vh_list

  def GetVhBaseUrl(self, vh_url, vh_ssl):
    return vh_url

  def _GetVhNameToBaseUrlDict(self):
    """Builds a dictionary mapping a virtual host name to a base URL.
    Returns:
      dictionary mapping a virtual host name to a base URL
      (scheme://host[:port]) -  {vh_name: vh_base_url}
    """
    vh_list = self.QueryVhList()
    vhname_to_baseurl_dct = dict(
        (vh_name,[]) for (vh_name, vh_url, vh_ssl) in vh_list)
    for (vh_name, vh_url, vh_ssl) in vh_list:
      vhname_to_baseurl_dct[vh_name].append(self.GetVhBaseUrl(vh_url, vh_ssl))

    return vhname_to_baseurl_dct
```

File `q.py`:
```Python
class PublishManagerHelper(object):
  """Mock class with the _GetVhNameToBaseUrlDict() method on the 'master'
  branch."""

  def __init__(self, query_vh_list):
    self.query_vh_list = query_vh_list

  def QueryVhList(self):
    return self.query_vh_list

  def GetVhBaseUrl(self, vh_url, vh_ssl):
    return vh_url

  def _GetVhNameToBaseUrlDict(self):
    """Builds a dictionary mapping a virtual host name to a base URL.
    Returns:
      dictionary mapping a virtual host name to a base URL
      (scheme://host[:port]) -  {vh_name: vh_base_url}
    """
    vh_list = self.QueryVhList()
    vhname_to_baseurl_dct = {
        vh_name: self.GetVhBaseUrl(vh_url, vh_ssl) for (
            vh_name, vh_url, vh_ssl) in vh_list}

    return vhname_to_baseurl_dct
```

Compare old method and updated method:
```Python
>>> import p
>>> import q
>>> u = p.PublishManagerHelper([('localhost', 'http://127.0.0.1', False), ('thanksgivinghost', 'https://bay.king.it', True)])
>>> u._GetVhNameToBaseUrlDict()
{'localhost': ['http://127.0.0.1'], 'thanksgivinghost': ['https://bay.king.it']}
# This method returns a dictionary with values which are lists.
>>> old = q.PublishManagerHelper([('localhost', 'http://127.0.0.1', False), ('thanksgivinghost', 'https://bay.king.it', True)])
>>> old._GetVhNameToBaseUrlDict()
{'localhost': 'http://127.0.0.1', 'thanksgivinghost': 'https://bay.king.it'}
# This method returns a dictionary with values which are not lists (strings in this case).
```

earth_enterprise/src/SConstruct
tst-lsavoie: I'm OK with leaving it as is.

earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
```diff
@@ -1417,9 +1417,11 @@ def _GetVhNameToBaseUrlDict(self):
       (scheme://host[:port]) -  {vh_name: vh_base_url}
     """
     vh_list = self.QueryVhList()
-    vhname_to_baseurl_dct = {
-        vh_name: self.GetVhBaseUrl(vh_url, vh_ssl) for (
-            vh_name, vh_url, vh_ssl) in vh_list}
+    vhname_to_baseurl_dct = dict(
```
tst-ppenev: :stuck_out_tongue: You don't need a for loop to create an empty dictionary, but it will work.

It could be just:

```Python
    vh_list = self.QueryVhList()
    vhname_to_baseurl_dct = {}
  
    for (vh_name, vh_url, vh_ssl) in vh_list:
      vhname_to_baseurl_dct[vh_name] = self.GetVhBaseUrl(vh_url, vh_ssl)
    return vhname_to_baseurl_dct
```

tst-awankhede: vhname_to_baseurl_dct = {}   throws error in 2.6, hence this hack

tst-ppenev: I don't have a problem with it with Python 2.6.6.

tst-ppenev: You could also use `vhname_to_baseurl_dct = dict()`.

tst-awankhede: for first suggestion , as I mentioned this is the error:
>>> import p
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "p.py", line 22
    vhname_to_baseurl_dct = dict{}
                                ^
SyntaxError: invalid syntax

tst-awankhede: Since its approved, I am not planning to make any more changes although seemingly vhname_to_baseurl_dct = dict() this is not throwing error, but not tested, so not changing it.

Issue #271, opened 2017-06-26
Rewrite legal notices


tst-lsavoie:
Help page 2987876.html includes a number of legal notices.  We may need to rewrite these notices to reflect the fact that GEE is open source (for example, removing the restriction on reverse engineering, reproduction, distribution, etc).  We also should remove any packages that are no longer used from the list of packages.

tst-lsavoie:
Closed via PR #478.

PR #478, opened 2017-09-05
#271 Updates legal notices page and adds license files 

tst-ccamp:
Implements: #271 
For preview: https://tst-ccamp.github.io/earthenterprise/geedocs/answer/2987876.html

### DIFF
--- a/docs/geedocs/answer/2987876.html
+++ b/docs/geedocs/answer/2987876.html
@@ -17,15 +17,36 @@
 <h1><a href="../index.html">Google Earth Enterprise Documentation Home</a> | Install GEE Server</h1>
 <h2>Legal notices</h2>
 
-<p>Copyright © 2008, 2009, 2010, 2011, 2012, 2013, 2014 Google, Inc.<br/>
+<p>Copyright &copy; 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017 Google, Inc.<br/>
+Copyright &copy; 2017 Open GEE Contributors<br/>
 All rights reserved.</p>
-<p>Google, Google Earth, and Keyhole are trademarks or registered trademarks of Google, Inc. in the United States and elsewhere. Other trademarks not owned by Google, Inc. are the property of their respective owners.</p>
-<p>All Google and Keyhole logos and graphics are copyrighted. The content of this publication is furnished for informational use only, is subject to change without notice, and should not be construed as a commitment by Google, Inc.</p>
-<p>This publication is provided pursuant to and subject to the terms of Google Earth Enterprise. This publication and the information herein may not be reverse engineered, reproduced, duplicated, modified, altered, disclosed, published, printed, or distributed to any third party whatsoever without the express written permission of Google, Inc.</p>
-<p>Google Earth Enterprise may utilize the following products as integrated components, linked libraries, or optional modules. No usage rights, expressed or implied, are provided for said products independent of Google Earth Enterprise, unless otherwise stated in the product documentation. All trademarks are the property of their respective owners.</p>
+<p>
+  Google, Google Earth, and Keyhole are trademarks or registered trademarks of Google, Inc. in the United States and
+  elsewhere. Other trademarks not owned by Google, Inc. are the property of their respective owners.
+</p>
+<p>
+  All Google and Keyhole logos and graphics are copyrighted. The content of this publication is furnished for 
+  informational use only, is subject to change without notice, and should not be construed as a commitment by 
+  Google, Inc. or Open GEE Contributors
+</p>
+<p>
+  Open Google Earth Enterprise is licensed under the 
+  <a href="https://raw.githubusercontent.com/google/earthenterprise/master/LICENSE">Apache License v2.0</a>
+</p>
+<p>
+  Open Google Earth Enterprise may utilize the following products as integrated components, linked libraries, or
+  optional modules. No usage rights, expressed or implied, are provided for said products independent of Open Google 
+  Earth Enterprise, unless otherwise stated in the product documentation. All trademarks are the property of their
+  respective owners. Each product is released under their own respective license. For bundled products, license 
+  information is available under the 
+  <a href="https://github.com/google/earthenterprise/tree/master/earth_enterprise/third_party">third_party directory</a>
+  of the repository.
+</p>
 <ul>
 <li>Apache</li>
+<li>APR</li>
 <li>eXpat</li>
+<li>Firefox</li>
 <li>GDAL</li>
 <li>GEOS</li>
 <li>gif</li>
@@ -45,6 +66,7 @@ All rights reserved.</p>
 <li>MrSID</li>
 <li>numpy</li>
 <li>OGDI</li>
+<li>OpenJPEG</li>
 <li>OpenLDAP</li>
 <li>OpenSSL</li>
 <li>PIL</li>
@@ -58,5 +80,5 @@ All rights reserved.</p>
 <li>Tornado</li>
 <li>Xerces-c</li>
 </ul>
-<div class="footer"><p class="BackToTop"><a href="#top_of_file">Back to top</a> <hr /></p> <p class="copyright">&copy;2015 Google</p></div>
+<div class="footer"><p class="BackToTop"><a href="#top_of_file">Back to top</a> <hr /></p> <p class="copyright">&copy;2015 Google; &copy;2017 Open GEE Contributors</p></div>
 </div></body> </html>
--- a/docs/geedocs/answer/2987876.html
+++ b/docs/geedocs/answer/2987876.html
@@ -29,11 +29,11 @@ All rights reserved.</p>
   <a href="https://raw.githubusercontent.com/google/earthenterprise/master/LICENSE">Apache License v2.0</a>
 </p>
 <p>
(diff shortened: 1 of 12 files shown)


docs/geedocs/answer/2987876.html
```diff
@@ -17,15 +17,36 @@
 <h1><a href="../index.html">Google Earth Enterprise Documentation Home</a> | Install GEE Server</h1>
```
tst-ccamp: I'd like to move this somewhere else. It doesn't seem to belong in this section and it is also difficult to find through navigating the menu.

tst-ccamp: I moved the page to its own section under the main menu and removed it from the old section.

docs/geedocs/answer/2987876.html
```diff
@@ -4,28 +4,44 @@
...
-<p>Google Earth Enterprise may utilize the following products as integrated components, linked libraries, or optional modules. No usage rights, expressed or implied, are provided for said products independent of Google Earth Enterprise, unless otherwise stated in the product documentation. All trademarks are the property of their respective owners.</p>
+<p>
+  Google, Google Earth, and Keyhole are trademarks or registered trademarks of Google, Inc. in the United States and
+  elsewhere. Other trademarks not owned by Google, Inc. are the property of their respective owners.
+</p>
+<p>
+  All Google and Keyhole logos and graphics are copyrighted. The content of this publication is furnished for 
+  informational use only, is subject to change without notice, and should not be construed as a commitment by 
+  Google, Inc. or Open GEE Contributors
```
tst-lsavoie: Missing a period on the end of this sentence.

docs/geedocs/answer/2987876.html
```diff
@@ -58,5 +75,5 @@
 <li>Tornado</li>
 <li>Xerces-c</li>
 </ul>
-<div class="footer"><p class="BackToTop"><a href="#top_of_file">Back to top</a> <hr /></p> <p class="copyright">&copy;2015 Google</p></div>
+<div class="footer"><p class="BackToTop"><a href="#top_of_file">Back to top</a> <hr /></p> <p class="copyright">&copy;2015 Google; &copy;2017 Open GEE Contributors</p></div>
```
tst-lsavoie: Should we change Google's copyright date from 2015 to 2017 since we did so at the top?

docs/geedocs/answer/2987876.html
```diff
@@ -4,28 +4,44 @@
...
+  optional modules. No usage rights, expressed or implied, are provided for said products independent of Open Google 
+  Earth Enterprise, unless otherwise stated in the product documentation. All trademarks are the property of their
+  respective owners. Each product is released under their own respective license. For bundled products, license 
+  information is available under the 
+  <a href="https://github.com/google/earthenterprise/tree/master/earth_enterprise/third_party">third_party directory</a>
+  of the repository.
+</p>
 <ul>
 <li>Apache</li>
+<li>APR</li>
```
tst-lsavoie: I'm going to assume you went through all the third party libraries and verified that there was a LICENSE file available somewhere in each of them.  I'm not sure it's worthwhile for a second person to do that.

docs/geedocs/answer/2987876.html
tst-ccamp: In this case, I assumed the footer refers to the date on the page, which is why I didn't change it. At the top I assumed it applies to everything in the repo.

docs/geedocs/answer/2987876.html
tst-ccamp: Yep, I did. Twice :)

docs/geedocs/answer/2987876.html
```diff
@@ -4,28 +4,44 @@
...
   </div>
   <div class="content">
 <a name="top_of_file"></a>
 <p><img src="../art/common/googlelogo_color_260x88dp.png" width="130" height="44" alt="Google logo" /></p>
-<h1><a href="../index.html">Google Earth Enterprise Documentation Home</a> | Install GEE Server</h1>
+<h1><a href="../index.html">Google Earth Enterprise Documentation Home</a></h1>
 <h2>Legal notices</h2>
 
-<p>Copyright © 2008, 2009, 2010, 2011, 2012, 2013, 2014 Google, Inc.<br/>
+<p>Copyright &copy; 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017 Google, Inc.<br/>
```
tst-nfarah: Could this be simplified to 2008-2017 Google, Inc.   ?

docs/geedocs/answer/2987876.html
```diff
@@ -45,6 +61,7 @@
 <li>MrSID</li>
```
tst-nfarah: MrSid binaries are still shipped in 5.2.0 ?

docs/geedocs/answer/2987876.html
```diff
@@ -45,6 +61,7 @@
...
 <li>numpy</li>
 <li>OGDI</li>
```
tst-nfarah: This list has corresponding 3rd party licenses in https://github.com/google/earthenterprise/tree/master/earth_enterprise/third_party ?

I don't see few of them. Ex: OGDI, gif, etc.

docs/geedocs/answer/2987876.html
tst-ccamp: No, it isn't, and neither is Kakadu. I went back and forth on this for a bit. The first sentence is: "Open Google Earth Enterprise may utilize the following products as integrated components, linked libraries, or optional modules." Under that definition it seems that MrSID would still apply, but it wouldn't necessarily be provided. So, I can either change the sentence to exclude products that aren't bundled, indicate products that would need to be provided separately (maybe with *), or just leave it as is. Do you have a preference?

docs/geedocs/answer/2987876.html
tst-ccamp: See the above answer as to why

docs/geedocs/answer/2987876.html
tst-lsavoie: I'd suggest changing it to only mention the things that are included with GEE.  I don't think we need to say anything about stuff users add themselves.

tst-nfarah: Since we don't bundle nor ship them, it seems to me, that we would not need to mention them. It would be up to the end user if they add extra components to ensure that they are in compliance with any licensing requirements. It may be confusing to list them as GDAL for example has its license mentioned some where else. Also it's possible to include other libraries that we don't know about. I would probably not listed them or otherwise list only the one that we don't bundle and mention that the list if only listing some of the modules not all of them. My preference is to not list them.

docs/geedocs/answer/2987876.html
```diff
@@ -4,59 +4,64 @@
...
+<li>mod_wsgi</li>
+<li>OpenJPEG</li>
 <li>OpenLDAP</li>
 <li>OpenSSL</li>
+<li>PCRE</li>
 <li>PIL</li>
 <li>PostGIS</li>
 <li>PostgreSQL</li>
-<li>PROJ</li>
 <li>psychopg2</li>
```
tst-lsavoie: Not your change, but there's no h in psycopg2.

docs/geedocs/answer/2987876.html
```diff
@@ -4,59 +4,64 @@
...
+  Earth Enterprise, unless otherwise stated in the product documentation. All trademarks are the property of their
+  respective owners. Each product is released under their own respective license. For bundled products, license 
+  information is available under the 
+  <a href="https://github.com/google/earthenterprise/tree/master/earth_enterprise/third_party">third_party directory</a>
+  of the repository.
+</p>
 <ul>
 <li>Apache</li>
+<li>APR</li>
 <li>eXpat</li>
```
tst-lsavoie: I don't see this one in third_party.  Are we using eXpat?

tst-ccamp: I thought I removed that one

docs/geedocs/answer/2987876.html
```diff
@@ -4,59 +4,63 @@
...
+</p>
+<p>
+  Open Google Earth Enterprise may utilize the following products as integrated components, linked libraries, or
+  optional modules. No usage rights, expressed or implied, are provided for said products independent of Open Google 
+  Earth Enterprise, unless otherwise stated in the product documentation. All trademarks are the property of their
+  respective owners. Each product is released under their own respective license. For bundled products, license 
+  information is available under the 
+  <a href="https://github.com/google/earthenterprise/tree/master/earth_enterprise/third_party">third_party directory</a>
+  of the repository.
+</p>
```
tst-nfarah: Does the first phrase '.... may utilize the following products'  refers to the un-bundled products (that are now unlisted) as later we say 'For bundled products, license information...' ? If yes, may want to remove 'the following' so it read '...may utilize products as ...'.

tst-ccamp: I don't understand the reasoning for the request. I can remove the word "may". In all likelihood it will utilize this list. I guess I can say "utilizes the following bundled products"

tst-nfarah: yes: i would remove 'may' as these products are bundled. I believe 'may' was referring to the other products  (like MrSID, etc.) that you removed.

tst-ccamp: OK, updated

Issue #456, opened 2017-08-15
Inconsistent behavior of vector layers after upgrade


tst-lsavoie:
On RHEL 7, after upgrading from 5.1.3 to 5.2.0 and rebuilding databases, vector layers sometimes have unexpected behavior.  For example, as you continuously zoom in, the vector layer may appear, then disappear, then reappear.  Sometimes the layer only disappears for a certain area (perhaps a single tile).

Steps to reproduce:
1. Install 5.1.3
1. Follow the instructions at https://github.com/google/earthenterprise/tree/master/earth_enterprise/src/fusion/tools/gee_test to set up and run the run_tutorial.sh script.
1. After the tutorial files are completely built and pushed, uninstall 5.1.3 and install 5.2.0.
1. Open and close all the resource files, saving them again as necessary.
1. Rebuild and push a database that contains a vector layer.
1. Load the database in a client and zoom in and out to see the layers appear and disappear.

PR #480, opened 2017-09-06
Updated release notes for Open GEE 5.2.0

tst-nfarah:
Updated release notes for 5.2.0 by adding resolved issues (after 5.2.0 beta) and known issues.

### DIFF
--- a/docs/geedocs/answer/7160000.html
+++ b/docs/geedocs/answer/7160000.html
@@ -160,18 +160,18 @@ current asset root without stopping Fusion.</p>
         </tr>
         <tr>
           <td>375</td>
-          <td>unbundle psycopg2</td>
-          <td>On Ubuntu 16.04 switch to python-psycopg2' instead of 'python2.7-psycopg2'.</td>
+          <td>Invalid version of psycopg2 on Ubuntu 16.04</td>
+          <td>On Ubuntu 16.04 switch to 'python-psycopg2' instead of 'python2.7-psycopg2'.</td>
         </tr>
         <tr>
           <td>380</td>
           <td>Provider field in resource-view is blank</td>
-          <td>No current work around</td>
+          <td>Open the individual resource to see the provider</td>
         </tr>
         <tr>
           <td>401</td>
           <td>GEE commands are not in the path for sudo. </td>
-          <td>Add /opt/google/bin to the path for all users, including the super user</td>
+          <td>Specify the full path when running commands or add /opt/google/bin to the path for all users, including the super user</td>
         </tr>
         <tr>
           <td>402</td>
@@ -196,11 +196,11 @@ current asset root without stopping Fusion.</p>
         <tr>
           <td>407</td>
           <td>Corrupt data warning when starting fusion</td>
-          <td>No current work around</td>
+          <td>No current work around but Fusion loads and runs correctly.</td>
         </tr>
         <tr>
           <td>423</td>
-          <td>Improve JPEG 2000 format support performance</td>
+          <td>Slower JPEG2000 performance than 5.1.3/td>
           <td>Use Geotiff or other image formats.</td>
         </tr>
         <tr>
@@ -224,9 +224,14 @@ current asset root without stopping Fusion.</p>
           <td>No current work around</td>
         </tr>
         <tr>
+          <td>456</td>
+          <td>Inconsistent behavior of vector layers after upgrade</td>
+          <td>No current work around</td>
+        </tr>
+        <tr>
           <td>474</td>
           <td>Running gee_check on some supported platforms reports that the platform is not supported</td>
-          <td>No current work around</td>
+          <td>You can ignore the failed test if using a supported platform (Ubuntu 14.04, Ubuntu 16.04, RHEL 7, and CentOS 7).</td>
         </tr>
         <tr>
           <td>476</td>
--- a/docs/geedocs/answer/7160000.html
+++ b/docs/geedocs/answer/7160000.html
@@ -307,21 +307,6 @@ current asset root without stopping Fusion.</p>
           <td>Needed to declare the coding in headers of python script:# -- coding: utf-8 -</td>
         </tr>
         <tr>
-          <td>382</td>
-          <td>Update documentation to bring it consistent with Fusion 5.2.0</td>
-          <td>Updated links and edited multiple documents to correct or add missing steps</td>
-        </tr>
-        <tr>
-          <td>413</td>
-          <td>Documentation on clearing the cache is wrong</td>
-          <td>Added correct steps to clear the cache in Google Eart Entreprise Client</td>
-        </tr>
-        <tr>
-          <td>414</td>
-          <td>Errors in tutorial on search tabs</td>
-          <td>Reduced the use of 'search tab' in answer document</td>
-        </tr>
-        <tr>
           <td>431</td>
           <td>portable build failure - CentOS 7 - old version of pexpect. </td>
           <td>Try loading fdpexpect for version 3.* and above. If not successful, fallback to fdpexpect from older pexpect package.</td>
--- a/docs/geedocs/answer/7160000.html
+++ b/docs/geedocs/answer/7160000.html
@@ -81,6 +81,7 @@ current asset root without stopping Fusion.</p>
         </tr>
       </tbody>
     </table>
+  <p>If you're upgrading from GEE 5.1.3, you may have to reinstall the prerequisite libraries after uninstalling 5.1.3 but before installing 5.2.0.</p>
   <h5>
   <h5>
     <p id="known_issues_5.2.0">Known Issues</p>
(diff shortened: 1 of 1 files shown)


tst-nfarah:
I signed it

docs/geedocs/answer/7160000.html
```diff
@@ -148,10 +149,95 @@
...
+        </tr>
+        <tr>
+          <td>380</td>
+          <td>Provider field in resource-view is blank</td>
+          <td>No current work around</td>
+        </tr>
+        <tr>
+          <td>401</td>
+          <td>GEE commands are not in the path for sudo. </td>
+          <td>Add /opt/google/bin to the path for all users, including the super user</td>
```
tst-lsavoie: Minor: another workaround is to specify the full path when running commands.  Not sure if it's worth adding that that.

docs/geedocs/answer/7160000.html
```diff
@@ -148,10 +149,95 @@
...
+        </tr>
+        <tr>
+          <td>405</td>
+          <td>Vector layer preview not cleared in some situations</td>
+          <td>Reset the preview window to the correct state by either clicking on it or previewing another vector layer</td>
+        </tr>
+        <tr>
+          <td>407</td>
+          <td>Corrupt data warning when starting fusion</td>
+          <td>No current work around</td>
```
tst-lsavoie: May want to state here that Fusion still operates correctly (as far as we can tell) so you can ignore the warning.

docs/geedocs/answer/7160000.html
```diff
@@ -148,10 +149,95 @@
...
+        </tr>
+        <tr>
+          <td>423</td>
+          <td>Improve JPEG 2000 format support performance</td>
```
tst-lsavoie: Minor: I'd reword this to something like "JPEG 2000 performance is slow" or something like that.

docs/geedocs/answer/7160000.html
```diff
@@ -148,10 +149,95 @@
...
+        </tr>
+        <tr>
+          <td>453</td>
+          <td>Improve `check_server_processes_running` detection for uninstall</td>
+          <td>No current work around</td>
+        </tr>
+        <tr>
+          <td>474</td>
+          <td>Running gee_check on some supported platforms reports that the platform is not supported</td>
+          <td>No current work around</td>
```
tst-lsavoie: Maybe should say that they can ignore the failed test if they are using a supported platform and list the supported platforms (Ubuntu 14.04, Ubuntu 16.04, RHEL 7, and CentOS 7).

docs/geedocs/answer/7160000.html
```diff
@@ -148,10 +149,95 @@
...
+        </tr>
+        <tr>
           <td>342</td>
           <td>Fusion crashes when opening an unsupported file type</td>
           <td>Re-open fusion and avoid opening unsupported file types.</td>
         </tr>
+        <tr>
+          <td>375</td>
+          <td>unbundle psycopg2</td>
+          <td>On Ubuntu 16.04 switch to python-psycopg2' instead of 'python2.7-psycopg2'.</td>
```
tst-lsavoie: Missing tick mark before python-psycopg2.

Also, I think it would be good to make the title more descriptive: maybe "Invalid version of psycopg2 on Ubuntu 16.04".

docs/geedocs/answer/7160000.html
```diff
@@ -210,6 +296,41 @@
           <td>Portable Globe Cutter fails with a Python error</td>
           <td>Fixed import errors in the cutter scripts</td>
         </tr>
+        <tr>
+          <td>381</td>
+          <td>Error running POI Search</td>
+          <td>Needed to declare the coding in headers of python script:# -- coding: utf-8 -</td>
+        </tr>
+        <tr>
+          <td>382</td>
```
tst-lsavoie: I think these three documentation fixes (382, 413, and 414) can be removed.  We have a statement at the top (in the paragraph about open sourcing) that says a number of documentation changes were made.  I think that covers this, though I'd be open to other's thoughts.

docs/geedocs/answer/7160000.html
```diff
@@ -81,6 +81,7 @@
         </tr>
       </tbody>
     </table>
+  <p>If you're upgrading from GEE 5.1.3, you may have to reinstall the prerequisite libraries after uninstalling 5.1.3 but before installing 5.2.0.</p>
```
tst-lsavoie: This was not in the original ticket, but perhaps we should also point out that we made a number of testing updates.  That could go in the section near the top on Open Sourcing.

docs/geedocs/answer/7160000.html
```diff
@@ -148,10 +149,95 @@
...
         </tr>
+        <tr>
+          <td>375</td>
+          <td>unbundle psycopg2</td>
+          <td>On Ubuntu 16.04 switch to python-psycopg2' instead of 'python2.7-psycopg2'.</td>
+        </tr>
+        <tr>
+          <td>380</td>
+          <td>Provider field in resource-view is blank</td>
+          <td>No current work around</td>
```
tst-lsavoie: You can work around this by opening the resource to view the provider.  It might make it clearer to change the title to indicate that the provider field is blank for resources in the asset manager.  If you open the individual resource, the provider is shown.

docs/geedocs/answer/7160000.html
```diff
@@ -148,10 +149,95 @@
...
+          <td>Fusion installer does not upgrade the asset root on RHEL 7</td>
+          <td>Upgrade the asset root manually by running the command that is printed when you try to start the fusion service</td>
+        </tr>
+        <tr>
+          <td>453</td>
+          <td>Improve `check_server_processes_running` detection for uninstall</td>
+          <td>No current work around</td>
+        </tr>
+        <tr>
+          <td>474</td>
```
tst-lsavoie: We should also add issue #456 in the known bugs.

docs/geedocs/answer/7160000.html
tst-nfarah: Sure. I'll add it.

docs/geedocs/answer/7160000.html
tst-nfarah: ok. How about 'No current work around but Fusion loads and runs correctly' ?

docs/geedocs/answer/7160000.html
tst-nfarah: Even if the original PR (423) title is 'Improve JPEG 2000 format support performance' ?

docs/geedocs/answer/7160000.html
tst-nfarah: ok. How about 'You can ignore the failed test if using a supported platform (Ubuntu 14.04, Ubuntu 16.04, RHEL 7 and CentOS 7)' ?

docs/geedocs/answer/7160000.html
tst-nfarah: ok for the '.' 
about the title, same question as before. We want to edit the title here but keep it as it is in the PR ? (375). The  PR title seems to point to a solution and not necessarily the issue.

docs/geedocs/answer/7160000.html
tst-nfarah: To clarify, you are proposing to move the 'If you're upgrading' to a new section near 'Open Sourcing' section and in that section add 'we made a number of testing updates' ? We should probably explain 'number of testing updates'.

docs/geedocs/answer/7160000.html
tst-nfarah: Ok

docs/geedocs/answer/7160000.html
tst-nfarah: Ok. I'll add it as well to the ticket.

docs/geedocs/answer/7160000.html
tst-lsavoie: Sounds good.

docs/geedocs/answer/7160000.html
tst-lsavoie: I haven't necessarily been keeping the issue title and the wording in the release notes the same, though that's not a bad idea.  The point I want to get across is that jp2 performance will be worse in 5.2.0 than it was in 5.1.3.

docs/geedocs/answer/7160000.html
tst-lsavoie: Looks good.

docs/geedocs/answer/7160000.html
tst-lsavoie: I think the "If you're upgrading..." sentence can stay here.  I only put the comment here because github doesn't let you comment on code that doesn't change.

On second thought I'm going to suggest not mentioning the testing since that's something end users don't really care about.  Which means you don't need to make any changes.

docs/geedocs/answer/7160000.html
tst-lsavoie: It currently says `On Ubuntu 16.04 switch to python-psycopg2'`.  It should say `On Ubuntu 16.04 switch to 'python-psycopg2'`.  The only change is to add an apostrophe.

The reason I'd suggest changing the title is that "unbundle psycopg2" sounds like an enhancement, but there is a known issue where some things don't work on Ubuntu 16.04 with the bundled psycopg2.  Maybe we should change the title here and on the issue.

docs/geedocs/answer/7160000.html
tst-nfarah: yeah, the title does not indeed reflect that. may be we leave it as is and if one interested to learn more they would look at the PR and understand the issue ? or we can change the title here ?

docs/geedocs/answer/7160000.html
tst-nfarah: ok

docs/geedocs/answer/7160000.html
tst-lsavoie: I'd like to change the title here - maybe "Slower JPEG2000 performance than 5.1.3" and then if they want to know more they can look at the issue on github.

docs/geedocs/answer/7160000.html
tst-nfarah: Added the missing '.
Changed the title in the release notes and added a comment/question in the 375 PR to eventually change its title.

docs/geedocs/answer/7160000.html
tst-nfarah: ok

docs/geedocs/answer/7160000.html
tst-nfarah: Yeah if these doc cleanup 'bugs' are not worth mentioning, i can remove them.

tst-lsavoie: I vote to remove them.

tst-nfarah: Done.

tst-lsavoie: I could also see leaving just 382 as a sort of catch all, but you might want to update the issue number to "382 and others" or something like that.  If you try to list them all it will be a long list I think.

PR #491, opened 2017-09-08
Script for building release archives

tst-lsavoie:
The automatically generated archives produced by github do not
include the git lfs files, so they are not suitable for building
the release.  Thus, we make our own archive files.  This script
provides an easy way to do that.

### DIFF
--- a/build_release_archive.sh
+++ b/build_release_archive.sh
@@ -1,10 +1,13 @@
 #!/bin/bash
 
 # This script creates archive files that can be distributed as a release. The
-# files that github provides do not include files stored in git lfs, so this
-# process is required to create working archives that we can publish on github.
-# The resulting archives will be placed in the current directory.  Because this
-# script downloads and packages a large repo it may take a long time to run.
+# automatically generated files on the github releases page do not include the
+# files stored in git lfs, so this process is required to create working
+# archives that we can publish on the github releases page.
+
+# This script places the generated archives in the current directory.  Because
+# this script downloads and packages a large repo it may take a long time to
+# run.
 
 # This script works under /tmp and it assumes that previous artifacts under /tmp
 # (such as previously downloaded repos or previously created archives) can be
--- a/build_release_archive.sh
+++ b/build_release_archive.sh
@@ -37,7 +37,7 @@ CLONE_DIR=`mktemp -d`
 
 # Download the repository
 cd "${CLONE_DIR}"
-git clone -b "${1}" --depth 1 https://github.com/google/${REPO_NAME}.git || exit
+git clone -b "${1}" --depth 1 "https://github.com/google/${REPO_NAME}.git" || exit
 
 # Remove the git-related files from the repo
 find "${REPO_NAME}" -name ".git*" -exec rm -Rf {} +
--- a/build_release_archive.sh
+++ b/build_release_archive.sh
@@ -20,6 +20,14 @@ if [ "$#" -ne 1 ]; then
   exit 1
 fi
 
+# Verify that git lfs is installed
+GIT_LFS_ENV=`git lfs env`
+SMUDGE_HOOK=`expr match "${GIT_LFS_ENV}" '.*git config filter\.lfs\.clean = "\(.*\)".*'`
+if [ -z "${SMUDGE_HOOK}" ]; then
+  echo "You must install git lfs before running this script."
+  exit
+fi
+
 REPO_NAME="earthenterprise"
 CLONE_DIR="/tmp"
 
@@ -30,9 +38,9 @@ RUN_DIR=`pwd`
 # Download the repository
 cd ${CLONE_DIR}
 rm -Rf ${REPO_NAME}
-git clone git@github.com:google/${REPO_NAME}.git
-cd ${REPO_NAME}
-git checkout ${1}
+git clone git@github.com:google/${REPO_NAME}.git || exit
+cd ${REPO_NAME} || exit
+git checkout ${1} || exit
 
 # Remove the git-related files from the repo
 FILES=`find . -name ".git*"`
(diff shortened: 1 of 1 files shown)


build_release_archive.sh
```diff
@@ -0,0 +1,50 @@
...
+CLONE_DIR="/tmp"
+
+TAR_FILE=${REPO_NAME}.tar.gz
+ZIP_FILE=${REPO_NAME}.zip
+RUN_DIR=`pwd`
+
+# Download the repository
+cd ${CLONE_DIR}
+rm -Rf ${REPO_NAME}
+git clone git@github.com:google/${REPO_NAME}.git
```
asmitaw: Do we need to make sure git lfs is installed?

tst-lsavoie: I thought about that, but scons already does that and this script will always be run by an expert, so I didn't feel it was important enough to check.

build_release_archive.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+# This script creates archive files that can be distributed as a release. The
+# files that github provides do not include files stored in git lfs, so this
```
tst-ppenev: Minor: I'd capitalize GitHub and Git LFS.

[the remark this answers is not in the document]
tst-ppenev: We could be more specific, and mention things like "The source archive that GitHub generates automatically for a release does not include . . .".

build_release_archive.sh
```diff
@@ -0,0 +1,50 @@
...
+
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 <committish>"
+  echo "where <committish> is a commit, branch, or tag that defines the release"
+  echo "to be archived."
+  exit 1
+fi
+
+REPO_NAME="earthenterprise"
+CLONE_DIR="/tmp"
```
tst-ppenev: Suggestion: You can use `mktemp -d` and not worry about overwriting, or deleting something by accident under `/tmp`.

build_release_archive.sh
```diff
@@ -0,0 +1,50 @@
...
+
+TAR_FILE=${REPO_NAME}.tar.gz
+ZIP_FILE=${REPO_NAME}.zip
+RUN_DIR=`pwd`
+
+# Download the repository
+cd ${CLONE_DIR}
```
tst-ppenev: Reminder: Quote variable expansion on the following 5 lines.

build_release_archive.sh
tst-ppenev: Suggestion: Use `git clone -b <branch_name> --depth 1 <repository_url>` instead of checking out the whole history on all branches.  I think it saves you downloading about a GB.

build_release_archive.sh
```diff
@@ -0,0 +1,50 @@
...
+rm -Rf ${REPO_NAME}
+git clone git@github.com:google/${REPO_NAME}.git
+cd ${REPO_NAME}
+git checkout ${1}
+
+# Remove the git-related files from the repo
+FILES=`find . -name ".git*"`
+for FILE in ${FILES}; do
+  rm -Rf ${FILE}
```
tst-ppenev: Reminder: Quote variable expansion.

build_release_archive.sh
```diff
@@ -0,0 +1,50 @@
...
+done
+
+# Create the archives
+cd ${CLONE_DIR}
```
tst-ppenev: Reminder: Quote variable expansion on all following lines.

build_release_archive.sh
tst-lsavoie: On second thought, this script doesn't run scons.  I'll look at adding a check.

build_release_archive.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Make a temporary directory
+CLONE_DIR=`mktemp -d`
+
+# Download the repository
+cd "${CLONE_DIR}"
+git clone -b "${1}" --depth 1 git@github.com:google/${REPO_NAME}.git || exit
+
+# Remove the git-related files from the repo
+cd "${REPO_NAME}"
```
tst-ppenev: Minor suggestion: These 5 lines could be just:
```BASH
find "${REPO_NAME}" -name ".git*" -exec rm -Rf {} +
```

tst-ppenev: The code looks fine to me.

build_release_archive.sh
tst-awankhede: +1

asmitaw:
I tried to run it myself with the branch release_5.2.0 and it fails because the pop-up that we regularly see for  public key password is supressed by the script and it always fails with the "permission denied" error.

However  the same command works by itself.
git clone -b release_5.2.0 --depth 1 git@github.com:google/earthenterprise.git

Also, can you please add example in the comments section, at the start of the script.

build_release_archive.sh
```diff
@@ -0,0 +1,55 @@
...
+TAR_FILE=${REPO_NAME}.tar.gz
+ZIP_FILE=${REPO_NAME}.zip
+RUN_DIR=`pwd`
+
+# Make a temporary directory
+CLONE_DIR=`mktemp -d`
+
+# Download the repository
+cd "${CLONE_DIR}"
+git clone -b "${1}" --depth 1 git@github.com:google/${REPO_NAME}.git || exit
```
tst-ppenev: As @asmitaw mentioned, you may want to use an `https` URL here, in case you're running this on a machine without enrolled SSH keys.

Also, you could quote the last command-line argument.

tst-ppenev:
I'd add some quotes to the `git clone` line, but otherwise, this is already approved.

tst-lsavoie:
@tst-awankhede I think your problem is fixed now - give it another try.

PR #494, opened 2017-09-11
Issue 204 experiment with travis ci

tst-eclamar:


.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+
+before_script:
+  - cd earth_enterprise
+
+script:
+  - scons -j3 release=1 build >build.log
+  - cd src/NATIVE-REL-x86_64/bin/tests
+  - ./RunAllTests.pl
+
+cache:
```
asmitaw: What is this cache about?
Seems like the notification is working in OpenGEE channel, however , it failed last time I saw the status.

can you attach a couple of screenshots of the results and Travis CI admin page or build output page?

tst-ccamp: It looks like it's configured to run all the pull requests, including the ones that aren't properly configured for Travis CI. Those will always fail :)

But, the last time it ran from this branch, it got this error:
```
geRelativePathTester succeeded
Fusion Notice:	Closing file /tmp/khk5B0bO/big.txt.
Fusion Notice:	Copying created file: /tmp/khk5B0bO/big.txt
Fusion Warning:	Unable to write 1048576 bytes to /tmp/khk5B0bO/big.txt: No space left on device
ERROR: expected "kBuffSize * kTestFileSize" == "(size_t) statBuf.st_size
       actual  "kBuffSize * kTestFileSize" = 4294967296 , "(size_t) statBuf.st_size" = 3557388288
" at 
  common/tests/geFileUtils_test.cpp:121
ERROR: expected "num_read" == "kBuffSize
       actual  "num_read" = 0 , "kBuffSize" = 4096
" at 
  common/tests/geFileUtils_test.cpp:129
```

That might be something that should be fixed

.travis.yml
```diff
@@ -0,0 +1,31 @@
+sudo: required
+dist: trusty
```
tst-ccamp: Do they only allow certain distributions here? From looking at the logs, they are are using docker. It would be awesome if we could use our own Docker image. That would save a lot of time in the setup and build stage.

.travis.yml
```diff
@@ -0,0 +1,31 @@
+sudo: required
```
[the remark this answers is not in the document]
tst-eclamar: Both.

.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+before_install:
+- sudo apt-get update -qq
+- sudo apt-get install -qq libperl4-corelibs-perl libtool xorg-dev doxygen python-dev alien swig libgtest-dev libstdc++6 libxml2-dev gettext libxinerama-dev libxft-dev libxrandr-dev libxcursor-dev libgdbm-dev libc6 libc6-dev libmng-dev zlib1g-dev libcap-dev libpng12-0 libpng12-dev freeglut3-dev flex libx11-dev bison++ bisonc++ libjpeg-dev libjpeg8-dev python2.7 python2.7-dev python2.7-psycopg2 libogdi3.2-dev libgif-dev libxerces-c-dev libgeos-dev libgeos++-dev libfreetype6 libfreetype6-dev python-imaging libproj-dev python-setuptools libgif-dev libxerces-c-dev libcap-dev libpq-dev openssl libxml2-utils libxmu-dev
+
+
+before_script:
+  - cd earth_enterprise
+
+script:
+  - scons -j3 release=1 build >build.log
```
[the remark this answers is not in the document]
tst-eclamar: Actually, the "-j" parameter seems to be ignored or overridden by Travis CI.  A job only gets two cores and the elapsed time for a compile is consistent with a "-j2" parameter.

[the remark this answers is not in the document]
tst-eclamar: Yes.

.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+
+cache:
+  directories:
+    - $HOME/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party
+
+notifications:
+  slack: opengee:YjzzzTnc7pgITPsrQPATIGwc
+  email:
+    recipients:
+      - elamar@t-sciences.com
```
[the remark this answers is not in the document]
tst-eclamar: The notification mechanism is hard-coded;  multiple address can be specified, but there doesn't seem a way to discover who submitted a job and send an e-mail to that person.

.travis.yml
tst-eclamar: Travis CI only allows 'precise' and 'trusty'.

A google search of "travis ci docker" yields some interesting results (https://docs.travis-ci.com/user/docker/).  I don't know enough Docker to comment on it's usability.

.travis.yml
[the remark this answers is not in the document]
tst-nfarah: I just found this https://docs.travis-ci.com/user/notifications/ about notifications. See section 'How is the build email receiver determined? '. It seems that by default it send an email to the committer (if it can find their email) Or one has to hard code the emails (as you mentioned). I would want to have both options so an 'admin' would get notification but also the committer. This would avoid the 'admin' to pocking team members about failed builds.

.travis.yml
tst-eclamar: Travis CI only allows 'precise' and 'trusty'.

A google search of "travis ci docker" yields some interesting results (https://docs.travis-ci.com/user/docker/). I don't know enough Docker to comment on it's usability.

tst-ccamp: I wrote a follow-up ticket to explore this possibility

.travis.yml
```diff
@@ -0,0 +1,31 @@
...
+  - cd earth_enterprise
+
+script:
+  - scons -j3 release=1 build > build.log
+  - cd src/NATIVE-REL-x86_64/bin/tests
+  - ./RunAllTests.pl
+
+cache:
+  directories:
+    - $HOME/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party
```
tst-ccamp: I think I see part of the reason why the cache isn't working, it's actually specifying the wrong path. This is the path it's caching:
/home/travis/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party
But, this is the actual path to where the third_party directory is being built:
/home/travis/build/tst-eclamar/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party/

tst-ccamp: But, to be honest, even if it was caching the right directory, I don't think it would work. I tried an experiment and the build fails because of missing files. So, I think you will need to cache a lot more directories than just this one. It may be better to create a custom docker image

tst-ccamp: I suggest removing the cache for now. I suspect that it may actually be slowing us down instead of speeding us up. At any rate, it certainly isn't speeding anything up at the moment :)

tst-eclamar: Ok. Will do.

README.md
```diff
@@ -1,4 +1,4 @@
-# Earth Enterprise [![Chat on Slack](https://img.shields.io/badge/chat-on%20slack-ff69b4.svg)](http://slack.opengee.org)
+# Earth Enterprise [![Chat on Slack](https://img.shields.io/badge/chat-on%20slack-ff69b4.svg)](http://slack.opengee.org) [![build](https://travis-ci.org/tst-eclamar/earthenterprise.svg?branch=master)](https://travis-ci.org/tst-eclamar/earthenterprise/builds)
```
tst-lsavoie: We'll need to change tst-eclamar to google in two places on this line, though we may want to hold off on the badge until we get Google to turn it on.

.travis.yml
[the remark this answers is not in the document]
tst-lsavoie: @tst-eclamar is there any way to get the email to work without putting in your personal email address?

Issue #328, opened 2017-07-12
Allow users to specify a decimation value from the UI


tst-ccamp:
As a follow up to #327, allow users to specify the decimation value in the fusion UI for a terrain resource.

AC:
1. Users can specify the decimation value for a terrain project from the fusion UI to be passed to the terrain resource builder

PR #499, opened 2017-09-13
#328 Allow users to specify a decimation value on the command line

tst-mswartz:
Added support for decimation threshold value on the command line as per #328

### DIFF
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
@@ -58,7 +58,7 @@ TmeshGenerator::Generate(float *srcSamples,
                          etArray<uchar> &compressed,
                          const size_t reserve_size,
                          bool decimate,
-                         float decimation_threshold
+                         double decimation_threshold
                         )
 {
   // reset my temporaries - they live in the class so I don't have to
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
@@ -57,9 +57,9 @@ TmeshGenerator::Generate(float *srcSamples,
                          const khTileAddr &tmeshAddr,
                          etArray<uchar> &compressed,
                          const size_t reserve_size,
-			 bool decimate,
-			 float decimation_threshold
-			)
+                         bool decimate,
+                         float decimation_threshold
+                        )
 {
   // reset my temporaries - they live in the class so I don't have to
   // constantly reallocate them.
@@ -77,21 +77,16 @@ TmeshGenerator::Generate(float *srcSamples,
   elev.resize(SizeX * SizeY);
 
   // Error thresholds for decimation - set these to 0 to skip decimation
-  // jdjohnso - These numbers came from empirical testing?
-  // jdjohnso - The allowable error gets smaller as the resolution increases.
-  // jdjohnso - The allowable error on border tiles is 1/10 that of interior
+  // Question - Did these numbers come from empirical testing?
+  // The allowable error gets smaller as the resolution increases.
+  // The allowable error on border tiles is 1/10 that of interior
   // tiles. This is because we really want the border tiles to match. If not
   // we get rips between mesh tiles in the client.
   //int error_level = tmeshAddr.level+1;
-  //double error = 0.009 * pow(2, 11-error_level);
   // Disable decimation for spike evaluation:
-  //int error_level = tmeshAddr.level+1;
 
-//FIXME TODO for debugging
-decimation_threshold = 0.5;
-decimate = true;
-
-  double error = (double) decimation_threshold * pow(2, 11 - (tmeshAddr.level + 1));
+  double threshold_modifier = pow(2, 11 - (tmeshAddr.level + 1));
+  double error = decimation_threshold * threshold_modifier;
   double borderError = decimate ? error / 10.0 : 0.0;
 
   // calculate position of this quadnode WRT the larger heightmap tile.
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
@@ -56,7 +56,10 @@ TmeshGenerator::Generate(float *srcSamples,
                          const khOffset<uint32> &wantOffset,
                          const khTileAddr &tmeshAddr,
                          etArray<uchar> &compressed,
-                         const size_t reserve_size)
+                         const size_t reserve_size,
+			 bool decimate,
+			 float decimation_threshold
+			)
 {
   // reset my temporaries - they live in the class so I don't have to
   // constantly reallocate them.
@@ -79,9 +82,18 @@ TmeshGenerator::Generate(float *srcSamples,
   // jdjohnso - The allowable error on border tiles is 1/10 that of interior
   // tiles. This is because we really want the border tiles to match. If not
   // we get rips between mesh tiles in the client.
-  int error_level = tmeshAddr.level+1;
-  double error = 0.009 * pow(2, 11-error_level);
-  double borderError = error / 10.0;
+  //int error_level = tmeshAddr.level+1;
+  //double error = 0.009 * pow(2, 11-error_level);
+  // Disable decimation for spike evaluation:
+  //int error_level = tmeshAddr.level+1;
+
+//FIXME TODO for debugging
+decimation_threshold = 0.5;
+decimate = true;
+notify(NFY_NOTICE, "Have just reset decimation to %d and decimation_threshold is %f\n", decimate, decimation_threshold);
+
+  double error = (double) decimation_threshold * pow(2, 11 - (tmeshAddr.level + 1));
+  double borderError = decimate ? error / 10.0 : 0.0;
 
   // calculate position of this quadnode WRT the larger heightmap tile.
   // xoff and yoff are in units of samples, and represent the offset
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
@@ -90,7 +90,6 @@ TmeshGenerator::Generate(float *srcSamples,
 //FIXME TODO for debugging
 decimation_threshold = 0.5;
 decimate = true;
-notify(NFY_NOTICE, "Have just reset decimation to %d and decimation_threshold is %f\n", decimate, decimation_threshold);
 
   double error = (double) decimation_threshold * pow(2, 11 - (tmeshAddr.level + 1));
   double borderError = decimate ? error / 10.0 : 0.0;
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
@@ -82,9 +82,6 @@ TmeshGenerator::Generate(float *srcSamples,
   // The allowable error on border tiles is 1/10 that of interior
   // tiles. This is because we really want the border tiles to match. If not
   // we get rips between mesh tiles in the client.
-  //int error_level = tmeshAddr.level+1;
-  // Disable decimation for spike evaluation:
-
   double threshold_modifier = pow(2, 11 - (tmeshAddr.level + 1));
   double error = decimation_threshold * threshold_modifier;
   double borderError = decimate ? error / 10.0 : 0.0;
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
@@ -27,7 +27,6 @@
 class TmeshGenerator
 {
   const khTilespaceFlat &sampleTilespace;
-  bool decimate;
 
   // temporaries used by generate. Hold them in the class so they don't
   // need to be constantly reallocated.
@@ -37,10 +36,8 @@ class TmeshGenerator
   std::vector<etVec3f>      elev;
 
  public:
-  TmeshGenerator(const khTilespaceFlat &sampleTilespace_,
-                 bool decimate_) :
-      sampleTilespace(sampleTilespace_),
-      decimate(decimate_)
+  TmeshGenerator(const khTilespaceFlat &sampleTilespace_) :
+      sampleTilespace(sampleTilespace_)
   { }
 
   void Generate(float *srcSamples,
@@ -48,7 +45,10 @@ class TmeshGenerator
                 const khOffset<uint32> &wantOffset,
                 const khTileAddr &tmeshAddr,
                 etArray<uchar> &compressed,
-                const size_t reserve_size);
+                const size_t reserve_size,
+		bool  decimate,
+		float decimation_threshold
+ 	      );
 };
 
 
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
@@ -47,7 +47,7 @@ class TmeshGenerator
                 etArray<uchar> &compressed,
                 const size_t reserve_size,
                 bool  decimate,
-                float decimation_threshold
+                double decimation_threshold
                );
 };
 
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
@@ -46,9 +46,9 @@ class TmeshGenerator
                 const khTileAddr &tmeshAddr,
                 etArray<uchar> &compressed,
                 const size_t reserve_size,
-		bool  decimate,
-		float decimation_threshold
- 	      );
+                bool  decimate,
+                float decimation_threshold
+               );
 };
 
 
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
@@ -171,7 +171,7 @@ TmeshPackgenTraverser::TmeshPackgenTraverser(
   notify(NFY_NOTICE, "    from %u heightmap tile(s)",
          productCoverage.extents.width() * productCoverage.extents.height());
   notify(NFY_NOTICE, "    %s",
-         config.decimate ? "with decimation" : "no decimation");
+         decimate_ ? "with decimation" : "no decimation");
 }
 
 void TmeshPackgenTraverser::Traverse(uint numcpus) {
--- a/earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
+++ b/earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
@@ -52,13 +52,16 @@ ExtraPixelsImpl::ExtraPixelsImpl(
 TmeshWorkItem::TmeshWorkItem(PacketFileWriter &_writer,
                              const AttributionByExtents &_attributions,
                              const khTilespaceFlat& _sampleTilespace,
-                             bool decimate, 
+                             bool decimate,
                              float decimation_threshold):
     writer(_writer),
     attributions(_attributions),
     sampleTilespace(_sampleTilespace),
     numPiecesUsed(0),
-    generator(sampleTilespace) {
+    generator(sampleTilespace),
+    decimate_(decimate),
+    decimation_threshold_(decimation_threshold) {
+
   uint ratio = RasterProductTilespaceBase.tileSize / sampleTilespace.tileSize;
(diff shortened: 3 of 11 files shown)


earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
```diff
@@ -74,14 +77,17 @@ TmeshGenerator::Generate(float *srcSamples,
...
-  // jdjohnso - The allowable error on border tiles is 1/10 that of interior
+  // Question - Did these numbers come from empirical testing?
+  // The allowable error gets smaller as the resolution increases.
+  // The allowable error on border tiles is 1/10 that of interior
   // tiles. This is because we really want the border tiles to match. If not
   // we get rips between mesh tiles in the client.
-  int error_level = tmeshAddr.level+1;
-  double error = 0.009 * pow(2, 11-error_level);
-  double borderError = error / 10.0;
+  //int error_level = tmeshAddr.level+1;
```
tst-lsavoie: We should remove this line and the next.

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
```diff
@@ -74,14 +77,17 @@ TmeshGenerator::Generate(float *srcSamples,
...
+  // Disable decimation for spike evaluation:
+
+  double threshold_modifier = pow(2, 11 - (tmeshAddr.level + 1));
+  double error = decimation_threshold * threshold_modifier;
+  double borderError = decimate ? error / 10.0 : 0.0;
```
tst-lsavoie: What is the purpose of the `decimate` boolean here?  If we set the threshold to 0 the borderError will come out to 0 as well, so I'm not sure what it's doing for us.

tst-mswartz: the decimate boolean is the one that comes from PacketLevelConfig in autoingest

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
tst-mswartz: Good point..  Just removed them

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
```diff
@@ -150,6 +154,15 @@ TmeshPackgenTraverser::TmeshPackgenTraverser(
       writer_(file_pool_, output, true /* overwrite */),
       extra_cache_(20000 /* apx 160M */),
       not_covered_tiles_exist_(false) {
+  if(decimation_threshold == 0.0)
+  {
+    decimate_ = false;
```
tst-lsavoie: I think we should set `decimate_ = config.decimate ` to maintain the same behavior as previous versions.  That does result in the odd situation where `decimate_` could be false while `decimation_threshold` is non-zero or vice versa, but I'm OK with that at this point.  We may want to do some additional research into what the decimate_ flag is used for, but that would be a task for a different ticket.

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.h
```diff
@@ -75,10 +75,14 @@ class TmeshWorkItem {
...
   uint numPiecesUsed;
   TmeshGenerator generator;
+  float decimation_threshold_;
+  float decimate_;
 
   TmeshWorkItem(PacketFileWriter &_writer,
                 const AttributionByExtents &_attributions,
-                const khTilespaceFlat &_sampleTilespace, bool decimate);
+                const khTilespaceFlat &_sampleTilespace, 
+  bool decimate,
```
tst-lsavoie: Inconsistent indentation here.

earth_enterprise/src/fusion/rasterfuse/gepackgen.cpp
```diff
@@ -37,7 +37,7 @@ void usage(const std::string &progn, const char *msg = 0, ...) {
 
   fprintf(stderr,
           "\n"
-          "usage: %s [options] {--imagery|--terrain} --config <configfile>"
+          "usage: %s [options] {--imagery|--terrain} --config <configfile> --decimation <decimation_threshhold>"
```
tst-lsavoie: Should we mention that the decimation threshold is only valid for terrain?

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
```diff
@@ -150,6 +157,15 @@ TmeshPackgenTraverser::TmeshPackgenTraverser(
       writer_(file_pool_, output, true /* overwrite */),
       extra_cache_(20000 /* apx 160M */),
       not_covered_tiles_exist_(false) {
+  if(decimation_threshold == 0.0)
```
tst-nfarah: should we handle invalid decimation_threshold values as either an error or keep processing as 'no decimation' ? I noticed this parameter validation is done in the command line that end-up calling this class, but it would be also to have validation at this entry point since it may be called from other clients (that might not did validation) ...

tst-mswartz: So you think I should check to see if decimation_threshold is <= 0.0  here?

earth_enterprise/src/fusion/rasterfuse/gepackgen.cpp
tst-mswartz: Addressing this in the commit I am about to make

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
tst-nfarah: It does not hurt since any client code calling this might not check.

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
```diff
@@ -150,6 +157,13 @@ TmeshPackgenTraverser::TmeshPackgenTraverser(
       writer_(file_pool_, output, true /* overwrite */),
       extra_cache_(20000 /* apx 160M */),
       not_covered_tiles_exist_(false) {
+
+  if((decimation_threshold <= 0.0) || (config.decimate == false)) {
```
tst-nfarah: minor: float comparison. can you pass a very small value (ex. 0.000000001) and verify that comparison. likely we may not have to worry about very small numbers.

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
```diff
@@ -56,7 +56,10 @@ TmeshGenerator::Generate(float *srcSamples,
                          const khOffset<uint32> &wantOffset,
                          const khTileAddr &tmeshAddr,
                          etArray<uchar> &compressed,
-                         const size_t reserve_size)
+                         const size_t reserve_size,
+                         bool decimate,
+                         float decimation_threshold
```
asmitaw: why not "double" ?  everything else in the error calculation is double

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
```diff
@@ -74,14 +77,14 @@ TmeshGenerator::Generate(float *srcSamples,
...
+  // Question - Did these numbers come from empirical testing?
+  // The allowable error gets smaller as the resolution increases.
+  // The allowable error on border tiles is 1/10 that of interior
   // tiles. This is because we really want the border tiles to match. If not
   // we get rips between mesh tiles in the client.
-  int error_level = tmeshAddr.level+1;
-  double error = 0.009 * pow(2, 11-error_level);
-  double borderError = error / 10.0;
+  double threshold_modifier = pow(2, 11 - (tmeshAddr.level + 1));
+  double error = decimation_threshold * threshold_modifier;
```
asmitaw: decimation_threshold - double calls for higher precision than float plus everything else is double in this equation.

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
tst-mswartz: But a double takes up twice as much memory .. I think I forgot I was writing server code again

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
asmitaw: its easier to understabd  11- error_level  vs 11 - (tmeshAddr.level + 1), I suggest keep as is

tst-mswartz: We really don't need it though, do we? Wouldn't it make more sense to change everything to floats?

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
```diff
@@ -37,18 +36,19 @@ class TmeshGenerator
   std::vector<etVec3f>      elev;
 
  public:
-  TmeshGenerator(const khTilespaceFlat &sampleTilespace_,
-                 bool decimate_) :
-      sampleTilespace(sampleTilespace_),
-      decimate(decimate_)
```
asmitaw: Why take out from constructor ?

tst-mswartz: It's now being passed in to the Generate() function directly along with decimation_threshold

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
asmitaw: +1

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
tst-mswartz: So you want me to have a 4th variable just for storing (tmeshAddr.level + 1)  ?  Why?

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
tst-mswartz: Are you suggesting that 0.0 is not the same as 0.000000000 ?

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
asmitaw: saw that, we can override in generate, but not sure why it cannot be kept as is in the constrcutors to initialise it there like it used to be before.

tst-mswartz: I don't see the point in keeping it in the constructor when it isn't being used from there.  Can you explain please? We have to call Generate() with the decimation_value regardless as it is now no longer hardcoded into the Generate() function.

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
tst-nfarah: I guess my ex did not apply to float size, but the idea stands, though it is minor. comparing float or double is not an exact thing like integer. Usually it is done with comparing the diff of the abs of the values with some very small value (epsilon) to know for sure the values are less/greater/equal. I think here it is not an issue unless if we expect a very tiny decimation value :)

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
tst-mswartz: I think you posted this comment in the wrong thread.

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
```diff
@@ -91,7 +95,10 @@ void TmeshWorkItem::DoWork(TmeshPrepItem *prep) {
                           subcol * sampleTilespace.tileSize),
          piece->targetAddr,
          piece->compressed,
-         kCRC32Size /* reserve size */);
+         kCRC32Size /* reserve size */,
+         decimate_,
+         decimation_threshold_
```
asmitaw: would be good to have comments to explain the decimate_ 
 and decimate_threshold_

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.h
tst-mswartz: As far as keeping the decimate value in the constructor goes, I genuinely don't see the point. Can you explain please? It is not being used from there. As 0.009 is now no longer hardcoded into the Generate() function, we are now calling Generate() with decimation_threshold value regardless.

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.h
```diff
@@ -75,13 +75,18 @@ class TmeshWorkItem {
...
   TmeshGenerator generator;
+  bool decimate_;
+  float decimation_threshold_;
 
   TmeshWorkItem(PacketFileWriter &_writer,
                 const AttributionByExtents &_attributions,
-                const khTilespaceFlat &_sampleTilespace, bool decimate);
+                const khTilespaceFlat &_sampleTilespace,
+                bool decimate,
+                float decimation_threshold);
```
asmitaw: can decimation_threshold change per TmeshWorkItem ?  Its already part of constructor  TmeshPackgenTraverser , so if it need to remain same thruout the class, we dont need it here, which is my understanding.

tst-mswartz: How else do you propose passing it to Generate() if not through the class that calls Generate()?  TmeshPackgenTraverser is the middleman between the command line and this point in the code.

earth_enterprise/src/fusion/rasterfuse/gepackgen.cpp
```diff
@@ -63,6 +66,7 @@ int main(int argc, char *argv[]) {
   options.flagOpt("?", help);
   options.opt("config", configFile,  khGetopt::FileExists);
   options.opt("output", output);
+  options.opt("decimation", decimation_threshold);
```
asmitaw: would be good to have some range validator like we have for numcpus.

earth_enterprise/src/fusion/rasterfuse/gepackgen.cpp
```diff
@@ -85,6 +89,9 @@ int main(int argc, char *argv[]) {
   if (!imagery && !terrain) {
     usage(progname, "--imagery or --terrain is required");
   }
+  if (imagery && (decimation_threshold != 0.009)) {
```
asmitaw: if (imagery && (decimation_threshold != defaultDecimationThreshold ))

tst-mswartz: This one I agree with, I forgot to change this when i added defaultDecimationThreshold a couple hours ago

earth_enterprise/src/fusion/rasterfuse/gepackgen.cpp
tst-mswartz: Can't until we decide on what a valid range might be . Lee discussed making that into a separate task.

earth_enterprise/src/server/gehttpd_init
```diff
@@ -95,6 +95,31 @@ if grep -q -s 'Ubuntu' /etc/lsb-release ; then
       RETVAL=$?
     }
 
+elif grep -q -s 'LinuxMint' /etc/lsb-release ; then
```
tst-lsavoie: If this is a copy of the Ubuntu block, I'd recommend having a single block and checking for both distributions in the previous block.

earth_enterprise/src/fusion/rasterfuse/gepackgen.cpp
tst-lsavoie: This is really picky, but it is possible for someone to specify `--imagery --decimation_threshold=0.009` and avoid this validation check.  You could set the decimation threshold to -1 by default and only set it to 0.009 after doing this check.  I'll leave that up to you - I'm happy to approve this either way.

earth_enterprise/src/server/gehttpd_init
tst-mswartz: Ah yeah good point, I will fix it

earth_enterprise/src/fusion/rasterfuse/gepackgen.cpp
tst-mswartz: Well, it's a valid point, although having decimation set while generating imagery packets doesn't seem to have any effect .  I skimmed khGetOpt.h earlier for other validation possibilities and I can do it like this:  options.setExclusive("imagery", "decimation")

tst-mswartz: Although now I sort of want to do that to some of the other exclusive option pairs .. I tested and it seems useful

tst-lsavoie: As I said, it's up to you, though the setExclusive thing does seem handy.

tst-mswartz: sample output from using setExclusive():

$ ./gepackgen --imagery --decimation 0.1234
Error: The following options are exclusive:
	--decimation
	--imagery

usage: ./gepackgen [options] {--imagery|--terrain} --config <configfile> -o <output>
   Supported options are:
      --help | -?:     Display this usage message
      --numcpus <num>: Num CPUs to use (default 1)
      --decimation <num>: error threshold for terrain decimation (default 0.009000)

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
tst-ccamp: I'm not sure that issue applies to 0.0. It's pretty much zero or not. Since this value doesn't come from any sort of calculation, but from direct user input, I don't think we have to worry about rounding error. If someone types 0.00000000000001 into the command line, then I would assume they meant for it to be non-zero and should trigger the decimation code.

tst-lsavoie:
Progress update: I've reviewed the code and it looks fine to me.  I'm working on testing it now but probably won't finish until tomorrow.

earth_enterprise/src/fusion/rasterfuse/TmeshPackgenTraverser.cpp
tst-nfarah: Yes it was very minor. We can disregard this.

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
```diff
@@ -74,14 +77,14 @@ TmeshGenerator::Generate(float *srcSamples,
...
-  // jdjohnso - The allowable error on border tiles is 1/10 that of interior
+  // Question - Did these numbers come from empirical testing?
+  // The allowable error gets smaller as the resolution increases.
+  // The allowable error on border tiles is 1/10 that of interior
   // tiles. This is because we really want the border tiles to match. If not
   // we get rips between mesh tiles in the client.
-  int error_level = tmeshAddr.level+1;
-  double error = 0.009 * pow(2, 11-error_level);
-  double borderError = error / 10.0;
+  double threshold_modifier = pow(2, 11 - (tmeshAddr.level + 1));
```
tst-nfarah: If "(tmeshAddr.level + 1)" computes an 'error_level', I would restore the variable 'error_level' as it was. This way we would follow how 'thereshold_modifier' is being calculated using an 'error_level' instead of having one complex equation not easy to follow.

tst-lsavoie:
I've reviewed the code and tested the changes and everything looks good.  I built the same database with decimation at 0.09 and decimation at 0 and I can clearly see the terrain changing as I zoom in and out on the 0.09 version but not the 0 version.

FYI, I cheated a bit in testing.  I modified earth_enterprise/src/fusion/autoingest/plugins.src/PacketLevel.src so that the fusion UI would pass a hard coded decimation value to gepackgen.  That allowed me to test all of the code changes in this PR without having to call gepackgen myself a bunch of times.

earth_enterprise/src/fusion/rasterfuse/TmeshGenerator.cpp
tst-mswartz: I disagree.  I am already using a separate variable to hold threshold_modifier, it seems very easy to understand.  The previous way, with error = 0.009 * pow(2, 11-error_level)   seems much more complicated.

tst-mswartz: I mean at that point you have to wonder why 2, why 11, why divide by 10.0

tst-nfarah: Yes it's already convoluted, that is why i suggested to make it less by restoring 'error_level'. It will not explain still the equation but it will at least keep the history that 'tmeshAddr.level+1' is some kind level of an error.

Issue #501, opened 2017-09-13
Update Website Links with Release 5.2.0


tst-awankhede:
Link to the release: https://github.com/google/earthenterprise/releases/tag/5.2.0-3.final
Link to the release notes: http://www.opengee.org/geedocs/answer/7160000.html
Link to the current docs for the release: http://www.opengee.org/geedocs/
AC:
Add a link to the release from opengee.org
Add a link to the release from the github README.md
Add a link to the docs for the release from the github README.md
Also add a link to opengee.org from the README.md page.

tst-lsavoie:
Closed via PR #503.

PR #503, opened 2017-09-14
Update Website Links with Release 5.2.0 #501

tst-awankhede:
For the Website changes to be tested, I am in process of setting up Jekyll.

For README.md , I have tested  in https://dillinger.io/
screenshot from 2017-09-13 17-30-40

### DIFF
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -8,6 +8,8 @@
       path: index.html#usecases
     - name: Features
       path: index.html#opengeefeatures
+    - name: Release 5.2.0
+      path: https://github.com/google/earthenterprise/releases/tag/5.2.0-3.final
 - name: Documentation
   path: geedocs
   secion_id: geedocs
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -8,8 +8,6 @@
       path: index.html#usecases
     - name: Features
       path: index.html#opengeefeatures
-    - name: Release 5.2.0
-      path: https://github.com/google/earthenterprise/releases/tag/5.2.0-3.final
 - name: Documentation
   path: geedocs
   secion_id: geedocs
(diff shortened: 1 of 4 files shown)


docs/_data/nav.yml
```diff
@@ -8,6 +8,8 @@
       path: index.html#usecases
     - name: Features
       path: index.html#opengeefeatures
+    - name: Release 5.2.0
+      path: https://github.com/google/earthenterprise/releases/tag/5.2.0-3.final
```
tst-lsavoie: This link doesn't work - it goes to https://tst-lsavoie.github.io/earthenterprise/https://github.com/google/earthenterprise/releases/tag/5.2.0-3.final.  There is probably some way to indicate that it is an absolute URL, not a relative one.

tst-lsavoie: By the way, I pushed these changes up to my fork, so you can see them at https://tst-lsavoie.github.io/earthenterprise.

tst-awankhede:
I had to do some testing on github pages so there are testing related commits can be deleted if thats annoying. the present changes links are as per screen shot, see the "Releases" link at the top menu.

here is screenshot of changes, see the releases link at the top menu before github cat.

screenshot from 2017-09-14 18-06-35

Also, I tried initially to add it to child in the fist menu - along with "why Open GEE and rest ", it would not take the absolute/ external links even with the following change in default.html

  {% for child in item.children %}
                        <li>
                          {% if child.path contains "http" %}
                            <a href="{{ child.path }}">{{ child.name }}</a>
			  {% else %}
                            <a href="{{ site.baseurl}}/{{ child.path }}">{{ child.name }}</a>
                          {% else %}
                        </li>

tst-ccamp:
I see two approvals, so I'm going to merge

PR #517, opened 2017-09-25
Add QT Creator files and tutorial files to .gitignore

tst-lsavoie:
Includes some updates to the .gitignore file to ignore QT Creator files and Tutorial files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,8 @@ earth_enterprise/tutorial/FusionTutorial/KML
 .settings
 
 # QT Creator project files
-earthenterprise.config
-earthenterprise.creator
-earthenterprise.creator.user
-earthenterprise.files
-earthenterprise.includes
+/*.config
+/*.creator
+/*.creator.user
+/*.files
+/*.includes
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,22 @@
 earth_enterprise/.sconsign.dblite
 earth_enterprise/tutorial/.sconsign.dblite
+
+# Tutorial files
 earth_enterprise/tutorial/FusionTutorial-Full.tar.gz
+earth_enterprise/tutorial/FusionTutorial/Imagery
+earth_enterprise/tutorial/FusionTutorial/Vector
+earth_enterprise/tutorial/FusionTutorial/Terrain
+earth_enterprise/tutorial/FusionTutorial/KML
 
 # Eclipse project files
 .cproject
 .project
 .pydevproject
 .settings
+
+# QT Creator project files
+earthenterprise.config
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,9 +1,22 @@
...
+earth_enterprise/tutorial/FusionTutorial/KML
 
 # Eclipse project files
 .cproject
 .project
 .pydevproject
 .settings
+
+# QT Creator project files
+earthenterprise.config
```
tst-ccamp: This filename is dependent on the project name, right? Would it be better to use a wildcard here (*.config, *.creator, etc.) in case someone decides to name their project different?

tst-lsavoie: Yes, you're right.  I'll make the change.

In my defense, I really created this so that Travis CI would have something to build, but I am hoping to get it checked in eventually.

tst-lsavoie: Looking into it a bit more, there are multiple files that end in .config or .files, so I can't simply ignore *.config and *.files.  For that we may just have to assume everyone will use earthenterprise as the project name.

tst-ccamp: Unless you specify "**/*.config" it will only match files in the same directory as the .gitignore file. Since there are no files with the .config extension in this directory, I think you're safe to do so. And for the record, I called my project OpenGEE instead of earthenterprise. :)

tst-lsavoie: In mine it's ignoring files from the location of the .gitignore down.  I could ignore these files in a specific directory, but then we're assuming that everyone is putting their QT Creator project at the same point in the directory structure, which is also not necessarily the case.  I suppose I could put them in several .gitignore files in the most likely places.

tst-ccamp: According to [this table](https://www.atlassian.com/git/tutorials/gitignore#git-ignore-patterns), it shouldn't ignore files in subdirectories. I wonder why you're getting different behavior. Maybe it's a git config setting.

tst-ccamp: Oh, looks like I read it wrong. You could do ./*.config I believe.

tst-lsavoie: Yeah, but that assumes we know where people will put their project files.  I guess I could put the .gitignore in earthenterprise, earth_enterprise, and src, which covers all the likely candidates.

tst-ccamp: You have to assume you know where they're putting it or you know what they're calling it :) But, that seems like a good compromise. 

Or we can just instruct devs to do a custom exclude for their particular project name: https://www.atlassian.com/git/tutorials/gitignore#personal-git-ignore-rules

tst-ppenev:
I don't use QT Creator, but this looks OK.

PR #518, opened 2017-09-25
I247 Default Imagery and Terrain file filters ignore tiff

### LINKED ISSUE #247: Default Imagery and Terrain file filters ignore files with the ".tiff" extension
When adding a source file that may be in Tiff format, the file dialog will display files with a ".tif" extension, but not files with a ".tiff" extension.

Expected: Files with either ".tif" or ".tiff" will be displayed.

tst-eclamar:
Fixes #247


tst-eclamar:
Note: this PR is just copying Asmita's PR #495

tst-lsavoie:
We also need documentation changes along with this - anywhere in the documentation where it lists supported file types it should also list .tiff/.TIFF.

PR #522, opened 2017-09-26
Added travis_wait to prevent timeout on Travis CI builds

tst-ccamp:
Splits build into 2 jobs (third_party & fusion build/tests)

This ensures that the build doesn't hit the 50 minute/job timeout. Total build time doesn't have a timeout value.

Test run: https://travis-ci.org/google/earthenterprise/builds/282885682


tst-lsavoie:
It timed out with this message at the bottom of the log before it got to the unit tests:

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

So we are hitting some maximum time limit.  Still, I think this change is useful.  I went ahead and restarted the build to see if we can get it to finish this time.

tst-lsavoie:
It passed on the second run.  Maybe we need to add this to the build step as well but have it run for something like an hour.

tst-ccamp:
I don't think that will work since the timeout happened while it was running. The only thing the `travis_wait` command does is output a line of text every minute to avoid the 10 minute "no output" timeout. That's clearly not what's causing our timeout, so I guess it's back to the drawing board.

tst-lsavoie:
I was thinking it's possible the build is not generating output for 10 minutes, though that seems unlikely.  In any case, I think this change is still valuable.

tst-ccamp:
I'm doing a little experiment with build stages that might help out

tst-ccamp:
The latest test run has caching completely working and the build is split into 2 jobs to ensure that it doesn't hit the 50 minute/job timeout (total build time doesn't have a timeout value). 

https://travis-ci.org/google/earthenterprise/builds/282885682

PR #542, opened 2017-10-12
Fix out of bounds error in box filter

### LINKED ISSUE #515: gepolymaskgen segfaults with specific inputs
I get a segmentation fault when running gepolymaskgen. 

I'm trying to create a custom mask like the one shown in [this](http://www.opengee.org/geedocs/answer/4525555.html#h.frsdlq-dl14oj) tutorial. The coastline mask that I use is a full-resolution world coastline shapefile named GSHHS_f_L1.shp located on the Data/ folder. The 44-tile mosaic of .jp2 files is located on the Imagery/ folder. The compressed input files (mosaic .khvr and world coastlines file) are located in [this](https://drive.google.com/open?id=0B1_3KrIKN1M1aHM5OWtzRmd5ZWc) Google Drive link. I'm running the command on a machine with 24 cores and 64 GB of RAM, since when I tried this earlier on a machine with 8 GB I was getting out of memory bad_alloc errors. However, after moving to the more powerful machine, this is my output:

```
user@gee:/gevol/src/Imagery/San_Diego_2014_NAIP_JP2$ time /opt/google/bin/gepolymaskgen --base_image san_diego_2014_mosaic_naip_jpg2000.khvr --invert --feather -300 --feather_border 1 --and_neg_mask /gevol/src/Data/gshhg-shp-2.3.6/GSHHS_shp/f/GSHHS_f_L1.shp  --invert --output_mask san_diego_2014_naip-custom-mask.tif
Fusion Notice:  Base image: san_diego_2014_mosaic_naip_jpg2000.khvr
Fusion Notice:  san_diego_2014_mosaic_naip_jpg2000.khvr width: 55666 height: 57853
Fusion Notice:  north: 3.878727e+06 south: 3.820874e+06 east: -1.300350e+07 west: -1.305917e+07
Fusion Notice:  File type: KHM/Keyhole Mosaic
Fusion Notice:  Projection : 'PROJCS["WGS_1984_Web_Mercator_Auxiliary_Sphere",GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137.0,298.257223563]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.017453292519943295]],PROJECTION["Mercator_Auxiliary_Sphere"],PARAMETER["False_Easting",0.0],PARAMETER["False_Northing",0.0],PARAMETER["Central_Meridian",0.0],PARAMETER["Standard_Parallel_1",0.0],PARAMETER["Auxiliary_Sphere_Type",0.0],UNIT["Meter",1.0],EXTENSION["PROJ4","+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext  +no_defs"]]'
Fusion Notice:  Invert
Fusion Notice:  Feather -300
Fusion Notice:  Feather border 1
Fusion Notice:  AND mask: /gevol/src/Data/gshhg-shp-2.3.6/GSHHS_shp/f/GSHHS_f_L1.shp
Fusion Notice:  Invert
Fusion Notice:  Output mask: san_diego_2014_naip-custom-mask.tif
Fusion Notice:  Setting base image san_diego_2014_mosaic_naip_jpg2000.khvr.
Fusion Notice:  Inverting current mask.
Fusion Notice:  Setting feather -300.
Fusion Notice:  Feather border is on.
Fusion Notice:  Writing alpha file san_diego_2014_naip-custom-mask.tif
Fusion Notice:  Executing: gdal_rasterize -b 1 -burn 255 -l GSHHS_f_L1 /gevol/src/Data/gshhg-shp-2.3.6/GSHHS_shp/f/GSHHS_f_L1.shp san_diego_2014_naip-custom-mask.tif
0...10...20...30...40...50...60...70...80...90...100 - done.
Total tiles to process: 1
Segmentation fault (core dumped)

real    10m5.050s
user    6m58.480s
sys     0m26.832s
```

tst-lsavoie:
Fixes #515.

The box filter uses a summed area table (SAT) in which each value in a matrix is based on values above and to the left of that value. However, there is no code to handle the left of the matrix (which has no values to the left).  This PR adds code to handle the left column which prevents writing outside the bounds of the matrix, which in turn prevents a seg fault in some cases.  There is no need to handle this for the top of the matrix because the code includes an optimization that wraps around to the bottom of the matrix and thus doesn't write out of bounds.

To test this change:
- Download the input data from https://drive.google.com/open?id=0B1_3KrIKN1M1aHM5OWtzRmd5ZWc
- Unzip the file and move the Data and Imagery folders to /gevol/src.
- Run the command `time /opt/google/bin/gepolymaskgen --base_image /gevol/src/Imagery/San_Diego_2014_NAIP_JP2/m_3211609_nw_11_1_20140530_20140721.jp2 --invert --feather -300 --feather_border 1 --and_neg_mask /gevol/src/Data/gshhg-shp-2.3.6/GSHHS_shp/f/GSHHS_f_L1.shp  --invert --output_mask san_diego_2014_naip-custom-mask.tif`.  If you run this on the master branch, it will seg fault after a few minutes.  With the changes in this PR, it will not seg fault.

Lots of the changes in this PR are just whitespace.  To see the PR with whitespace ignored, add `?w=1` to the end of the URL: https://github.com/google/earthenterprise/pull/542/files?w=1.

### DIFF
--- a/earth_enterprise/src/fusion/tools/boxfilter.cpp
+++ b/earth_enterprise/src/fusion/tools/boxfilter.cpp
@@ -104,14 +104,26 @@ const uchar *BoxFilterTiledImage::GetImageTile(int tile_x, int tile_y,
   }
 }
 
+// Computes the left SAT element of a given row and advances the pointer
+inline void ComputeLeftSAT(int **elem,  // pointer to element in SAT table
+                           uchar value,  // corresponding image value
+                           int prev_row_offset) {  // offset to previous row elem
+  **elem =
+      + *(*elem + prev_row_offset)
+      + value;
+  ++(*elem);
+}
 
 // Computes a single SAT element given current values and an offset to
 // the previous row of values in the table.
 inline void ComputeSAT(int *elem,  // pointer to element in SAT table
                        uchar value,  // corresponding image value
                        int prev_row_offset) {  // offset to previous row elem
-  *elem = - *(elem + prev_row_offset - 1) + *(elem + prev_row_offset)
-          + *(elem - 1)                   + value;
+  *elem =
+      - *(elem + prev_row_offset - 1)
+      + *(elem + prev_row_offset)
+      + *(elem - 1)
+      + value;
 }
 
 // Computes a horizontal span of entries in the SAT with a single input
@@ -290,8 +302,10 @@ void BoxFilterTiledImage::BoxFilter(int box_width, int box_height,
           // (box_halfw+1 tile padding on left, box_halfw inside image
           // tile).
           if (tile_col == 0) {
-            if (tile_row == 0) {  // Compute top tile row.
-              ComputeSATSpanValue(&psat, box_halfw + 1, border,
+            if (tile_row == 0) {
+              // Compute top tile row.
+              ComputeLeftSAT(&psat, border, sat_prev_row_offset);
+              ComputeSATSpanValue(&psat, box_halfw, border,
                                    sat_prev_row_offset);
               ComputeSATSpanValues(&psat, box_halfw, &pimg,
                                    sat_prev_row_offset);
@@ -325,7 +339,8 @@ void BoxFilterTiledImage::BoxFilter(int box_width, int box_height,
           // Handle case where row >= box_height.
           // Take care of the first box_width elements of SAT row.
           if (tile_col == 0) {  // Leftmost tile column: compute.
-            ComputeSATSpanValue(&psat, box_halfw + 1, border,
+            ComputeLeftSAT(&psat, border, sat_prev_row_offset);
+            ComputeSATSpanValue(&psat, box_halfw, border,
                                 sat_prev_row_offset);
             ComputeSATSpanValues(&psat, box_halfw, &pimg, sat_prev_row_offset);
           } else {  // Copy box_width elements from vertical band.
--- a/earth_enterprise/src/fusion/tools/boxfilter.cpp
+++ b/earth_enterprise/src/fusion/tools/boxfilter.cpp
@@ -108,10 +108,9 @@ const uchar *BoxFilterTiledImage::GetImageTile(int tile_x, int tile_y,
 // don't have a similar function for the top row of the SAT because the
 // calculation will just wrap around the horizontal band which is set to 0,
 // so the extra elements will have no effect.
-static inline
-void ComputeLeftSAT(int **elem,  // pointer to element in SAT table
-                    uchar value,  // corresponding image value
-                    int prev_row_offset) {  // offset to previous row elem
+inline void ComputeLeftSAT(int **elem,  // pointer to element in SAT table
+                           uchar value,  // corresponding image value
+                           int prev_row_offset) {  // offset to previous row elem
   **elem =
       + *(*elem + prev_row_offset)
       + value;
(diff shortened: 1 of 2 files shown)


tst-lsavoie:
@asmitaw your comment somehow got disconnected from the code, but the point is that ComputeLeftSAT computes a SAT entry and also advances the pointer to the next entry.  That's why we pass a pointer to a pointer.  ComputeSATSpanValue and ComputeSATSpanValues both do the same things.  That's why it looks like I'm passing the same pointer to all of them, but really each of them changes the pointer.

earth_enterprise/src/fusion/tools/boxfilter.cpp
```diff
@@ -104,39 +104,82 @@ const uchar *BoxFilterTiledImage::GetImageTile(int tile_x, int tile_y,
...
+  ComputeSATSpanValues(psat, box_halfw, pimg, prev_row_offset);
+}
+
 
-// Main function to do the box filtering using a Summed Area Table.
+// Main function to do box filtering.  It is used, for example, when feathering
+// a mask - in that case, each output pixel is the average of all the input
+// pixels within a box centered on the pixel being calculated.
+//
+// This function is heavily optimized.  It uses a Summed Area Table (SAT) to
```
tst-nfarah: can we reference SAT background info  ? may be: https://en.wikipedia.org/wiki/Summed-area_table or some thing similar ?

earth_enterprise/src/fusion/tools/boxfilter.cpp
```diff
@@ -104,39 +104,82 @@ const uchar *BoxFilterTiledImage::GetImageTile(int tile_x, int tile_y,
...
 
 // Computes a single SAT element given current values and an offset to
 // the previous row of values in the table.
-inline void ComputeSAT(int *elem,  // pointer to element in SAT table
-                       uchar value,  // corresponding image value
-                       int prev_row_offset) {  // offset to previous row elem
-  *elem = - *(elem + prev_row_offset - 1) + *(elem + prev_row_offset)
-          + *(elem - 1)                   + value;
+static inline
+void ComputeSAT(int *elem,  // pointer to element in SAT table
```
tst-nfarah: can we unit test this function/method 'ComputeSAT' with positive cases and negative ones to simulate going outside the matrix bounds ?

earth_enterprise/src/fusion/tools/boxfilter.cpp
```diff
@@ -104,39 +104,82 @@ const uchar *BoxFilterTiledImage::GetImageTile(int tile_x, int tile_y,
...
+// calculation will just wrap around the horizontal band which is set to 0,
+// so the extra elements will have no effect.
+static inline
+void ComputeLeftSAT(int **elem,  // pointer to element in SAT table
+                    uchar value,  // corresponding image value
+                    int prev_row_offset) {  // offset to previous row elem
+  **elem =
+      + *(*elem + prev_row_offset)
+      + value;
+  ++(*elem);
```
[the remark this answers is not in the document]
tst-lsavoie: If this is called correctly it will never go outside the allocated memory.  I patterned it after the existing functions ComputeSATSpanValue(s).

earth_enterprise/src/fusion/tools/boxfilter.cpp
tst-lsavoie: I could write some positive test cases, but this function doesn't handle going outside the matrix bounds, so I'm not sure what the point of negative test cases would be.

earth_enterprise/src/fusion/tools/boxfilter.cpp
tst-lsavoie: Done.

earth_enterprise/src/fusion/tools/boxfilter.cpp
tst-lsavoie: My thinking here is that these functions are not part of the interface of this module.  Since they are declared inline, they cannot be called from other files.  Since they're not part of the interface, I decided they are covered sufficiently by the existing unit tests.  In any case, it's impossible to unit test these functions without removing the inline modifier or refactoring some other way, which I think is not justified in this case.

earth_enterprise/src/fusion/tools/boxfilter.cpp
```diff
@@ -135,8 +147,32 @@ inline void ComputeSATSpanValues(int **psat,  // ptr to first SAT element
...
+// computed from the image.
+inline void ComputeLeftSATSpan(int **psat,  // ptr to the row in the SAT
+                               uchar border,  // border vlaue
+                               int prev_row_offset,  // offset to the prev row
+                               int box_halfw,  // (width of box -1) / 2
+                               const uchar **pimg) {  // ptr to image row
+  // Compute the far left value
+  ComputeLeftSAT(psat, border, prev_row_offset);
+  // Compute the padding outside the image
+  ComputeSATSpanValue(psat, box_halfw, border, prev_row_offset);
```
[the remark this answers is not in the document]
tst-lsavoie: Yes, ComputeLeftSAT now computes the first element in the row and ComputeSATSpanValue computes the others, so it computes one less than before.

tst-ccamp:
@tst-lsavoie The code looks good to me. Have you considered how we might augment our tests for cases like this? Should we add this data to our regression tests?

tst-lsavoie:
@tst-ccamp I think the best solution is to run a lot of test cases under valgrind or address sanitizer or something like that to detect buffer overflows and memory leaks.  I don't think we're quite ready for that at this point because there are likely other memory errors that need to be fixed before those tests will pass.  I think that's best do that as a separate task.

Issue #540, opened 2017-10-10
Write script to clone code


tst-basilhuffman:
As a corollary to PR #475 a script should be written to clone code. It should handle both the case of cloning the main repo as well as a user-forked repo.

tst-ppenev:
:stuck_out_tongue: There, actually, is such a script among the Docker scripts, too.

tst-basilhuffman:
Is this pull-gee-repository.sh ?

tst-ppenev:
Yes, it is.  Although, this one doesn't contain calculating the repository URL from a Git Hub username.  We could make it a bit more user-friendly, and get it out of the Docker scripts directory.

tst-basilhuffman:
I can do this when time comes.

PR #553, opened 2017-10-16
script to clone code

tst-basilhuffman:
re-write of a similar script of pavel's in the docker build, but a bit
more user friendly. low priority. addresses #540. might be useful
for new folks

### DIFF
--- a/scripts/pull-gee-repository.sh
+++ b/scripts/pull-gee-repository.sh
@@ -1,5 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
+# Copyright 2017 OpenGee Contributors
 
 #===============================================================================
 #
--- a/scripts/pull-gee-repository.sh
+++ b/scripts/pull-gee-repository.sh
@@ -40,9 +40,7 @@ IS_SSH=1
 IS_HTTPS=2
 IS_MAIN=3
 IS_FORK=4
-SSH_DIR=".ssh"
-SSH_KEY_PRIV="id_rsa"
-SSH_KEY_PUB="$SSH_KEY_PRIV.pub"
+
 CLONE_REPO=$IS_MAIN
 CLONE_TYPE=""
 
@@ -75,7 +73,6 @@ function get_type ()
     return $IS_MAIN
 }
 
-
 #===  FUNCTION  ================================================================
 #          NAME:  clone_ssh
 #   DESCRIPTION:  will cone the repo over ssh
@@ -89,17 +86,11 @@ function clone_ssh ()
         echo "Error: no USER_NAME given!"
         exit 1
     fi
-    WD="$(pwd)"
-    cd
-    if [ ! -f "$SSH_DIR/$SSH_KEY_PRIV" ] || [ ! -f "$SSH_DIR/$SSH_KEY_PUB" ]
-    then
-        echo "SSH Key(s) missing!"
-        exit 1
-    fi
-    cd $WD
+
     echo "Attempting to clone fork $USER_NAME..."
     echo "git clone git@github.com:$USER_NAME/earthenterprise.git $DIR_NAME"
     git clone git@github.com:$USER_NAME/earthenterprise.git $DIR_NAME
+    echo $?
 }
 
 #===  FUNCTION  ================================================================
--- a/scripts/pull-gee-repository.sh
+++ b/scripts/pull-gee-repository.sh
@@ -32,7 +32,9 @@
 #===============================================================================
 #       Constants  
 #===============================================================================
+THIS=$(basename "$0")
 USER_NAME=""
+PASSWORD=""
 DIR_NAME="earthenterprise"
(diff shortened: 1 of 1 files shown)


tst-basilhuffman:
refactored to allow for interactive and non interactive mode

to test:

1) ./pull-gee-repository.sh
        clones main repo into "earthenterprise"

2) ./pull-gee-repository.sh -i
        runs script in user-interactive mode

3) ./pull-gee-repository.sh -d <DIR>
        clones main repo into "<DIR>"

4) ./pull-gee-repository.sh -f <USERNAME>
        clones fork <USERNAME> via ssh. can combine with -d

5) ./pull-gee-repository.sh -f <USERNAME> -p <PASSWORD>
        clones fork <USERNAME> via https with password <PASSWORD>

6) ./pull-gee-repository.sh -p <PASSWORD>
        FAILS due to not have a username i.e. fork

7) ./pull-gee-repository.sh -d
        FAILS due to not specifying a directory

8) ./pull-gee-repository.sh -h
        displays help info

scripts/pull-gee-repository.sh
```diff
@@ -0,0 +1,370 @@
...
+
+#===  FUNCTION  ================================================================
+#          NAME:  get_type 
+#   DESCRIPTION:  returns the repo to clone, main or fork 
+#    PARAMETERS:  n/a
+#       RETURNS:  IS_FORK, if fork; IS_MAIN, if main
+#===============================================================================
+function get_type ()
+{
+    printf "Do you want to clone the (m)ain repo or a (f)ork (default is main)? "
```
tst-jblair: This reads to me like "main" would be a valid input here... but the check a couple of lines down will *only* match if someone *exactly* enters an "m".  I'm not terribly proficient with bash, is there a smoother way to do this?

scripts/pull-gee-repository.sh
```diff
@@ -0,0 +1,370 @@
...
+THIS=$(basename "$0")
+USER_NAME=""
+PASSWORD=""
+DIR_NAME="earthenterprise"
+IS_SSH=1
+IS_HTTPS=2
+IS_MAIN=3
+IS_FORK=4
+SSH_DIR=".ssh"
+SSH_KEY_PRIV="id_rsa"
```
tst-jblair: While I'd believe that RSA is the dominant key type choice, it's not the only choice... I'd expect ECDSA to be a non-zero proportion of the user base?

scripts/pull-gee-repository.sh
```diff
@@ -0,0 +1,370 @@
...
+    CLONE_REPO=$?
+
+    if is_cloned
+    then
+        echo "$DIR_NAME is already cloned!"
+        exit 1 
+    fi
+    
+    # if cloning main, just use standard https
+    if [ $CLONE_REPO -eq $IS_MAIN ]
```
tst-jblair: I'm curious why "cloning main" ==> "want to use https"?  Personally, I only ever want to use ssh, to save on the number of passwords I need to keep up with.

Also, why does the path a little lower go through get_password, but this one doesn't?

tst-basilhuffman: This was talked about it slack, current build instructions list as ssh but it was figured that https would be best if you are only cloning the main (google/earthenterprise) repo.  Just covering the case where the user does not have a github account and just wants to pull down the repo.  Thus, no password is required.

Cloning a fork is a different case

scripts/pull-gee-repository.sh
tst-basilhuffman: I suppose that the most appropriate validation would be to check that ~/.ssh is present and contains files?

scripts/pull-gee-repository.sh
tst-jblair: Ah, interesting - I hadn't recalled that "anonymous https" was an option ;).

scripts/pull-gee-repository.sh
tst-ppenev: I don't see the benefit of checking what SSH keys a user has.  I think it's sufficient to use an SSH Git URL, and let `git` handle the rest.

tst-jblair: The specific keys aren't being checked at all; this section is really aimed at seeing *if* the user has keys. My comment really came down to that "id_rsa" is too specific.

tst-basilhuffman: Thats fine, I just figured the standard git error message here is a bit cryptic, doesn't really explain much at first glance

tst-ppenev: I know. And I don't see how checking if a user has keys gives us anything. It seems to be used only to replace the error message that `git` would give you, if you tried to clone over SSH, and Git couldn't find your keys.

tst-basilhuffman: Fair enough. If you think thats sufficient than I can take it out

PR #554, opened 2017-10-16
Build RPM packages for libraries common to Fusion and Server

tst-ppenev:
* Added Gradle build tasks for building Post GIS and `common` RPMs (`openGeePostGisRpm`, and `openGeeCommonRpm`).

The common RPM is not necessarily only what is shared between Fusion and Server, but trying to modify all the `scons` targets that build `common`, `fusion` and `server` would require significantly greater amount of time.  When more packages are unbundled from Open GEE in the future, that effort will decrease.  Therefore, that clean-up is left for the future.

To build the RPMs:

1. cd to earth_enterprise/rpms
1. Run `./gradlew openGeeRpms`


earth_enterprise/rpms/build.gradle
```diff
@@ -9,123 +9,265 @@ buildscript {
...
+    summary = 'Post GIS extension for Postgre SQL server used by Open GEE'
+    packageDescription = 'Post GIS extension for Postgre SQL server used by Open GEE.'
+    license = 'GPL 2'
+    packager = 'gee-oss@googlegroups.com'
+    url = 'https://github.com/google/earthenterprise/tree/master/earth_enterprise/third_party/postgis'
+    arch = 'X86_64'
+    os = 'LINUX'
+    type = 'BINARY'
+
+    from (postGisInstallDir_opt) {
```
tst-lsavoie: Any reason PostGIS isn't included in the same RPM as the other 3rd party packages.

tst-jcupp:
Is the `.gradle` folder and its files necessary?

earth_enterprise/rpms/build.gradle
```diff
@@ -9,123 +9,265 @@ buildscript {
...
-        into package_install_path
+    from(stagedInstallDir_fusion) {
+        into packageInstallRootDir
     }
 }
 
-task packageRpms(dependsOn: ['serverRpm','fusionRpm','commonRpm'])
-task packageDebs(dependsOn: ['serverDeb','fusionDeb','commonDeb'])
-task osPackage(dependsOn: ['packageRpms','packageDebs'])
+task opengeeRpms(
```
tst-lsavoie: Not sure if this matters but there's a capitalization inconsistency here - the other targets start with openGee, not opengee.  Same with opengeeDebs below.

tst-lsavoie:
I'm getting an error when installing the common RPM:

```
Running Transaction
  Installing : opengee-common-5.2.0-1.x86_64                                1/1 
Error unpacking rpm package opengee-common-5.2.0-1.x86_64
error: unpacking of archive failed on file ./var/opt/google/openssl/misc: cpio: Archive file not in header
  Verifying  : opengee-common-5.2.0-1.x86_64                                1/1 

Failed:
  opengee-common.x86_64 0:5.2.0-1  
```

tst-lsavoie:
I'm going to merge this to the rhel6-rpms branch to simplify future development and fixing the install problems.

earth_enterprise/rpms/build.gradle
tst-ppenev: It has a GPL license, which usually implies that derived works have to be licensed under GPL, as well.  We're packaging separately, so as not to produce a package that's a derived work.

tst-ppenev:
@tst-jcupp: Hm. `.gradle` should probably not be in the repository. I was having problems running the wrapper that came with @tst-dsheremata's clone, so I ended up including everything.

We should try removing it, and making sure it works when cloned by developers.

PR #559, opened 2017-10-17
WIP - Rpm experiments with symlinks

husf-dsheremata:
The symlink code that I have in here now seems to fix the CPIO issues - but it needs more cleaning up (i.e., a single call to a function instead of the cut-copy-paste code)


earth_enterprise/rpms/build.gradle
```diff
@@ -5,127 +5,313 @@
...
+import com.netflix.gradle.plugins.utils.JavaNIOUtils
+
 buildscript {
     repositories {
         jcenter()
     }
+
     dependencies {
-        classpath "com.netflix.nebula:gradle-ospackage-plugin:4.4.2"
+        classpath 'com.netflix.nebula:gradle-ospackage-plugin:3.6.0'
```
tst-ppenev: This seems to be what fixes the RPM for Cent OS 6.  The symlinks don't seem to make a differenc.

tst-dsugar: In a way this makes sense to me...I am guessing there is something incompatible in cpio/rpm support changes between what is in older and newer rpms...does this then still work for centos7/rhel7 though?

tst-ccamp: We should probably file a bug against the plugin. You'd expect the newer version to work better :)

tst-ppenev: There, actually, already was such an issue opened for `ospackage`.  I added some comments with the versions that work and the ones that don't work with Cent OS 6.

tst-ppenev:
@tst-dsheremata: I've incorporated the changes from this pull request into #562.

PR #561, opened 2017-10-17
I560 Create RPM for GEE Common for RHEL 6 - initial ingest

tst-eclamar:
Scripts for pre-/post- install/uninstall of the common RPM.


tst-ppenev:
I think we can close this in favor of #562.

PR #562, opened 2017-10-18
RPM common packages optional scons and build fixes

tst-ppenev:
* Made compiling Open GEE as a dependency of the RPM build tasks optional. (Can be enabled from the command line.)

* Rolled back to a version of OS Package that produces a `common` RPM which installs on Cent OS 6.

* Removed the `.gradle` directory.


tst-dsheremata:
I've tested and it looks good.  Small fixes will be needed as part of the subsequent server and fusion RPM stories, as agreed to at scrum.  (These should also include the doc updates regarding build environment and documentation updates. )

earth_enterprise/rpms/build.gradle
```diff
@@ -1,17 +1,26 @@
-/*
-    See documentation on the osPackage plugin here: https://github.com/nebula-plugins/gradle-ospackage-plugin
+// Command-line parameters:
+//
+//     ./gradlew -PbuildOpenGee:
+// Compile and install Open GEE in a temporary directory before packaging.
+// By default no compilation, or installation are run.  In that case, you must
+// ensuce the main scons "stage_install" target has been called before building
```
tst-lsavoie: I'm pretty sure 'ensuce' is not a word.

earth_enterprise/rpms/build.gradle
```diff
@@ -1,17 +1,26 @@
-/*
-    See documentation on the osPackage plugin here: https://github.com/nebula-plugins/gradle-ospackage-plugin
+// Command-line parameters:
+//
+//     ./gradlew -PbuildOpenGee:
+// Compile and install Open GEE in a temporary directory before packaging.
+// By default no compilation, or installation are run.  In that case, you must
```
tst-lsavoie: This line is confusing to me.  The previous line says that this command compiles and installs Open GEE, and this line says that it doesn't.  I think maybe most of this paragraph is meant to apply to the case when you don't specify `-PbuildOpenGEE`, but that wasn't really clear to me.

tst-lsavoie:
If you just run `./gradlew` nothing happens.  We either need to document the task to run or add some default tasks.

earth_enterprise/rpms/opengee-common/post-install.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright 2017 Google Inc.
```
tst-ccamp: This is a new file, so shouldn't it be copyright "Open GEE Contributors"?

tst-lsavoie:
On the commits by other authors: @tst-dsheremata has approved this PR and @tst-eclamar has submitted his commits in a separate PR, so both are OK with their commits being included in this repo.

tst-lsavoie:
Also, I'm OK with my commits being included.

PR #563, opened 2017-10-19
** WIP ** #547: RHEL6 Fusion RPM

tst-ppenev:
**This pull request is a work-in-progress for review by other developers. It is not ready to be merged.**

* Merged @tst-eclamar's Fusion install scripts.

* Converted `gevars.sh` into a Groovy template expanding the Open GEE version.

* Added a key-value store that can be used to communicate values from one install script to another.

* Added an `installUtils` template that includes the expanded `gevars.sh`, and the key-value store library.

* Rewrote some `opengee-common` and `opengee-fusion` install scripts on the new infrastructure.


earth_enterprise/rpms/opengee-common/pre-uninstall.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
tst-dsugar: I wonder if this really should be a post-uninstall rather than a pre-uninstall.  Things still exist under the group ownership until the actual remove is executed...

tst-ppenev: Yeah. The only issue used to be, if the script was trying to access `/etc/init.d/gevars.sh` which wouldn't exist in the post-uninstall. However, I moved `gevars.sh` to the `common` package, so that shouldn't be a limitation here.

I have not fixed the Fusion install scripts, yet.  There also needs to be the `gefusionuser` creation and removal in them.

PR #564, opened 2017-10-19
** WIP ** Work in Progress Branch for RHEL6 RPMs

tst-ppenev:
**This pull request is for reference to developers.  It is not intended to be merged.**

This pull request collects commits from various developers, so that each individual developer doesn't have to collect the various commits.  New development can branch off of this branch, and it will be periodically merged in, so that other developers can benefit, even though the code is not, yet, ready for review and testing.


tst-ppenev:
I don't think there's a way for an RPM to 'use the "names" for these users and groups'.  The files and directories in an RPM are stored as a GZip'ped CPIO archive.  CPIO doesn't store user and group names for files, only their IDs, as I think most archives do.

tst-ppenev:
Work-in-progress no longer needed, since the branches that provide basic functionality have been merged in the upstream feature branch.

Issue #439, opened 2017-08-14
Uninstalling fusion without stopping it results in unexpected error message


tst-lsavoie:
If you don't stop the Fusion service and run `./uninstall_fusion.sh`, you get an error message starting with:

```
mkdir: missing operand
Try 'mkdir --help' for more information.
./uninstall_fusion.sh: line 472: $UNINSTALL_LOG: ambiguous redirect
```

Later the message informs you that you should stop fusion before uninstalling.  However, the spurious error messages should be eliminated to prevent confusion.

tst-lsavoie:
Closed via PR #565

PR #565, opened 2017-10-19
Fusion uninstall error messages

tst-mswartz:
#439  Uninstalling fusion without stopping fusion service results in unexpected error message


tst-basilhuffman:
Ubuntu 16.04:
Won't uninstall gefusion while running and ambiguous error message has been removed.  Will not install fusion while gefusion is running. Looks good.

Perhaps outside of scope:
geserver has status "Active (exited)" when running. I think this is a systemd vs sysv issue? 
I was able to run uninstall_server.sh while geserver was running.  When trying to run install_server.sh, I am prompted to stop init.d script, which is not present, preventing reinstallation.

tst-eclamar:
I have reviewed and tested this PR and it looks good.  I approve.

PR #567, opened 2017-10-23
WIP rpm-rhel6 mods for server file collisions and pre-install scripts. 

tst-dsheremata:
Work in progress - for now the only way I can get the duplicate files not to collide is to enumerate them explicitly.  Any thoughts?


tst-dsheremata:
OBE by PR 582.

PR #574, opened 2017-10-25
Enabling MrSid raster format if MrSID DSDK is installed on platform

tst-nfarah:
If MrSID DSDK (Decoding SDK) is installed on machine, GDAL will be built with MrSID raster format and Fusion would support such format.


earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -26,11 +26,28 @@
...
 # kakadu_version = 'v5_2_6-00418C'
 # geoexpress_version = 'MrSID_DSDK-8.5.0.3422'
+
+# Check if MrSID decoding library is installed and get the decoding SDK path
+mrSIDLibName = 'libltidsdk.so'
+mrSIDDSDKPath = ""
+findLibraryCommand = "/sbin/ldconfig -p | grep %s | tr ' ' '\n' | grep /" %mrSIDLibName
+mrSIDlibraryFullPath=os.popen(findLibraryCommand).read()
+if not mrSIDlibraryFullPath:
+   print("mrSID library was not found ! GDAL will be built without mrSID raster format support.")
```
tst-lsavoie: I don't know if this print statement is necessary.  Building without MrSID is the default, so I'm not sure if we want to notify people about it.

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -145,6 +162,9 @@ openjpeg_extract = gdal_env.Command(
 #             kakadu_version, kakadu_version, kakadu_version, gdal_target))])
 
 
+# [8]
```
tst-lsavoie: Probably ought to remove this since it's not used.

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -26,11 +26,28 @@
...
+mrSIDDSDKPath = ""
+findLibraryCommand = "/sbin/ldconfig -p | grep %s | tr ' ' '\n' | grep /" %mrSIDLibName
+mrSIDlibraryFullPath=os.popen(findLibraryCommand).read()
+if not mrSIDlibraryFullPath:
+   print("mrSID library was not found ! GDAL will be built without mrSID raster format support.")
+else:
+   print "mrSID library was found (%s)! GDAL will be built with mrSID raster format support." %mrSIDlibraryFullPath
+   getFolderCommand = "dirname %s" %mrSIDlibraryFullPath
+   mrSIDLibraryPath=os.popen(getFolderCommand).read()
+   mrSIDDSDKPath = os.path.normpath(mrSIDLibraryPath + os.sep + os.pardir)
```
tst-lsavoie: There's a slightly simpler way to do this with `os.path.join(mrSIDLibraryPath, os.pardir`

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -26,11 +26,28 @@
...
+# Check if MrSID decoding library is installed and get the decoding SDK path
+mrSIDLibName = 'libltidsdk.so'
+mrSIDDSDKPath = ""
+findLibraryCommand = "/sbin/ldconfig -p | grep %s | tr ' ' '\n' | grep /" %mrSIDLibName
+mrSIDlibraryFullPath=os.popen(findLibraryCommand).read()
+if not mrSIDlibraryFullPath:
+   print("mrSID library was not found ! GDAL will be built without mrSID raster format support.")
+else:
+   print "mrSID library was found (%s)! GDAL will be built with mrSID raster format support." %mrSIDlibraryFullPath
+   getFolderCommand = "dirname %s" %mrSIDlibraryFullPath
```
tst-lsavoie: Python includes os.path.dirname, though I'm not sure if it will do exactly the same thing.

earth_enterprise/src/third_party/gdal/SConscript
tst-lsavoie: os.path.dirname might also work.

earth_enterprise/src/third_party/gdal/SConscript
asmitaw: I think its helpful if people actually have the MrSIDDSK in some other path and expect that to work, unless that is absolutely not a case ever( meaning do we gurantee that there would always be same directory/ filepath or a symlink regardless of OS platform ?)

tst-nfarah: MrSID DSDK can be 'installed' in any folder as the user choose to. The only requirement is that location is discover-able by 'ldconfig'. This means that 'ldconfig' cache need to be refreshed after installing MrSID DSK.

tst-nfarah: @tst-lsavoie : i can remove the print statement when mrSID DSDK library was not found and keep the one when the library is found.

earth_enterprise/src/third_party/gdal/SConscript
tst-nfarah: yes

earth_enterprise/src/third_party/gdal/SConscript
tst-nfarah: you mean: os.path.normpath(os.path.join(mrSIDLibraryPath, os.pardir)) ?

tst-nfarah: done

earth_enterprise/src/third_party/gdal/SConscript
tst-nfarah: done

earth_enterprise/src/third_party/gdal/SConscript
tst-nfarah: done

tst-dsheremata:
When some of the GEE tools that were built on a machine with the MrSid SDK and then intalled and executed on a machine *without* the MrSid SDK, they fail. (example: "[root@geec7 lib]# geselectassetroot --assetroot /gevol/assets
geselectassetroot: error while loading shared libraries: libltidsdk.so: cannot open shared object file: No such file or directory
"

I think that they should either execute without the capabilities or we should force fusion server to fail earlier with a more descriptive error that the MrSid SDK should be installed.

tst-nfarah:
@tst-dsheremata: This is a default behavior of a platform when an app is built with a library as a dependency: it fails with that error when the library is not found on the target machine. This is said, there may be a way to work around this: building GDAL with mrSID as plugin and not a 'hard embedded' format. This may, however, change/augment the scope of this PR.

tst-nfarah:
@tst-dsheremata : To ensure that MrSID is supported on the end machine, MrSID library need to be there. We would ensure of that through an RPM. So for now we would not change the scope of this PR.

earth_enterprise/src/third_party/gdal/README.txt
```diff
@@ -35,3 +35,11 @@ on a SLES 9 box (compatible with SLES 9,10, RHEL 4,5 and Ubuntu 6,8)
 
 To Build, you will need to build on a SLES 9 box using the scripts found in
 //depot/googleclient/geo/fusion/RPMS/build_scripts
+
+To build GDAL with MrSID Raster Format, MrSID library needs to be installed on your platform:
+1- Get MrSID Decoding SDK library
+2- Copy MrSDK library into a folder. Ex: \opt
```
tst-lsavoie: Should be `/opt`, not `\opt`

earth_enterprise/src/third_party/gdal/README.txt
```diff
@@ -35,3 +35,11 @@ on a SLES 9 box (compatible with SLES 9,10, RHEL 4,5 and Ubuntu 6,8)
...
+3- Ensure that 'libltidsdk.so' can be disovered by 'ldconfig' command:
+	3.1- Add the full path of 'libltidsdk.so' to ‘/etc/ld.so.conf’. You can edit that file as root admin only.
```
tst-lsavoie: Do we want a specific MrSID library named here or should we leave the name generic?

tst-lsavoie:
I've tested this on all 6 platforms, including the latest changes on at least one platform.  Everything works fine.  I made a couple of small comments about the readme change but overall this looks good.

earth_enterprise/src/third_party/gdal/README.txt
tst-nfarah: yes.

earth_enterprise/src/third_party/gdal/README.txt
tst-nfarah: yes we do want that specific library name.

earth_enterprise/src/third_party/gdal/README.txt
tst-nfarah: done

earth_enterprise/src/third_party/gdal/README.txt
```diff
@@ -35,3 +35,11 @@ on a SLES 9 box (compatible with SLES 9,10, RHEL 4,5 and Ubuntu 6,8)
 
 To Build, you will need to build on a SLES 9 box using the scripts found in
 //depot/googleclient/geo/fusion/RPMS/build_scripts
+
+To build GDAL with MrSID Raster Format, MrSID library needs to be installed on your platform:
+1- Get MrSID Decoding SDK library
```
tst-ccamp: I happen to know where to get this library, so I'm speaking on behalf of the engineers who will be coming along after us :)

Can you add a link to the LizardTech page where you can sign up to request and download the SDK? (https://www.lizardtech.com/developer) I searched for "MrSID Decoding SDK library" and I didn't see that page come up in the first page of Google results.

earth_enterprise/src/third_party/gdal/README.txt
```diff
@@ -35,3 +35,11 @@ on a SLES 9 box (compatible with SLES 9,10, RHEL 4,5 and Ubuntu 6,8)
...
+2- Copy MrSDK library into a folder. Ex: /opt
```
tst-ccamp: Would it be better to put this in /usr/local/lib?

earth_enterprise/src/third_party/gdal/README.txt
```diff
@@ -35,3 +35,11 @@ on a SLES 9 box (compatible with SLES 9,10, RHEL 4,5 and Ubuntu 6,8)
...
+3- Ensure that 'libltidsdk.so' can be disovered by 'ldconfig' command:
+	3.1- Add the full path of 'libltidsdk.so' to ‘/etc/ld.so.conf’. You can edit that file as root admin only.
+	3.2- Rebuild the libraries path cache (‘/etc/ld.so.cache’): $sudo ldconfig
```
tst-ccamp: These commands should be surrounded by backticks "`" so that they will be properly formatted for markdown and helps them to stand out.

earth_enterprise/src/third_party/gdal/README.txt
tst-nfarah: sure

earth_enterprise/src/third_party/gdal/README.txt
tst-nfarah: I contemplated that idea, but the way GDAL expect MrSid library is a folder where there is a 'lib' sub-folder and 'include' sub-folder. Using 'usr/local/lib' might indicate that the library is stored there while the headers are store in 'usr/local/include'.

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -26,11 +26,26 @@
...
 
 gdal_version = 'gdal-2.1.2'
 openjpeg_version = 'openjpeg-v2.1.2-linux-x86_64'
 # kakadu_version = 'v5_2_6-00418C'
 # geoexpress_version = 'MrSID_DSDK-8.5.0.3422'
+
+# Check if MrSID decoding library is installed and get the decoding SDK path
+mrSIDLibName = 'libltidsdk.so'
+mrSIDDSDKPath = ""
+findLibraryCommand = "/sbin/ldconfig -p | grep %s | tr ' ' '\n' | grep /" %mrSIDLibName
```
tst-ccamp: There is also a "scons" way to check for the library. If you take a look at the [gTest Sconscript](https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/third_party/gtest/SConscript), you'll see what I mean. This works similarly to the "configure" step in building with make. It has the added benefit of not just checking ldconfig, but also verifies that it can be loaded with the compiler that scons is configured to use. But, I'll leave it up to you whether you think it's worth the trouble to convert to that method.

tst-nfarah: hmm.. you mean:
conf = Configure(env)
if not conf.CheckLib('gtest'):
?
I don't see a way from there to get the path of the lib.

earth_enterprise/src/third_party/gdal/README.txt
tst-ccamp: Oh, I see, this isn't actually a markdown file, so the formatting doesn't work. I guess it doesn't matter either way then.

earth_enterprise/src/third_party/gdal/README.txt
```diff
@@ -35,3 +35,11 @@ on a SLES 9 box (compatible with SLES 9,10, RHEL 4,5 and Ubuntu 6,8)
 
 To Build, you will need to build on a SLES 9 box using the scripts found in
```
tst-ccamp: Wow, most of the info in this file extremely out of date and doesn't apply to Open GEE at all.

tst-ccamp: If you decide to clean up the file, you can also rename it to README.md

tst-nfarah: I just did clean up the README.txt and renamed it to README.md

earth_enterprise/src/third_party/gdal/README.md
```diff
@@ -0,0 +1,25 @@
+Project: gdal-ge
```
tst-ccamp: If you change to this, it should correct the formatting:

```
Project: gdal-ge

Description:

This project builds a custom/sandboxed version of GDAL libraries and utilities
for Google Earth Enterprise Fusion.
It includes a number of extensions:
* Open jpeg2000 decoder
* MrSID decoder

and patches:
* gdal-xerces-conftest.patch
* gdal-mrsiddataset.patch
* gdal-kakopen.patch
* gdal-GBdatum.patch
* gdal-performance.patch


Install Location: /opt/google/{bin,lib}

To build GDAL with MrSID Raster Format, MrSID library needs to be installed on your platform:

1. Get MrSID Decoding SDK library [Lizardtech Developer Page](https://www.lizardtech.com/developer/)
1. Copy MrSDK library into a folder. Ex: /opt
1. Ensure that 'libltidsdk.so' can be disovered by `ldconfig` command:
   1. Add the full path of 'libltidsdk.so' to ‘/etc/ld.so.conf’. You can edit that file as root admin.
   1. Rebuild the libraries path cache (‘/etc/ld.so.cache’): `$sudo ldconfig`
   1. Verify that `ldconfig` command can now detect 'libltidsdk.so' library: `$ldconfig -p | grep libltidsdk.so`
```

earth_enterprise/src/third_party/gdal/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Description:
+This project builds a custom/sandboxed version of GDAL libraries and utilities
+for Google Earth Enterprise Fusion.
+It includes:
+  open jpeg2000 decoder
```
tst-ccamp: This part is also formatted incorrectly, see my changes in previous comment

PR #576, opened 2017-10-25
Key value store unit tests

tst-lsavoie:
This PR adds unit tests for the key value store that is used by the RPMs.  It uses the shunit2 testing library - the majority of the changes are from including this library.  shunit2 is an xUnit style testing library for shell scripts (https://github.com/kward/shunit2).  While writing the tests I noticed a few things about the key value store that may need to be addressed:

1. We are using flock when updating keys in the file.  We should probably use it for all the get and set functions involving a file.
1. keyvalue_file_escape_value tries to handle strings that include null characters.  I'm not sure this is necessary - if a string has a null character, it ends at the null character.  There is a failing test case because of this, so we need to decide what to do about that.

I also wrote a top level test.sh file that calls the RPM tests and the Fusion unit tests.  I've also included a build_and_install.sh script that I've been using locally in case anyone else finds it useful.


tst-lsavoie:
Based on conversation about the two points above:

1. We should probably use flock everywhere for consistency but since this runs in RPM scripts we're already guaranteed only one instance is running, so it's not a high priority.
1. Handling nulls in value strings is potentially a problem, but it's very unlikely so we're not going to worry about it for now.  As a short term solution I'm going to comment out the failing test and include an explanation.

earth_enterprise/test.sh
```diff
@@ -0,0 +1,66 @@
...
+# should be added here as they are created.  To add a new test:
+#   1. Add any code necessary to call the test script from this directory.
+#   2. Call test_banner with the name of the test right before the test script
+#      is called (it makes the output nicer).
+#   3. Ensoure that FAILURES is incremented if the test script fails.
+
+FAILURES=0
+
+BANNER="============================================================"
+
```
tst-ccamp: Set working directory

earth_enterprise/test.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
tst-ccamp: This should be moved to a different directory and/or given a more descriptive filename

earth_enterprise/tools/shunit2-2.1.6/bin/gen_test_results.sh
```diff
@@ -0,0 +1,62 @@
+#! /bin/sh
```
tst-ccamp: Is this not available as an OS level package we could add as a prereq? I really hate bundling new stuff

earth_enterprise/test.sh
tst-ppenev: I vote for creating, e.g. a `test` subdirectory under the `scripts` directory to host scripts that are for testing, but are not otherwise required for running Open GEE.

earth_enterprise/rpms/test/key-value_test.sh
```diff
@@ -0,0 +1,208 @@
...
+
+test_keyvalue_file_escape_value() {
+  value_escape_test "Basic value escape failed" "abcde" "abcde"
+  value_escape_test "New line value escape failed" 'abc
+def' 'abc\ndef'
+  value_escape_test "Carriage return value escape failed" $'abc\rdef' 'abc\rdef'
+  value_escape_test "Backslash value escape failed" 'back\slash' 'back\\slash'
+  value_escape_test "Backspace value escape failed" $'abc\bdef' 'abc\bdef'
+
+  # This test fails because bash assumes the string ends at the null value.
```
tst-ppenev: According to Server Fault, and Stack Overflow (e.g., https://stackoverflow.com/questions/6570531/assign-string-containing-null-character-0-to-a-variable-in-bash), BASH doesn't support variable values or command-line arguments containing null characters, because it uses null-terminated C strings internally.

tst-ppenev: I think we can convert this commented out unit test to just a comment stating that the current versions of BASH don't support null characters in variable values.

earth_enterprise/tools/shunit2-2.1.6/bin/gen_test_results.sh
tst-ppenev: Well, there is https://github.com/bpkg/bpkg#usage. I couldn't tell if it supports updating already installed packages, and we might have to contribute a packaging for ShUnit 2 to be able to use it with `bpkg`.

Otherwise, we could just add the ShUnit 2 repository as a submodule of the Open GEE repository. It may be less instrumentalized, but that way it would be automatically fetched when someone clones the Open GEE repository, and it would be kept up to date.

tst-lsavoie: Looks like it's included in the default yum and apt repos.  I'll see about switching to those.

earth_enterprise/rpms/shared/lib/key-value.sh
```diff
@@ -46,6 +46,8 @@ keyvalue_file_escape_key()
 }
 
 # Escapes a value to a string that can be stored in the key-value store.
+# Note: this function does not escape null characters because they are not
```
tst-ppenev: Minor: I'd say, more specifically, e.g. `because current versions of BASH don't support null characters in variable values`, since null characters can appear in other places, such as being piped to command input, or from command output.

earth_enterprise/rpms/test/key-value_test.sh
```diff
@@ -0,0 +1,203 @@
...
+  # $3 - expected output
+  function_with_output_test "keyvalue_file_escape_key" "$1" "$2" "$3"
+}
+
+test_keyvalue_file_escape_key() {
+  key_escape_test "Basic key escape failed" "abcde" "abcde"
+  key_escape_test "Space key escape failed" "hello world" "hello\ world"
+  key_escape_test "Dollar key escape failed" 'Owed $20' 'Owed\ \$20'
+  key_escape_test "Backslash key escape failed" 'back\slash' 'back\\slash'
+  key_escape_test "Multiple key escape failed" ' \$' '\ \\\$'
```
tst-ppenev: Note: You may wont to test the case of a key ending with one, or more new-line characters.

earth_enterprise/rpms/test/key-value_test.sh
```diff
@@ -0,0 +1,203 @@
...
+test_keyvalue_file_escape_value() {
+  value_escape_test "Basic value escape failed" "abcde" "abcde"
+  value_escape_test "New line value escape failed" 'abc
+def' 'abc\ndef'
+  value_escape_test "Carriage return value escape failed" $'abc\rdef' 'abc\rdef'
+  value_escape_test "Backslash value escape failed" 'back\slash' 'back\\slash'
+  value_escape_test "Backspace value escape failed" $'abc\bdef' 'abc\bdef'
+  value_escape_test "Form feed value escape failed" $'abc\fdef' 'abc\fdef'
+  value_escape_test "Vertical tab value escape failed" $'abc\vdef' 'abc\vdef'
+  value_escape_test "Multiple value escape failed" $'\r\n\\\b\f\v' '\r\n\\\b\f\v'
```
tst-ppenev: Note: You may want to test the case of a value ending with one, or more new-line characters.

scripts/build_and_install.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# This is a simple script that builds and installs Open GEE Fusion and Server.
+# It's useful for development.
+
+# You can customize this script by changing VERSION to "release", "optimize",
+# or (empty) and by changing CORES to the number of concurrent processes you
+# want to run.
+
+VERSION="release"
+CORES=8
```
tst-ppenev: Please, use `$[ 2 * $(nproc) ]`.  It's already an inconvenience that it's been hard-coded in other scripts.

scripts/build_and_install.sh
```diff
@@ -0,0 +1,55 @@
...
+
+SCRIPT_DIR=`dirname $0`
```
tst-ppenev: Reminder: Escape `$0`.

scripts/build_and_install.sh
```diff
@@ -0,0 +1,55 @@
...
+BASE_DIR_DOTS="${SCRIPT_DIR}/../earth_enterprise"
```
tst-ppenev: Suggestion: This seems like a variable that's used in only one place. Consider substituting its value, and removing the variable.

scripts/build_and_install.sh
```diff
@@ -0,0 +1,55 @@
...
+# functions.
+
+# This is a simple script that builds and installs Open GEE Fusion and Server.
+# It's useful for development.
+
+# You can customize this script by changing VERSION to "release", "optimize",
+# or (empty) and by changing CORES to the number of concurrent processes you
+# want to run.
+
+VERSION="release"
```
tst-ppenev: Minor: I wouldn't call `release` a version. I'd call that that, e.g. `RELEASE` or `BUILD_TYPE`.

scripts/build_and_install.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
tst-ppenev: This seems to be rather close to what `build-gee.sh && install-gee.sh` under `docker/image-definition/scripts` do. We could consider reducing code duplication.

scripts/build_and_install.sh
```diff
@@ -0,0 +1,55 @@
...
+
+VERSION="release"
+CORES=8
+
+SCRIPT_DIR=`dirname $0`
+BASE_DIR_DOTS="${SCRIPT_DIR}/../earth_enterprise"
+BASE_DIR=$(cd "${BASE_DIR_DOTS}"; pwd)
+
+if [ -z "${VERSION}" ]; then
+  VER_STR=""
```
tst-ppenev: Minor: As above, I'd call this something like `RELEASE_ARGUMENT`, or `BUILD_TYPE_ARGUMENT`.

scripts/run_all_unit_tests.sh
```diff
@@ -0,0 +1,68 @@
...
+  echo "$1"
+  echo "$BANNER"
+  echo
+}
+
+test_banner() {
+  banner "Running $1"
+}
+
+SCRIPT_DIR=`dirname $0`
```
tst-ppenev: Reminder: Escape `$0`.

scripts/run_all_unit_tests.sh
```diff
@@ -0,0 +1,68 @@
...
+BASE_DIR_DOTS="${SCRIPT_DIR}/../earth_enterprise"
```
tst-ppenev: Suggestion: Consider removing a `BASE_DIR_DOTS` variable used in only one place.

scripts/run_all_unit_tests.sh
```diff
@@ -0,0 +1,68 @@
...
+BASE_DIR=$(cd "${BASE_DIR_DOTS}"; pwd)
+
+# RPM tests
+test_banner "RPM Unit Tests"
```
tst-ppenev: Minor: We may consider calling this `Packaging Unit Tests`, since the scripts can produce `.deb`s, as well.

earth_enterprise/rpms/test/key-value_test.sh
tst-lsavoie: Newlines apparently aren't escaped for keys.  I'm not sure if it's worth adding support for that.

tst-ppenev: Really? They are escaped in my BASH interpreter.

tst-ppenev:
```BASH
. shared/lib/key-value.sh 
printf -v K "%b" "New ln:\\n"
keyvalue_file_escape_key "$K" K_ESCAPED
echo "$K_ESCAPED"
# Output: $'New ln:\n'
```

tst-lsavoie: The syntax is funny.  I found a way that works.

tst-ppenev: :sun_with_face:

earth_enterprise/rpms/test/key-value_test.sh
```diff
@@ -0,0 +1,209 @@
...
+  function_with_output_test "keyvalue_file_escape_key" "$1" "$2" "$3"
+}
+
+test_keyvalue_file_escape_key() {
+  key_escape_test "Basic key escape failed" "abcde" "abcde"
+  key_escape_test "Space key escape failed" "hello world" "hello\ world"
+  key_escape_test "Dollar key escape failed" 'Owed $20' 'Owed\ \$20'
+  key_escape_test "Backslash key escape failed" 'back\slash' 'back\\slash'
+  key_escape_test "Multiple key escape failed" ' \$' '\ \\\$'
+  key_escape_test "test" 'abc
```
tst-ppenev: Reminder: You may wont to rename this test from `test` to something more descriptive, like `Line feed at end of key escape failed`.

tst-lsavoie: Yes, that wasn't supposed to make it into a commit.

earth_enterprise/tools/shunit2-2.1.6/bin/gen_test_results.sh
tst-lsavoie: So this is available for RHEL and CentOS 6 but not 7.  I could switch to another framework but I really like the way this one works.  Perhaps this is something we should discuss.

tst-lsavoie:
@tst-ccamp @tst-ppenev I think I've addressed all the comments.  I tested the new get-gee-build-environment.sh script on all platforms.  I also tested building a docker image, though it's hard to tell if it actually ran the scripts I changed.

earth_enterprise/rpms/test/key-value_test.sh
```diff
@@ -0,0 +1,212 @@
...
+}
+
+key_escape_test() {
+  # $1 - message
+  # $2 - input value
+  # $3 - expected output
+  function_with_output_test "keyvalue_file_escape_key" "$1" "$2" "$3"
+}
+
+test_keyvalue_file_escape_key() {
```
tst-nfarah: may be add test cases for lower and upper cases if the used function is case sensitive

earth_enterprise/docker/image-definition/scripts/build-gee.sh
```diff
@@ -1,4 +0,0 @@
-#! /bin/bash
```
tst-ppenev: I don't think that's going to work in a Docker build.  The `docker/image-definition/scripts/build-gee.sh` script will be copied to the image, but the `../../../../scripts/build-gee.sh` one won't be.

You'd want to change the `Dockerfile.stage-1.*` scripts to use `build-gee.sh` from the cloned repository, and remove `build-gee.sh` from `image-definition/scripts`.

The same could be done for `build-portable-server.sh`.

earth_enterprise/docker/image-definition/scripts/common-build-install.sh
```diff
@@ -0,0 +1 @@
+../../../../scripts/common-build-install.sh
```
tst-ppenev: Note: This won't work in Docker, either, like `build-gee.sh` above.

earth_enterprise/docker/image-definition/scripts/install-gee.sh
```diff
@@ -1,29 +0,0 @@
-#! /bin/bash
```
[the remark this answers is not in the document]
tst-ppenev: Reminder: this script won't work in Docker, anyway.

scripts/common-build-install.sh
```diff
@@ -0,0 +1,41 @@
...
+  BUILD_TYPE="${BUILD_TYPE}=1"
+fi
+
+# Default number of processes is twice the number of cores
+PROCS=${2:-$[ 2 * $(nproc) ]}
+
+# Find the earth_enterprise directory
+REL_SCRIPT_DIR=`dirname "$0"`
+SCRIPT_DIR=$(cd "${REL_SCRIPT_DIR}"; pwd)
+if [ `basename "${SCRIPT_DIR}"` = "scripts" ]; then
```
tst-ppenev: Reminder: escape command-substitution. (This one would fail for a script path that contains any shell special characters.)

E.g.:

```BASH
if [ "$(basename "$SCRIPT_DIR")" = "scripts" ]; then
```

tst-ppenev: I don't think it should matter if you're building a Docker image, or not.  I think we should remove this logic.

scripts/install-gee.sh
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Installs Open GEE.  You must have built Open GEE previously.  Accepts two
+# paramters: the build type (release, optimize, or debug - defaults to release)
+# and the number of processes to run (defaults to twice the number of cores).
+
+source common-build-install.sh
```
tst-ppenev: Use path to the sourced file that doesn't depend on the current directory.

tst-lsavoie:
I ended up removing the build_and_install.sh script.  It was getting to be too much of a headache and it's not the point of this PR.

earth_enterprise/rpms/test/key-value_test.sh
tst-lsavoie: I added some mixed case characters to some of the tests.

PR #577, opened 2017-10-26
Add Google Analytics to OpenGEE

tst-mswartz:
This is going to be confusing to test, because in order for it to work in master the analytics ID number and property info have to be pointing to opengee.org, but in order for it to test it has to be pointing to tst-mswartz.github.io/earthenterprise... it's checked in pointing to the former, but if I change the ID UA—108632131-2  to UA—108632131-1  it will be pointing to the latter.  You could always add a property for your own github pages address and update the ID number for that in your fork, and test your own analytics.


tst-basilhuffman:
Looks ok to me

PR #578, opened 2017-10-26
Automated build environment scripts

tst-basilhuffman:
Addressed issues vis a vis auto accept


scripts/get-gee-build-environment.sh
```diff
@@ -419,9 +419,9 @@ done
 check_root
 
 if [ -n "$ASSUME_YES" ]; then
-    ASSUME_YES_PACKAGE_MANAGER_PARAMETER="-y"
```
tst-ppenev: I don't think this fixes the issue.  This makes it so that, if you specify `--assume-yes` you don't get assume-yes, and if you disable assume-yes, you get `--assume-yes`.

Issue #566, opened 2017-10-19
Upgrade PostgreSQL to 9.6


tst-ccamp:
We should unbundle Postgres because it should be upgraded for security purposes and to reduce future maintenance efforts.

tst-ccamp:
Since work has already been done under this ticket, I changed it to just the be the upgrade. We can open a follow up ticket when/if we decide it should be unbundled.

tst-ccamp:
I'm working on the data upgrade path now

tst-lsavoie:
Closed via PR #641.

PR #579, opened 2017-10-26
#566 Update PostgreSQL command interfaces and clean up SQL files

tst-ccamp:
Partial implementation for #566 

Verification Steps:

1. Remove /tmp/fusion_os_install if it exists
1. Make sure at least one fusion database is published
1. Test running install_server.sh while server is still installed
    * It should backup and restore the postgres databases
    * All published fusion databases should still be published
1. Test running install_server.sh without server installed
1. With server installed, publish a database
1. Test running `sudo -u gepguser /opt/google/bin/geresetpgdb` with the following parameters
    1. soft: It should back up the databases and output the location of the backup; it will also replace the databases with initial blank versions
    1. restore <path of backup>: Using the backup path output from the "soft" command, this should restore the databases to their previous state before running the "soft" command, fusion databases should be published
    1. hard: It should completely wipe out the data in the database and start fresh. No databases will be published

Note: I didn't add any code to remove the sql files from /opt/google/bin


earth_enterprise/src/support/geecheck/lib/FusionUtils.pm
```diff
@@ -454,7 +454,9 @@ sub check_httpd_services() {
 
 # Check the postgres installation.
 sub check_postgres() {
-  my $max_stack_depth = readpipe "grep max_stack_depth /var/opt/google/pgsql/data/postgresql.conf";
+  my @max_stack_depth_list = readpipe "sudo -u gepguser /opt/google/bin/psql  postgres -c \"show max_stack_depth\"";
```
tst-lsavoie: Minor: there's a double space before postgres.

earth_enterprise/src/server/geresetpgdb
```diff
@@ -181,15 +181,21 @@ function do_dump {
...
 # Deletes old data and recreates db instance.
 function do_reset {
-  if [ -f $VARPGDIR/data/postmaster.pid ]; then
-    $BINDIR/pg_ctl -D $VARPGDIR/data -l $VARPGDIR/logs/pg.log stop
-  fi
 
-  # delete all the db files. not sure if we should do this.
-  rm -Rf $VARPGDIR/data
+  # Drop gee databases
+  if [ `$BINDIR/psql -lqt | cut -d \| -f 1 | grep -w gestream | wc -l` == 1 ]; then
```
tst-lsavoie: If I list databases, I see the following:

```
 geendsnippet  | gepguser | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 geplaces      | geuser   | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 gepoi         | gepguser | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 gesearch      | gepguser | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 gestream      | gepguser | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 postgres      | gepguser | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 searchexample | geuser   | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 template0     | gepguser | UTF8     | en_US.UTF-8 | en_US.UTF-8 | =c/gepguser
                                                                 : gepguser=CTc/
gepguser
 template1     | gepguser | UTF8     | en_US.UTF-8 | en_US.UTF-8 | =c/gepguser
                                                                 : gepguser=CTc/
```

There are apparently other databases that should be dropped.  Hopefully we aren't missing any that are created dynamically.

tst-lsavoie:
A couple of points about geresetpgdb:

- 'soft' doesn't restore the database, it apparently just backs it up so you can restore it later.
- Some of the text needs to be changed.  For example, it says "WARNING: 'geresetpgdb hard' will completely delete postgresql data directory, and then recreate the gestream, gesearch, gepoi, gesnippets databases in their initial state." This is no longer accurate.

tst-lsavoie:
One other note: to get gerestpgdb to work I had to su to gepguser and then cd to a directory where gepguser had permissions.  Do you know of an easier way to run it?

earth_enterprise/src/server/geresetpgdb
tst-ccamp: This is the list of dbs that were always dropped in the "soft" reset. geplaces and searchexample are only dropped in the "hard" reset. I just moved it around so that the do_reset function is called in both soft and hard reset, but the example databases are only dropped for hard.

tst-ccamp:
@tst-lsavoie That's not the behavior that I saw in my testing with the script. When I did a soft reset the databases were always restored. Did you see that behavior in your testing?

tst-ccamp:
@tst-lsavoie I use `sudo -u gepguser ./geresetpgdb hard`. I'll update the verification steps

tst-ccamp:
@tst-lsavoie I updated the verification steps to reflect the proper behavior of "soft" and added a verification step for "restore"

earth_enterprise/src/server/geresetpgdb
```diff
@@ -83,8 +84,8 @@ else
 fi
 
 if [ $type == "hard" ]; then
-  echo "   WARNING: 'geresetpgdb $type' will completely delete postgresql data"
-  echo "directory, and then recreate the gestream, gesearch, gepoi, gesnippets"
+  echo "   WARNING: 'geresetpgdb $type' will completely delete all GEE postgresql"
```
tst-nfarah: minor: all **Open** GEE ...

earth_enterprise/src/server/geresetpgdb
```diff
@@ -201,27 +207,49 @@ if [ -e $VARPGDIR/data/postmaster.pid ]; then
 fi
 
 if [ $type == "backup" ]; then
-  if [ ! -e $VARPGDIR/data ]; then
-    echo "WARNING: There is no data to back up."
```
tst-nfarah: Any reason that info about no data to back up was removed ? It would be, still, useful to post info to a user if they asked for backup and there was nothing to back up.

earth_enterprise/src/server/geresetpgdb
```diff
@@ -201,27 +207,49 @@ if [ -e $VARPGDIR/data/postmaster.pid ]; then
...
-  else
-    do_dump
-  fi
-  exit
+  do_dump
 fi
 
 # if the data dir does not exist then this has to be a hard reset.
+# TODO: This will need to be removed when we unbundle
```
tst-nfarah: minor: unbundle **PostgreSQL**
so it a bit more clear to any one else not familiar with effort to unbundle ... ?

earth_enterprise/src/server/geresetpgdb
```diff
@@ -201,27 +207,49 @@ if [ -e $VARPGDIR/data/postmaster.pid ]; then
...
+    # Start postmaster.
+    if [ ! -f $VARPGDIR/data/postmaster.pid ]; then
+      $BINDIR/pg_ctl -D $VARPGDIR/data -l $VARPGDIR/logs/pg.log start -w
+    fi
+
+    # Do reset and recreate database instance.
+    echo Removing geserve-databases...
+    do_reset
+
+    # Also drop example databases
```
tst-nfarah: Any reason why 'geplaces; and 'searchexample' dropping was not made part of 'do_reset' ?

earth_enterprise/src/SConstruct
```diff
@@ -350,6 +350,7 @@ installdirs = {
     'server_root': Dir('server/', installdir),
     'server_share': InstServerDir('share'),
     'server_search': InstServerDir('search'),
+    'server_sql': InstServerDir('share/opengee-server/sql'),
 
     'httpd': InstServerDir('gehttpd'),
     'httpdmodules': InstServerDir('gehttpd/modules'),
```
tst-nfarah: 'src/installer/products/server/GEServerBase.pm' and 'src/installer/lib/Users.pm' were removed because they were 
obselete or was it part of this PR ?

We would need to ensure that RPM work is in sync with any files removal.

earth_enterprise/src/server/geresetpgdb
tst-ccamp: The check is no longer valid as the directory will always be there after the unbundle (this change is in anticipation of that). If we need to, we can do a different check. I'm not entirely sure it's necessary.

tst-nfarah: ok

earth_enterprise/src/server/geresetpgdb
tst-ccamp: To match the same behavior as before, that is done as part of the hard reset, not the soft reset. My guess is performance is the reason behind it. Soft reset is used with the upgrade target and it takes up to a minute to reinstall the example databases.

earth_enterprise/src/server/geresetpgdb
tst-ccamp: Hmmm, technically the script would work with older version of GEE databases also. I'm inclined to leave off Open

earth_enterprise/src/SConstruct
tst-ccamp: It's specified to delete them if possible in the ticket. I checked to make sure the files are not directly referenced on the RPM branch and also tested the installation without them present and saw no issues.

tst-nfarah: ok

tst-ccamp:
I got two approvals, so merging

Issue #423, opened 2017-08-02
Improve JPEG 2000 format support performance


tst-ccamp:
When compared to GeoTiff, building an imagery resource based on a JPEG 2000 source image takes many times longer. Investigate possible bottlenecks in the way the library is used. If there isn't something that can be optimized, consider converting the JPEG 2000 file temporarily to GeoTiff before running the rest of the build steps.

tst-awankhede:
with numCPU =  4,

I got the blue marble jp2  =  15 min 
Blue marble tiff = 3 min

Will update the thread once I post it to GDAL group.

tst-nfarah:
A new version of Open Jpeg library was released (http://www.openjpeg.org/2017/08/10/OpenJPEG-2.2.0-released). I tested it in comparison with Open Jpeg 2.1.2 and did see performance improvements that requires setting an env variable to enable multi-threading when decoding the data. The performance improvement (in OpenJpeg 2.2.0) seems to targets mostly tiled jp2 files. Using the new version on 2 CPU, 20 GB RAM and 8 enabled threads, i got:

bluemarble_4km.jp2  (10800 x 5400 pixels). 37.3 mb file size. Single tiled: **7.6 mn** (OpenJpeg 2.2.0) vs **13.8 mn** (OpenJpeg 2.1.2)
usgsLanSat.jp2 (8627 x 7745 pixels). 48 mb file size. Single tiled: **9.9 mn** vs **17.4 mn**

There is ongoing work to speed up more OpenJpeg performance for singled tiled images.
Will be looking forward to re-test the newer version of OpenJpeg once it will be released.

PR #580, opened 2017-10-27
I423 - Update openjp2 library to v2.3.0 from v2.1.2

tst-eclamar:
Fixes: #423 
Update openjp2 library from v2.1.2 to v.2.3.0
* Create new library to replace old
* Modify SConscript to point to new library version
* Modify gdal configure script to find new location for header file


tst-ccamp:
The changes made to the GDAL library should be done through the patch process like the various other patches we make to GDAL. The reason for this is so that they don't have to keep being made manually every time we do minor updates to the library. Please look into making the required changes using a patch. This isn't an urgent change, so we might as well do it right the first time around instead of as a follow-up.

tst-nfarah:
Doing a test to compare openJpeg 2.1.2 vs openJpeg 2.3.0 trough 'gdal_translate' of two jp2000 files we have in the tutorial (bluemarble_4km.jp2 and usgsLanSat.jp2) to Geotiff, i got (on Ubuntu 16.04) the following:
bluemarble_4km.jp2: 
------------------------
OpenJpeg 2.1.2: 1381 seconds. 23 minutes
OpenJpeg 2.3.0: 20 seconds

usgsLanSat.jp2:
-----------------
OpenJpeg 2.1.2: 1370 seconds. 22.8 minutes
OpenJpeg 2.3.0:  23 seconds

These numbers are pretty good. 

This is said, it appears that we may need to patch GDAL a little bit more to be in sync with enabling multi-threading when decoding jp2 files. See last two revisions made to 'openjpegdataset.cpp':
[https://trac.osgeo.org/gdal/log/trunk/gdal/frmts/openjpeg/openjpegdataset.cpp?rev=40607](url)

as well as revision made part of this ticket: [https://trac.osgeo.org/gdal/ticket/7074](url)

Eventually we would upgrade our GDAL and get these changes, but for the time being we, probably, need to address them through a patch.

tst-nfarah:
@tst-eclamar : I'm fine with creating a follow up ticket to patch GDAL more as mentioned in my previous comment. It appears that we get some performance results, but i think we can leverage multi-threaded decoding more by enabling it. The way to enable it is to patch 'openjpegdataset.cpp' though an API call to OpenJpeg library.

earth_enterprise/src/third_party/gdal/SConscript
```diff
@@ -26,26 +26,11 @@
 # lines below.
 
 Import('third_party_env')
-import os
```
tst-nfarah: These mrSID changes are now in master. Eventually you want to get them by updating your fork  or just ensuring that we get them when this PR is merged.

tst-ccamp:
Changes look good and I see 2 approvals, so I'm merging this

PR #583, opened 2017-10-30
feat #547 (rpms): RHEL6 Fusion RPM

### LINKED ISSUE #547: Create RPM for GEE Fusion for RHEL 6

tst-ppenev:
* Added read lock to key-value store reads.
* Fixed dependent templates.
* Switched Docker build scripts to install from packages.
* Added JRE installation to Docker `get-gee-build-environment.sh`.
* Fixed bugs in the Docker build scripts.
* Migrated some configuration file creation to template expansion during package building, instead of file creation in post-install scripts.
* Added configuration files created by the post-install script to package management.
* Migrated file mode changing from the post-install script to the package building phase.
* Updated node content extraction with `xmllint` for compatibility with the older version in RHEL 6.
* Updated asset root volume free space calculation for compatibility with the `df` version in RHEL 6.

Fixes #547.


earth_enterprise/rpms/shared/lib/read-xml.sh
```diff
@@ -0,0 +1,340 @@
+# Copyright 2017 the Open GEE Contributors.
```
tst-lsavoie: @tst-ppenev can we look for a solution that doesn't involve parsing XML in bash?  Surely there's a tool that can do it for us that's available on all platforms.  I just worry that writing our own XML parsing library will lead to trouble down the road.

tst-ppenev: We can have a spike to evaluate XML tools, if you want.

What can of trouble are you expecting? I'm not sure it makes that much of a difference what an XML parser is implemented in.

tst-dsugar: @tst-ppenev  From my perspective, I simply want to make sure what we choose to do is both maintainable and consistent on all the platforms it is needed on.  I am not particularly concerned what it is written in, but if there is something we can already use, then that is something we do not have to also actively maintain.

tst-ppenev: Well, it doesn't make much of a difference to me.

I know this would work to finish the install scripts, and it has a minimal amount of dependencies.  I don't know what the best tool we can use on the needed platforms is.

I think we may have enough to finish the packaging tasks.  We're only using basic features, which I don't expect to change in the XML standard soon.  So, we may be able to collect all the concerns to address, and add a follow-up task.

tst-ccamp: My proposal is to use this function. It calls xmllint, which is available on all our target platforms, and should be compatible with the older version of xmllint installed on rhel 6 and centos 6. I tested on Ubuntu 16.04 and RHEL6, so hopefully that is a representative sample. 
```
#!/bin/bash

get_xpath()
{
  local FILE="$1"
  local XPATH="$2"
  
  local OUTPUT=$(echo "cat $XPATH" | xmllint --shell "$FILE" | sed '/^\/ >/d' | sed 's/<[^>]*.//g')
  echo $OUTPUT 
}

echo $(get_xpath '/opt/google/etc/systemrc' '//Systemrc/assetroot/text()')
```

tst-dsugar: This is most consistent with what we had used before...

tst-dsheremata: I like it!

tst-dsugar: I already used  get_xpath in my pull request (#585), so we can either merge my branch into this, or approve this and then rebase and take mine out of wip...

earth_enterprise/docker/image-definition/scripts/build-gee.sh
```diff
@@ -1,4 +1,88 @@
...
+    fi
+
+    if [ -z "$BUILD_DEBS" ] && type apt-get >/dev/null 2>&1; then
+        BUILD_DEBS="yes"
+    fi
+fi
+
+
+cd earthenterprise/earth_enterprise || exit 1
+scons "-j$[ 2 * $(nproc) ]" release=1 build || exit 1
```
tst-lsavoie: Very minor - it seems like the `[ 2 * $(nproc) ]` should be extracted out to a variable to make it easier to change.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -196,13 +191,14 @@ final_assetroot_configuration()
...
         "$BASEINSTALLDIR_OPT/bin/geconfigureassetroot" --addvolume \
             "opt:$BASEINSTALLDIR_OPT/share/tutorials" --noprompt --nochown
         if [ $? -eq 255 ]; then
-            echo -e "The geconfigureassetroot utility has failed on attempting"
-            echo -e "to add the volume 'opt:$BASEINSTALLDIR_OPT/share/tutorials'."
-            echo -e "This is probably due to a volume named 'opt' already exists."
+            cat <<END
+The geconfigureassetroot utility has failed on attempting
+to add the volume 'opt:$BASEINSTALLDIR_OPT/share/tutorials'.
+This is probably due to a volume named "opt"'s already existing.
```
tst-lsavoie: Minor: I think it would be less awkward to say "This is probably because a volume named "opt" already exists."

earth_enterprise/rpms/opengee-server/post-install.sh
```diff
@@ -48,7 +47,7 @@ main_postinstall()
     setup_geserver_daemon
 
     # 5) Register publish root
-    $BASEINSTALLDIR_OPT/bin/geconfigurepublishroot --noprompt --path=$PUBLISHER_ROOT
+    "$BASEINSTALLDIR_OPT/bin/geconfigurepublishroot" --noprompt "--path=$PUBLISHER_ROOT"
```
tst-lsavoie: Minor: would it make more sense to have `--path="$PUBLISHER_ROOT"` rather than `"--path=$PUBLISHER_ROOT"`?  I moved the open quote.

earth_enterprise/docker/image-definition/scripts/install-gee.sh
```diff
@@ -7,23 +7,30 @@
...
+# cd earthenterprise/earth_enterprise || exit 1
+# scons -j8 release=1 stage_install || exit 1
+
+
 
-# Install Fusion:
-cd src/installer &&
-(echo C; echo C) | ./install_fusion.sh -hnmf || exit 1
+if type apt-get >/dev/null 2>&1; then
+    # Use RPM packages:
```
tst-ppenev: :stuck_out_tongue: Fix comments to correspond to the type of packages being installed.

earth_enterprise/rpms/opengee-server/post-install.sh
tst-ppenev: That depends on your style.  I'm fine with it either way.

tst-lsavoie:
I got a few errors installing this on a RHEL 6 machine with GEE prereqs:

```
Total size: 119 M
Installed size: 119 M
Downloading Packages:
Running rpm_check_debug
Running Transaction Test
Transaction Test Succeeded
Running Transaction
/var/tmp/rpm-tmp.P1NbEs: line 367: [[: /dev/mapper/vg_rhel6-lv_root
37711436: syntax error: operand expected (error token is "/dev/mapper/vg_rhel6-lv_root
37711436")
  Installing : opengee-fusion-5.2.0-1.x86_64                                1/1 
Setting up the Fusion daemon...
Fusion daemon setup ... Done
Fusion Warning:	Unable to chmod(/gevol/src): Operation not permitted
Fusion Warning:	Unable to open permissions on /gevol/src
Switching to (gefusionuser, gegroup)
Validating assetroot
Making new assetroot ....
Configured /gevol/assets.
Switching to (gefusionuser, gegroup)
Validating assetroot
Volumes modified
Configured /gevol/assets.
Starting System Manager... Done
Starting Resource Provider... Done
  Verifying  : opengee-fusion-5.2.0-1.x86_64                                1/1 

Installed:
  opengee-fusion.x86_64 0:5.2.0-1                                               

Complete!
```

The install seemed to work fine, though.

tst-ppenev:
@tst-lsavoie: Can you include the command to reproduce? I haven't seen that, and I don't know what checks are being run before installation.

tst-lsavoie:
It's pretty much the default stuff.  I started with a default RHEL 6 install, ran the get-gee-build-environment.sh script, then did `sudo yum -y install` on opengee-postgis, opengee-common, and opengee-fusion, in that order.

I also cloned the earthenterprise repo somewhere in there but I didn't build it - I build the RPMs on a separate RHEL 6 RPM.

tst-ppenev:
Well, I don't get these errors in a Docker container, so I'm trying to find out how to get them.

You built RPMs on one VM, and installed on another?

tst-lsavoie:
Yes, I built the RPMs on one VM and installed on anther.  I may have time later today to dig into this more.

tst-dsugar:
To me it seems like it may be leaking something about the build environment into the rpm pre script.  /dev/mapper path is related to how docker/linux containers map file systems.  I suspect /gevol/src may not exist, and this is actually kinda harmless in itself though the rpm build.gradle should at least create an empty one so long as other things expect it to exist.

tst-lsavoie:
Here are the scripts produced from `rpm -q --scripts`.
[opengee-fusion-scripts.txt](https://github.com/google/earthenterprise/files/1458590/opengee-fusion-scripts.txt)

tst-lsavoie:
One other problem I noticed - the download_tutorial.sh script is missing.  As we discussed, it's probably best to package the tutorial files as a separate RPM.

Overall fusion is working fine.  I was able to build a database with imagery, vector, and terrain layers and it works fine in EC.

tst-ppenev:
@tst-lsavoie: I added a fix for the Fusion post-install script error you were seeing.  Can you retest?

earth_enterprise/rpms/shared/install-utils.sh.template
```diff
@@ -32,22 +32,43 @@ ASSET_ROOT="/gevol/assets"
...
+
+xml_file_get_xpath()
+{
+    local FILE="$1"
+    local XPATH="$2"
+
+    # Warning: `xmllint --noent` doesn't recognize named entities like
+    # "&lt;" and "&gt;":
+    echo "cat $XPATH" | xmllint --noent --nocdata --shell "$FILE" |
+    # Skip the first and the last line:
```
tst-lsavoie: Minor: this comment is now wrong.

tst-ppenev: Which comment is wrong?

tst-lsavoie: "Skip the first and the last line" since we're now skipping the last 2 lines.  It's a small thing.

tst-lsavoie:
I'm still getting the same error with the latest changes.  I'm trying a clean build next.

tst-lsavoie:
Yep, even with a clean build I'm still seeing the same errors.  Fusion still seems to work, however.

earth_enterprise/rpms/shared/install-utils.sh.template
tst-ppenev: The comment is correct. The code wasn't running correctly before.

earth_enterprise/rpms/opengee-fusion/pre-install.sh
```diff
@@ -14,31 +14,26 @@
...
 # Main Functions
 #-----------------------------------------------------------------
 main_preinstall()
 {
-    service gefusion stop
+    if [ -f /etc/init.d/gefusion ]; then
+        service gefusion stop
+    fi
 
     load_systemrc_config
```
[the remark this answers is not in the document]
tst-ppenev: I think all install scripts are designed to read `systemrc` before using the values that are stored there.  This seems to allow system administrators to use custom user and group names.

I don't think it's possible to store names of users and groups in a CPIO archive, so that can't be done from Gradle.

The usual practice is to allocate UIDs and GIDs for system users and groups in each distribution.  We could submit a request for `gefusionuser` and `gegroup`.  But, the need for the other users and groups would be eliminated if we unbundled the software components they are associated with.

On the other hand, I think it's a bad idea idea to customize user and group names.  Actually, the Fedora packaging guide recommends that a package never delete user and group names, even if it created them on installation.  User and group removal is left as a manual step for administrators.  That's because there's no good way to ensure there are no files or directories that were created with these UIDs and GIDs that may be used again on that host.

earth_enterprise/rpms/opengee-fusion/pre-uninstall.sh
```diff
@@ -20,6 +20,10 @@ set +x
 # Main Function
 #-----------------------------------------------------------------
 
-/etc/init.d/gefusion stop
+service gefusion stop
+
+if [ -f "$CHKCONFIG" ]; then
+    "$CHKCONFIG" --del gefusion
+fi
```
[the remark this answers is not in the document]
tst-ppenev: We may also have another code section that gets executed on Ubuntu, and use the same scripts on all distributions.  At least, until we rewrite them as templates.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -150,20 +144,20 @@ install_or_upgrade_asset_root()
...
 
-    if [ "$IS_NEWINSTALL" == true ]; then
+    if [ ! -d "$ASSET_ROOT/.config" ]; then
         "$BASEINSTALLDIR_OPT/bin/geconfigureassetroot" --new --noprompt \
             --assetroot "$ASSET_ROOT" --srcvol "$SOURCE_VOLUME"
         chown -R "$GEFUSIONUSER:$GEGROUP" "$ASSET_ROOT"
     else
         # upgrade asset root -- if this is a master
         if [ "$IS_SLAVE" == false ]; then
             # TODO: Verify this logic -- this is what is defined in the
```
tst-ccamp: Does this TODO still apply?

tst-ppenev: I don't know what "the installer documentation" is, and why different versions of GEE have different file owners.  So, as far as I know, it still applies.

tst-dsheremata: I know it's not very clean, but I don't think that this should be holding up the merge.

PR #584, opened 2017-10-31
#547: RHEL6 RPMs: Add nested templates and key-value read lock

tst-ppenev:
Partially implements: #547 
* Switched to using nested templates for the `openGeeSharedFiles` task, so that a specific order of visiting files is not required for the `Copy` task.
* Added a read lock for reading from the install script key-value store.


earth_enterprise/rpms/shared/lib/key-value.sh
```diff
@@ -149,9 +149,12 @@ keyvalue_file_get()
...
     if [ -f "$_KFG_STORE_PATH" ]; then
-        keyvalue_file_get_from_stdin \
-            "$_KFG_KEY" "$_KFG_OUTPUT_VARIABLE_NAME" \
-            < "$_KFG_STORE_PATH"
+        (
+            flock --shared -n 9 || exit 1
+            keyvalue_file_get_from_stdin \
+                "$_KFG_KEY" "$_KFG_OUTPUT_VARIABLE_NAME" \
+                < "$_KFG_STORE_PATH"
+        ) 9> "$STORE_PATH.lock"
```
tst-lsavoie: `STORE_PATH` doesn't exist in this function.  I think this should be `_KFG_STORE_PATH`.

earth_enterprise/rpms/shared/lib/key-value.sh
```diff
@@ -149,9 +149,12 @@ keyvalue_file_get()
     local _KFG_OUTPUT_VARIABLE_NAME="$3"
 
     if [ -f "$_KFG_STORE_PATH" ]; then
-        keyvalue_file_get_from_stdin \
-            "$_KFG_KEY" "$_KFG_OUTPUT_VARIABLE_NAME" \
-            < "$_KFG_STORE_PATH"
+        (
+            flock --shared -n 9 || exit 1
```
tst-lsavoie: I merged this with my unit tests (PR #576) and ran them and one is failing.
```
ASSERT:Get from file failed expected:<value2> but was:<>
```
It appears the output variable is being set but it's not escaping from the sub shell.

tst-ppenev: Thanks, @tst-lsavoie. I was rushing this through without giving it enough attention. It should be fixed now.

earth_enterprise/rpms/shared/lib/key-value.sh
tst-ppenev: :P You're right. It should be fixed now.

PR #585, opened 2017-10-31
WIP: rpm fusion scripting

tst-dsugar:
Not yet clean, but is at least now debug-able...

### DIFF
--- a/earth_enterprise/rpms/opengee-fusion/post-install.sh
+++ b/earth_enterprise/rpms/opengee-fusion/post-install.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 #
-# Copyright 2017 the Open GEE Contributors
+# Copyright 2017 The Open GEE Contributors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -19,13 +19,12 @@
 set +x
 
 #------------------------------------------------------------------------------
-
-# Versions and user names:
+# Definitions
 GEE="Google Earth Enterprise"
 GEEF="$GEE Fusion"
-CHKCONFIG="/sbin/chkconfig"
-INITSCRIPTUPDATE="/usr/sbin/update-rc.d"
-BININSTALLPROFILEDIR="/etc/profile.d"
+CHK_CONFIG="/sbin/chkconfig"
+INITSCRIPT_UPDATE="/usr/sbin/update-rc.d"
+PROFILE_DIR="/etc/profile.d"
 
 #------------------------------------------------------------------------------
 # Get system info values:
@@ -62,8 +61,8 @@ main_postinstall()
     echo "$GEE_VERSION" > "$BASEINSTALLDIR_ETC/fusion_version"
 
     chmod 755 "$BININSTALLROOTDIR/gefusion"
-    chmod 755 "$BININSTALLPROFILEDIR/ge-fusion.csh"
-    chmod 755 "$BININSTALLPROFILEDIR/ge-fusion.sh"
+    chmod 755 "$PROFILE_DIR/ge-fusion.csh"
+    chmod 755 "$PROFILE_DIR/ge-fusion.sh"
     chown root:$GEGROUP $BASEINSTALLDIR_VAR/run
     chown root:$GEGROUP $BASEINSTALLDIR_VAR/log
 
@@ -79,14 +78,24 @@ main_postinstall()
 #-----------------------------------------------------------------
 # Post-install Functions
 #-----------------------------------------------------------------
+
+get_xpath()
+{
+  local FILE="$2"
+  local XPATH="$1"
+  
+  local OUTPUT=$(echo "cat $XPATH" | xmllint --shell "$FILE" | sed '/^\/ >/d' | sed 's/<[^>]*.//g')
+  echo $OUTPUT 
+}
+
 setup_fusion_daemon()
 {
     # setup fusion daemon
     printf "Setting up the Fusion daemon...\n"
 
-    test -f "$INITSCRIPTUPDATE" && "$INITSCRIPTUPDATE" -f gefusion remove
-    test -f "$INITSCRIPTUPDATE" && "$INITSCRIPTUPDATE" gefusion start 90 2 3 4 5 . stop 10 0 1 6 .
-    test -f "$CHKCONFIG" && "$CHKCONFIG" --add gefusion
+    test -f "$INITSCRIPT_UPDATE" && "$INITSCRIPT_UPDATE" -f gefusion remove
+    test -f "$INITSCRIPT_UPDATE" && "$INITSCRIPT_UPDATE" gefusion start 90 2 3 4 5 . stop 10 0 1 6 .
+    test -f "$CHK_CONFIG" && "$CHK_CONFIG" --add gefusion
     
     printf "Fusion daemon setup ... Done\n"
 }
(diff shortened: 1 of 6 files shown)


earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -150,7 +157,7 @@ install_or_upgrade_asset_root()
     chmod 644 "$SYSTEMRC"
     chown "$GEFUSIONUSER:$GEGROUP" "$SYSTEMRC"
 
-    if [ "$IS_NEWINSTALL" == true ]; then
```
tst-ppenev: Hm. Is this replacing using `gegroup_existed` and `gefusionuser_existed`?

earth_enterprise/rpms/opengee-fusion/pre-install.sh
```diff
@@ -68,18 +71,16 @@ END
...
 }
 
-create_users_and_groups()
-{
-    # Add user if it does not exist:
-    if [ -z "$(getent passwd "$GEFUSIONUSER")" ]; then
-		mkdir -p "$BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER"
-		useradd --home "$BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER" \
-            --system --gid "$GEGROUP" "$GEFUSIONUSER"
-        keyvalue_file_set "$GEE_INSTALL_KV_PATH" gefusionuser_existed "false"
```
tst-ppenev: `gefusionuser_existed` is still used in `install_or_upgrade_asset_root()`. I'm not sure replacing this with `check_username` is going to work there.

earth_enterprise/rpms/opengee-fusion/pre-uninstall.sh
```diff
@@ -14,12 +14,15 @@
...
+
 set +x
 
 #-----------------------------------------------------------------
 # Main Function
 #-----------------------------------------------------------------
 
-/etc/init.d/gefusion stop
+service gefusion stop
+test -f $CHKCONFIG && $CHKCONFIG --del gefusion
```
tst-ppenev: That's not a good way to write this test. You'll end up exiting with a failure error code, if `gefusion` is not registered as a service.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -61,6 +64,8 @@ main_postinstall()
     chmod 755 "$BININSTALLROOTDIR/gefusion"
     chmod 755 "$BININSTALLPROFILEDIR/ge-fusion.csh"
     chmod 755 "$BININSTALLPROFILEDIR/ge-fusion.sh"
+    chown root:$GEGROUP $BASEINSTALLDIR_VAR/run
```
tst-ppenev: I know `GEGROUP` is not supposed to contain any special characters, but I still like quotes. `BASEINSTALLDIR_VAR` may be less guaranteed to be safe.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -23,6 +23,9 @@ set +x
 # Versions and user names:
 GEE="Google Earth Enterprise"
 GEEF="$GEE Fusion"
+CHKCONFIG="/sbin/chkconfig"
+INITSCRIPTUPDATE="/usr/sbin/update-rc.d"
```
tst-ppenev: Minor: Can we use some word separators in variable names? I have to stare too long to read these.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -108,19 +113,21 @@ compare_asset_root_publishvolume()
...
-                IS_SLAVE=false
-                ;;
-            *)
-                IS_SLAVE=true
-
-                echo -e "\nThe asset root [$ASSET_ROOT] is owned by another Fusion host:  $EXISTING_HOST"
-                echo -e "Installing $GEEF $GEE_VERSION in Grid Slave mode.\n"
-                ;;
-        esac
+        #TODO: xmllint does not have --xpath option
```
tst-lsavoie: We'll need to bring this logic back eventually (I see you've marked it as a TODO already).

earth_enterprise/rpms/opengee-fusion/post-install.sh
tst-dsugar: Because these get used in both server and fusion, they should migrate to gevars.sh, and yes, we_can_add a separator when doing so, it does not hurt at all...

earth_enterprise/rpms/opengee-fusion/post-install.sh
tst-dsugar: I can add quoting.

earth_enterprise/rpms/opengee-fusion/pre-uninstall.sh
tst-dsugar: If it is not registered, that maybe is a failure condition...at least it is presumed that it should have been, since it is registered in the post-install, and this is logic we have used before.

earth_enterprise/rpms/opengee-fusion/pre-install.sh
tst-dsugar: well, that's why it's a wip...I must have missed that.

tst-dsugar: I believe that is now tested for differently, removing that requirement from the script.

tst-ppenev: You replaced in one place, but not in another, so it's still not removed as a requirement.

Also, can you describe what the tests looking for whether `gegroup` and `gefusionuser` are trying to determine, and how you're trying to determine that without the need for checking for `gegroup` and `gefusionuser`?  I don't quite understand what we're looking for, and what the difference between an existing and a new installation of Fusion and Server are.  There's possibly other information we can use, like checking whether there exist files to upgrade, or whether we're installing a new RPM package, or upgrading an existing one.

earth_enterprise/rpms/opengee-fusion/pre-uninstall.sh
tst-ppenev: Whether we've used this logic before, or if it's a failure condition, I think we may still want not to fail in a pre-uninstall script, unless there's a really serious condition that can cause data loss, or some other condition that's hard to recover from.

A user should usually not unregister services manually, but there really ins't anything preventing that from happening, and I don't think that's a good enough reason to prevent a package from being able to uninstall.

earth_enterprise/rpms/opengee-fusion/pre-install.sh
tst-dsugar: If we really need a strong determinant on whether we are doing a new vs upgrade install, there also is a mechanism for doing this by examining the $1 argument of the pre/post scripts, per:
  
https://fedoraproject.org/wiki/Packaging:Scriptlets

Debian pre/post install scripts support something somewhat similar, but I think passes the version of the old package if there is one, and the version of the current package, which can then be compared.

tst-ppenev: I think that may not be the appropriate determinant we want.  I think a lot of this code was added to make the Open GEE 5.2 installers work with GEE installations that existed before open sourcing and used different installers.

tst-dsugar:
This pr will likely become part of and delivered in #583 and if so, this will be closed...

tst-ppenev:
Merged into #583.

PR #586, opened 2017-11-03
I423 - Update openjp2 library to V2.3.0 - take two

tst-eclamar:
Add updated openjp2 library
	Modify SConscript to patch gdal to include V2.3.0
	Add patch file

### DIFF
--- a/earth_enterprise/src/third_party/gdal/gdal-jp2-V2.3.0.patch
+++ b/earth_enterprise/src/third_party/gdal/gdal-jp2-V2.3.0.patch
@@ -1,0 +1,45 @@
+diff -Naur gdal-2.1.2/configure gdal-2.1.2_OJ2-v2.3.0/configure
+--- gdal-2.1.2/configure	2016-10-24 04:00:31.000000000 -0700
++++ gdal-2.1.2_OJ2-v2.3.0/configure	2017-10-26 20:20:31.000000000 -0700
+@@ -23677,12 +23677,12 @@
+         LIBS="-lopenjp2 $LIBS"
+     fi
+   else
+-    for ac_header in openjpeg-2.1/openjpeg.h
++    for ac_header in openjpeg-2.3/openjpeg.h
+ do :
+-  ac_fn_c_check_header_mongrel "$LINENO" "openjpeg-2.1/openjpeg.h" "ac_cv_header_openjpeg_2_1_openjpeg_h" "$ac_includes_default"
++  ac_fn_c_check_header_mongrel "$LINENO" "openjpeg-2.3/openjpeg.h" "ac_cv_header_openjpeg_2_1_openjpeg_h" "$ac_includes_default"
+ if test "x$ac_cv_header_openjpeg_2_1_openjpeg_h" = xyes; then :
+   cat >>confdefs.h <<_ACEOF
+-#define HAVE_OPENJPEG_2_1_OPENJPEG_H 1
++#define HAVE_OPENJPEG_2_3_OPENJPEG_H 1
+ _ACEOF
+ 
+ fi
+@@ -23743,11 +23743,11 @@
+   HAVE_OPENJPEG=yes
+   if test -r $with_openjpeg/include/openjpeg-2.0/openjpeg.h ; then
+     EXTRA_INCLUDES="-I$with_openjpeg/include $EXTRA_INCLUDES"
+-  elif test -r $with_openjpeg/include/openjpeg-2.1/openjpeg.h ; then
++  elif test -r $with_openjpeg/include/openjpeg-2.3/openjpeg.h ; then
+     OPENJPEG_VERSION=20100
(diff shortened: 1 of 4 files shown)


earth_enterprise/src/third_party/gdal/gdal-jp2-V2.3.0.patch
```diff
@@ -0,0 +1,45 @@
...
+ _ACEOF
+ 
+ fi
+@@ -23743,11 +23743,11 @@
+   HAVE_OPENJPEG=yes
+   if test -r $with_openjpeg/include/openjpeg-2.0/openjpeg.h ; then
+     EXTRA_INCLUDES="-I$with_openjpeg/include $EXTRA_INCLUDES"
+-  elif test -r $with_openjpeg/include/openjpeg-2.1/openjpeg.h ; then
++  elif test -r $with_openjpeg/include/openjpeg-2.3/openjpeg.h ; then
+     OPENJPEG_VERSION=20100
```
tst-nfarah: Should this be changed to to 2.3 instead of 2.0 ?

earth_enterprise/src/third_party/gdal/gdal-jp2-V2.3.0.patch
```diff
@@ -0,0 +1,45 @@
...
+     EXTRA_INCLUDES="-I$with_openjpeg/include $EXTRA_INCLUDES"
+   else
+-    as_fn_error $? "openjpeg.h not found in $with_openjpeg/include/openjpeg-2.0 or $with_openjpeg/include/openjpeg-2.1" "$LINENO" 5
++    as_fn_error $? "openjpeg.h not found in $with_openjpeg/include/openjpeg-2.0 or $with_openjpeg/include/openjpeg-2.3" "$LINENO" 5
+   fi
+ 
+   { $as_echo "$as_me:${as_lineno-$LINENO}: checking for opj_stream_set_user_data_length in -lopenjp2" >&5
+diff -Naur gdal-2.1.2/frmts/openjpeg/openjpegdataset.cpp gdal-2.1.2_OJ2-v2.3.0/frmts/openjpeg/openjpegdataset.cpp
+--- gdal-2.1.2/frmts/openjpeg/openjpegdataset.cpp	2016-10-24 04:00:35.000000000 -0700
++++ gdal-2.1.2_OJ2-v2.3.0/frmts/openjpeg/openjpegdataset.cpp	2017-10-27 16:32:51.000000000 -0700
```
tst-nfarah: openjpegdataset.cpp appears to be patched more as indicated in [https://github.com/google/earthenterprise/pull/580](url)

See my previous comment in that PR refering to [https://trac.osgeo.org/gdal/log/trunk/gdal/frmts/openjpeg/openjpegdataset.cpp?rev=40607](url) changes.

earth_enterprise/src/third_party/gdal/gdal-jp2-V2.3.0.patch
tst-eclamar: I don't understand your question.  It's being changed from '2.1' to '2.3'...  I'm leaving 'OPENJPEG_VERSION' alone.  Is it that to which you are referring?

tst-eclamar:
Opened in error - please see PR #580

earth_enterprise/src/third_party/gdal/gdal-jp2-V2.3.0.patch
tst-nfarah: I was referring to line 26 as its appear to have the version number: 20100.

tst-eclamar:
Noureddine: Can you add the comment to the current PR?  For some reason, I
cannot see it when I go to github.com/google/earthenterprise.git.  Is it
not appearing because you commented in a close PR?

The OPENJPEG_VERSION variable is used, and only has meaning, inside of the
configure script.  Also, it is set but never used.

tst-nfarah:
If its not used, then may be no issue. In the original PR [https://github.com/google/earthenterprise/pull/580](url) I did add comment that 'openjpegdataset.cpp' needs to be patched more. See my comment there.

tst-eclamar:
What, specifically, needs to be patched?  I've taken a look at the diffs
for the two that you mentioned and they are quite extensive.  Is there a
pressing need for this?  Can we limit the scope to something small?

tst-ccamp:
@tst-nfarah I think I agree with @tst-eclamar here. Those changes do look pretty substantial. My original request was to not modify the gdal source archive. I think the additional changes could be made in a follow-up ticket. If we still get excellent performance without these changes, they may not be needed. However, if they are bug fixes, then that's something we can look into doing sooner.

tst-nfarah:
@tst-ccamp @tst-eclamar 
let's move this discussion to the open PR: https://github.com/google/earthenterprise/pull/580
As this PR is closed.

I'll add a comment there.

PR #587, opened 2017-11-03
fix #359 (fusion): Gracefully handle errors when saving

### LINKED ISSUE #321: RasterAsset :Adding more source files does not trigger IsModified or "Save"
1. Create a imagery resource by fusion -> asset manager-> new imagery resource-> provide source file
2. save( note that  none of the other fields are changed)
3.  Add another source files to same asset - make sure not to close the New asset window(dialog) 
4.  Now see that Save option is still grayed( the model is not updated with the new source file added , hence no indication that the Imagery resource has modified)
5. Now close the resource - > notice that "Changes have been made to the resource , do you want to save the resource" is not presented to the user

### LINKED ISSUE #335: Second change in the same Raster Asset UI does not trigger "Save"
1. Create a Imagery Resource with valid file from correct source volume , specify the date on the imagery 
2. Now Save( this should be the first Save on this UI)
3. now edit the date field to change some value 
4. Close
5. Note two issues 
   a) Save is disabled even after a change to already Saved resource and
   b) Close does not evaluate IsModified to be TRUE hence no dialog " Resource is modified, do you want to save the changes ?"

### LINKED ISSUE #359: Save option is disabled in the fusion UI when there is an error saving
The save option is disabled in the resource creation form when there is an error saving.  Thus, if you fix the error, you still cannot save the resource without closing the form and opening it again.  Some manifestations of this ticket were fixed previously in #167.  This ticket is to fix the problem for file errors (unsupported file types or no file).  Thanks to @protodonuts for the original bug report.

1. In Fusion, open "Asset Manager", click "New Imagery Resource", add some invalid file(s) (e.g. unsupported file, or Tiff file without SRS, or no file at all) to the "Source File(s)" list.
2. Select File menu -> Save. Specify a resource name, then click "Save".

After an error message shows up as expected, there are problems:

- "Save" in File menu is disabled.
- "Save As..." looks OK, but any attempt to correct the "Source File(s)" list would be futile. The resource cannot be saved because the same error is raised again.

To proceed, I have to close the window and open a new one.

(Built on Ubuntu 14)

tst-jblair:
Fixes #359 
- Canceling a Save no longer disables saving.
- Save error state can be updated from within the UI (and thus cleared when the sources list is modified).
- Fixes a bug where Saving generates a new window class, which is completely disconnected from the actual window that the User is interacting with.

Fixes #335 
- Changing a value in the Resource dialog now registers as needing a save.

Fixed #321 
- Adding/removing images in the Resource dialog now registers as needing a save.

### DIFF
--- a/earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
+++ b/earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
@@ -112,8 +112,6 @@ template <class Defs, class FinalClass>
     // Re-assemble request after updating model-object.
     request = AssembleEditRequest();
   }
-  std::string buf, ref;
-  request.SaveToString(buf, ref);
 
 
   if (Defs::kSubmitFunc(request, *error_msg, 0 /* timeout */)) {
--- a/earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
+++ b/earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
@@ -97,12 +97,12 @@ bool AssetDerived<Defs, FinalClass>::IsModified() {
 }
 
 template <class Defs, class FinalClass>
-  bool AssetDerived<Defs, FinalClass>::SubmitEditRequest(QString* error_msg, bool save_error_) {
+  bool AssetDerived<Defs, FinalClass>::SubmitEditRequest(QString* error_msg, bool last_save_error) {
   typename Defs::Request request = AssembleEditRequest();
-  if (!save_error_ && request == saved_edit_request_ )
+  if (!last_save_error && request == saved_edit_request_ )
     return true;
 
-  if (save_error_ || request.assetname != saved_edit_request_.assetname) {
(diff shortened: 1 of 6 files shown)


earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
```diff
@@ -127,11 +128,13 @@ template <class Defs, class FinalClass>
...
+void AssetDerived<Defs, FinalClass>::Init(bool re_init) {
   // defer to final class
   final()->FinalPreInit();
 
-  InstallMainWidget();
+  if (re_init == false) {
+    InstallMainWidget();
+  }
   SetName(saved_edit_request_.assetname);
   SetMeta(saved_edit_request_.meta);
```
tst-dsugar: I do not believe main_widget_->Prefill() should be called if we do not call InstallMainWidget(), and this seems to be what is adding duplicate resources in the imagery project in the issue I was looking at.  I need to do some more testing, since this is also tied to behavior of other asset project types.  @tst-lsavoie @tst-jblair

Issue #528, opened 2017-09-29
scons file inconsistencies vis a vis Ubuntu


tst-basilhuffman:
Effects **earth_enterprise/src/SConstruct**, starting on line **152**:

    # setup platform specific operations
    is_hardy = True
    # if non-zero return status it is not hardy
    #  Warning: assume your Ubuntu version is 10.04.02, otherwise need change
    #the version number in order to complie
    if os.WEXITSTATUS(os.system('awk \'/Ubuntu/ { split($2, array,"."); \
            if(!(array[1]>=10 && array[2]>=04)) exit(1) }\'  /etc/issue')):
       print 'Unknown Unbuntu version, require 10.04.x or higher.\
       Please check /etc/issue for version number'
       Exit(1)
       is_hardy = False

There are a number of issues with this:

1. This script validates on anything 10.04 or higher. Nothing before 14.04 is supported now.
2. Some newer Ubuntu flavors have ditched convention and changed the contents of /etc/issue from e.g. **Ubuntu X.Y \n \l** to **Ubuntu \<VERSION NAME\> \n \l**
3. Misspelling of Ubuntu (spelled Unbuntu)

Perhaps there needs to be a better way to approach this, as support will be expanded in the future.

**This only effects Ubuntu**

tst-jcupp:
Closed per: https://github.com/google/earthenterprise/pull/592

PR #592, opened 2017-11-07
issue #528: chore (build): address scons file inconsistencies vis a vis ubuntu

tst-basilhuffman:
addresses issue #528

### DIFF
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -151,11 +151,6 @@ else:
 
 # setup platform specific operations
 is_min_ubuntu = True
-# if non-zero return status it is not hardy
-# Warning: assume your Ubuntu version is 10.04.2, otherwise need change 
-#the version number in order to compile
-
-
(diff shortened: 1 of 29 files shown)


earth_enterprise/src/SConstruct
```diff
@@ -150,16 +150,27 @@ else:
   python_include = '/usr/include/python%s' % python_major_version
 
 # setup platform specific operations
-is_hardy = True
+is_min_ubuntu = True
 # if non-zero return status it is not hardy
```
tst-jblair: This comment block is out of sync with the code, now. Presumably it should just be removed?

tst-basilhuffman: I dont follow

tst-basilhuffman: Sorry Jim, I misunderstood your comment. No issues, will change.

tst-mswartz:
This looks ok to me ... As an aside though (and I know we don't support Mint) but my `lsb_release` returns 18.2

Issue #356, opened 2017-07-19
Update to 5.2.1 Version Number


tst-ccamp:
Update version number and create release candidate branch.

tst-awankhede:
more details would help, may be a link to release process or the set of steps.
 or a previous release issue number

Issue #132, opened 2017-05-04
Differentiate between development builds and release builds


tst-ccamp:
Anywhere the version number is displayed to the user (installer, admin console, fusion UI) it should indicate whether the product is an official release or a development build. There isn't currently a mechanism in place to do this.

tst-ccamp:
Closing in favor of #632

PR #597, opened 2017-11-09
build: Improve version strings

tst-jblair:
Remove redundant version-string-generating functions, and set it up so that they all derive from one place.

This one place uses git's tag history to find the most recent tag that looks like a version number, and the number of commits since that tag occurred. From this information, a commit-specific version number is generated, and stored in two "base" files (one is plain text, one is C code).

Future modifications could potentially remove some redundant files that contain little more than the version strings themselves.


tst-ccamp:
I'm a little concerned about this after learning that it ends up tagging the asset root with the full version number. Traditionally it doesn't try to upgrade the asset root for every simple little patch. That needs to stick with the 5.2.0 or 5.2.1 and not include the full version string. 

We need to determine where we should use the full version string vs. where we need to stick with the simple version string.

tst-jcupp:
will fix #132

tst-jcupp:
will also fix: #356

tst-jcupp:
@tst-ccamp is the asset root the only place where are more "bare version" e.g. 5.2.1 needs to be applied instead of the full version string?

tst-jcupp:
got two approvals, waiting on a final review from @tst-ccamp

tst-jcupp:
@tst-jblair we will need to support the use-case of building without .git, plus other caveats w.r.t. deb/rpm filenames...

tst-ppenev:
The Portable Server build script should also be updated to use the new version files.

tst-jblair:
There are src/NATIVE*/gee_version* files with the short and long version names in them, which are created during a build.

There is also src/version.txt which gets updated during a build to have the short version number; but, it is also intended to be checked into the repository, and it will be queried for version information if git is not available.

So, src/version.txt is effectively taking the place of src/fusion_version.txt (which I don't believe has any auto-update capability). But, I'm not removing fusion_version.txt from the repo at this point, just in case there are places that I'm not aware of that might still query it (as the scripts tst-ppenev reference evidently do).

tst-jblair:
Implementation details have been updated.
- Without git, a src/version.txt file is referenced to git the latest known build number.
- With git, src/version.txt is updated so that the value can be stored in the repository.
- If you are on a version-style tag commit, that tag will be returned.
- If you have a dirty repository, the datestamp is appended to the latest tag.
- Otherwise, the datestamp and current commit (short) hash are appended.
- Revision and build numbers and alpha -> beta -> release cycles are updated as appropriate.

Recommended testing:
- Tag your local repo with a version-like string such as "5.2.1-1.beta".
- Build and check output in src/NA*/gee*txt and src/version.txt
- Make a file change to make the repo dirty.
- Build and check output files.
- Commit that change (git status should be clean, except maybe for src/version.txt)
- Build and check output files.

- Tag your repo with a release style tag, such as "5.2.1-2.release".
- Repeat the above remaining steps.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/geecheck_tests/common.py
```diff
@@ -20,7 +20,7 @@
 import platform
 import socket
 
-LATEST_VERSION = '5.2.0'
+LATEST_VERSION = '5.2.1'
```
tst-ppenev: Are we going to update this for every version of GEE?

earth_enterprise/src/portableserver/build.py
```diff
@@ -140,7 +140,8 @@ def get_version(self):
         """Parses version information, and sets member variables."""
 
         if self.base_version is None:
-            path = os.path.join(self.source_dir, 'fusion_version.txt')
+            # This was fusion_version.txt; file has moved.
```
tst-ppenev: I don't think comments about non-existent files are very helpful to future developers.  I think we can do without this one.

earth_enterprise/src/earthplugin/SConscript
```diff
@@ -20,47 +20,18 @@ import os
...
-WriteBuildInfoFile(build_info, server_version_file_path)
-
 # Copy files from '../earth_enterprise/src/earthplugin/..' directory.
 env.installDirExcluding('earth', Dir('#earthplugin').abspath, ['SConscript'])
 
 # Copy files from '../earth_enterprise/src/{REL_XXX|OPT_XXX}/earthplugin/..'
-# directory. There is only one file 'version.txt' in this folder.
-env.installDirExcluding('earth', '.', [])
+# directory.
+env.install('earth', File('../builddate.txt').abspath)
```
tst-ppenev: Do we need 3 files to store a version?  Are the places that read it unable to extract specific information from one file?

earth_enterprise/src/earthplugin/SConscript
```diff
@@ -20,47 +20,18 @@ import os
...
+env.install('earth', File('../gee_long_version.txt').abspath)
```
tst-ppenev: Can we get some comments clarifying the difference between `version` and `long_version`?

tst-jcupp:
Are we using UTC datestamps and can we use Seconds in that datestamp to avoid CI collisions?

tst-dsugar:
This may need to be reconciled with #606

tst-ccamp:
These changes are now being addressed on #630, so I'm closing this one.

Issue #595, opened 2017-11-09
Move OpenJPEG 2.3.0 into third party build


tst-eclamar:
As a Open GEE user, I want to have OpenJPEG 2.3.0 built from official sources, so that I have confidence about the software being used in the product.
AC
add Open JPEG to the third party builds
remove the .so file from git lfs sources, tarball up the sources instead of the .so
verify jenkins can build it with existing build jobs

PR #598, opened 2017-11-10
refactor #595 (build): move open jpeg 2.3.0 into third party build

tst-eclamar:
1. Remove openjpeg binary tarball from inclusion into gdal.
2. Add openjpeg source tarball;  add to compile and install process.


tst-eclamar:
Why does Travis continue to fail?  It compiles correctly on my system and under jenkins...

tst-jcupp:
@tst-eclamar do we need to modify the travis-ci pre-build setup?

tst-jcupp:
will fix #595

PR #600, opened 2017-11-10
feat #566: Upgrade postgres and postgis versions

tst-nfarah:
Goal of this PR is to upgrade PostgreSQL to version 9.6.5 from 8.4.22 and upgrade PostGIS to version 2.3.4 from 1.5.8.

This will partially help resolving #566 

## Made Changes:
1- How PostGIS is enabled inside PostgreSQL . In the used (old) PostGIS version it's done through the execution of two  .sql files (postgis.sql and spatial_ref_sys.sql). With the new PostGIS version this is done through the execution of the command 'CREATE EXTENSION postgis'
2- Removed the installation of 'PL/pgSQL' since starting PostgreSQL 9.0 it is now installed by default
("createlang...r plpgsql ..." command is not needed any more)
3- Updated used PostGIS syntax for spatial functions. Many needed to have "ST_" pefix and few their names were changed.
4- Updated the 'geplaces' and 'searchexample' databases (compressed in .gz) to be compatible with how geometry are now handled in postGIS (views and not tables any more).

## Verification Steps:

**Basic verification:**
- Get the PR, build OpenGEE and install it
- Verify the now installed version of PostgreSQL: $postgres -V
--> This should yield "9.6.5"
- Verify the installed version of PostGIS: using psql try the command "SELECT PostGIS_Full_Version();". Verify that is "2.3.4".

**Functionality verification:**
- Create a vector resource enabling a search on it (Layer properties)
- Create a 3D database with a vector project
- Push that database
- Publish the database with 'POI search', 'Places' and 'Search example' search tabs
- Using Google Earth Enterprise Client, verify that published database display ok
- Try the 'POI Search', 'Places' and 'Search Example'.


earth_enterprise/third_party/postgresql/README.google
```diff
@@ -1,5 +1,5 @@
 URL: http://www.postgresql.org/developer/
-Version: 8.4.22
```
tst-lsavoie: Similar changes need to be made in the README.google file for postgis.

tst-ppenev: There are also the `README.txt`s under `src/third_party` that also have version numbers in them, and are probably outdated.  I think we can remove the `.spec` files from `src/third_party`.

tst-nfarah: Good points. I, now, made those changes: updated readme files and removed stale and not needed .spec files.

tst-ccamp:
In my testing on RHEL 6 (probably occurs on other platforms as well), when I upgrade an existing installation of GEE Server, PGSQL fails to start and displays this error in the log file: 
```LOG:  skipping missing configuration file "/var/opt/google/pgsql/data/postgresql.auto.conf"
FATAL:  database files are incompatible with server
DETAIL:  The data directory was initialized by PostgreSQL version 8.4, which is not compatible with this version 9.6.5.
LOG:  skipping missing configuration file "/var/opt/google/pgsql/data/postgresql.auto.conf"
```

Any databases that already exist should be upgraded to be compatible with the new version in the installer.

PR #606, opened 2017-11-13
feat #605 (rpms): Target Platform in RPM File Names

### LINKED ISSUE #605: Indicate Target Platform in RPM File Names
As an Open GEE RPM user, I want to be able to tell what platform a given RPM would work on from its file name.

For example, shared libraries used by Open GEE have different versions on RHEL 6 and RHEL 7.  Therefore, an installation of an RPM containing Open GEE compiled for RHEL 6 would fail to run on RHEL 7, because the linked would not find the referenced shared libraries.  As a user, I want to be able to tell whether I can install a given RPM on my platform from its file name.

## Notes:

Conventionally, Cent OS 6 and RHEL 6 user the `el6` ID in RPM file names, and Cent OS 7 and RHEL 7 use `el7`.  Use the conventional names that users are already familiar with.

For example RPM packages may have names like these:

```
opengee-common-5.2.0-1.el6.x86_64.rpm
opengee-fusion-5.2.0-1.el6.x86_64.rpm
opengee-postgis-1.5.8-1.el6.x86_64.rpm
opengee-server-5.2.0-1.el6.x86_64.rpm
```

## AC:

* The RPM build tasks should produce RPM file names that identify the platform the RPMs can run on.

tst-ppenev:
Fixes #605.

**Note: This pull request is from a branch off of #583.  It's intended to be reviewed after that pull request is merged.**

* Added target platform ID (`el6` for Cent OS and RHEL 6 and `el7` for Cent OS and RHEL 7) to RPM file names.


tst-dsugar:
I think this is generally correct, though may touch upon other versioning changes.  I think somewhere we should also disable rpm package generation on ubuntu, and selectively disable deb generation on rhel/centos.

tst-ppenev:
The Docker build scripts only build packages for the package manager that exists on the guest OS, so only RPMs are built on RHEL and CentOS, and only Deb's are built on Ubuntu.

tst-ppenev:
I'm also considering moving this to a standalone Groovy class, since that build file has been getting large.

tst-dsugar:
@tst-ppenev  Ah...while I actually built RPMs on ubuntu at one point thru the install_package, I also explicitly installed rpm with apt to do it.

tst-ppenev:
The branch I was working on to fix RPM dependencies, actually, breaks building RPMs on Ubuntu. It requires `rpm-build`, as well as the database that lists what RPM package which capabilities.

The Deb package dependency detection is similar, and I don't think it would work on CentOS or RHEL.

earth_enterprise/rpms/buildSrc/src/main/groovy/com/tsciences/os/Platform.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+class Platform {
+    static private String osId = null;
+    static private String osVersionId = null;
+
+    static def obtanOsAndVersionIds() {
+        osId = null
+        osVersionId = null
+
+        if (new File('/etc/os-release').exists()) {
```
tst-dsugar: os-release actually is a good choice to use.  While not yet a concern, I did test and it does produce a valid result on ubuntu.

earth_enterprise/rpms/buildSrc/src/main/groovy/com/tsciences/os/Platform.groovy
```diff
@@ -0,0 +1,64 @@
...
+                    osVersionId = osVersionId.find(/[0-9]+/)
+                }
+            }
+        } else if (new File('/etc/system-release-cpe').exists()) {
+            def osReleaseComponents = new File('/etc/system-release-cpe').text.split(':')
+            // The extract only the number from version strings like '6server':
+            osId = osReleaseComponents[2]
+            osVersionId = osReleaseComponents[4].find(/[0-9]+/)
+        }
+    }
```
tst-dsugar: minor note: of course if we have neither, I presume osVersionId and osId remain null, which means getOsVersionId() & getOsId() will check again later in vein on the next rpm.  My suggestion, have it default to something like "unknown" and "0" if there is no result.

earth_enterprise/rpms/buildSrc/src/main/groovy/com/tsciences/os/Platform.groovy
```diff
@@ -0,0 +1,64 @@
+package com.tsciences.os
+
+class Platform {
+    static private String osId = null;
+    static private String osVersionId = null;
+
+    static def obtanOsAndVersionIds() {
```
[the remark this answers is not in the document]
tst-ppenev: :stuck_out_tongue: Yes, it was.  I'll fix that.

earth_enterprise/rpms/buildSrc/src/main/groovy/com/tsciences/os/Platform.groovy
tst-ppenev: If they're `null`, they'll expand to the `null` string, which somewhat like `unknown`, or `0`.  However, that can only happen on unsupported platforms where we don't have a defined build behavior.

tst-lsavoie:
Two approvals.  Merging.

PR #615, opened 2017-11-21
test: Add publish step to test scripts

### LINKED ISSUE #616: Add publish step to testing scripts
The `publish_tutorial.sh` and `publish_historical.sh` scripts push their respective databases but don't actually publish them.  We should add the publish step to those scripts to make testing easier.

tst-lsavoie:
This adds a missing step to some of the test scripts.  I haven't finished testing these changes but I'm creating the PR now so that people can start reviewing them.

Fixes #616.


tst-lsavoie:
I've now tested the code and fixed one minor bug.  It works fine.

earth_enterprise/src/fusion/tools/gee_test/publish_historical.sh
```diff
@@ -23,15 +23,20 @@ echo "Using asset root: $ASSET_ROOT"
 
 /opt/google/bin/geserveradmin --adddb "$ASSET_ROOT/Tutorial/Databases/SFinset_1.kdatabase/gedb.kda/ver001/gedb"
```
tst-nfarah: This works just fine, but may be use a simpler path for the database:
/opt/google/bin/geserveradmin --adddb Tutorial/Databases/SFinset_1.kdatabase

So we don't have to point to the low level file ... If you need that specific version, i would use:
/opt/google/bin/geserveradmin --adddb Tutorial/Databases/SFinset_1.kdatabase?version=1

PR #617, opened 2017-11-21
Opgee 265 mr sid in build

tst-eclamar:
Add a script to be run by jenkins to add MrSID to the build.

### DIFF
--- a/scripts/get-MrSID-environment.sh
+++ b/scripts/get-MrSID-environment.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+
+#
+# Copyright 2017 Open GEE Contributors
+#
+# Licensed under the Apache license, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the license.
+#
+
+# This scripts installs packages, downloads and builds software used to
+# download the Open GEE source code and build it.
+#   It runs on all supported platforms.
+#   This script does not share code with other shell scripts so that it can be
+# distributed as a single file that can be used to obtain a build environment
+# and source code for Open GEE.
+
+SELF_NAME=$(basename "$0")
+
+is_string_value_false()
+{
+    case $(echo "$1" | tr '[:upper:]' '[:lower:]') in
+        false|off|no|0)
+            return 0
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+}
+
+
+#===  FUNCTION  ================================================================
+#          NAME:  check_root
+#   DESCRIPTION:  checks to see if the user has root permissions
+#    PARAMETERS:  none
+#       RETURNS:  nothing
+#===============================================================================
+function check_root ()
+{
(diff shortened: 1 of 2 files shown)


scripts/get-gee-build-environment.sh
```diff
@@ -32,14 +32,14 @@ MIN_GCC_VERSION="4.8"
 : ${PACKAGE_BUILD_DIR:=""}
 
 # Whether to delete the package build directory before exiting:
-: ${CLEAN_PACKAGE_BUILD_DIR:="yes"}
+: ${CLEAN_PACKAGE_BUILD_DIR:="no"}
```
tst-ppenev: Is changing the default values necessary for building MrSID?

scripts/get-gee-build-environment.sh
```diff
@@ -32,14 +32,14 @@ MIN_GCC_VERSION="4.8"
 : ${PACKAGE_BUILD_DIR:=""}
```
tst-ppenev: This file has been removed from this repository.  This change should not be part of this pull request.

scripts/get-MrSID-environment.sh
```diff
@@ -0,0 +1,65 @@
...
+        echo "$SELF_NAME Error: Must be root!"
+	echo
+	exit 1
+    fi
+}
+
+check_root
+
+MRSIDTGZ="MrSID_DSDK-9.5.4.4703-rhel6.x86-64.gcc482.tar.gz"
+echo "MRSIDTGZ=$MRSIDTGZ"
```
tst-ppenev: What is the purpose of this `echo` command? Is it just for debugging?

scripts/get-MrSID-environment.sh
```diff
@@ -0,0 +1,65 @@
...
+# This scripts installs packages, downloads and builds software used to
+# download the Open GEE source code and build it.
+#   It runs on all supported platforms.
+#   This script does not share code with other shell scripts so that it can be
+# distributed as a single file that can be used to obtain a build environment
+# and source code for Open GEE.
+
+SELF_NAME=$(basename "$0")
+
+is_string_value_false()
```
tst-ppenev: These functions look like code may be used elsewhere, and that can be put in a shell library and re-used.

scripts/get-MrSID-environment.sh
```diff
@@ -0,0 +1,65 @@
...
+    fi
+}
+
+check_root
+
+MRSIDTGZ="MrSID_DSDK-9.5.4.4703-rhel6.x86-64.gcc482.tar.gz"
+echo "MRSIDTGZ=$MRSIDTGZ"
+wget http://bin.lizardtech.com/download/developer/$MRSIDTGZ -O /tmp/$MRSIDTGZ
+tar -xzf /tmp/$MRSIDTGZ -C /opt
+echo /opt/MrSID_DSDK-9.5.4.4703-rhel6.x86-64.gcc482/Raster_DSDK/lib >>/etc/ld.so.conf
```
tst-ppenev: This will keep adding lines to `/etc/ld.so.conf`, and there's no counterpart that removes them.

I think it's better to specify a directory on the `ldconfig` command line, or, alternatively, add a file to `/etc/ld.so.conf.d/`.

scripts/get-gee-build-environment.sh
tst-basilhuffman: When running the script on centos6, the build died before the MrSID script was even executing. It looked to be dying around the time it was getting to cleanup of the gtest rpm. It wasn't happening in any of the other platforms so I suggested turning that flag off and seeing if it had an effect.

I assume it must have had an effect. Were there any issues vis a vis centos6 initially? Rhel6?

I'm a bit confused as to why -c exists if it defaults to yes also...

PR #618, opened 2017-11-27
doc #620: Add cmake to the build process

tst-eclamar:
Just add 'cmake' to the Markdown READMEs.


tst-jcupp:
Fixes #620

tst-nfarah:
could we update the ReadMe part of this fix ? so we have one PR and also since they seems to be one issue: adding cmake to list of pre-requists and documenting it.

tst-nfarah:
so #620 is already fixed ? as it's status is currently open.

PR #619, opened 2017-11-27
Opgee 265 mr sid in build

tst-eclamar:
Just remove old/erroneously included script.


tst-ccamp:
We don't need this additional script

PR #621, opened 2017-11-28
add (build):  scons clean targets

tst-mswartz:
deletes scons metadata with -c distclean  option, deletes all build data with -c buildclean option ..  in order to not delete third_party build data, khEnvironment.py must be modified, or else the lines in the third_party SConstructs that call ExecuteOnClean() have to be removed or changed, because right now the third_party data is deleted intentionally and I wasn't sure about changing that as part of this ticket


earth_enterprise/SConstruct
```diff
@@ -180,7 +180,24 @@ env.PhonyTargets(**this_dict)
...
+env.Clean("distclean", [
+  ".sconsign.dblite",
+  ".sconf_temp",
+  "config.log",
+  "src/.sconsign.dblite",
+  "src/.sconf_temp",
+  "src/config.log"
+  ])
+
+if 'buildclean' in COMMAND_LINE_TARGETS:
```
tst-basilhuffman: Is the -c option necessary if 'buildclean' is present at all? Looks like "scons buildclean" and "scons -c buildclean" are equivalent?

tst-mswartz: Ok it's actually necessary for distclean but not buildclean

tst-basilhuffman: Do we want to have consistency? I suppose so, right?

tst-mswartz: I think it's just easiest to call it with the -c ...  if i were to stick the buildclean stuff inside a single Clean() target it wouldn't be able to call the next SConstruct down the line (in the src directory), which does require a -c for buildclean

tst-basilhuffman: So are you saying, don't worry about buildclean being after the -c option? Not really following what you're saying here...

tst-mswartz: I'm saying, it is consistent as long as you use the -c option. Was that your criticism? It wasn't very clear.

tst-jcupp:
What are the advantages of only cleaning scons metadata using `distclean` over the full clean?

earth_enterprise/SConstruct
tst-basilhuffman: If its the only way to do it, then so be it. It just seems more logical to me to only allow buildclean with -c and not on its own

tst-mswartz: The -c in scons denotes something that literally has a Clean target defined in the scons sense of the word, which seems to require that it fit inside a list or list comprehension that produces the files which are to be deleted as part of that Clean. If we want to do things other than delete files in a list, it has to be defined in a different way, which means it is not a scons Clean target, which means it doesn't get a -c.

tst-basilhuffman: Sounds ok. Approve.

tst-jcupp:
Can we update the original description with usage: an example command, etc. I tried to issue one command with both targets -c cleanall build, but it didn't work as expected (I'm used to gradle behavior).

tst-mswartz:
The original description on this page?   I think it works more like make than like gradle, at least in that aspect.

tst-eclamar:
Can you explain what I'm seeing?  The 'cleanall' doesn't seem to do just that - what is the agreed upon scope of the clean?

I did a diff of the directory, before-and-after, and get a very small delta:
diff -qr earth_enterprise/ before/earth_enterprise/
diff: before/earth_enterprise/docs/landing_page/manual: No such file or directory
Files earth_enterprise/src/config.log and before/earth_enterprise/src/config.log differ
Only in before/earth_enterprise/src/.sconf_temp: conftest_0
Files earth_enterprise/src/.sconf_temp/conftest_0.o and before/earth_enterprise/src/.sconf_temp/conftest_0.o differ
Files earth_enterprise/src/.sconf_temp/conftest_1.o and before/earth_enterprise/src/.sconf_temp/conftest_1.o differ
Files earth_enterprise/src/.sconsign.dblite and before/earth_enterprise/src/.sconsign.dblite differ

What am I missing?

tst-mswartz:
Can you please explain what command you ran and from which directory?  scons release=1 -c cleanall will delete  /tmp/fusion_os_install, scons metadata in the current directory and in src, also src/NATIVE-REL-x86_64 contents (except for a couple things in third_party)

tst-mswartz:
Oh actually it looks like it's stopped cleaning the scons metadata since i changed it from distclean to cleanall... i will look into it now. I wonder what happened, it's definitely in that block.

tst-mswartz:
The reason it wasn't deleting all the scons metadata (but the files Eric mentioned were different) was because it performed some clean actions after deleting those files and thus recreated them with data about deleting things.   Similarly, i can't use Clean() or Delete() to delete the files because those are scons functions and recreate the metadata files after removing them.

PR #625, opened 2017-12-06
Postgres upgrade

dilykat:


dilykat:
update

tst-ccamp:
@dilykat This branch isn't ready to be merged to master yet, because there is still work to do to handle the case of upgrading the data. I'm working on that currently.

Issue #58, opened 2017-04-08
how to install and deploy them (gee fusion and gee server)


bugyak:

tst-ccamp:
Hi @bugyak, the build instructions are located on the GitHub repo wiki. https://github.com/google/earthenterprise/wiki/Build-Instructions

The installation scripts are in work and being tracked under separate issues. In the meantime there is an experimental branch that you are welcome to try out. 
https://github.com/google/earthenterprise/tree/fusion_installer

When the install scripts are finalized, the wiki will be updated with instructions on using them.

PR #627, opened 2017-12-07
groovy...passing build info from scons

### LINKED ISSUE #626: scons installdir=xxx option not passed to RPM build
If installdir is used in scons, the rpm's still try to populate from the default /tmp/fusion_os_install, producing mostly empty rpms.

tst-dsugar:
Closes #626 by passing config state from scons to gradle build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.creator.user
 /*.files
 /*.includes
+
+# Internal build support
+.buildinfo
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,3 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.files
 /*.includes
 
-# properties files
-*.properties
-
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,6 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.files
 /*.includes
 
-# Internal build support
-.buildinfo
+# properties files
+*.properties
+
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -26,8 +26,8 @@ EnsureSConsVersion(0, 96, 92)
 
 # process commandline arguments
 installdir = ARGUMENTS.get('installdir', '/tmp/fusion_os_install')
-with open('.buildinfo', 'w') as f:
-    print >> f, str('PREFIX="')+installdir.strip()+str('"')
+with open('rpms/build.properties', 'w') as f:
+    print >> f, str('prefix=') + installdir.strip()
 
 installdir = Dir(installdir)
 def dir_cat(installdir, subdir1, subdir2):
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -27,7 +27,9 @@ EnsureSConsVersion(0, 96, 92)
 # process commandline arguments
 installdir = ARGUMENTS.get('installdir', '/tmp/fusion_os_install')
 with open('rpms/build.properties', 'w') as f:
-    print >> f, str('prefix=') + installdir.strip()
+    print >> f, "## Auto-generated file from scons."
+    print >> f, "## Manual changes will be overwritten."
+    print >> f, str('staged=') + installdir.strip()
 
 installdir = Dir(installdir)
 def dir_cat(installdir, subdir1, subdir2):
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -26,6 +26,8 @@ EnsureSConsVersion(0, 96, 92)
 
 # process commandline arguments
 installdir = ARGUMENTS.get('installdir', '/tmp/fusion_os_install')
+with open('.buildinfo', 'w') as f:
+    print >> f, str('PREFIX="')+installdir.strip()+str('"')
 
 installdir = Dir(installdir)
 def dir_cat(installdir, subdir1, subdir2):
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -55,8 +55,19 @@ def openGeeVersion =
         it // The first string that remains is the version string.
     }
 
+def stagedInstall_path = '/tmp/fusion_os_install'
+if (new File('../.buildinfo').exists()) {
+    new File('../.buildinfo').readLines().each {
+        if(it.startsWith('PREFIX=')) {
+            stagedInstall_path = it.substring('PREFIX='.length())
+            if(stagedInstall_path.startsWith('"')) {
+                stagedInstall_path = stagedInstall_path.substring(1, stagedInstall_path.length() - 1)
+            }
+        }
+    }
+}
 
-def stagedInstallDir = new File('/tmp/fusion_os_install')
+def stagedInstallDir = new File(stagedInstall_path)
 def stagedInstallDir_common = new File(stagedInstallDir, 'common')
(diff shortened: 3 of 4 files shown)


earth_enterprise/SConstruct
```diff
@@ -26,6 +26,8 @@ EnsureSConsVersion(0, 96, 92)
 
 # process commandline arguments
 installdir = ARGUMENTS.get('installdir', '/tmp/fusion_os_install')
+with open('.buildinfo', 'w') as f:
```
tst-ppenev: I think we should pass this on the command line, or use a Gradle settings file, rather than implementing our own.

tst-dsugar: @tst-ppenev  This is an entirely valid point of view.  I would be happy to discuss this.

tst-ppenev: The way it used to work before is that the `stageOpenGeeInstall` task would copy the install files under the build directory.

Passing a value in a file creates a number of possible problems.  E.g., you can get undefined behavior if you try to run two tasks that use the file concurrently.

You can just add another project property.  E.g., `stageInstallDirectory`, and pass it on the command line like the `buildOpenGee` one.

On another note, I'm not sure it's a good idea to have tasks for both compiling and packaging in the same Scons file.  It may be contributing to poor isolation.

.gitignore
```diff
@@ -20,3 +20,7 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.creator.user
 /*.files
 /*.includes
+
+# properties files
+*.properties
```
tst-ccamp: Sorry, we can't blanket ignore all ".properties" files. Plus, this is a common type and it might frustrate devs who need to add a new file and can't figure out why  I've found two legitimate ones that are checked in. You'll need to figure out how to more specifically filter out the properties files you don't want included.
```
$ find . | grep properties
./src/server/wsgi/conf/postgres.properties
...
./rpms/gradle/wrapper/gradle-wrapper.properties
```

tst-dsugar: We can just ignore this specific one in rpms/build.properties then...or call it something else, like I originally did with .buildinfo.

earth_enterprise/SConstruct
```diff
@@ -26,6 +26,8 @@ EnsureSConsVersion(0, 96, 92)
 
 # process commandline arguments
 installdir = ARGUMENTS.get('installdir', '/tmp/fusion_os_install')
+with open('rpms/build.properties', 'w') as f:
+    print >> f, str('prefix=') + installdir.strip()
```
tst-ccamp: Is there any mechanism that would avoid adding this over and over if you build more than once without cleaning?

earth_enterprise/rpms/build.gradle
```diff
@@ -55,8 +55,16 @@ def openGeeVersion =
         it // The first string that remains is the version string.
     }
 
+def stagedInstall_path = '/tmp/fusion_os_install'
+if (new File('build.properties').exists()) {
```
tst-ccamp: Minor: Why instantiate this file object more than once instead of assigning it to a variable. Probably minuscule as far as performance, just curious if this is a preferred style for Gradle or something like that.

earth_enterprise/SConstruct
tst-dsugar: Well, you get it every time you run scons, because you may change installdir in any run.  I also wanted to push the version info and maybe other things thru properties later, which also may change.

tst-ccamp: Minor: I'd probably name this something a little more descriptive than "prefix", which is pretty generic. Internally it's called "stagedInstall_path", so that would probably be better.

tst-dsugar: My original thought is we might later have "prefix=", "version=", and "release=", so I was thinking in terms of short words....

tst-dsugar: "staged=" may be clearer than prefix= though...

earth_enterprise/rpms/build.gradle
```diff
@@ -55,8 +55,16 @@ def openGeeVersion =
...
+    def props = new Properties()
+    new File('build.properties').withInputStream {
+        props.load(it)
+    }
+    stagedInstall_path = props["prefix"]
```
tst-ccamp: Minor-ish: Is it possible that there might be other ways to create build.properties or other values in it in the future? Is it clear that prefix is always required to be there? I think it might be better to either check to see if "prefix" is present or specify a default.

.gitignore
tst-dsugar: Actually I found there is a .gitignore in the rpms/ directory so I could move it to there...

earth_enterprise/rpms/build.gradle
tst-dsugar: There is a default set if there is no properties file at all, but yes,I can not change if it is not defined.

tst-ccamp: Just wondering if that might cause an error if the build.properties is present but prefix isn't set.

earth_enterprise/SConstruct
tst-ccamp: Right, my point is that you are only appending this string to the file rather than recreating the file from scratch each time, so you'll end up with something like:
```
prefix=/tmp/install_os_install
prefix=/some/other/path
```
I'm not sure what gradle will do when it reads that. So, I'd either delete the file every time, but then it would be good to add a message that says it's autogenerated and shouldn't be edited by hand. Or, actually replace the line with prefix rather than append.

tst-dsugar: Actually, it opens it for 'w' at the beginning, so it never appends to existing file, it rewrites, so there is always only one prefix= line.  Append would make sense for adding more vars though once it is open write....

.gitignore
tst-ccamp: Unfortunately, that won't quite do the trick since one of the valid properties files is under the rpms directory. "./rpms/gradle/wrapper/gradle-wrapper.properties" Probably best to just ignore the specific build.properties file instead of all of them.

earth_enterprise/SConstruct
tst-ccamp: This one is minor, so it's really a matter of opinion. If you want to keep it to a single word, prefix or staged would be fine.

tst-ccamp: Ah, I am not fluent in the details of Python behavior it seems. After consulting the documentation, I see that it will truncate the file. But, you're right about it being a little awkward if you wanted to add more values later. But, as long as it's within the "with" block you should be fine. I'd still suggest adding a comment stating it's an auto-generated file though. If someone saw that sitting in one of the directories they may think it's a valid place to set properties.

earth_enterprise/rpms/build.gradle
tst-dsugar: Yes, it could, I am adding a check to make sure it is defined....otherwise the original default will be used from line #58

earth_enterprise/SConstruct
tst-dsugar: I am renaming it to staged= as I think it is clearer given how it is used later.

tst-ccamp:
The most recent changes address my concerns. Approved and merging!

earth_enterprise/SConstruct
tst-ppenev: I vote for using the same name that's used in the `scons` scripts, and Gradle scripts.  It just makes it worse to rename a variable just to pass it from one place to another.  It makes it more tedious to track where a value is used.

tst-ppenev: I hope we don't add `version` and `release` to this file.  Gradle can read the already existing version files from where they are.

Issue #607, opened 2017-11-13
Fix RPM Dependencies


tst-ppenev:
As an Open GEE RPM user, I want the each Open GEE RPM to depend on all the packages that it requires at run-time, so that package management systems can resolve dependencies and automatically install required packages.

## AC

* Installing each Open GEE RPM on a minimal (e.g. `atomic`) distribution installation should automatically install all required packages for its operation.
* Installing `opengee-common`, `opengee-postgis`, `opengee-server` and `opengee-fusion` should work on RHEL 6, Cent OS 6, RHEL 7 and Cent OS 7.

tst-lsavoie:
Closed via PR #628

PR #628, opened 2017-12-07
#607: Fix RPM Dependencies

tst-ppenev:
* Moved `libpq` from the `opengee-server` to the `opengee-postgis` package.

* Added resolution of shell commands to packages that provide them for RPM (and Deb) packages.

* Added detection of provided and required capabilities using `rpm-build`'s `find-requires` and `find-provides`.

* Added support for `Requires(pre)` dependencies by rebuilding the RPM output by OS Package with `rpmrebuild`.

* Updated documentation.

## Verification Steps

* Build the OpenGEE RPMs.
    * Compile OpenGEE.
    * Stage install OpenGEE.
    * Package OpenGEE. (E.g.: `cd earthenterprise/earth_enterprise/rpms && ./gradlew openGeeRpms`)
* Transfer the RPMs to a machine with a minimum of packages installed.
* Enable the EPEL Yum repository.
* Install all RPMs in one command, such as `yum install opengee-*.rpm`.
* Verify the installation of all packages succeeds, and Server and Fusion start.  (You can make sure they pass a basic smoke test.)
* Uninstall OpenGEE RPMs. (E.g., `yum remove 'opengee-*'`.)
* Verify uninstallation succeeds.


earth_enterprise/rpms/build.gradle
```diff
@@ -272,15 +318,26 @@ task openGeeCommonDeb (type: Deb) {
     }
 }
 
-task openGeeServerRpm (type: Rpm, dependsOn: openGeeSharedFiles) {
+task openGeeServerRpm (type: TstRpm, dependsOn: openGeeSharedFiles) {
```
tst-ccamp: Why do these begin with Tst? Shouldn't it be something like GeeRpm?

tst-ppenev: This a class derived from `com.netflix.gradle.plugins.rpm.Rpm` from the OS Package plug-in.  It extends the `Rpm` Gradle task to be able to add commands as dependencies, and scan files that go into the RPM for shared library, and scripting language module dependencies, as well as list shared libraries that go into the package as `provides`.

tst-ppenev: We could submit those in a pull request to OS Package, except the latest version of OS Package fails to produce RPMs that install on Cent OS 6, so we'd have to get it into an earlier version, and propagated to the latest version for whet it, hopefully, gets fixed to work with Cent OS / RHEL 6.

earth_enterprise/rpms/build.gradle
```diff
@@ -41,6 +41,40 @@ buildscript {
 
 apply plugin: "nebula.ospackage"
 
+def osId = null
+def osVersionId = null
+
+if (new File('/etc/os-release').exists()) {
```
tst-dsugar: Doesn't this conflict with pr/606 which relocated this to a Platform class?

tst-ppenev: The other pull request needs to be merged into this one.  Or, if the other is merged into `master`, `master` would be merged into this one.  This one is currently behind.

earth_enterprise/rpms/build.gradle
tst-ppenev: @tst-ccamp: Oh, you're right.  We may not want to have that in OpenGEE.  Do we have reverse domain name prefix we can use for OpenGEE packages?

earth_enterprise/rpms/build.gradle
tst-dsugar: Indeed, the other one has been merged into master, and now breaks this one...

earth_enterprise/rpms/build.gradle
```diff
@@ -236,10 +242,21 @@ task openGeeCommonRpm(type: Rpm, dependsOn: openGeeSharedFiles) {
...
+
+    requireCommands(packageSharedCommands +
+        ['cat', 'cut', 'getent', 'grep', 'groupadd', 'groupdel'])
+    
+    /* Since some libraries (i.e., MrSid)  cannot be redistributed without a 
+       license and are not packaged as an RPM, we need to filter these out from the
+       automatic 'requires' dependency generation. */
+    
+    def filterCapabilities = ['libltidsdk.so()(64bit)']
+    autoFindRequiresFilter = { filterCapabilities.contains( 'libltidsdk.so()(64bit)') ? null : it }
```
tst-ppenev: @tst-dsheremata: This condition will be always true, and you'll drop all requires.  I think you meant to write:

```Groovy
filterCapabilities.contains(it)
```

earth_enterprise/rpms/build.gradle
```diff
@@ -236,10 +242,21 @@ task openGeeCommonRpm(type: Rpm, dependsOn: openGeeSharedFiles) {
...
+
+    requireCommands(packageSharedCommands +
+        ['cat', 'cut', 'getent', 'grep', 'groupadd', 'groupdel'])
+    
+    /* Since some libraries (i.e., MrSid)  cannot be redistributed without a 
+       license and are not packaged as an RPM, we need to filter these out from the
+       automatic 'requires' dependency generation. */
+    
+    def filterCapabilities = ['libltidsdk.so()(64bit)']
+    autoFindRequiresFilter = { filterCapabilities.contains( it ) ? null : it }
```
tst-ppenev: This should be `contians(it.capabilityName)`.

earth_enterprise/rpms/opengee-common/post-install.sh
```diff
@@ -24,12 +24,15 @@ check_group()
     # Add group if it does not exist:
     if [ -z "$GROUP_EXISTS" ]; then
         groupadd -r "$GEGROUP" &> /dev/null
-        keyvalue_file_set "$GEE_INSTALL_KV_PATH" gegroup_existed "false"
+        echo "Added group $GEGROUP"
```
tst-lsavoie: Minor: do we want to leave this in?  It looks like debugging output.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -64,6 +66,32 @@ main_postinstall()
...
+    # Add group if it does not exist.  Note that it is also created by the common installer, 
+    # and common will remove it if necessary. 
+    if [ -z "$GROUP_EXISTS" ]; then
+        groupadd -r "$GEGROUP" &> /dev/null
+        echo "Fusion install added group $GEGROUP"
+    fi
+
+    # Add user if it does not exist:
+    if [ -z "$(getent passwd "$GEFUSIONUSER")" ]; then
+		mkdir -p "$BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER"
```
tst-lsavoie: Minor: there's some funny indentation in this if/else statement.  It would be nice to clean it up.

tst-ccamp:
Looks like this needs to be rebased. I don't really think you made 62 commit for this change 😆

tst-ppenev:
@tst-ccamp: I thought we just squash on merge.  How many commits should this be reduced to?

earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy
```diff
@@ -0,0 +1,66 @@
...
+                (it.exitCode == null || it.exitCode == exitCode) &&
+                (it.stdOut == null || it.stdOut == stdoutBuffer) &&
+                (it.stdErr == null || it.stdErr == stderrBuffer)
+            ) {
+                // Suppress result:
+                return null;
+            }
+        }
+        if (exitCode != 0) {
+            throw new IllegalArgumentException("""\
```
tst-ccamp: Why IllegalArgumentException? Is that the only way there could be an issue running the command or just the most likely?

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -64,6 +64,7 @@ main_postinstall()
     final_fusion_service_configuration
 }
 
+
```
tst-ccamp: Minor: unnecessary line

earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy
```diff
@@ -0,0 +1,66 @@
...
+            this.stdErr = stdErr
+        }
+    }
+
+    // Returns the standard output that results from executing a given command.
+    //     To pipe to standard input set `stdInAction` to a lambda that
+    // accepts the standard input writer as an parameter.
+    //     To suppress known command results (return `null`), use the
+    // `suppressedResults` parameter.
+    static def expand(
```
tst-ccamp: Expand seems like an odd name for this method. Isn't it executing the command?

tst-ccamp:
@tst-ppenev It looks like this branch diverged from master at some point in time and there could be double commits in it. We could just squash, it's just not very clean. It's not a huge deal I don't guess.

earth_enterprise/rpms/build.gradle
```diff
@@ -57,13 +57,17 @@ def openGeeVersion =
         it // The first string that remains is the version string.
     }
 
-def stagedInstall_path = '/tmp/fusion_os_install'
+def stagedInstall_path = project.hasProperty('stageInstallPath') ?
```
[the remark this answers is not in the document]
tst-ppenev: Yeah, I'll use some elvis.

earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy
tst-ppenev: Yeah, it's executing it, and returning the standard output.

earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy
tst-ppenev: No, that would be just for the case when the command couldn't be found.  I'll look for a better exception.

earth_enterprise/rpms/build.gradle
```diff
@@ -236,10 +257,25 @@ task openGeeCommonRpm(type: Rpm, dependsOn: openGeeSharedFiles) {
         'OpenSSL and ' +
         'Skia 5.1.2'
     type = BINARY
+    autoFindProvides = true
+    autoFindRequires = true
 
     requires('opengee-postgis', '1.5.8', GREATER | EQUAL)
```
tst-ccamp: This will be an issue with the PostGIS upgrade, along with the conflict below. Is this something that can be addressed here or will we need a separate PR on the postgres_upgrade branch?

tst-ppenev: What is the issue and the conflict below?

If PostGIS is unbundled, we should be able to just remove this dependency.  If it's not unbundled on `master`, we should do that when we have a pull request that will unbundle it.

tst-dsugar:
This is what I am currently getting testing this...

Do we need to pass a --define for _topdir?

* What went wrong:
Execution failed for task ':openGeeServerRpm'.
> java.lang.Exception:                 Rebuilding RPM to add Requires(pre) failed!
                  Command: [/usr/bin/rpmrebuild, --package, --change-spec-requires=/bin/bash /home/dsugar/src/earthenterprise/earth_enterprise/rpms/build/rpmRebuild/requires-filter.sh, --define, _rpmdir /home/dsugar/src/earthenterprise/earth_enterprise/rpms/build/rpmRebuild, 

/home/dsugar/src/earthenterprise/earth_enterprise/rpms/build/distributions/opengee-server-5.2.0-1.el7.x86_64.rpm]
                  Exit code: 1.
                  Error output:
                  error: failed to create directory %{_topdir}: /rpmbuild: Permission denied
  /usr/lib/rpmrebuild/rpmrebuild.sh: ERROR: package '/home/dsugar/src/earthenterprise/earth_enterprise/rpms/build/distributions/opengee-server-5.2.0-1.el7.x86_64.rpm' build failed
  
                  Standard output:

earth_enterprise/rpms/build.gradle
tst-ccamp: It's not yet unbundled, but it is upgraded. Although, during my testing I didn't have any issues actually running it. I think it may just be reporting the wrong version. I believe the work done for upgrading it may have taken place before this was merged to master. I can update the existing PR #638 to update the version number on the RPM package, or I can create another subtask to handle that separately.

tst-ppenev:
@tst-dsugar: Do you have directories created by a different user than the one you're running the build with?

earth_enterprise/rpms/build.gradle
tst-ppenev: I still see `postgis-1.5.8.tar.gz` on `master`.

tst-dsugar:
I do not believe so.  I also checked, in that I did have a ~/.rpmmacros file, so to be safe I killed that. Why would it go to /rpmbuild?  I will check further, as it may well be something odd I do have in my env/configured on that machine.

earth_enterprise/rpms/build.gradle
tst-ccamp: Right, I've agreed that the change doesn't need to go in this PR. It can go in a separate PR to the postgres_upgrade branch.

tst-dsugar:
I was able to reproduce the issue on another, known good, centos 7 system where I had produced rpms before, and @tst-dsheremata had somewhat similar results building on centos 6, as shown here:

* What went wrong:
Execution failed for task ':openGeeServerRpm'.
> java.lang.Exception: Rebuilding RPM to add Requires(pre) failed!
  Command: [/usr/bin/rpmrebuild, --package, --change-spec-requires=/bin/bash /workspace/earthenterprise/earth_enterprise/rpms/build/rpmRebuild/requires-filter.sh, --define, _rpmdir /workspace/earthenterprise/earth_enterprise/rpms/build/rpmRebuild, /workspace/earthenterprise/earth_enterprise/rpms/build/distributions/opengee-server-5.2.0-1.el6.x86_64.rpm]
  Exit code: 1.
  Error output:
  mkdir: cannot create directory `/.tmp': Permission denied

  Standard output:

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeRpm.groovy
```diff
@@ -0,0 +1,218 @@
...
+cat <<RequiresEND
+${
+    requiresPre.collect { "Requires(pre): ${it}" }.join("\n")
+ }
+RequiresEND
+""")
+
+        GeeCommandLine.expand(
+            [
+                "/usr/bin/rpmrebuild", "--package",
```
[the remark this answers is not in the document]
tst-ppenev: We can submit a pull request to the RedLine library to add support for the `Requires(pre)` header. In the mean time, we can use `rpmrebuild`, or, maybe, `rpmbuild`, if we rewrite a bit of how OS Package works.

tst-dsugar:
Unfortunately I have a new problem trying to verify an empty centos 7 with epel-release installed, after building on a centos 7 dev machine, and I am not sure yet why...

bash-4.2# yum install opengee-postgis-1.5.8-1.el7.x86_64.rpm 
Loaded plugins: fastestmirror
Examining opengee-postgis-1.5.8-1.el7.x86_64.rpm: opengee-postgis-1.5.8-1.el7.x86_64
Marking opengee-postgis-1.5.8-1.el7.x86_64.rpm to be installed
Resolving Dependencies
--> Running transaction check
---> Package opengee-postgis.x86_64 0:1.5.8-1.el7 will be installed
--> Processing Dependency: libcrypto.so.1.0.0()(64bit) for package: opengee-postgis-1.5.8-1.el7.x86_64
Loading mirror speeds from cached hostfile
 * base: mirror.cogentco.com
 * epel: fedora-epel.mirrors.tds.net
 * extras: mirror.trouble-free.net
 * updates: mirrors.centos.webair.com
--> Processing Dependency: libssl.so.1.0.0()(64bit) for package: opengee-postgis-1.5.8-1.el7.x86_64
--> Finished Dependency Resolution
Error: Package: opengee-postgis-1.5.8-1.el7.x86_64 (/opengee-postgis-1.5.8-1.el7.x86_64)
           Requires: libssl.so.1.0.0()(64bit)
Error: Package: opengee-postgis-1.5.8-1.el7.x86_64 (/opengee-postgis-1.5.8-1.el7.x86_64)
           Requires: libcrypto.so.1.0.0()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest

tst-dsugar:
Okay, it appears libssl.so.1.0.0 (and crypto) is in the opengee-common rpm. It means opengee-postgis cannot be installed alone.  I think it is a bug, but not critical.  I was doing verification by installing each rpm with yum in order to verify they each pull in all required dependencies.

PR #630, opened 2017-12-08
build #632: improve version strings

tst-basilhuffman:
Fork of Jim's branch: improved-version-strings

Code review to verify:

https://github.com/google/earthenterprise/pull/597

Comments edited to address previous reviews. To test, per Jim:

Implementation details have been updated.

Without git, a src/version.txt file is referenced to git the latest known build number.
With git, src/version.txt is updated so that the value can be stored in the repository.
If you are on a version-style tag commit, that tag will be returned.
If you have a dirty repository, the datestamp is appended to the latest tag.
Otherwise, the datestamp and current commit (short) hash are appended.
Revision and build numbers and alpha -> beta -> release cycles are updated as appropriate.
Recommended testing:

Tag your local repo with a version-like string such as "5.2.1-1.beta".
Build and check output in src/NA*/gee*txt and src/version.txt
Make a file change to make the repo dirty.
Build and check output files.
Commit that change (git status should be clean, except maybe for src/version.txt)
Build and check output files.
Then:

Tag your repo with a release style tag, such as "5.2.1-2.release".
Repeat the above remaining steps.


earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +162,192 @@ def EmitBuildDateStrfunc(target, build_date):
...
+
+    return final
+  
+
+def CheckGitAvailable():
+    """Try the most basic of git commands, to see if there is
+       currently any access to a repository."""
+    
+    try:
+        subprocess.check_output(['git', 'status'])
```
tst-ppenev: Consider using GitPython, or write a comment about why it's not suitable for the environments we support.

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +162,192 @@ def EmitBuildDateStrfunc(target, build_date):
...
+        return raw
+
+    # Tear apart the information in the version string.
+    components = ParseRawVersionString(raw)
+  
+    # Determine how to update, since we are *not* on tagged commit.
+    if components['isFinal']:
+        components['patch'] = 0
+        components['patchType'] = "alpha"
+        components['revision'] = components['revision'] + 1
```
tst-ppenev: Can you explain how we know that the `revision` part of the version should be increased?  How do we know we shouldn't increase the `major` or `minor` version number instead?

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +162,192 @@ def EmitBuildDateStrfunc(target, build_date):
...
+    """Break apart a raw version string into its various components,
+    and return those entries via a dictionary."""
+
+    components = { }    
+    rawComponents = raw.split("-")
+    
+    base = rawComponents[0]
+    patchRaw = rawComponents[1]
+    components['numCommits'] = rawComponents[2]
+    components['hash'] = rawComponents[3]
```
tst-ppenev: Q: Are we including a hash as part of a Git tag?  Doesn't Git already know what commit it's at?

tst-ppenev:
Some general comments:

* We should know what release we're working on (e.g., alpha, hot fix, next minor release, next major release) before starting to make commits for it.  We should tag the first commit towards the new release, or use a branch for it (which would allow, e.g., working on hot fixes and a next major or minor version simultaneously).

* We should avoid using hyphens in version strings.  They'll have to be converted to something else in order to be used in package versions, and there aren't many options for other characters to use.

* We should consider having a version component for hot fixes that are released between scheduled releases, e.g. when a vulnerability is found and needs to be fixed quickly.  We may be using the 'revision' component of the version number for scheduled releases, rather than the 'minor' number, and not have a version number for quick fixes, or, perhaps, we don't have any scheduled releases and we're considering everything we're adding a quick fix.

* We should consider not putting a commit ID in a version string, but recording it as part of the build context.  They don't allow you tell which version is higher than another, and are probably not anything that someone except a developer debugging a problem would want to know.  They seem to be just packing extra information in a version string, because we didn't find a better place for it.

* Timestamps help in determining what version is newer, but it's still possible to build an older commit at a later time.  We can consider using a number connected to the latest commit in the repository before the timestamp.  (E.g., number of commits since release tag.).

earth_enterprise/src/SConstruct
```diff
@@ -150,22 +151,16 @@ else:
...
-    Exit(1)
-    is_min_ubuntu = False
-# lsb_release is not present, must be RHEL or CentOS, move on
-except:
-  print "OS is not Ubuntu"
+is_hardy = True
+# if non-zero return status it is not hardy
+# Warning: assume your Ubuntu version is 10.04.2, otherwise need change 
+#the version number in order to compile
+if os.WEXITSTATUS(os.system('awk \'/Ubuntu/ { split($2, array,"."); \
```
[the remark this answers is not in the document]
tst-basilhuffman: This was fixed in PR 528, updated to all Python. Jim's branch is very old and requires merging.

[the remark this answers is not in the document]
tst-ppenev: I see.  It doesn't have to be fixed in this pull request if it's already existing code, but it didn't show up as moved from somewhere else in this pull request.

This is minor, I wouldn't hold up a pull request for it, if it works.

earth_enterprise/src/fusion/autoingest/FusionConnection.cpp
```diff
@@ -188,7 +188,7 @@ FusionConnection::ConnectTokhAssetManager(void)
                                   (khAssetManagerAddr())));
 
   handle->NoRetRequest("ValidateProtocolVersion",
-                       std::string(FUSION_VERSION));
+                       std::string(GEE_VERSION));
```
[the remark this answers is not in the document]
tst-jblair: The description of the task as I understood it was: there are too many version strings floating around the code base and we need to standardize them. So, I thought we were actively wanting to make the Fusion and Server version strings the same.

earth_enterprise/src/scons/khEnvironment.py
tst-jblair: There were no rules in the Version documentation for changing the major/minor version numbers. Those changes should come from someone tagging the repository to a newer version.

earth_enterprise/src/scons/khEnvironment.py
tst-jblair: No, none of the git tags should have a hash.

earth_enterprise/src/earthplugin/SConscript
```diff
@@ -20,47 +20,23 @@ import os
...
 # Copy files from '../earth_enterprise/src/earthplugin/..' directory.
 env.installDirExcluding('earth', Dir('#earthplugin').abspath, ['SConscript'])
 
 # Copy files from '../earth_enterprise/src/{REL_XXX|OPT_XXX}/earthplugin/..'
-# directory. There is only one file 'version.txt' in this folder.
-env.installDirExcluding('earth', '.', [])
+# directory.
+
+# contains the build date in YYYY-MM-DD format
+env.install('earth', File('../builddate.txt').abspath)
```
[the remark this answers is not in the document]
tst-jblair: A couple of these files exist purely to get the needed information into the intro web pages.  And a couple are because the existing code wants the data available as C++ macros.

A general "this is the build information" config file would be cleaner, though. It will naturally add the need for additional parsing calls in various places, though.

earth_enterprise/src/fusion/autoingest/FusionConnection.cpp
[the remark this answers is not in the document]
tst-ppenev: I see this as two issues.  There wasn't a well centralized place for setting version information that gets propagated to all other places that use it, which this pull request is solving.  On the other hand, I think it's better when projects (and classes and methods) have limited tasks, and isolation, so they can evolve without having to consider how they affect all other components.

We can have a shared library between Server and Fusion that provides a C++ data structure with all the version components, and a SCons share library to populate it from Git.

earth_enterprise/src/earthplugin/SConscript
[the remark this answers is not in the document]
tst-ppenev: Well, if the version is a string, you have to parse it to unpack the various parts of it.  It's not very complicated parsing, but there are serializers and deserializers in both Python and Groovy so you can put a data structure in a file, e.g. from Python, and read it back, e.g. in Groovy.  You could expand a template to write a Python structure as a C++ class, or generate C macros.

What kind of expansion is used to get it into Web pages?

earth_enterprise/src/scons/khEnvironment.py
tst-ppenev: I'm confused.  Why is `revision` increased for a `final` version, and why does it become `alpha`?

earth_enterprise/src/earthplugin/SConscript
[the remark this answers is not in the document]
tst-jblair: For the web pages, there's a small javascript routine that just grabs the contents of the file, and appends it to the appropriate text box.

earth_enterprise/src/scons/khEnvironment.py
tst-jblair: When the repo is tagged as something like "5.2.1-2.final" (or "5.2.1-2.release", now), that's indicating "we declare 5.2.1 to be done, and this is something that can be released to the public".

If we've started making commits and changes from that point, then we have, by definition, moved on past 5.2.1, and to 5.2.2 (operating under the scenario where there's no clear indicator that we should be moving on to 5.3 instead - I believe someone should tag a "5.3.0-0.beta" commit in order to make that transition, lacking any better ideas).

earth_enterprise/src/earthplugin/SConscript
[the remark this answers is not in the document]
tst-ppenev: Well, Java Script is pretty good reading JSON. :stuck_out_tongue:
And I think it'd be fine with other formats, as well.

earth_enterprise/src/scons/khEnvironment.py
tst-ppenev: I see. It doesn't seem like a good idea to be making commits towards an unknown version.  I may prefer an error, but I'm OK with this.

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +161,192 @@ def EmitBuildDateStrfunc(target, build_date):
   return 'EmitBuildDate(%s, %s)' % (target, build_date)
 
 
+def EmitVersionHeaderFunc(target, backupFile):
+  """Emit version information to the target file."""
+
+  versionStr = GetVersion(backupFile)
+  longVersionStr = GetLongVersion(backupFile)
+
+  fp = open(target, 'w')
```
tst-ppenev: Minor: You can use a `with`statement for writing to files.  It handles exceptions, and is shorter.

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +161,192 @@ def EmitBuildDateStrfunc(target, build_date):
...
+def EmitVersionHeaderStrfunc(target, backupFile):
+  return 'EmitVersionHeader(%s, %s)' % (target, backupFile)
+  
+  
+def EmitVersionFunc(target, backupFile):
+  """Emit version information to the target file."""
+
+  versionStr = GetVersion(backupFile)
+
+  fp = open(target, 'w')
```
tst-ppenev: Minor: Use a `with` statement.

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +161,192 @@ def EmitBuildDateStrfunc(target, build_date):
...
+  fp.write(versionStr)
+  fp.close()
+  
+  fp = open(backupFile, 'w')
```
tst-ppenev: Minor: `with`.

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +161,192 @@ def EmitBuildDateStrfunc(target, build_date):
...
+def EmitVersionStrfunc(target, backupFile):
+  return 'EmitVersion(%s, %s)' % (target, backupFile)
+  
+  
+def EmitLongVersionFunc(target, backupFile):
+  """Emit version information to the target file."""
+
+  versionStr = GetLongVersion(backupFile)
+
+  fp = open(target, 'w')
```
tst-ppenev: Minor: You know.

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -159,6 +161,192 @@ def EmitBuildDateStrfunc(target, build_date):
...
+    return (len(str) > 0)
+    
+
+def ReadBackupVersionFile(target):
+    """There should be a file checked in with the latest version
+       information available; if git isn't available to provide
+       information, then use this file instead."""
+    
+    try:
+        fp = open(target, 'r')
```
tst-ppenev: Minor: `with`.

tst-ccamp:
@tst-basilhuffman Can you create an issue that summarizes the changes that have been made to link with this PR?

tst-jcupp:
fixes #632

PR #631, opened 2017-12-11
umask fix for opengee-server rpms

tst-dsheremata:
Fixed permissions in post-install scripts where umask is aggressive on some systems (i.e. imask 077)

### DIFF
--- a/earth_enterprise/rpms/opengee-server/post-install.sh
+++ b/earth_enterprise/rpms/opengee-server/post-install.sh
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 set +x
+umask 002
 
 #-----------------------------------------------------------------
 # Definitions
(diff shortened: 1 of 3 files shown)


tst-dsheremata:
Updated to include the umask fix for install_fusion.ss and install_server.sh as well.

earth_enterprise/rpms/opengee-server/post-install.sh
```diff
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 set +x
+umask 002
```
tst-dsugar: I am assuming this is the only rpm script that currently requires this added, and I see both install_xx scripts have it now too.

tst-dsheremata: Correct - the fusion post_install.sh does not create any files.  While it does sometimes create /gevol, I don't believe this is much of an issue as there's a lot of user administration around /gevol management that will probably require documentation and/or modification to other scripts (i.e., geconfigureassetroot) if un-sane umask causes a problem there.

PR #638, opened 2017-12-22
566 postgres install rpm data upgrade

tst-ccamp:
## Verification Steps

1. Create a VM (RHEL or CentOS) with GEE Server 5.2.0 installed using install_server.sh script
2. Publish one or more globes with search tabs
3. Verify the globe(s) load using Earth EC
4. Do a search or two and take note of the results
5. Take a snapshot for ease of going back to this state
6. Remove any old /tmp/fusion_os_install directories
7. Build, create RPM files, and install the feature branch
8. The geserver service should start without issues
9. Log into the admin console and verify that the database(s) are still present
10. Verify the globe(s) load using Earth EC with "disable cache" set
11. Do the same searches as before and make sure the results are the same
12. Go back to step 7. You can skip build, and just run stage install, and create RPM files again. There should be additional RPM files with a new version.
13. Continue with the same verification once more to verify that the install upgrade also works with PostgreSQL 9.6 installed

Optional (because there are known issues with updating RPM files from master)

1. Install RPM files built from master
1. Run steps 2-11 from above

Issues you may encounter:
1. When the RPMs are upgraded they will remove the GEE users and the service configuration
    * Solution: reinstall the RPM files `sudo yum reinstall opengee*<version>*.rpm"
1. The opengee-common RPM may fail to be removed with an error such as: 
    `opengee-common-5.2.0-1.el6.x86_64 was supposed to be removed but is not!`
    
    * Solution: remove the package manually and disable running the script: `sudo yum remove --setopt=tsflags=noscripts "opengee-common-5.2.0-1.el6.x86_64" ` (The exact package name should match the one in the error message)

I merged the latest master into this branch and postgres_upgrade for testing. Still targeting postgres_upgrade for now so these changes can be reviewed independently.


earth_enterprise/rpms/build.gradle
```diff
@@ -84,9 +70,26 @@ def stagedInstallDir_server_geplaces =
     new File(stagedInstallDir, 'geplaces/opt/google/share/geplaces')
 def stagedInstallDir_server_user_magic =
     new File(stagedInstallDir, 'server/user_magic')
+def stagedInstallDir_server_version =
+    new File(stagedInstallDir, 'server/opt/google/gehttpd/htdocs/earth/gee_long_version.txt')
 def stagedInstallDir_fusion = new File(stagedInstallDir, 'fusion')
 def packageInstallRootDir = new File('/')
 
+def openGeeVersion =
```
tst-dsugar: Is any of this going to conflict with work in #628 ?

tst-ccamp: I needed this change to test RPM upgrades. If #628 fixes the issue, I'm more than happy to remove my changes

earth_enterprise/src/server/gepgcommon
```diff
@@ -0,0 +1,97 @@
...
+
+# Runs a command a specific user
+# If it's the same user, then just execute directly, otherwise use su or sudo
+function run_as_user
+{
+  if [ "$( whoami )" == "$1" ]; then
+    >&2 echo "sh -c \"$2\""
+    sh -c "$2"
+  else
+    local use_su=`su $1 -c 'echo -n 1' 2> /dev/null  || echo -n 0`
```
tst-ppenev: Suggestion: This test for `su` will break depending on what shell profile/resource script executes.  It can also get stuck, because `su` waits for the user to enter a password.

Unless `su` gets stuck, you can assume using `su` succeeded. You can use a test like:
```
if su "$1" -c 'echo 1' >/dev/null 2>&1; then
    # something
fi
```

earth_enterprise/src/server/gepgcommon
```diff
@@ -0,0 +1,97 @@
...
+    if [ "$use_su" -eq 1 ] ; then
+        >&2 echo "cd / ;su $1 -c \"$2\""
+        ( cd / ;su $1 -c "$2" )
+    else
+        >&2 echo "cd / ;sudo -u $1 $2"
+        ( cd / ;sudo -u $1 $2 )
```
tst-ppenev: Suggestion: Add quoting to the `sudo` arguments, and `--` before the command.

earth_enterprise/src/server/gepgcommon
```diff
@@ -0,0 +1,97 @@
...
+
+# Retrieves the gepoi table names from the gepoi database
+function get_gepoi_tables {
+  echo $( run_as_user $GEPGUSER "$BINDIR/psql gepoi -c \"\dt gepoi*\" | grep table | awk -F\| '{ print \$2 }'" )
+}
+
+# Dumps geserve-databases.
+function do_dump {
+  # Start postmaster.
+  if [ ! -f $VARPGDIR/data/postmaster.pid ]; then
```
tst-ppenev: Lint: Quote path.

earth_enterprise/src/server/geresetpgdb
```diff
@@ -79,6 +79,10 @@ VARPGDIR=/var/opt/google/pgsql
 SHAREDIR=/opt/google/share/postgresql/contrib
 BINDIR=/opt/google/bin
 SQLDIR=/opt/google/share/opengee-server/sql
+SCRIPTDIR=`dirname $0`
```
tst-ppenev: Lint: Quote `"$0"`.

earth_enterprise/src/server/geresetpgdb
```diff
@@ -79,6 +79,10 @@ VARPGDIR=/var/opt/google/pgsql
...
+
+# Import utility functions
+. $SCRIPTDIR/gepgcommon
```
tst-ppenev: Lint: Quote path.

earth_enterprise/rpms/build.gradle
```diff
@@ -84,9 +70,26 @@ def stagedInstallDir_server_geplaces =
     new File(stagedInstallDir, 'geplaces/opt/google/share/geplaces')
 def stagedInstallDir_server_user_magic =
     new File(stagedInstallDir, 'server/user_magic')
+def stagedInstallDir_server_version =
```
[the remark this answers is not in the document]
tst-ccamp: The convention seemed to be that all these would start with previously concatenated path names, so I just shortened the end into one word. I could call it stagedInstallDir_server_gee_long_version, but seems too long to me.

earth_enterprise/rpms/opengee-server/post-install.sh
```diff
@@ -170,7 +171,15 @@ fix_postinstall_filepermissions()
...
-    echo 2 | run_as_user "$GEPGUSER" "/opt/google/bin/geresetpgdb upgrade"
+    DB_BACKUP=""
+    DB_BACKUP_LATEST="$(ls -td -- $PGSQL/data.backup_dump*/ | head -n 1)"
+    echo "Latest backup: $DB_BACKUP_LATEST"
+    if [ -d "$DB_BACKUP_LATEST" ]; then
+        DB_BACKUP="$DB_BACKUP_LATEST"
+        echo "Restoring data backup from: $DB_BACKUP"
+    fi
+
+    echo 2 | run_as_user "$GEPGUSER" "/opt/google/bin/geresetpgdb upgrade $DB_BACKUP"
```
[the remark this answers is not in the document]
tst-ccamp: It defaults to the previous behavior in that case and prints a warning. In practice, I don't think that should happen with our installers from here on out.

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-ppenev: The purpose of this file is to contain the version string.  It doesn't have much to do with install paths, despite its unfortunate location.

earth_enterprise/rpms/opengee-fusion/pre-uninstall.sh
```diff
@@ -20,7 +20,9 @@ set +x
 # Main Function
 #-----------------------------------------------------------------
 
-service gefusion stop
-remove_service gefusion
+if [ "$1" = "0" ]; then
+	service gefusion stop
```
tst-ppenev: Lint: Equalize indentation.

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-ccamp: "stagedInstallDir" is prepended to all these path variables. It refers to whatever the "installdir" parameter is when running "stage_install". If I changed it here, it would no longer follow the pattern of all the rest. I don't want to change all of them, so I don't want to change this one.

earth_enterprise/src/SConstruct
```diff
@@ -336,6 +336,8 @@ installdirs = {
     'common_fonts': InstCommonDir('share/fonts'),
     'common_geecheck': InstCommonDir('share/support/geecheck'),
     'common_geecheck_lib': InstCommonDir('share/support/geecheck/lib'),
+    'common_version_file' : InstCommonDir('gee_version.txt'),
```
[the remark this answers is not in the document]
tst-ccamp: I was going to move the location, but ended up just using the one that already exists. I'll remove these

earth_enterprise/rpms/build.gradle
tst-ccamp: I ran out of time last week, so I left it in here. I can split it up if it's deemed necessary.

tst-ppenev: I think there's a ticket that @tst-dsugar may be working on for this purpose.  It may involve a little more than what is here, such as differentiating between 'alpha', 'beta' and 'release' versions, and adding more dependency headers, such as `provides`.

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-ppenev: You can leave it as it is, but these variables depend on each other, and define the staged install directory hierarchy.  The version file has nothing to do with that hierarchy, and where it is located makes no difference, so the variable doesn't need to reflect that.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,20 +43,6 @@ apply plugin: "nebula.ospackage"
 
 def rpmPlatformString = com.tsciences.os.Platform.rpmPlatformString
```
tst-nfarah: do we want 'tsciences' to be part of the platform string ?

earth_enterprise/rpms/build.gradle
```diff
@@ -43,20 +43,6 @@ apply plugin: "nebula.ospackage"
...
-        commentIndex < 0 ? it : it.substring(0, commentIndex)        
-    }.
-    collect { it.trim() }. // Trim whitespace.
-    findAll { !it.isEmpty() }. // Skip empty lines.
-    findResult {
-        it // The first string that remains is the version string.
-    }
-
 def stagedInstall_path = '/tmp/fusion_os_install'
 if (new File('build.properties').exists()) {
```
tst-nfarah: what happen is '/tmp/fusion_os_install' does not exist ? should we check and post an error ?

earth_enterprise/src/server/gepgcommon
```diff
@@ -0,0 +1,96 @@
+#! /bin/bash
+#
+# Copyright 2017 Open GEE Developers.
```
tst-nfarah: update the year to 2018 ? may want to do incrementally in other places as we touch them.

tst-ppenev: You, generally, use the earliest year something was copyrighted.  Alternatively, you can explicitly include all years for which you have a copyright, in case it may be transferred.  We can go either way, although I don't enjoy updating all file comments every year.

tst-ccamp: I will update the year in the files as I make changes

earth_enterprise/rpms/build.gradle
tst-ccamp: I think that's being addressed in a different PR

tst-ppenev: Hm. Maybe you'd want to merge some commits before the PR with these changes gets squashed and merged into `master`.  Or, you can resolve merge conflicts afterwards.

tst-ccamp: Depending on timing I can either resolve the conflicts with master on this PR or with the merge from postrgres_upgrade to master. I think it makes sense to wait until the other PR is merged before trying to address those issues.

tst-nfarah: I tested the rpm on centos 7 to verify that it would upgrade my postgreSQL databases and it did. The run searchs returned the same results as before the upgrade using the rpm but i had one issue. When the databases were being restored from dumps, i get errors/warning about "plpgsql":

```Restoring geserve-databases from dump /var/opt/google/pgsql/data.backup_dump.26393/ ...
Adding PostGIS extension
CREATE EXTENSION
Restoring geendsnippet.dump...
Restoring gepoi.gepoi_1.dump...
Restoring gesearch.dump...
 setval 
--------
      1
(1 row)

 setval 
--------
      1
(1 row)

 setval 
--------
      6
(1 row)

Restoring gestream.dump...
 setval 
--------
      1
(1 row)

 setval 
--------
      1
(1 row)

 setval 
--------
     29
(1 row)

 setval 
--------
      4
(1 row)

 setval 
--------
      4
(1 row)

 setval 
--------
      3
(1 row)

psql:/opt/google/share/opengee-server/sql/gesearch_tables.sql:20: ERROR:  language "plpgsql" already exists
 create_tables 
---------------
 
(1 row)

 run_searchdef_upsert 
----------------------
 
(1 row)

psql:/opt/google/share/opengee-server/sql/gesearch_tables.sql:167: ERROR:  cannot drop language plpgsql because extension plpgsql requires it
HINT:  You can drop extension plpgsql instead.
psql:/opt/google/share/opengee-server/sql/gestream_tables.sql:20: ERROR:  language "plpgsql" already exists
 create_or_update_tables 
-------------------------
 
(1 row)

psql:/opt/google/share/opengee-server/sql/gestream_tables.sql:147: ERROR:  cannot drop language plpgsql because extension plpgsql requires it
HINT:  You can drop extension plpgsql instead.
Done.
waiting for server to shut down.... done
server stopped
upgrade done
The PostgreSQL component is successfully installed.```

Comment (tst-ccamp) [reply earth_enterprise/rpms/build.gradle]: I think it's an incompatibility between dumps created in 8.4 and restored in 9.6. As far as I've been able to tell they don't cause any problems. I'm afraid that trying to fix it so it doesn't print errors could cause more issues since we'd have to modify the dump files. So I'm recommending just living with the errors.

Comment (tst-nfarah) [reply earth_enterprise/rpms/build.gradle]: Yes, indeed, those errors do not seem to cause any issues as far as databases restoration goes. They seem to be  caused by change how postgreSQL does treat 'language "plpgsql"'. In recent versions of postgreSQL (starting in 9.0) ""plpgsql" is installed by default, so we don't need to try to install or drop it in our '.../src/server/gesearch_tables.sql' and '.../src/server/gestream_tables.sql' any more since we are using postgreSQL 9.6.5.

For reference: https://www.postgresql.org/docs/current/static/release-9-0.html
See " PL/pgSQL is now installed by default, ...".

I made local changes in both mentioned .sql files, re-built the rpm and did an install. Those errors were not occurring any more, my databases were restored ok and the search results returned same results as before rpm upgrade execution.

Here are the needed changes:
- Delete or comment out lines 20 and 167 in '../src/server/gesearch_tables.sql'
- Delete or comment out lines 20 and 147 in '../src/server/gestream_tables.sql'

Comment (tst-ccamp) [reply earth_enterprise/rpms/build.gradle]: Removed those lines as instructed and it does fix the errors. I didn't realize those sql files were still be run on the upgrade, but that's good to know :)

File: earth_enterprise/rpms/build.gradle (hunk 12 above)
Comment (tst-ccamp) [reply earth_enterprise/rpms/build.gradle]: This is the error that's returned 
```

* Where:
Build file '/shared_dev/earthenterprise/earth_enterprise/rpms/build.gradle' line: 79

* What went wrong:
> /tmp/fusion_os_install/server/opt/google/gehttpd/htdocs/earth/gee_long_version.txt (No such file or directory)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
```
It doesn't directly say that /tmp/fusion_os_install doesn't exist, but upon investigation, the developer should realize pretty quick as they try to look for the missing file. :) It wouldn't hurt to add a check, but I think it's beyond the scope of this issue as that problem would exist whether or not we're doing the Postgres upgrade. If you think it should be addressed, let's write a follow-up ticket.

earth_enterprise/rpms/build.gradle
tst-nfarah: For re-verification i rebuilt the RPM and executed the installers. I did not see the errors any more and my functional tests for the published databases and search worked just fine. 

I noticed however that we always dump and restore the databases even if we are dealing with the same PostgreSQL version (9.6.5 --> 9.6.5). This is fine for now. We can add some smart into it, eventually later, to skip the dump/restore if the same version of PostgreSQL is already installed.
This can be done in a later ticket.

earth_enterprise/rpms/build.gradle
tst-nfarah: I'm fine with how it works for now. The error would give the developer a hint of what went wrong.

tst-ccamp:
I see two approvals, so merging!

PR #639, opened 2017-12-24
issue with duplicate resources in imagery projects

tst-dsugar:
This is preliminary because where this is may break other things.  It may be that the re_init flag has to be passed down thru prefill and handled explicitly in the derived imagery asset class, and ignored in others.

### DIFF
--- a/earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
+++ b/earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
@@ -134,11 +134,12 @@ void AssetDerived<Defs, FinalClass>::Init(bool re_init) {
 
   if (re_init == false) {
     InstallMainWidget();
-  }
-  SetName(saved_edit_request_.assetname);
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/fusion/fusionui/AssetDerivedImpl.h
```diff
@@ -134,11 +134,12 @@ void AssetDerived<Defs, FinalClass>::Init(bool re_init) {
 
   if (re_init == false) {
```
tst-nfarah: what 're-init' means here ?

tst-nfarah: This changed class 'AssetDerivedImpl' seems to be used in multiple places: imagery (resource, project), vector (resource, project), terrain (resource, project), database, KML project, Map layer, Map Project, Map database, Mercator image (resource, project), Mercator map database. We need to test all UI components that use those items to ensure that they work as designed.

PR #642, opened 2017-12-29
Fixes #646: rpm scripting fixes to correctly support package upgrading

### LINKED ISSUE #646: RPMs fail to upgrade with future releases
Build and install a set of opengee rpm's, called 5.2.x-1.

Manually modify the build.gradle file, change line release = "1.$

{rpmPlatformString}" to release = "2.${rpmPlatformString}".

Build a new set of opengee rpms, which will have 5.2.x-2 as their version number.

Use yum install or rpm -Uhv to try and update your previously installed 5.2.x-1 packages to this new 5.2.x-2 version.

Expected Result:

Upgrade completes without errors, you now are using 5.2.x-2.

Actual Result:

opengee-common will fail in pre-uninstall, and block updating from completing. System state may even become bad.

tst-dsugar:
Fixes #646


tst-dsugar:
@tst-ccamp this is what I came up with...

earth_enterprise/rpms/opengee-common/post-uninstall.sh
```diff
@@ -22,18 +22,23 @@ BININSTALLROOTDIR="/etc/init.d"
 
 #------------------------------------------------------------------------------
 # Group names:
-GEGROUP_NAME=$(cat "$BININSTALLROOTDIR/gevars.sh" | grep GEGROUP | cut  -d'=' -f2)
-
 
 remove_users_groups()
```
tst-ppenev: Suggestion: Consider replacing this script with just printing a message that the system administrator may want to remove `gegroup`, if there are now files or directories using it.

earth_enterprise/rpms/opengee-common/pre-install.sh
```diff
@@ -24,9 +24,6 @@ check_group()
     # Add group if it does not exist:
     if [ -z "$GROUP_EXISTS" ]; then
```
tst-ppenev: Suggestion: You can replace this function with `groupadd -f -r "$GEGROUP"`.

earth_enterprise/rpms/opengee-fusion/post-uninstall.sh
```diff
@@ -26,4 +26,7 @@ remove_users_and_groups()
 # Main Function
 #-----------------------------------------------------------------
 
-remove_users_and_groups
+# remove users only if actually uninstalling
+if [ "$1" = "0" ] ; then
+    remove_users_and_groups
```
tst-ppenev: Suggestion: We may want to just print a message that the system administrator can delete `gefusionuser`, if no files or directories are using it.

earth_enterprise/rpms/opengee-fusion/pre-install.sh
```diff
@@ -70,11 +81,11 @@ create_users_and_groups()
 		mkdir -p "$BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER"
 		useradd --home "$BASEINSTALLDIR_OPT/.users/$GEFUSIONUSER" \
             --system --gid "$GEGROUP" "$GEFUSIONUSER"
-        keyvalue_file_set "$GEE_INSTALL_KV_PATH" gefusionuser_existed "false"
 	else
```
tst-ppenev: Lint: Fix indentation.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -148,11 +157,7 @@ install_or_upgrade_asset_root()
     else
         # upgrade asset root -- if this is a master
         if [ "$IS_SLAVE" = false ]; then
-            # TODO: Verify this logic -- this is what is defined in the
-            # installer documentation, but needs confirmation
-            keyvalue_file_get "$GEE_INSTALL_KV_PATH" gegroup_existed NEW_GEFUSIONUSER
-            keyvalue_file_get "$GEE_INSTALL_KV_PATH" gefusionuser_existed NEW_GEFUSIONUSER
-            if [ "$NEW_GEGROUP" = true ] || [ "$NEW_GEFUSIONUSER" = true ]; then
+            if [ "$NEW_INSTALL" = false ] ; then
```
tst-ppenev: Q: How does this test if installing the `opengee-common` package created a new `gegroup` or not?

Q: Can we test whether to 'upgrade' by checking if one of the top-level directories has the wrong permissions?  Do we know upgrading from what version of GEE to what version of GEE requires changing file ownership?

earth_enterprise/rpms/opengee-server/post-install.sh
```diff
@@ -17,6 +17,12 @@
 set +x
 umask 002
 
+# we are not using this yet, but is probably good to have if we need to
```
tst-ppenev: Minor: I don't think we should have code that's not used anywhere.  We can add it when we need it.

tst-dsugar: I have killed all the post-install scripts in gradle now, because they only were involved in removing users and groups.  I could nuke the files as well...

earth_enterprise/rpms/opengee-fusion/post-uninstall.sh
tst-dsugar: We could.  For now I killed the post-uninstalls entirely in gradle...

earth_enterprise/rpms/opengee-server/post-install.sh
tst-ppenev: I think it's useful to print a message reminding a system administrator that they can remove the users and groups, if they want to.

earth_enterprise/rpms/opengee-common/post-uninstall.sh
```diff
@@ -22,18 +22,23 @@ BININSTALLROOTDIR="/etc/init.d"
...
 {
-    if [ -n "$(getent group "$GEGROUP_NAME")" ]; then
-        groupdel "$GEGROUP_NAME"
+    local GROUP_EXISTS=$(getent group "$GEGROUP")
+
+    # Add group if it does not exist:
```
tst-ppenev: Minor: I don't think the command below is adding a group.

tst-dsugar: Yes, I was trying to make add and delete consistent (they were originally very differently coded), and that is a copy-paste error.  But it may not matter since the post-uninstall is no longer in gradle.

earth_enterprise/rpms/opengee-server/post-install.sh
tst-dsugar: You have a good point.  I can re-activate the post-uninstalls and use it for that.

earth_enterprise/rpms/opengee-common/pre-install.sh
```diff
@@ -24,9 +24,6 @@ check_group()
...
         groupadd -r "$GEGROUP" &> /dev/null
-        keyvalue_file_set "$GEE_INSTALL_KV_PATH" gegroup_existed "false"
```
tst-ppenev: Old comment: This used to part of determining whether to change ownership and 'upgrade' a Fusion asset root. Now it seems to be simply omitted.  Are we sure that a user can't bring in an old asset root with the wrong group?

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -183,7 +189,7 @@ END
 
 final_assetroot_configuration()
 {
-    if [ "$IS_SLAVE" = true ]; then
+    if [ "$IS_SLAVE" == "true" ]; then
```
tst-ppenev: Lint: `==` is not a POSX operator, '=' is.  In case we care about POSIX compliance.

earth_enterprise/rpms/opengee-common/pre-install.sh
tst-dsugar: is it valid for both rhel6 and 7?

earth_enterprise/rpms/opengee-fusion/post-install.sh
tst-dsugar: We now know that opengee-common creates in it's pre-install now, and does not remove it all (or only in a post-uninstall when not upgrading), so we assume it's there based on this.  There are other/other kinds of tests we could do, but I was trying to keep consistent with practices common to other packages.

tst-dsugar: There are things we also probably still over do, and could do more checks for, but I did not want to invalidate all prior testing, and so focused on the issues around package updating itself only rather than about some of the extra and misc warning messages we still do get.

earth_enterprise/rpms/opengee-common/pre-install.sh
tst-ppenev: Yes.

earth_enterprise/rpms/opengee-common/pre-install.sh
tst-dsugar: There is a different way that this is checked for now, and the post-install.sh treats the rpm fresh install as an "upgrade" install for this if there was a prior non-rpm install found.  This should produce the same behavior as before, because gegroup_existed detection was actually used indirectly to determine install/upgrade behavior in post-install.sh prior.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -147,12 +157,8 @@ install_or_upgrade_asset_root()
...
     else
         # upgrade asset root -- if this is a master
-        if [ "$IS_SLAVE" = false ]; then
-            # TODO: Verify this logic -- this is what is defined in the
-            # installer documentation, but needs confirmation
-            keyvalue_file_get "$GEE_INSTALL_KV_PATH" gegroup_existed NEW_GEFUSIONUSER
-            keyvalue_file_get "$GEE_INSTALL_KV_PATH" gefusionuser_existed NEW_GEFUSIONUSER
-            if [ "$NEW_GEGROUP" = true ] || [ "$NEW_GEFUSIONUSER" = true ]; then
+        if [ "$IS_SLAVE" = "false" ]; then
+            if [ "$NEW_INSTALL" = "false" ] ; then
```
tst-ppenev: What I'm saying is that this is modifying the existing behavior by no longer testing if `gegroup` was created.

I don't think the question is whether `gegroup` exists after it is created by installing `opengee-common`.  What I understand the code is trying to do is:

* If asset data existed before `gegroup` or `gefusionuser` were created in the install scripts, change file ownership of the data.
* If `gegroup` and `gefusionuser` already existed before starting an installation, assume ownership is correct, and don't change it.

This change skips the test for whether `gegroup` existed.  I think we could write the tests better, but I'm wondering if it's a good idea to skip one of them.

tst-dsugar: And what I am saying is I deduce this from rather different means, based on both the rpm install/upgrade state, and also on how NEW_INSTALL gets set if there was a prior non-rpm install.

tst-ppenev: The test for 'a prior non-rpm install' seems to be whether `/etc/init.d/gefusion` exists.  So, if a user had an old version of GEE and Fusion with some data, uninstalled them, and installed RPMs, that assert root owner update will no longer happen, and Fusion would complain.  (I don't think it would work to install RPMs while the old version is installed anyway, at least until we implement the upgrade scripts from old versions.)

This is not very hard to fix, so I don't think it's a great problem, but we also don't have to introduce it.

tst-dsugar: I believe we already had this problem possible in the older script.

tst-ppenev: :stuck_out_tongue: The older script wouldn't have that problem, since it relied on whether the group and user existed, not on whether `/etc/init.d/gefusion` existed. Anyway, we can tell an administrator how to fix that after an upgrade.

tst-dsugar:
@tst-nfarah https://github.com/google/earthenterprise/issues/646 describes the issue and steps to reproduce that is pr will resolve.  I should have added the ticket earlier.

earth_enterprise/rpms/opengee-fusion/post-install.sh
tst-ppenev: Could you add a small Markdown file on upgrading that includes the command to update an asset root from an old install?

tst-ppenev:
Q: Why does this pull request seem to include a lot of unrelated changes from other pull requests, like adding `rpmrebuild` and `GitPython` to build environment scripts, and adding RPM and Deb build classes?

tst-ccamp:
@tst-dsugar I think merging master into your feature branch will fix those weird problems with unrelated commits

tst-dsugar:
It should be clean now

earth_enterprise/rpms/opengee-common/post-uninstall.sh
```diff
@@ -22,18 +22,18 @@ BININSTALLROOTDIR="/etc/init.d"
...
 # Group names:
-GEGROUP_NAME=$(cat "$BININSTALLROOTDIR/gevars.sh" | grep GEGROUP | cut  -d'=' -f2)
-
 
 remove_users_groups()
 {
-    if [ -n "$(getent group "$GEGROUP_NAME")" ]; then
-        groupdel "$GEGROUP_NAME"
-    fi
+    echo "opengee group $GEGROUP may now be safely removed."
```
tst-ppenev: Style: 'opengee' -> 'OpenGEE'

Minor: The group can be safely removed, only if the system administrator can confirm there are no more files or directories owned by it.

earth_enterprise/rpms/opengee-fusion/post-uninstall.sh
```diff
@@ -16,14 +16,15 @@
 
 remove_users_and_groups()
 {
-    if getent passwd "$GEFUSIONUSER" > /dev/null; then
-        userdel "$GEFUSIONUSER"
-    fi
+    echo "opengee user $GEFUSIONUSER may now be safely removed."
```
tst-ppenev: Style: 'opengee' -> 'OpenGEE'

Minor: The user can be removed safely only if the system administrator can confirm there are no files owned by that user that can appear on the machine.

earth_enterprise/rpms/opengee-fusion/pre-install.sh
```diff
@@ -15,13 +15,27 @@
...
 #-----------------------------------------------------------------
 # Main Functions
 #-----------------------------------------------------------------
 main_preinstall()
 {
+    # cleanup of special non-rpm upgrade case flag...
+    rm -f "$BASEINSTALLDIR_OPT/.upgrade-migration"
+
+    # Whether rpm or non-rpm upgrade...
     if [ -f /etc/init.d/gefusion ]; then
```
tst-ppenev: Q: How did you test this code?  Were you able to do a non-RPM upgrade with these RPMs?

earth_enterprise/rpms/opengee-server/post-uninstall.sh
```diff
@@ -21,20 +21,14 @@ set +x
...
-    # Fedora recommends to never delete users that a package has created:
-    #   https://fedoraproject.org/wiki/Packaging:UsersAndGroups
-    # TODO: Consider whether we want to remove this:
-    if getent passwd "$GEAPACHEUSER" >/dev/null; then
-        userdel "$GEAPACHEUSER"
-    fi
-    if getent passwd "$GEPGUSER" >/dev/null; then
-        userdel "$GEPGUSER"
-    fi
+    echo "opengee users $GEAPACHEUSER and $GEPGUSER may now safely be removed."
```
tst-ppenev: Same comments as above.

Style: 'opengee' -> 'OpenGEE'

Minor: The message would be more accurate, if it said the users and groups can be removed, if there are no files or directories owned by them.

earth_enterprise/rpms/opengee-common/pre-install.sh
tst-dsugar: So you are suggesting we add a chgrp -r gegroup /gevol if we are doing a non-rpm upgrade and there is an existing /gevol directory?  If so, that would be easy enough to add...

earth_enterprise/rpms/opengee-fusion/post-install.sh
tst-dsugar: If this is only a matter of permissions and assuring the group/ownership match up, it might be easier to do a chgrp/chown...

earth_enterprise/rpms/opengee-server/post-uninstall.sh
tst-dsugar: true...let me do that now.

earth_enterprise/rpms/opengee-common/pre-install.sh
tst-ppenev: I think that's what the code in `opengee-fusion/post-install.sh` does, e.g. when determining whether to pass the no-chown option to `geupgradeassetroot`.  The issue here is to determine when to do that ownership change.

earth_enterprise/rpms/opengee-fusion/pre-install.sh
tst-dsugar: This is based on the idea that you install the rpms without first removing the non-rpm install (that is, you do not run uninstall-server.sh or uninstall-fusion.sh first).  If you think this is not a valid scenario, then I could remove the migration checking entirely and maybe that would be a better way to address the concern you are stating.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -148,12 +158,8 @@ install_or_upgrade_asset_root()
...
                 NOCHOWN=""
                 UPGRADE_MESSAGE="The upgrade will fix permissions for the asset root and source volume. This may take a while."
```
tst-dsugar: This talks about "upgrade", but not with the same meaning as rpm upgrade.  In the case of rpm's, we have nailed down the gegroup and users for the life of the rpm installs.  Hence, "upgrading", but from one rpm to a newer one, should I believe normally be done with --nochown, and the original NEW_INSTALL test a few lines about should actually be if true, as the full chown should be done if it is a new install and yet there is a prior asset root.  I think this misunderstanding is the real issue, and the upgrade-migration test then does actually become unnecessary.

earth_enterprise/rpms/opengee-common/pre-install.sh
tst-dsugar: And now I do understand the concern completely, as being about when and why to set no-chown.

earth_enterprise/rpms/opengee-fusion/post-install.sh
```diff
@@ -148,12 +152,9 @@ install_or_upgrade_asset_root()
...
     else
         # upgrade asset root -- if this is a master
-        if [ "$IS_SLAVE" = false ]; then
-            # TODO: Verify this logic -- this is what is defined in the
-            # installer documentation, but needs confirmation
-            keyvalue_file_get "$GEE_INSTALL_KV_PATH" gegroup_existed NEW_GEFUSIONUSER
-            keyvalue_file_get "$GEE_INSTALL_KV_PATH" gefusionuser_existed NEW_GEFUSIONUSER
-            if [ "$NEW_GEGROUP" = true ] || [ "$NEW_GEFUSIONUSER" = true ]; then
+        if [ "$IS_SLAVE" = "false" ]; then
+            OWNERSHIP=`find /gevol/assets -maxdepth 0 -printf "%g:%u"`
```
tst-dsugar: @tst-ppenev so I came up with a new and more direct way to test asset file ownership. If ownership is same as rpm expects, it then sets nochown...

tst-ppenev: Yeah, this is what I was suggesting earlier.  I have two comments:

* Use `"$ASSET_ROOT"` instead of hard coding `/gevol/assets`.
* I'm wondering if `-maxdepth 0` is going to be enough.  You could use something like `find -maxdepth 2 \( \! -user "$GEFUSIONUSER" \) -o \( \! -group "$GEGROUP" \)` to test whether you should `chown`.

tst-dsugar:
I tested with -maxdepth 0 doing it this way...but yes, it should use "$ASSET_ROOT" to be most accurate.

earth_enterprise/rpms/opengee-server/pre-uninstall.sh
```diff
@@ -17,16 +17,12 @@
...
-main_preuninstall()
-{
-    service geserver stop
-    remove_service geserver
-}
 
-#-----------------------------------------------------------------
-# Pre-install Main
-#-----------------------------------------------------------------
+service geserver stop
```
tst-ccamp: You don't need to stop the service if this is an upgrade. You should only be doing things that don't effect a running service at this point. Or if you're supper worried about that, you can stop it, do your work, and then start it again.

earth_enterprise/rpms/opengee-fusion/pre-uninstall.sh
```diff
@@ -21,6 +21,10 @@ set +x
 #-----------------------------------------------------------------
 
 service gefusion stop
```
tst-ccamp: You don't need to stop the service if this is an upgrade. You should only be doing things that don't effect a running service at this point. Or if you're supper worried about that, you can stop it, do your work, and then start it again.

earth_enterprise/rpms/opengee-server/pre-uninstall.sh
tst-dsugar: You do have to stop it if it is an upgrade, because you are replacing the current executables with new ones. Well, you could do condrestart, but I wasnt sure if the current scripts supported that.

PR #652, opened 2018-01-05
Issue 651: Remove unnecessary gtest dependency from libgeprotobuf

### LINKED ISSUE #651: libgeprotobuf depends on gtest-devel
On some platforms, libgeprotobuf is built such that it depends on gtest-devel.  Since gtest is only used for testing this should not be a requirement for run time libraries.

Steps to reproduce:

1. Build RPMs on CentOS 7
1. Install the RPMs on another machine on which gtest-devel is not installed
1. Symptom 1: run `sudo /opt/google/bin/geserveradmin`.  You will get an error message indicating that there was a problem loading libgtest.
1. Symptom 2: run `/etc/init.d/geserver restart`. You will get an error because a file named "stream_space" does not exist and the server will fail to start.

tst-lsavoie:
Fixes #651 

This issue removes the dependency of gtest from libgeprotobuf.  It's a bit hacky since it's not clear why gtest is being included as a dependency in the first place, but it does the job.

Verification steps:
1. Build RPMs on a CentOS 7 box.
1. Install the RPMs on a separate CentOS 7 box that doesn't have gtest or gtest-devel installed.
1. Run `sudo /etc/init.d/geserver restart`.  There should be no failures.
1. Run `sudo /opt/google/bin/geserveradmin`.  It should print out a help message and not complain about missing libraries.

### DIFF
--- a/earth_enterprise/src/google/protobuf/SConscript
+++ b/earth_enterprise/src/google/protobuf/SConscript
@@ -70,6 +70,18 @@ stub_files = [
   'substitute.cc'
 ]
 
+# Remove gtest from the list of required libraries. For some reason the LIBS
+# variable includes "gtest" here, which makes libgeprotobuf unnecessarily
+# depend on gtest-devel. This is a hack to fix that problem, but it's a fairly
+# safe hack.
+try:
+  local_libs = local_env['LIBS']
+  local_libs = [lib for lib in local_libs if lib is not "gtest"]
+  local_env['LIBS'] = local_libs
+except:
+  # No libraries were defined; nothing to do
+  pass
+
 # Compile the cc files and add them to the "proto.a" lib.
 geprotobuf = local_env.sharedLib('geprotobuf', cpp_files
                     + map(lambda f: 'io/' + f, io_files)
--- a/earth_enterprise/src/google/protobuf/SConscript
+++ b/earth_enterprise/src/google/protobuf/SConscript
@@ -74,13 +74,10 @@ stub_files = [
 # variable includes "gtest" here, which makes libgeprotobuf unnecessarily
 # depend on gtest-devel. This is a hack to fix that problem, but it's a fairly
 # safe hack.
-try:
+if 'LIBS' in local_env:
   local_libs = local_env['LIBS']
   local_libs = [lib for lib in local_libs if lib is not "gtest"]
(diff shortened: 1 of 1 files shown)


tst-lsavoie:
Yep, somehow gtest gets into LIBS.  I tried tracing how that happens but I wasn't able to come up with anything.  scons is a bit too opaque about how things happen.

tst-ppenev:
@tst-lsavoie: Run `find-requires` on each file that goes into the `server` package.

E.g.:
```bash
#! /bin/bash

: ${SEARCH_PATH:=/tmp/fusion_os_install}

find "$SEARCH_PATH" |
    while read f; do
        reqs=$(echo "$f" | /usr/lib/rpm/find-requires)
        if [ -n "$reqs" ]; then
            echo "-----------------------------"
            echo "File: $f"
            echo "---------"
            echo "$reqs"
            echo "-----------------------------"
            echo
        fi
    done
```

I don't think testing that `geserver` and `geserveradmin` start is enough to confirm whether a library is linked against.  My guess is that we may be including a unit test in the final package which depends on GTest.

However, we also have diagnostic tests that can be run from the Server UI or the command line, and I don't know if any of them might use GTest.

earth_enterprise/src/google/protobuf/SConscript
```diff
@@ -70,6 +70,18 @@ stub_files = [
...
 
+# Remove gtest from the list of required libraries. For some reason the LIBS
+# variable includes "gtest" here, which makes libgeprotobuf unnecessarily
+# depend on gtest-devel. This is a hack to fix that problem, but it's a fairly
+# safe hack.
+try:
+  local_libs = local_env['LIBS']
+  local_libs = [lib for lib in local_libs if lib is not "gtest"]
+  local_env['LIBS'] = local_libs
+except:
```
tst-ppenev: Minor: You might want to catch a specific exception, if all you're doing is check whether `local_libs` is defined, or use an `if` statement. This implementation will prevent you from seeing any exception that may occur in that block.

tst-lsavoie:
@tst-dsugar @tst-ppenev I made a small change to address Pavel's comment about error handling.  Take another look.

tst-lsavoie:
Found additional places where gtest is a dependency.  Closing - new PR coming soon.

PR #655, opened 2018-01-06
Issue 654 - add missing proj-devel rpm dependency, remove circular dependencies in rpms

tst-dsheremata:
Tested clean packaging builds on a clean CentOS 6 docker image and ran tests on a fresh minimal CentOS 6 container.

### DIFF
--- a/earth_enterprise/BUILD_RPMS.md
+++ b/earth_enterprise/BUILD_RPMS.md
@@ -27,6 +27,12 @@ java-1.7.0-openjdk-devel```.  There are no other special prerequisites, as the
 SCons build target pulls down osPackage and dependencies as needed before
 creating RPMs.
 
+Ensure that all required Debian or Yum repositories are enabled on your 
+platform.  For EL6 and EL7 systems, the EPEL repository must be installed 
+and enabled. 
+RHEL satellite based systems also require the *optional* repository ( i.e., rhui-DISTRO-rhel-server-releases-optional) repository to provide some libraries 
+including xerces-c and some perl libraries.
+
 ## Creating RPMs
 
 A new SCons target was added to make RPMs, **package_install**, that can be run
@@ -54,19 +60,19 @@ of the Git repo checkout directly.
 After the SCons **package_install** target completes the rpm command can
 install and test the RPMs created on CentOS or RHEL.  To do this, change to the
 *rpms/build/distributions* directory.  Use ```sudo rpm -Uhv
-opengee-postgis-*.rpm``` and ```sudo rpm -Uhv opengee-common-*.rpm``` to get
+opengee-common-*.rpm``` and ```sudo rpm -Uhv opengee-postgis-*.rpm``` to get
 the base dependencies installed, in that order.  Afterward use ```sudo rpm -Uhv
 opengee-fusion-*.rpm```, and then ```sudo rpm -Uhv opengee-server-*.rpm``` to
 install and test the RPM packaged Open GEE Fusion and Server.
 
-Yum can also be used to install opengee RPM files, and will also automatically
-install required system dependencies.  However, since the opengee packages are
-not in a repository, using ```sudo yum install opengee-server-*.rpm``` or
+Use `yum` to install opengee RPM files so that it will also automatically
+install required system dependencies if they are in enabled repositories.  However, 
+if the opengee packages installed from local files, use ```sudo yum install *.rpm``` 
+in the `rpms/build/distributions` directory to install all rpms and dependencies
+in one step.  Note that using ```sudo yum install opengee-server-*.rpm``` or
 ```sudo yum install opengee-fusion-*.rpm``` will not automatically install
-**opengee-common** or **opengee-postgis**, and yum will refuse to install if
-they are not already installed.  All the rpm packages could be also installed
-together using ```sudo yum install *.rpm``` in the `rpms/build/distributions`
-directory.
+**opengee-common** or **opengee-postgis** if they are not also specified and yum 
+will refuse to install if they were not previously installed.  
 
 The RPMs install Open GEE to the same */opt/google* path that the non-package
 based install scripts did.  If Open GEE is already installed from the Git
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -440,8 +440,21 @@ built from raster, vector, and location properties data.
     license = 'ASL 2.0'
     url = 'https://github.com/google/earthenterprise'
     type = BINARY
+<<<<<<< HEAD
     autoFindProvides = true
     autoFindRequires = true
+=======
+
+    requires('bash')
+    requires('coreutils')
+    requires('findutils')
+    requires('grep')
+    requires('libxml2')
+    requires('net-tools')
+    requires('sed')
+    requires('util-linux')
+    requires('proj')
+>>>>>>> Added proj rpm dependency to opengee-fusion rpm
 
     requires('opengee-common', openGeeVersion, GREATER | EQUAL)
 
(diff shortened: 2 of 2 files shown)


tst-dsugar:
I think relocating the postgis dependency to resolve the circular dep issue helps a lot, especially since the documentation is again correct.

tst-nfarah:
@tst-dsheremata can we add verification steps ?

tst-dsugar:
I have a new issue too.  The documentation is still wrong, because now opengee-common has to be installed first if done one at a time like described, not opengee-postgis.

tst-dsugar:
@tst-nfarah the docs do need an add a warning that if you use rpm by itself (like rpm -Uhv) to install rpms, you are also manually responsible for installing all required dependencies.  I believe your issue exists for this reason also, and would that would have installed with yum install ./opngee-common-...

earth_enterprise/rpms/build.gradle
```diff
@@ -262,13 +263,14 @@ task openGeeCommonRpm(type: GeeRpm, dependsOn: openGeeSharedFiles) {
...
     autoFindRequires = true
 
-    requires('opengee-postgis', '1.5.8', GREATER | EQUAL)
     requires('chkconfig')
     requires('shadow-utils')
     requires('perl')
     requires('sudo')
-    conflicts('opengee-postgis', '2.0', GREATER | EQUAL)
-
+    requires('xerces-c')
```
tst-ppenev: I don't think this is necessary. It is auto-detected:

```bash
rpm -qpR opengee-common-5.2.0-1.el7.x86_64.rpm | grep xerces

libxerces-c-3.1.so()(64bit)
```

earth_enterprise/rpms/build.gradle
```diff
@@ -262,13 +263,14 @@ task openGeeCommonRpm(type: GeeRpm, dependsOn: openGeeSharedFiles) {
...
+    requires('ogdi')
```
tst-ppenev: I don't think this is necessary. It is auto-detected:

```bash
rpm -qpR opengee-common-5.2.0-1.el7.x86_64.rpm | grep ogdi

libogdi.so.3()(64bit)
```

earth_enterprise/rpms/build.gradle
```diff
@@ -262,13 +263,14 @@ task openGeeCommonRpm(type: GeeRpm, dependsOn: openGeeSharedFiles) {
...
+    requires('geos')
```
tst-ppenev: Unnecessary:

```bash
rpm -qpR opengee-common-5.2.0-1.el7.x86_64.rpm | grep geos

libgeos_c.so.1()(64bit)
```

earth_enterprise/rpms/build.gradle
```diff
@@ -262,13 +263,14 @@ task openGeeCommonRpm(type: GeeRpm, dependsOn: openGeeSharedFiles) {
     autoFindProvides = true
     autoFindRequires = true
 
-    requires('opengee-postgis', '1.5.8', GREATER | EQUAL)
     requires('chkconfig')
     requires('shadow-utils')
     requires('perl')
```
tst-ppenev: `perl` is unnecessarily added here. It's already auto-detected:

```bash
rpm -qpR opengee-common-5.2.0-1.el7.x86_64.rpm | grep perl

/bin/perl
/usr/bin/perl
perl >= 1:5
perl(Config)
perl(Exporter)
perl(Fcntl)
perl(File::Basename)
perl(File::Spec)
perl(File::stat)
perl(File::Temp)
perl(FindBin)
perl(Getopt::Long)
perl(Getopt::Std)
perl(IO::Handle)
perl(lib)
perl(strict)
perl(vars)
perl(warnings)
perl(WWW::Curl::Easy)
```

earth_enterprise/rpms/build.gradle
```diff
@@ -262,13 +263,14 @@ task openGeeCommonRpm(type: GeeRpm, dependsOn: openGeeSharedFiles) {
     autoFindProvides = true
     autoFindRequires = true
 
-    requires('opengee-postgis', '1.5.8', GREATER | EQUAL)
     requires('chkconfig')
```
tst-ppenev: `chkconfig` doesn't have to be added here. It's auto-detected:

```bash
rpm -qpR opengee-common-5.2.0-1.el7.x86_64.rpm | grep chkconfig

chkconfig
```

earth_enterprise/rpms/build.gradle
```diff
@@ -443,7 +448,11 @@ built from raster, vector, and location properties data.
     autoFindProvides = true
     autoFindRequires = true
 
+    requires('proj-devel') /* This dependency is not being picked up automatically by GeeRpm task */
+    requires('opengee-postgis', '1.5.8', GREATER | EQUAL)
```
tst-ppenev: I don't think Fusion needs PostGIS. Why is it a dependency?

earth_enterprise/rpms/build.gradle
tst-dsheremata: On different versions of RHEL6 with satellite servers these dependencies have not always been auto-detected.  More testing should be done before removing these explicit dependencies - I recommend that we leave them in for the time being as a safeguard.

earth_enterprise/rpms/build.gradle
tst-dsheremata: (see below)

earth_enterprise/rpms/build.gradle
tst-dsheremata: (see below)

earth_enterprise/rpms/build.gradle
tst-dsheremata: (see below)

earth_enterprise/rpms/build.gradle
tst-dsheremata: When the `scons release=1 package` target is run as a non-root user, scons does not include  `/sbin` in the path, so auto-detection of `chkconfig` fails.

earth_enterprise/rpms/build.gradle
tst-dsheremata: It was a transient dependency before these changes (by way of opengee-common).  I did not want to remove this without a complete regression test - but perhaps we can in a follow-on issue.

tst-dsheremata:
@tst-dsugar @tst-nfarah  - Per your comments, I also added some revisions to BUILD_RPMS.md to make installing local RPMs more clear to avoid dependency errors.

earth_enterprise/BUILD_RPMS.md
```diff
@@ -27,6 +27,12 @@ java-1.7.0-openjdk-devel```.  There are no other special prerequisites, as the
 SCons build target pulls down osPackage and dependencies as needed before
 creating RPMs.
 
+Ensure that all required Debian or Yum repositories are enabled on your 
+platform.  For EL6 and EL7 systems, the EPEL repository must be installed 
+and enabled. 
+RHEL satellite based systems also require the *optional* repository ( i.e., rhui-DISTRO-rhel-server-releases-optional) repository to provide some libraries
```
tst-ppenev: Minor: Can you include a complete command for enabling a repository with `subscription-manager`?  I don't think I've seen a `rhui-DISTRO` before.

tst-lsavoie:
@tst-dsheremata your last commit didn't get linked up to your tst-dsheremata account.  Most likely your email address isn't listed correctly.  Can you take a look and fix that?

earth_enterprise/BUILD_RPMS.md
tst-dsheremata: Yes - we should do this, but it's a little tricky because there are many variations.  The AWS GovCloud environment I tested on had `rhui-REGION-rhel-server-releases-optional`, RHEL6 has "rhel-6-server-optional-rpms", but that can vary depending on whether it's RHSN, RHN satellite, or "Classic".   I don't believe this should hold up this PR

tst-lsavoie:
@tst-nfarah confirmed that his comments have been addressed.

PR #659, opened 2018-01-09
#657: RPM: Fix Gradle Version Reading

### LINKED ISSUE #657: Run RPM Gradle Tasks without Requiring a SCons Install
As an OpenGEE user, and package maintainer, I would like to be able to run RPM Gradle tasks, such as `clean`, or others, without requiring a full build and staged install of OpenGEE.

## Steps to Reproduce:

1. Remove staged install directories, e.g.:

```bash
rm -Rf /tmp/fusion_os_install
```

2. Run a trivial Gradle task, such as `clean`:

```bash
cd earthenterprise/earth_enterprise/rpms
./gradlew clean
```

## Observed Behavior

```

* Where:
Build file '/home/pav/prod/bitbucket.com/t-sciences/geoplatform/earthenterprise_code/earthenterprise/earth_enterprise/rpms/build.gradle' line: 77

* What went wrong:
> /tmp/fusion_os_install/gee_long_version.txt (No such file or directory)
```

## Expected Behavior

```
BUILD SUCCESSFUL in 6s
1 actionable task: 1 up-to-date
```

tst-ppenev:
* Read `gee_long_version.txt` from the build directory, instead of the staged install directory.

* Call `scons` to build `gee_long_version.txt`, if it doesn't exist.
  
Fixes #657.


earth_enterprise/rpms/build.gradle
```diff
@@ -59,8 +62,19 @@ if (new File('build.properties').exists()) {
     }
 }
 
+def openGeeVersionFile = new File("../src/NATIVE-REL-x86_64/gee_long_version.txt")
+
+if (!openGeeVersionFile.exists()) {
+    GeeCommandLine.expand(
+            ["/usr/bin/scons", "release=1", "NATIVE-REL-x86_64/gee_long_version.txt"],
```
tst-lsavoie: Is there any way to make this more generic, i.e. by not specifying "release=1" and "NATIVE-REL-x86_64"?  Or is that not necessary since it will only build a single file?

tst-ppenev: If we modify the `scons` task to output the file to generic location, we could use that.  Otherwise, where the file ends up, depends on what `release` flag you pass.  So, you need some way to decide what that's going to be.

In general, RPM packaging uses only release builds.  You can try to trick it to use other builds, but you'll have to make some changes by hand.

This command only executes the targets necessary to produce that single file.

tst-lsavoie:
@tst-ppenev can we close this PR?  I think everything in it was included in PR #662 which is now merged.

tst-ppenev:
@tst-lsavoie: Yes. It can be closed, since changes have been already merged.

PR #665, opened 2018-01-18
 #375: Fix psycopg2 loading in Cutter.

### LINKED ISSUE #375: Invalid version of psycopg2 on Ubuntu 16.04
looks like cutter does not recognize the bundled python-psycopg2, hence when we install it  as prerequisite then only cutter would work( identifying it from system path)
However according to src/third_party/SConstruct, either we should un-bundle and test OR 
if there is still a reason to bundle it, then document and verify that  cutter works w/o system install of python-psycopg2

tst-ppenev:
* Added `PYTHONPATH` to Apache environment, so it gets passed on to CGI scripts like Cutter.

Resolves #375.

### DIFF
--- a/earth_enterprise/src/scons/khEnvironment.py
+++ b/earth_enterprise/src/scons/khEnvironment.py
@@ -20,6 +20,7 @@ Define build environments, builders, actions and helper methods in this
 central place and reuse it in all SConscripts as much as possible.
 """
 
+import errno
 import git
 import os
 import os.path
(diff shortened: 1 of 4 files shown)


earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -141,6 +142,31 @@ def WriteToFileStrfunc(file_name, strn):
   return 'WriteToFile(%s, %s)' % (file_name, strn)
 
 
+def StringExpandFileFunc(target, source, env):
```
tst-nfarah: we are creating a file and saving env variables into ? what is the reason ?

tst-nfarah: Should we check the passed arguments ? or they will be always valid ?

tst-ppenev: We're not saving environment variables in a file. This function expands variables in a template file.  The reason is in `cgi.conf.in`.  I think this is better than using `sed` from the command line to substitute substrings in a file as is done in other `scons` targets.

tst-ppenev: What kind of check do you mean?  You would get an exception if you pass a non-existent source file, or if the template contains an invalid format, etc.  Is there something you think should be checked?

tst-nfarah: target or/and source or/and env can be null ? or this would not happen ?

tst-nfarah: ok

tst-ppenev: If any of these is `None`, you'd get an exception.

PR #672, opened 2018-01-23
Fix links in installer

### LINKED ISSUE #671: Documentation link on server admin page is broken in release_5.2.1
It appears the "create_links" function was not included in the RPM installers.

tst-lsavoie:
Fixes #671.

Adds links that the installers create but the RPMs do not.  Also removes non-working links from the server admin page.


tst-dsheremata:
Looks good to me.   (Note - have not built and tested these changes, just a code review.

PR #675, opened 2018-01-24
fix #668 (fusion) dual config push fails

### LINKED ISSUE #668: Cannot push databases with search enabled in dual server configuration
Steps to reproduce:

1. Install GEE Fusion and Server on different machines
1. On the fusion machine build a database that includes a vector project with POI search enabled for at least one layer.
1. Configure the Server in the Server Associations Manager so Fusion can push to it.
1. Push the database to the remote server.

Expected result: The push completes without error.

Actual result: Near the end of the push, the error message "Error: Server-side Internal Error.  Could not get file paths from server. Search server push failed." appears.  The database is still pushed to the server but it won't publish unless you disable POI search.

This bug does not occur when fusion and server are installed on the same machine.  It's possible that it also does not occur when fusion and server are on separate machines but the asset root and publish root are located on a network share that is mounted at the same location on the two machines.

Note: this bug was introduced in 5.1.3 (before the code was open sourced).

**Details**

The push actually occurs by syncing the files between fusion and server.  Fusion makes a list of the files in the database, then asks the server for the list of files that it already has and then sends over the files not in that list.  This happens in earth_enterprise/src/fusion/gepublish/PublisherClient.cpp in the function called `PublisherClient::SyncDatabase`.  That function calls `SyncFiles` twice, once for the stream server and once for the search server (I'm not sure why it's divided like this).  The error actually happens on the server side, so in `PublisherClient::SyncFiles` fusion requests the list of files from the server and the server responds with an error, causing Fusion to error out.

The server side hits the error in earth_enterprise/src/server/wsgi/serve/push/search/core/search_push_manager.py in the function `__UpdatePoiStatus` (which is called from `__SynchronizeDb`).  While the server is building the list of files it has, it also tries to process the POI files it already has.  One of these files is called `poifile`, and it is an XML file that references another file called `poifile0` that has the actual POI data in it.  However, fusion never transfers `poifile0` to the server, so the server can't find the file and throws an exception.

From debugging I have determined that `poifile` is in the list of files generated by fusion to send to the server but `poifile0` does not.  Perhaps this file should be included in the manifest somehow.  Perhaps it should be included in the list some other way.  It's also possible that other files may need to be included in the list; for example, maybe there will be `poifile1/2/3` if the list of POIs is big enough.

One other note: poifile includes the full path to poifile0.  This path does not exist on the server, only on fusion.  However, it looks like there is already logic to handle this case such that the server is able to find the correct location of poifile0 on the server side by modifying the path.  Thus I don't think this is the source of the error.

tst-rwildes:
Fixes #668 

Adding support to make sure poi data files are included in the manifest when pushing a database to the server.

### DIFF
--- a/earth_enterprise/src/common/ManifestEntry.h
+++ b/earth_enterprise/src/common/ManifestEntry.h
@@ -25,12 +25,13 @@
 #include "common/khTypes.h"
 #include "common/khFileUtils.h"
 
+struct ManifestEntry;
 
 struct ManifestEntry {
   std::string orig_path;
   std::string current_path;
   uint64 data_size;
-  std::vector<std::string> dependents;
+  std::vector<ManifestEntry> dependents;
 
   ManifestEntry() : data_size(0) {}
 
--- a/earth_enterprise/src/common/ManifestEntry.h
+++ b/earth_enterprise/src/common/ManifestEntry.h
@@ -58,7 +58,8 @@ struct ManifestEntry {
   bool operator==(const ManifestEntry &other) const {
     return orig_path == other.orig_path
     && current_path == other.current_path
-        && data_size == other.data_size;
+        && data_size == other.data_size
+        && dependents == other.dependents;
   }
 };
 
--- a/earth_enterprise/src/common/ManifestEntry.h
+++ b/earth_enterprise/src/common/ManifestEntry.h
@@ -21,6 +21,7 @@
 #define COMMON_MANIFESTENTRY_H__
 
 #include <string>
+#include <vector>
 #include "common/khTypes.h"
 #include "common/khFileUtils.h"
 
@@ -29,8 +30,9 @@ struct ManifestEntry {
   std::string orig_path;
   std::string current_path;
   uint64 data_size;
+  std::vector<std::string> dependents;
 
-  ManifestEntry() {}
+  ManifestEntry() : data_size(0) {}
 
   explicit ManifestEntry(const std::string& orig_path_) :
       orig_path(orig_path_),
--- a/earth_enterprise/src/fusion/dbmanifest/SConscript
+++ b/earth_enterprise/src/fusion/dbmanifest/SConscript
@@ -29,4 +29,15 @@ dbmanifest = env.sharedLib(
           'gedbroot',
           'geautoingest'])
 
+dbmanifest_test = env.test('dbmanifest_unittest',
+                         ['dbmanifest_unittest.cpp'],
+                         LIBS=['gtest',
+                               'dbmanifest',
+                               'geutil',
+                               'gecommon',
+                               'xerces-c',
+                               'gexml',
+                               'qt-mt'])
+
 env.install('common_lib', [dbmanifest])
+# , 'geutil', 'gecommon', 'gexml', 'xerces-c'
--- a/earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
+++ b/earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
@@ -63,6 +63,7 @@
 #include "common/khStringUtils.h"
 #include "common/khstl.h"
 #include "common/notify.h"
+#include "common/khxml/khdom.h"
 
 DbManifest::DbManifest(std::string* db_path)
   : db_path_(*db_path),
@@ -271,6 +272,7 @@ void DbManifest::GetManifest(geFilePool &file_pool,
           stream_manifest->push_back(ManifestEntry(poi_file));
         }
         search_manifest->push_back(ManifestEntry(poi_file));
+        GetPoiDataFiles(poi_file, search_manifest);
       }
     }
     // These file paths are listed directly in the GedbFusionConfig.
@@ -314,6 +316,7 @@ void DbManifest::GetManifest(geFilePool &file_pool,
           stream_manifest->push_back(ManifestEntry(orig, curr));
         }
         search_manifest->push_back(ManifestEntry(orig, curr));
+        GetPoiDataFiles(curr, search_manifest);
       }
     }
 
@@ -327,6 +330,51 @@ void DbManifest::GetManifest(geFilePool &file_pool,
   GetXmlManifest(stream_manifest);
 }
 
+// Read POI file and extract the data file names from within it and then add
+// them to the manifest too
+void DbManifest::GetPoiDataFiles(const std::string& poi_file,
+                     std::vector<ManifestEntry>* search_manifest)
+{
+  notify(NFY_DEBUG,
+        "Parsing poi file %s looking for data files", poi_file.c_str());
+  khxml::DOMLSParser *parser = CreateDOMParser();
+  if (parser) {
+    khxml::DOMDocument *doc = ReadDocument(parser, poi_file);
+    if (doc) {
+      try {
+        if (khxml::DOMElement *root = doc->getDocumentElement()) {
+          if (khxml::DOMElement* el_search_table = GetFirstNamedChild(root, "SearchTableValues")) {
+            std::vector<std::string> data_files;
+            FromElementWithChildName(el_search_table, "SearchDataFile", data_files);
+            if (data_files.empty()) {
+              notify(NFY_WARN,
+                    "No data files defined for poi file %s", poi_file.c_str());
+            }
+            for (uint i = 0; i < data_files.size(); ++i) {
+              const std::string& data_file = data_files[i];
+              search_manifest->push_back(ManifestEntry(data_file));
+            }
+          } else {
+            notify(NFY_WARN,
+                  "No search table element found in poi file %s", poi_file.c_str());
+          }
+        } else {
+          notify(NFY_WARN,
+                 "No document element loading poi file %s", poi_file.c_str());
+        }
+      } catch(const std::exception &e) {
+        notify(NFY_WARN, "%s while loading poi file %s", e.what(), poi_file.c_str());
+      } catch(...) {
+        notify(NFY_WARN, "Unable to load poi file %s", poi_file.c_str());
+      }
+    } else {
+      notify(NFY_WARN, "Unable to read poi file %s", poi_file.c_str());
+    }
+    DestroyParser(parser);
+  } else {
+    notify(NFY_WARN, "Unable to get parser for poi file %s", poi_file.c_str());
+  }
+}
 
 std::string DbManifest::LocalesFilename() const {
   return khComposePath(db_path_, LocaleSet::LocalesFilename());
--- a/earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
+++ b/earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
@@ -273,7 +273,7 @@ void DbManifest::GetManifest(geFilePool &file_pool,
           stream_manifest->push_back(ManifestEntry(poi_file));
         }
         search_manifest->push_back(ManifestEntry(poi_file));
-        GetPoiDataFiles(poi_file, stream_manifest, search_manifest);
+        GetPoiDataFiles(&(stream_manifest->back()), &(search_manifest->back()));
       }
     }
     // These file paths are listed directly in the GedbFusionConfig.
@@ -318,7 +318,7 @@ void DbManifest::GetManifest(geFilePool &file_pool,
           stream_manifest->push_back(ManifestEntry(orig, curr));
         }
         search_manifest->push_back(ManifestEntry(orig, curr));
-        GetPoiDataFiles(curr, stream_manifest, search_manifest);
+        GetPoiDataFiles(&(stream_manifest->back()), &(search_manifest->back()));
       }
     }
 
@@ -333,15 +333,14 @@ void DbManifest::GetManifest(geFilePool &file_pool,
 }
 
 // Read POI file and extract the data file names from within it and then add
-// them to the manifest too
-void DbManifest::GetPoiDataFiles(const std::string& poi_file,
-                     std::vector<ManifestEntry>* stream_manifest,
-                     std::vector<ManifestEntry>* search_manifest)
+// them to the manifest entries as dependant files
+void DbManifest::GetPoiDataFiles(ManifestEntry* stream_manifest_entry,
+                                 ManifestEntry* search_manifest_entry)
 {
+  const std::string& poi_file = search_manifest_entry->current_path;
+  assert(stream_manifest_entry->current_path == search_manifest_entry->current_path);
   notify(NFY_DEBUG,
         "Parsing POI file %s looking for data files", poi_file.c_str());
-  size_t stream_poi_file_idx = stream_manifest->size() - 1;
-  size_t search_poi_file_idx = search_manifest->size() - 1;
   khxml::DOMLSParser *parser = CreateDOMParser();
   if (parser) {
     khxml::DOMDocument *doc = ReadDocument(parser, poi_file);
@@ -362,12 +361,12 @@ void DbManifest::GetPoiDataFiles(const std::string& poi_file,
               // poi data files have to be set aside until after AddDB processing is done
               // so we hide them in a special location now within the manifest and will
               // use them when we crate an 'upload' manifest during DBsync processing
-              if (stream_manifest != search_manifest) {
+              if (stream_manifest_entry != search_manifest_entry) {
                 notify(NFY_DEBUG, "Adding POI datafile as a dependent file to stream manifest: %s", data_file.c_str());
-                stream_manifest->at(stream_poi_file_idx).dependents.push_back(data_file);
+                stream_manifest_entry->dependents.push_back(data_file);
               }
-              search_manifest->at(search_poi_file_idx).dependents.push_back(data_file);
               notify(NFY_DEBUG, "Adding POI datafile as a dependent file to search manifest: %s", data_file.c_str());
+              search_manifest_entry->dependents.push_back(data_file);
             }
           } else {
             notify(NFY_WARN,
--- a/earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
+++ b/earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
@@ -263,7 +263,7 @@ void DbManifest::GetManifest(geFilePool &file_pool,
     // assetroot).
     geindex::IndexManifest::GetManifest(file_pool, fusion_config_.index_path_,
                                         *stream_manifest, tmp_dir, "", "");
-    notify(NFY_DEBUG, "Processing %ld POI files 1", fusion_config_.poi_file_paths_.size());
+    notify(NFY_DEBUG, "Processing %ld POI files with empty prefix", fusion_config_.poi_file_paths_.size());
     for (uint i = 0; i < fusion_config_.poi_file_paths_.size(); ++i) {
       const std::string poi_file = fusion_config_.poi_file_paths_[i];
       if (!khExists(poi_file)) {
@@ -303,7 +303,7 @@ void DbManifest::GetManifest(geFilePool &file_pool,
       }
     }
 
-    notify(NFY_DEBUG, "Processing %ld POI files 2", fusion_config_.poi_file_paths_.size());
+    notify(NFY_DEBUG, "Processing %ld POI files with prefix", fusion_config_.poi_file_paths_.size());
     // The *.poi file paths are in the GedbFusionConfig.
     for (uint i = 0; i < fusion_config_.poi_file_paths_.size(); ++i) {
       const std::string& orig = fusion_config_.poi_file_paths_[i];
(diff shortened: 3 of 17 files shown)


earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -327,6 +330,51 @@ void DbManifest::GetManifest(geFilePool &file_pool,
...
+        notify(NFY_WARN, "%s while loading poi file %s", e.what(), poi_file.c_str());
+      } catch(...) {
+        notify(NFY_WARN, "Unable to load poi file %s", poi_file.c_str());
+      }
+    } else {
+      notify(NFY_WARN, "Unable to read poi file %s", poi_file.c_str());
+    }
+    DestroyParser(parser);
+  } else {
+    notify(NFY_WARN, "Unable to get parser for poi file %s", poi_file.c_str());
```
tst-dsugar: Certainly it is possible, but not tragic, if we fail to read the poi_file into the parser (readDocument returns a nullptr).  It could be malformed, it could have wrong file perms, it could even be missing.  But if we actually fail to create a parser, that seems like something must be broken inside the app (corrupted memory, out of memory, etc).

tst-rwildes: Agreed.  Returning NULL (`nullptr` is C++11 BTW ;) ) from `CreateDOMParser` means something really bad is going on outside of the scope of this function.  Maybe something more harsh than a warning level debug print is inline for this?  There is no `NFY_ERROR` so the next step up is `NFY_FATAL` and that aborts the process.  Could do that or could raise an exception.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -262,6 +263,7 @@ void DbManifest::GetManifest(geFilePool &file_pool,
     // assetroot).
     geindex::IndexManifest::GetManifest(file_pool, fusion_config_.index_path_,
                                         *stream_manifest, tmp_dir, "", "");
+    notify(NFY_DEBUG, "Processing %ld POI files with empty prefix", fusion_config_.poi_file_paths_.size());
```
tst-lsavoie: Are these new debug statements intended to be permanent?

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -327,6 +332,64 @@ void DbManifest::GetManifest(geFilePool &file_pool,
...
 
+// Read POI file and extract the data file names from within it and then add
+// them to the manifest too
+void DbManifest::GetPoiDataFiles(const std::string& poi_file,
+                     std::vector<ManifestEntry>* stream_manifest,
+                     std::vector<ManifestEntry>* search_manifest)
+{
+  notify(NFY_DEBUG,
+        "Parsing POI file %s looking for data files", poi_file.c_str());
+  size_t stream_poi_file_idx = stream_manifest->size() - 1;
```
tst-lsavoie: Is the POI file guaranteed to be the last entry in the list of manifests?  Also, are we certain there is only one?

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-rwildes: Some I can remove if we think they are excessive.  But most of them did prove helpful and could be helpful later.  I *think* this one would be good to keep (but I could be persuaded otherwise).  But let me know if there are ones you think are excessive.  All the new tracing I added where debug level or  verbose level.  These levels don't log by default.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-rwildes: All the places that call this method currently call it right after adding the POI file to the manifest so right now yes both assumptions are true.  But I could see this being a less then obvious requirement and someone making a change that expectantly broke this requirement.  Any ideas to make it more obvious?  I could comment the assumptions to bring more attention them.  I could also move more of the code into this method and do the actual adding of the POI file in this method too.  Thoughts?

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-dsugar: They could also be commented out for now...though in theory they are already tagged as debug and so I would expect them to be runtime disableable.

tst-rwildes: Yes by default all these messages are disabled at run-time.  You have to start fusion with an environment variable (KH_NFY_LEVEL) set to a numeric value of 6 (for NFY_DEBUG) or 7 (for NFY_VERBOSE).  Otherwise the default is info (NFY_INFO/4).

tst-lsavoie: I didn't realize they were all DEBUG or VERBOSE statements. Given that, I think they're fine.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-lsavoie: Here are a couple of thoughts I had.  I'll leave it up to you to decide how you want to handle it.

1. Instead of passing the full `stream_manifest` and `search_manifest` vectors, just pass the new manifest entry.  So call it like `GetPoiDataFiles(curr, stream_manifest.back(), search_manifest.back())`. That makes the requirement more obvious - you add the element to the back and then get the back element in the next line, which is clear enough for me.

2. Have `GetPoiDataFiles` return the list of dependencies and call it first.  So the code ends up something like
```
         std::vector<std::string> deps = GetPoiDataFiles(curr);
         if (stream_manifest != search_manifest) {
           stream_manifest->push_back(ManifestEntry(orig, curr, deps));
         }
         search_manifest->push_back(ManifestEntry(orig, curr, deps));
```
Then you have to worry about the cost of copying a vector, though.

tst-rwildes: Now that I think about it your option 1 is better than what  I was proposing.  The only reason I'm passing the manifest vectors is because originally we where adding the data files to the manifest vectors.  But we had to change that and given that change passing the manifest entries is a much cleaner solution.

tst-lsavoie:
General comments - we should add/update some unit tests if possible.

tst-rwildes:
@tst-lsavoie Does the build (e.g. scons release=1 build) command run unit tests or is that a separate command?

tst-lsavoie:
It's a separate command:
```
cd src/NATIVE-REL-x86_64/bin/tests
./RunAllTests.pl
```

tst-lsavoie:
I tested this in a single and dual server configuration with multiple vector layers with search enabled and everything worked fine.  I'm going to hold off on approving, though, because I haven't finished my code review.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -327,6 +332,64 @@ void DbManifest::GetManifest(geFilePool &file_pool,
...
+                    "No data files defined for POI file %s", poi_file.c_str());
+            }
+            for (uint i = 0; i < data_files.size(); ++i) {
+
+              const std::string& data_file = data_files[i];
+              
+              // poi data files have to be set aside until after AddDB processing is done
+              // so we hide them in a special location now within the manifest and will
+              // use them when we crate an 'upload' manifest during DBsync processing
+              if (stream_manifest != search_manifest) {
```
tst-lsavoie: I guess if you don't pass the full vector you're going to run into trouble here, but you could do the check in the outer function and pass a boolean or something.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -327,6 +332,64 @@ void DbManifest::GetManifest(geFilePool &file_pool,
...
+                notify(NFY_DEBUG, "Adding POI datafile as a dependent file to stream manifest: %s", data_file.c_str());
+                stream_manifest->at(stream_poi_file_idx).dependents.push_back(data_file);
+              }
+              search_manifest->at(search_poi_file_idx).dependents.push_back(data_file);
```
tst-lsavoie: Very minor: you've got adding the file to the vector and printing the notification in opposite orders here and a couple of lines above.

earth_enterprise/src/fusion/gepublish/PublishHelper.cpp
```diff
@@ -156,8 +156,7 @@ bool PublishHelper::ProcessGetRequest(const std::string& url,
   else
     *args += host_name;
 
-  notify(NFY_DEBUG, "Request %s args: %s %s", url.c_str(), args->c_str(),
-         (username_.empty() ? "nouser" : "user"));
+  notify(NFY_DEBUG, "ProcessGetRequest %s Starting", url.c_str());
```
tst-lsavoie: It might be a good idea to reverse this change since the older version seems to give more information.

earth_enterprise/src/fusion/gepublish/PublisherClient.cpp
```diff
@@ -1221,40 +1269,38 @@ bool PublisherClient::UploadFiles(ServerType server_type,
...
-      // Many LocalTransfers in succession may sometimes overwhelm a server.
-      // To be safe we do a handful of retries with increasing delays between
-      // attempts.
-      int tries = 4;
-      int sleep_secs = 15;
-      while (!LocalTransfer(
-                 server_type, src_path, dest_path, prefer_copy)) {
-        if (--tries == 0) {
+
+      if (!LocalTransferWithRetry(server_prefix, host_root, server_type,
```
tst-lsavoie: I'd prefer brackets around this if statement since the conditional is so long.

earth_enterprise/src/fusion/gepublish/PublisherClient.cpp
```diff
@@ -1162,6 +1163,53 @@ bool PublisherClient::LocalTransfer(ServerType server_type,
   return ProcessGetRequest(server_type, &args, "", &empty_vec);
 }
 
+bool PublisherClient::LocalTransferWithRetry(const std::string& server_prefix,
```
tst-lsavoie: I like the fact that you broke this out into a function.

earth_enterprise/src/fusion/gepublish/PublisherClient.cpp
```diff
@@ -1221,40 +1269,38 @@ bool PublisherClient::UploadFiles(ServerType server_type,
...
+      if (!LocalTransferWithRetry(server_prefix, host_root, server_type,
+                                  tmpdir, entry.current_path, entry.orig_path))
+        return false;
+
+      for(size_t iDep = 0; iDep < entry.dependents.size(); ++iDep) {
+        notify(NFY_DEBUG, "Adding dependent file: %s", entry.dependents[iDep].c_str());
+        dependent_entries.push_back(ManifestEntry(entry.dependents[iDep]));
+      }
+
+      if (report_progress && progress_ != NULL) {
```
tst-lsavoie: I think there might be an error in the progress reporting. The first time I pushed it went to about 99%, then dropped back to about 50%, then went to 100%. It happened really quickly so I'm not completely sure of what I saw, but there may be some error in how this reports progress now.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-rwildes: Actually I thought about that and what I'll do is pass in the pointers to the entries.  If the vector pointers point to the same vector so will the entry pointers point to the same entry.

earth_enterprise/src/server/wsgi/serve/push/search/core/search_push_manager.py
```diff
@@ -522,11 +530,13 @@ def __UpdatePoiStatus(self, file_paths, file_sizes, client_host_name):
           serve_utils.FileSizeMatched(server_file_path, file_sizes[i])):
 
         existing_files.append(file_paths[i])
+        logger.debug("%s already exists", server_file_path)
 
         file_paths.pop(i)
         file_sizes.pop(i)
       else:
         i += 1
+        logger.debug("%s does not exists", server_file_path)
```
tst-lsavoie: Minor: take the ending 's' off of 'exists'

earth_enterprise/src/server/wsgi/serve/push/search/util/search_schema_parser.py
```diff
@@ -394,7 +412,8 @@ def main(argv):
   parser = SearchSchemaParser(db_updater)
 
   start = time.time()
-  (num_fields, sql_search, balloon_style) = parser.Parse(poifile, "test_poi")
+  # TODO(RAW): add prefix path to test this
```
tst-lsavoie: Still a TODO here.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-rwildes: Good catch.  I should be consistent.

earth_enterprise/src/fusion/gepublish/CurlRequest.cpp
```diff
@@ -193,14 +199,17 @@ bool UploadRequest::Start(bool report_progress, geProgress* progress) {
       return false;
     }
     Close();
+    processed_size += entry.data_size;
```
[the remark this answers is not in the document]
tst-rwildes: I moved it here because when I added the else clause to the if statement the code was redundant in both blocks of code.  So if I was going to the do the calculation no matter what I might as well always do it outside the if clause.  Should yield the same result because before the change we only did the calculation and traced the result if report_progress and progress where true.  And that would not change for the duration of the function so it always did it for each loop iteration or it never did it.  I changed things to always calculate and trace the result depending on trace level set.

earth_enterprise/src/fusion/gepublish/PublishHelper.cpp
tst-rwildes: Yes the older version did give more but the problem was the caller also traced out the args at NFY_DEBUG level.  Normally double tracing things at debug level isn't a terrible thing but that parameter is huge and just made reading the debug logs difficult having to wade through almost back to back dumps of that argument.  Maybe put it back here and remove it from the caller?

tst-rwildes: I do see I removed the user trace.  Should put that back.

earth_enterprise/src/fusion/gepublish/PublisherClient.cpp
tst-rwildes: I'll look into it.  There might be a pre-loop somewhere calculating total size.  I would need to fix it to add in the size of the data files.

earth_enterprise/src/server/wsgi/serve/push/search/core/search_push_manager.py
tst-rwildes: Good catch

earth_enterprise/src/server/wsgi/serve/push/search/util/search_schema_parser.py
tst-rwildes: Oh yes.  This looks like a unit test for this module.  Do the unit tests for python get run differently than the unit tests for C++?  I noticed that python .pyc files are not created when I run 'scons build' so I'm guessing the tests are run differently too.

earth_enterprise/src/fusion/gepublish/PublishHelper.cpp
tst-rwildes: I think I'm going to change the caller and put this one back the way it was.

tst-rwildes: I take that back.  This is the caller.  The callee below `request.Start(*args)` also traces the arguments out.  Probably should leave the lower level one in and take the arg trace out of this one as it adds no value.  Actually makes things more difficult.

earth_enterprise/src/fusion/gepublish/PublisherClient.cpp
tst-rwildes: This was actually a good find...  I think my code was double transferring files when for local server.  Ugh...  I'll clean that up.

tst-ccamp:
Since this is still open, be sure to check that this fixes the problem for Disconnected Send as well

earth_enterprise/src/server/wsgi/serve/push/search/util/search_schema_parser.py
tst-lsavoie: I'm not sure there are python unit tests for GEE server yet. I don't want to overload this PR with building a unit testing framework, so it's OK with me if we skip python testing for now.

earth_enterprise/src/common/ManifestEntry.h
```diff
@@ -29,8 +30,9 @@ struct ManifestEntry {
   std::string orig_path;
   std::string current_path;
   uint64 data_size;
+  std::vector<std::string> dependents;
```
tst-lsavoie: Should we add this to the `==` operator?

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -327,6 +332,63 @@ void DbManifest::GetManifest(geFilePool &file_pool,
...
+                    "No data files defined for POI file %s", poi_file.c_str());
+            }
+            for (uint i = 0; i < data_files.size(); ++i) {
+
+              const std::string& data_file = data_files[i];
+              
+              // poi data files have to be set aside until after AddDB processing is done
+              // so we hide them in a special location now within the manifest and will
+              // use them when we crate an 'upload' manifest during DBsync processing
+              if (stream_manifest_entry != search_manifest_entry) {
```
tst-lsavoie: I'm trying really hard to determine if this line is doing the right thing. I think the behavior is a little different than the previous version. `stream_manifest != search_manifest` will return false if the vectors are different but have identical contents.  Since this version is comparing pointers, it will return true for different vectors with identical contents. Of course, it's possible that there was a bug in the original version and this is better. The safest path is probably to do the comparison outside this function and pass a boolean in to indicate whether this line should be executed.

earth_enterprise/src/fusion/gepublish/PublisherClient.cpp
```diff
@@ -1271,36 +1271,13 @@ bool PublisherClient::UploadFiles(ServerType server_type,
...
-    for (; i < num_entries; ++i) {
+    for (size_t i = 0; i < num_entries; ++i) {
       const ManifestEntry& entry = entries[i];
 
       if (!LocalTransferWithRetry(server_prefix, host_root, server_type,
-                                  tmpdir, entry.current_path, entry.orig_path))
+                                  tmpdir, entry.current_path, entry.orig_path)) {
         return false;
-
-      for(size_t iDep = 0; iDep < entry.dependents.size(); ++iDep) {
```
tst-lsavoie: Any reason all of this was removed? Is it not necessary?

earth_enterprise/src/common/ManifestEntry.h
tst-rwildes: Oh...  I missed that.  Yes it should be.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-rwildes: These are pointers the line is not comparing that the values are the but that the instances are different.  Which is what we had before and what we still want.  If the instances are different we want to be sure to add the same data to both instances.  If they are the same then we want to to be sure to not to double add the data.  If we had shared_ptr<>'s we could optimize the code to have two manifests where one happened to point to a sub-set of of the others entries.  Doing that with raw pointers gets ugly....  This isn't the most space efferent solution but that was how the original code worked.

earth_enterprise/src/fusion/gepublish/PublisherClient.cpp
tst-rwildes: No it is not...  I got confused and this was double adding files.  It 'worked' but it re-copied the files twice.  If you are doing links it probably would not be a huge performance issue but if are are actually doing a copy it could be.  Your comment about the strange percent made be go back re-look at things and that is when I discovered the error.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
tst-ccamp: Would we be able to enable the c++11 flag on a per-executable or per CPP basis? I don't like the idea of mixing the project with some c++98 and some c++11 files, but it may be the only practical way for us to start upgrading.

earth_enterprise/src/fusion/dbmanifest/SConscript
```diff
@@ -29,4 +29,15 @@ dbmanifest = env.sharedLib(
...
+                         LIBS=['gtest',
+                               'dbmanifest',
+                               'geutil',
+                               'gecommon',
+                               'xerces-c',
+                               'gexml',
+                               'qt-mt'])
+
 env.install('common_lib', [dbmanifest])
+# , 'geutil', 'gecommon', 'gexml', 'xerces-c'
```
tst-lsavoie: Can this be removed?

tst-rwildes: Oops.  Yes.  I was trying to figure out a linker error I was getting.  I forgot to clean that up....

PR #679, opened 2018-01-25
Globe Assembly and Diassembly fails on Red Hat 6 (and CentOS 6)

### LINKED ISSUE #677: Trying to disassemble or assemble a portable globe or map GLC fails on RedHat 6
On Red hat 6, using ''Settings' → 'Glc Disassembly' in 'GEE Server' admin web page to disassemble or assemble an existing portable globe (with 'search') fails with this error: "FAILED 'module' object has no attribute 'check_output'".

tst-nfarah:
This fix #677 

'check_output' was introduced in python 2.7. Using that function for Red Hat 6 that uses python 2.6 give the reported error ("'module' object has no attribute 'check_output' ".).

- Enable 'cutter': sudo /opt/google/bin/gecutter enable
- Use an existing '.glc' or create one using 'Cutter' 
- Register that '.glc' in GEE Server (or simply copy it to '/opt/google/gehttpd/htdocs/cutter/globes/')
- Try to Disassemble the registered '.glc':
- Go to the GEE Server administrative page (https://server.host.name/admin).
- Go to 'Settings' → 'Glc Disassembly'.
- Select the GLC from the drop down.
- Enter a 'New output directory' that's different from the input directory.
- Click 'Dissassemble glc'

--> An error is posted: FAILED 'module' object has no attribute 'check_output'

Expected behavior: no error. Disassembling should work.

The same error is encountered when trying to assemble a globe:
Start with at least 2 glm or glb files registered in GEE Server
- Open GLC Assembler
- Add 2 glm or glb files as layers in the GLC
- Select "Use Search" on one of the globes
- Fill out the rest of the fields as normal, including KML for polygon
- Click Assemble GLC
Expected: GLC assembly completes without error
Actual: fails with: "FAILED 'module' object has no attribute 'check_output' ".

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
@@ -332,6 +332,11 @@ def RunCmd(os_cmd):
                          stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE)
     results = p.communicate()[1]
+    return_code = p.returncode
+    if return_code != 0:
+      errorMessage = "%s (return code %d)" % (results, return_code)
+      results = errorMessage
+
     return results.split("\n")
   except Exception, e:
     # print "FAILURE: %s" % e.__str__()
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
@@ -332,10 +332,9 @@ def RunCmd(os_cmd):
                          stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE)
     results = p.communicate()[1]
-    return_code = p.returncode
-    if return_code != 0:
-      errorMessage = "%s (return code %d)" % (results, return_code)
-      results = errorMessage
+    if p.returncode != 0:
+      results = "{0} (return code {1})".format(results, p.returncode)
+      return ["", results]
 
     return results.split("\n")
   except Exception, e:
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
@@ -331,12 +331,14 @@ def RunCmd(os_cmd):
     p = subprocess.Popen(os_cmd, shell=False,
                          stdout=subprocess.PIPE,
                          stderr=subprocess.PIPE)
-    results = p.communicate()[1]
-    if p.returncode != 0:
-      results = "{0} (return code {1})".format(results, p.returncode)
+    results = p.communicate()[0]
+    err_data = p.communicate()[1]
+    return_code = p.returncode
+    if return_code != 0:
+      results = "{0} (return code {1})".format(err_data, return_code)
       return ["", results]
-
-    return results.split("\n")
+    else:
+      return results.split("\n")
   except Exception, e:
     # print "FAILURE: %s" % e.__str__()
     return ["", e.__str__()]
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
@@ -328,9 +328,12 @@ def RunCmd(os_cmd):
   try:
     if isinstance(os_cmd, str):
       os_cmd = shlex.split(os_cmd)
-    results = subprocess.check_output(os_cmd)
+    p = subprocess.Popen(os_cmd, shell=False,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
+    results = p.communicate()[1]
     return results.split("\n")
-  except subprocess.CalledProcessError as e:
+  except Exception, e:
     # print "FAILURE: %s" % e.__str__()
     return ["", e.__str__()]
 


earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
```diff
@@ -328,9 +328,12 @@ def RunCmd(os_cmd):
   try:
     if isinstance(os_cmd, str):
       os_cmd = shlex.split(os_cmd)
-    results = subprocess.check_output(os_cmd)
+    p = subprocess.Popen(os_cmd, shell=False,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
+    results = p.communicate()[1]
```
tst-ppenev: This doesn't seem to check the command's exit status like `check_open` does.  I think you should raise an exception, or print a failure if the exit status is non-zero.

tst-ppenev: Q: Do we not want to capture the standard error output, and log it, or show it to the user when the command fails?

tst-nfarah: I, now, added support for it.

tst-nfarah: when the command fails, we would have captured  'results' and it will be returned to the calling clients as such. Does not the above comment and code changed address this (error handling) ?

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
```diff
@@ -328,9 +328,17 @@ def RunCmd(os_cmd):
...
+    return_code = p.returncode
+    if return_code != 0:
+      errorMessage = "%s (return code %d)" % (results, return_code)
+      results = errorMessage
```
tst-ppenev: Minor: The old function returned `["", errorMessage]` on error, this one returns `errorMessage.split()`.  I don't really know what the significance of the initial blank string is, nor does it seem a very readable convention, but I hope code that calls this function won't misinterpret when a command succeeds.

tst-ppenev: Suggestion: The user may benefit from seeing the standard error when a command fails.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
```diff
@@ -328,9 +328,17 @@ def RunCmd(os_cmd):
...
     if isinstance(os_cmd, str):
       os_cmd = shlex.split(os_cmd)
-    results = subprocess.check_output(os_cmd)
+    p = subprocess.Popen(os_cmd, shell=False,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
+    results = p.communicate()[1]
+    return_code = p.returncode
+    if return_code != 0:
+      errorMessage = "%s (return code %d)" % (results, return_code)
```
tst-ppenev: Style: Using string's `format` method is recommended over the `%` operator nowadays. I find it can be, also, more readable.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
tst-dsugar: Yes, I think this addresses the original question too...

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
tst-nfarah: Not sure you meant about the standard error ?

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
tst-nfarah: done

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
tst-nfarah: I restored the old code behavior in case of an error. It is now returning ["", errorMessage].

tst-nfarah: Not sure about what you meant by standard error ? when the comment fails because of an error or an exception an error is returned to the client code.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/utils.py
tst-ppenev: You're right. :stuck_out_tongue: The `results` variable seems like it would have the standard error of the process from the code above, but not the standard output.

tst-nfarah:
@tst-ppenev I created a separate ticket for the 'empty' folder issue: https://github.com/google/earthenterprise/issues/685

That issue is unrelated to this current PR as I was able to reproduce it on '5.2.1' branch where this PR changes are not present (yet).

tst-ppenev:
@tst-nfarah: The scope of the ticket seemed to me to be fixing Composite assembly and disassembly.

This may be a technical argument about Agile and similar methodology.  It doesn't seem to me that a change that fixes an error message, but still leaves the user unable to use the function provides much user value, but if the team is OK with limiting the scope of this pull request to a task for fixing an error message, and leaving the bug for another issue, I can approve.

I'm not sure exactly what you were trying to test on the '5.2.1' branch.  We know disassembly doesn't work on that branch on Cent OS / RHEL 6, so the question is what ticket we're going to fix it on.

I presume that assembly and disassembly work on Cent OS / RHEL 7 from the '5.2.1' branch, but it may be useful to confirm that from someone who's run the test.

tst-nfarah:
@tst-ppenev I tested '5.2.1' on Red Hat 7 to 'skip' the python 2.6 error. The scope of this ticket was to fix the reported error ("'module' object has no attribute 'check_output' ".). But now it looks like the 'Disassembly' is simply not working independently from the python 2.6. It appears to be a different issue from this PR. To track it separately I created a new ticket for it: #685.

tst-ppenev:
@tst-nfarah: Thanks for the clarification. I'm approving this one.

PR #681, opened 2018-01-29
Fix #680 (installation): can't restore old database dumps without poi searches

### LINKED ISSUE #680: Restoring a DB dump from earlier versions of GEE that does not have a search enabled will fail
Restoring a DB dump from earlier versions of GEE that does not have a search enabled will cause the gepoi table to not initialize correctly. This causes users to not be able to push databases with POI search layers. If there is at lease one POI search layer, then this issue error does not occur.

Test cases to run:
Backup from 5.1.2 or 5.2.0 (should be the same)/restore in 5.2.1 without POI search
Backup from 5.1.2 or 5.2.0/restore in 5.2.1 with POI search
Backup/restore on 5.2.1 without POI Search
Backup/restore on 5.2.1 with POI Search

tst-mreinert:
Fixes #680 modified geresetpgdb to ensure that postgres extensions are always added to the poi database when restoring/upgrading

### DIFF
--- a/earth_enterprise/src/server/geresetpgdb
+++ b/earth_enterprise/src/server/geresetpgdb
@@ -304,6 +304,10 @@ if [ "$type" == "upgrade" -o "$type" == "restore" ]; then
   "$BINDIR/createdb" -T template0 geendsnippet
   "$BINDIR/createdb" -T template0 gesearch
   "$BINDIR/createdb" -T template0 gepoi
+
+  echo Adding PostGIS extension
+  "$BINDIR/psql" -U geuser -d gepoi -c "CREATE EXTENSION postgis"
+
   echo "Done."
 elif [ "$type" != "backup" ]; then
   echo "Creating geserve-databases..."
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/server/geresetpgdb
```diff
@@ -279,6 +279,10 @@ if [ "$type" == "upgrade" -o "$type" == "restore" ]; then
   "$BINDIR/createdb" -T template0 geendsnippet
   "$BINDIR/createdb" -T template0 gesearch
   "$BINDIR/createdb" -T template0 gepoi
+
+  echo Adding PostGIS extension
```
tst-ccamp: Minor: We can remove this echo. I think I originally had it in there as debug info

tst-ccamp: Ah, nevermind. It's probably good to keep this here to verify that it's run

tst-jblair:
Reviewed and approved.

Issue #678, opened 2018-01-25
Detailed profiling data for terrain processing


tst-lsavoie:
For debugging purposes it would be useful to have detailed profiling information from the gecombineterrain executable. This will help us understand the performance of the executable and how we can optimize it.

tst-lsavoie:
Closed via PRs #683 and #695.

PR #683, opened 2018-01-30
#678: Profile gecombineterrain

tst-lsavoie:
This PR adds a basic profiling capability to gecombineterrain. The intention is to use the profiling data to debug performance issues.

Partially completes #678 - additional unit tests will be added in a follow up PR.

I've been testing by running
```
sudo /opt/google/bin/gecombineterrain --indexversion 7 --numcpus 64 --output /gevol/assets/Databases/CA_POIs.kdatabase/terrain.kta/ver002/terrain.pack /gevol/assets/Projects/Terrain/worldtopo.ktproject/geindex.kta/ver003/worldtopo-v3.geindex
```
You'll need to change the paths based on your databases and projects. You should see a bunch of "Fusion Notice" logging statements print out to `stderr` with the performance logging data.

### DIFF
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -19,7 +19,7 @@
 #include <sys/syscall.h>
 #include <time.h>
 
-#include "common/profiler.h"
+#include "common/performancelogger.h"
 #include "common/notify.h"
 #include "common/timeutils.h"
 
@@ -27,13 +27,13 @@ using namespace std;
 using namespace getime;
 
 // Initialize static members of Profiler class
-Profiler * const Profiler::_instance = new Profiler();
+PerformanceLogger * const PerformanceLogger::_instance = new PerformanceLogger();
 
 // Initialize static members of BlockProfiler class
-Profiler * const BlockProfiler::profiler = Profiler::instance();
+PerformanceLogger * const BlockPerformanceLogger::perfLogger = PerformanceLogger::instance();
 
 // Log a profiling message
-void Profiler::log(
+void PerformanceLogger::log(
     const string & operation,
     const string & object,
     const timespec startTime,
@@ -62,7 +62,7 @@ void Profiler::log(
 }
 
 // Begin profiling
-BlockProfiler::BlockProfiler(
+BlockPerformanceLogger::BlockPerformanceLogger(
     const string & operation,
     const string & object,
     const size_t size
@@ -76,7 +76,7 @@ BlockProfiler::BlockProfiler(
 }
 
 // Stop profiling and log results
-BlockProfiler::~BlockProfiler() {
+BlockPerformanceLogger::~BlockPerformanceLogger() {
   const timespec endTime = getMonotonicTime();
-  profiler->log(operation, object, startTime, endTime, size);
+  perfLogger->log(operation, object, startTime, endTime, size);
 }
--- a/earth_enterprise/src/common/performancelogger.h
+++ b/earth_enterprise/src/common/performancelogger.h
@@ -12,20 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef PROFILER_H
-#define PROFILER_H
+#ifndef PERFORMANCELOGGER_H
+#define PERFORMANCELOGGER_H
 
 #include <string>
 #include <time.h>
 
 /*
- * Singleton class for profiling events. The programmer can use this class
- * to log the timing of operations. This class is intended for performance
- * debugging.
+ * Singleton class for logging event performance. This class is intended for
+ * performance debugging.
  */
-class Profiler {
+class PerformanceLogger {
   public:
-    static Profiler * instance() { return _instance; }
+    static PerformanceLogger * instance() { return _instance; }
     void log(
         const std::string & operation, // The operation being timed
         const std::string & object,    // The object that the operation is performed on
@@ -33,47 +32,46 @@ class Profiler {
         const timespec endTime,        // The end time of the operation
         const size_t size = 0);        // The size of the object, if applicable
   private:
-    static Profiler * const _instance;
-    Profiler() {}
+    static PerformanceLogger * const _instance;
+    PerformanceLogger() {}
 };
 
 /*
- * A convenience class for profiling a block of code. Profiling begins when an
+ * A convenience class for timing a block of code. Timing begins when an
  * instance of this class is created and ends when the instance goes out of
  * scope.
  */
-class BlockProfiler {
+class BlockPerformanceLogger {
   public:
-    BlockProfiler(
+    BlockPerformanceLogger(
         const std::string & operation,
         const std::string & object,
         const size_t size = 0);
-    ~BlockProfiler();
+    ~BlockPerformanceLogger();
   private:
-    static Profiler * const profiler;
+    static PerformanceLogger * const perfLogger;
     const std::string operation;
     const std::string object;
     const size_t size;
     const timespec startTime;
 };
 
-// Programmers should use the macros below to profile code instead of using the
+// Programmers should use the macros below to time code instead of using the
 // classes above directly. This makes it easy to use compile time flags to
-// exclude the profiling code.
+// exclude the time code.
 
 // For all of the macros, the variadic arguments include the object name
 // (required) and the size (optional).
 
-// This macro will profile a block of code. The call to this macro should be
-// the first statement in the block you want to profile - it will not profile
-// anything that comes before it. It will continue profiling until it goes out
-// of scope.
-#define PROFILE_BLOCK(op, ...) BlockProfiler _block_prof_inst(op, __VA_ARGS__)
+// This macro will time a block of code. The call to this macro should be the
+// first statement in the block you want to time - it will not time anything
+// that comes before it. It will continue timinguntil it goes out of scope.
+#define PERF_LOG_BLOCK(op, ...) BlockPerformanceLogger _block_prof_inst(op, __VA_ARGS__)
 
-// You can use these macros if you want to profile a part of a block of code.
-// If you use multiple profiling statements in the same block you must give
-// each one a unique name.
-#define BEGIN_PROFILING(name, op, ...) BlockProfiler * name = new BlockProfiler(op, __VA_ARGS__)
-#define END_PROFILING(name) delete name
+// You can use these macros if you want to time a part of a block of code. If
+// you use multiple perf loggin statements in the same block you must give each
+// one a unique name.
+#define BEGIN_PERF_LOGGING(name, op, ...) BlockPerformanceLogger * name = new BlockPerformanceLogger(op, __VA_ARGS__)
+#define END_PERF_LOGGING(name) delete name
 
-#endif // PROFILER_H
+#endif // PERFORMANCELOGGER_H
--- a/earth_enterprise/src/common/profiler.cpp
+++ b/earth_enterprise/src/common/profiler.cpp
@@ -21,20 +21,19 @@
 #include <time.h>
 
 #include "common/notify.h"
-#include "combineterrain.h"
-#include "combineterrainprofiler.h"
+#include "common/profiler.h"
 
 using namespace std;
 
-// Initialize static members
-CombineTerrainProfiler * const CombineTerrainProfiler::_instance =
-    new CombineTerrainProfiler();
-CombineTerrainProfiler * const BlockProfiler::terrainProf =
-    CombineTerrainProfiler::instance();
+// Initialize static members of Profiler class
+Profiler * const Profiler::_instance = new Profiler();
+
+// Initialize static members of BlockProfiler class
+Profiler * const BlockProfiler::profiler = Profiler::instance();
 
 // Log a profiling message
-void CombineTerrainProfiler::log
-    (TerrainEvent event, const string operation, const string object, const size_t size) {
+void Profiler::log
+    (ProfileEvent event, const string & operation, const string & object, const size_t size) {
   stringstream message;
   
   message.setf(ios_base::fixed, ios_base::floatfield);
@@ -60,7 +59,7 @@ void CombineTerrainProfiler::log
   notify(NFY_NOTICE, "%s\n", message.str().c_str());
 }
 
-double CombineTerrainProfiler::getTime() const {
+double Profiler::getTime() const {
   struct timespec currTime;
   clock_gettime(CLOCK_MONOTONIC, &currTime);
   return static_cast<double>(currTime.tv_sec) + (currTime.tv_nsec * NSEC_TO_SEC);
--- a/earth_enterprise/src/common/profiler.cpp
+++ b/earth_enterprise/src/common/profiler.cpp
@@ -1,4 +1,4 @@
-// Copyright 2018 the Open GEE Contributors
+// Copyright 2018 The Open GEE Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,12 +17,14 @@
 #include <string>
 #include <unistd.h>
 #include <sys/syscall.h>
+#include <time.h>
 
 #include "common/profiler.h"
 #include "common/notify.h"
 #include "common/timeutils.h"
 
 using namespace std;
+using namespace getime;
 
 // Initialize static members of Profiler class
 Profiler * const Profiler::_instance = new Profiler();
@@ -34,11 +36,11 @@ Profiler * const BlockProfiler::profiler = Profiler::instance();
 void Profiler::log(
     const string & operation,
     const string & object,
-    const double startTime,
-    const double endTime,
+    const timespec startTime,
+    const timespec endTime,
     const size_t size) {
 
-  const double duration = endTime - startTime;
+  const timespec duration = timespecDiff(endTime, startTime);
   const pid_t tid = syscall(SYS_gettid);
   stringstream message;
 
@@ -75,6 +77,6 @@ BlockProfiler::BlockProfiler(
 
 // Stop profiling and log results
 BlockProfiler::~BlockProfiler() {
-  const double endTime = getMonotonicTime();
+  const timespec endTime = getMonotonicTime();
   profiler->log(operation, object, startTime, endTime, size);
 }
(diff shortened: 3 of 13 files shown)


earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.h
```diff
@@ -0,0 +1,78 @@
...
+
+/*
+ * Singleton class for profiling terrain combine events. The programmer can use
+ * this class to mark the beginning and end of events, and the timing of those
+ * events will be written out to a log file. This class is intended for
+ * performance debugging.
+ */
+class CombineTerrainProfiler {
+  public:
+    static CombineTerrainProfiler * instance() { return _instance; }
```
tst-nfarah: May be here check if '_instance' is null and create the instance and if it's not null, simply return the instance.
This way the client code would need to call only 'instance()' static method.

earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.h
```diff
@@ -0,0 +1,78 @@
...
+    inline void Begin(const std::string operation, const QuadtreePath & path, const size_t size) {
+      log(BEGIN, operation, path.AsString(), size);
+    }
+
+    inline void End(const std::string operation, const QuadtreePath & path) {
+      log(END, operation, path.AsString());
+    }
+
+  private:
+    enum TerrainEvent { BEGIN, END };
```
tst-nfarah: May be make 'TerrainEvent' generic as it does not have any thing special for 'terrain'. Also we could use that generic event, in other places where we may want to track the start and end of an operation.

earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.h
```diff
@@ -0,0 +1,78 @@
...
+    
+    static const double NSEC_TO_SEC = 1e-9;
+    static CombineTerrainProfiler * const _instance;
+    
+    CombineTerrainProfiler() {}
```
tst-nfarah: May be make 'CombineTerrainProfiler' generic by renaming the class and move it in a common folder in the source tree.

earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <time.h>
+
+#include "common/notify.h"
+#include "combineterrain.h"
+#include "combineterrainprofiler.h"
+
+using namespace std;
+
+// Initialize static members
+CombineTerrainProfiler * const CombineTerrainProfiler::_instance =
```
tst-nfarah: this might not be needed if we change the 'instance()' method implementation as suggested below.

earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.cpp
```diff
@@ -0,0 +1,67 @@
...
+  message << " " << operation << " " << object;
+  
+  if (size > 0) {
+      message << ", size " << size;
+  }
+  
+  notify(NFY_NOTICE, "%s\n", message.str().c_str());
+}
+
+double CombineTerrainProfiler::getTime() const {
```
tst-nfarah: this is a generic method, may be we can move it a 'TimeUtils' class ?

earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.h
tst-lsavoie: I actually considered that method but prefer the way I did it. If you create the `_instance` object here in an if statement it creates a race condition - if two threads call it at the same time they might create two separate instances. Instead, I opted to make `_instance` a static const member variable, which ensures that it will be instantiated once when the program starts.

earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.cpp
tst-lsavoie: See comment below.

earth_enterprise/src/fusion/gecombineterrain/combineterrainprofiler.h
tst-nfarah: you can create a thread safe singleton, but yeah the way you have will work too.

earth_enterprise/src/common/profiler.h
```diff
@@ -0,0 +1,78 @@
...
+// exclude the profiling code.
+
+// For all of the macros, the variadic arguments include the object name
+// (required) and the size (optional).
+
+// This macro will profile a block of code. The call to this macro should be
+// the first statement in the block you want to profile - it will not profile
+// anything that comes before it. It will continue profiling until it goes out
+// of scope.
+#define PROFILE_BLOCK(op, ...) BlockProfiler _block_prof_inst(op, __VA_ARGS__)
```
tst-nfarah: should these macro be no-op if one do not want to do profiling ? or is this is controlled some where else ?

earth_enterprise/src/common/profiler.h
```diff
@@ -0,0 +1,78 @@
...
+
+/*
+ * Singleton class for profiling events. The programmer can use this class
+ * to log the timing of operations. This class is intended for performance
+ * debugging.
+ */
+class Profiler {
+  public:
+    static Profiler * instance() { return _instance; }
+    void log(
```
tst-nfarah: can 'log' called from more than one thread ? should be thread safe in general ?

earth_enterprise/src/common/profiler.cpp
```diff
@@ -0,0 +1,80 @@
...
+          << "end time: " << endTime
+          << ", "
+          << "duration: " << duration
+          << ", "
+          << "thread: " << tid;
+  if (size > 0) {
+    message << ", size: " << size;
+  }
+
+  notify(NFY_NOTICE, "%s\n", message.str().c_str());
```
tst-nfarah: where log text are stored ? in a file or stdout ? or can this be controlled by passing a 'std::stringstream' as argument ?

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
```diff
@@ -88,6 +90,7 @@ void TerrainCombiner::CombineTerrainPackets(
 
 void TerrainCombiner::CombineChildren(const TerrainQuadsetGroup &quadset_group,
                                       const QuadtreePath even_path) {
+  BEGIN_PROFILING(combineChildrenProf, "CombineChildren", even_path.AsString());
```
tst-nfarah: may be "CombineTerrainChildren" ?

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
```diff
@@ -187,6 +191,7 @@ void TerrainCombiner::WriteCombinedTerrain(
     const QuadtreePath even_path,
     const std::vector<const TerrainPacketItem *> &items,
     uint progress_increment) {
+  BEGIN_PROFILING(calcReadSize, "calcReadSize", even_path.AsString());
```
tst-nfarah: may be use more descriptive name for the operation instead of short name "calcReadSize" ?

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
```diff
@@ -204,8 +209,13 @@ void TerrainCombiner::WriteCombinedTerrain(
     }
   }
 
-  // Create the new PacketInfo and initialize it's raw buffer.
+  END_PROFILING(calcReadSize);
+  BEGIN_PROFILING(readPackets, "read", even_path.AsString(), read_buffer_size);
```
tst-nfarah: i would use a more descriptive name for this operation. "read" sounds very generic.

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
```diff
@@ -234,7 +246,9 @@ void TerrainCombiner::CompressPacket(PacketInfo* packet) {
   std::string& buffer = packet->RawBuffer();
   std::string& compressed_buffer = packet->CompressedBuffer();
 
-    // Compress data
+  PROFILE_BLOCK("compress", packet->EvenPath().AsString(), buffer.size());
```
tst-nfarah: "CompressData" ?

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
```diff
@@ -253,6 +267,7 @@ void TerrainCombiner::CompressPacket(PacketInfo* packet) {
 
 void TerrainCombiner::WritePacket(PacketInfo* packet) {
   std::string& compressed_buffer = packet->CompressedBuffer();
+  PROFILE_BLOCK("write", packet->EvenPath().AsString(), compressed_buffer.size());
```
tst-nfarah: same comment as "read"..may be use more descriptive operation name...

tst-ccamp:
Somewhat minor comment overall. Let's make sure we get our terminology straight. I don't believe we should describe this code as a "profiler". We will have another task for profiling the code using valgrind. So, in order to not get the two mixed up, we should use a different term here like performance logging. I know it sounds trivial, but I've already seen an instance where the term profiler has confused someone, basically thinking we're already working on the code profiling part, which we haven't.

earth_enterprise/src/common/profiler.h
```diff
@@ -0,0 +1,78 @@
...
+    ~BlockProfiler();
+  private:
+    static Profiler * const profiler;
+    const std::string operation;
+    const std::string object;
+    const size_t size;
+    const double startTime;
+};
+
+// Programmers should use the macros below to profile code instead of using the
```
tst-ccamp: Are the compile flags to disable the macros going to be implemented in a follow-up? And a general comment with macros: be sure not to call anything within the macro call that has a side-effect (unless it's a simple cache, for example)

earth_enterprise/src/common/timeutils.h
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+// Basic time-related utilities
+
+#ifndef TIMEUTILS_H
+#define TIMEUTILS_H
+
+const double NSEC_TO_SEC = 1e-9;
+
+double getMonotonicTime();
```
tst-ccamp: Kind of scary this is in the default namespace. Consider adding to some kind of namespace, maybe here and for the Profiler class too.

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
tst-ccamp: When you disable the block, be sure to do it in such a way that it won't execute even_path.AsString()

earth_enterprise/src/common/timeutils.cpp
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <time.h>
+
+#include "timeutils.h"
+
+double getMonotonicTime() {
```
tst-ccamp: I'm not sure how often this is called, but maybe consider making it inline

earth_enterprise/src/common/timeutils.cpp
```diff
@@ -0,0 +1,23 @@
...
+  struct timespec currTime;
+  clock_gettime(CLOCK_MONOTONIC, &currTime);
+  return static_cast<double>(currTime.tv_sec) + (currTime.tv_nsec * NSEC_TO_SEC);
```
tst-ccamp: Do you think we might run into rounding errors by converting this to a double? Looks like we get about 15 digits safely, give or take. It wouldn't be hard to just store the timespec values until after you do the subtraction. It really depends on what they initialize the clock to. If the number of digits in a time for nanoseconds(9) + seconds(unknown) will exceed 15, then we should do the subtraction using the timespec values. The diff, at least should be within the threshold, but I don't think we can guarantee that an individual instance of a time returned from this function will be.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -339,8 +343,10 @@ int main(int argc, char **argv) {
 
     merger->Start();
     JOBSTATS_END(job_stats, MERGER_CREATED);
+    END_PROFILING(create_merger);
 
     // Feed this merge into a QuadsetGather operation
+    BEGIN_PROFILING(create_gatherer, "create", "gatherer");
```
tst-ccamp: Maybe we should put our profiler within the JOBSTATS_ block to avoid accidentally including their overhead within our overhead

earth_enterprise/src/common/profiler.cpp
tst-lsavoie: For now this prints to `stderr`. The idea is that this ticket includes the bare minimum of a logging framework and we can enhance it in future tickets as needed.

earth_enterprise/src/common/profiler.h
tst-lsavoie: Yes, in general this should be thread safe, but that will be implemented in future PRs.

earth_enterprise/src/common/profiler.h
tst-lsavoie: Yes, disabling profiling will be implemented in a follow up.

earth_enterprise/src/common/profiler.h
tst-lsavoie: Yes, that's slated for a follow up.

earth_enterprise/src/common/timeutils.cpp
tst-lsavoie: I wish we had C++11.

This is a good idea. I'll make the change.

tst-lsavoie:
Thanks for the comments; I believe I've addressed them all.

earth_enterprise/src/common/timeutils_unittest.cpp
```diff
@@ -0,0 +1,67 @@
+// Copyright 2018 The Open GEE Contributors
```
tst-lsavoie: FYI, this is an initial set of unit tests to handle the time subtraction code, which is the trickiest new code. Additional unit tests will be added in a follow up ticket.

earth_enterprise/src/common/timeutils.h
```diff
@@ -0,0 +1,34 @@
...
+#include <time.h>
+#include <ostream>
+
+namespace getime {
+
+timespec getMonotonicTime();
+double timespecToDouble(const timespec tm);
+timespec timespecDiff(const timespec x, const timespec y);
+bool operator<(const timespec & x, const timespec & y);
+bool operator>=(const timespec & x, const timespec & y);
```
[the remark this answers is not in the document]
tst-lsavoie: Yes, I thought about that. I decided not to to avoid writing unused code.

earth_enterprise/src/common/timeutils.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <ostream>
+
+#include "khTypes.h"
+#include "timeutils.h"
+
+using namespace std;
+
+namespace getime {
+
+const u_int32_t SEC_TO_NSEC = 1e9;
```
tst-basilhuffman: They're essentially equivalent, but isn't uint32_t generally considered "the standard way" to go about things?

tst-lsavoie: I was going based off of what is defined in khTypes.h, but I like uint32_t better. I'll switch to that.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -0,0 +1,82 @@
...
+// Begin profiling
+BlockPerformanceLogger::BlockPerformanceLogger(
+    const string & operation,
+    const string & object,
+    const size_t size
+  ) :
+    operation(operation),
+    object(object),
+    size(size),
+    startTime(getMonotonicTime())
```
tst-basilhuffman: One last minor style issue: maybe think about renaming member variables, or constructor parameters by prefacing by underscore.  

`Foo::Foo(const bar & _baz) : baz(_baz) { }`

Looks better imo.

tst-basilhuffman:
Output looks good

earth_enterprise/src/common/performancelogger.h
```diff
@@ -0,0 +1,77 @@
...
+// This macro will time a block of code. The call to this macro should be the
+// first statement in the block you want to time - it will not time anything
+// that comes before it. It will continue timinguntil it goes out of scope.
+#define PERF_LOG_BLOCK(op, ...) BlockPerformanceLogger _block_prof_inst(op, __VA_ARGS__)
+
+// You can use these macros if you want to time a part of a block of code. If
+// you use multiple perf loggin statements in the same block you must give each
+// one a unique name.
+#define BEGIN_PERF_LOGGING(name, op, ...) BlockPerformanceLogger * name = new BlockPerformanceLogger(op, __VA_ARGS__)
+#define END_PERF_LOGGING(name) delete name
```
tst-nfarah: hopefully no one would be passing  null name causing an exception. do we want to check ?

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -0,0 +1,82 @@
...
+  stringstream message;
+
+  message.setf(ios_base::fixed, ios_base::floatfield);
+  message << setprecision(9)
+          << operation << " " << object << ": "
+          << "start time: " << startTime
+          << ", "
+          << "end time: " << endTime
+          << ", "
+          << "duration: " << duration
```
tst-nfarah: may be add unit for duration message. seconds i gather ? later we may want this to have more smart into it to output the 'right' unit. ex, if the duration is more than one hour, it would make sense to convert that large second number to hour. etc. but this is for later PR.

tst-lsavoie:
I've found a bug in this code while writing unit tests - the BlockPerformanceLogger is sometimes destroyed before the end of the block. It's a fairly small change - we'll just have to explicitly start and end all the logging. Given that it's a small change I'd like to go ahead and get this ticket merged and fix the bug in the follow up with the unit tests.

PR #691, opened 2018-02-03
Added function to work around xpath issues in el6

tst-basilhuffman:
Added xml_file_get_xpath( ) to install_fusion.sh. Git-lfs is not being detected in src/SConscript 

`File "/home/basilh/workspace/ee_bh/earth_enterprise/src/SConstruct", line 507, in <module>
It appears that git lfs is not installed.  Please
install git lfs (instructions are available at
https://git-lfs.github.com).  Then run
'git lfs pull' to pull down large files before
building.
scons: *** [build] Error 1
`

When relevant portions are commented out, build still fails:

`set -x && cd /home/basilh/workspace/ee_bh/earth_enterprise/src/NATIVE-DBG-x86_64/third_party/libjs/mozilla && tar xzf`
`/home/basilh/workspace/ee_bh/earth_enterprise/third_party/libjs/js-1.8.0-rc1.tar.gz js/src/config` `js/src/editline && touch /home/basilh/workspace/ee_bh/earth_enterprise/src/NATIVE-DBG-x86_64/third_party/libjs/.extract_libjs`
`+ cd /home/basilh/workspace/ee_bh/earth_enterprise/src/NATIVE-DBG-x86_64/third_party/libjs/mozilla`
`+ tar xzf /home/basilh/workspace/ee_bh/earth_enterprise/third_party/libjs/js-1.8.0-rc1.tar.gz `js/src/config js/src/editline`

`gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
scons: *** Error 2
scons: *** [NATIVE-DBG-x86_64/third_party/libjs/.extract_libjs] Error 2
scons: *** [build] Error 2
`

### DIFF
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -225,18 +225,6 @@ show_fusion_running_message()
 	echo -e "To use this installer to upgrade, you must stop all fusion services.\n"	
 }
 
-xml_file_get_xpath()
-{
-    local FILE="$1"
-    local XPATH="$2"
-
-    # Warning: `xmllint --noent` doesn't recognize named entities like
-    # "&lt;" and "&gt;":
-    echo "cat $XPATH" | xmllint --noent --nocdata --shell "$FILE" |
-    # Skip the first and the last line:
-        tail -n +2 | head -n -1
-}
-
 load_systemrc_config()
 {
 	if [ -f "$SYSTEMRC" ]; then
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -225,13 +225,25 @@ show_fusion_running_message()
 	echo -e "To use this installer to upgrade, you must stop all fusion services.\n"	
 }
 
+xml_file_get_xpath()
+{
+    local FILE="$1"
+    local XPATH="$2"
+
+    # Warning: `xmllint --noent` doesn't recognize named entities like
+    # "&lt;" and "&gt;":
+    echo "cat $XPATH" | xmllint --noent --nocdata --shell "$FILE" |
+    # Skip the first and the last line:
+        tail -n +2 | head -n -1
+}
+
 load_systemrc_config()
 {
 	if [ -f "$SYSTEMRC" ]; then
 		# read existing config information
-		ASSET_ROOT=$(xmllint --xpath '//Systemrc/assetroot/text()' $SYSTEMRC)
-		GEFUSIONUSER_NAME=$(xmllint --xpath '//Systemrc/fusionUsername/text()' $SYSTEMRC)
-		GROUPNAME=$(xmllint --xpath '//Systemrc/userGroupname/text()' $SYSTEMRC)		
+		ASSET_ROOT=$(xml_file_get_xpath "$SYSTEMRC" "//Systemrc/assetroot/text()")
+		GEFUSIONUSER_NAME=$(xml_file_get_xpath "$SYSTEMRC" "//Systemrc/fusionUsername/text()")
+		GROUPNAME=$(xml_file_get_xpath "$SYSTEMRC" "//Systemrc/userGroupname/text()")
 	fi
 }
 
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -742,8 +742,9 @@ check_asset_root_volume_size()
 check_fusion_master_or_slave()
 {
     if [ -f "$ASSET_ROOT/.config/volumes.xml" ]; then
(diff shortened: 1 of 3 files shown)


tst-basilhuffman:
Issues above are due to cloning the repo with lfs files concurrently (method 1 from build instructions) and do not occur when splitting into two steps (method 2).

Testing should be done but cloning and pulling lfs files separately.

tst-dsugar:
Was this done in the newer RPM installer scripting, but not in the older shell script installer?

tst-basilhuffman:
Yes, I looked briefly at another way to implement it, but no reason to reinvent the wheel.

tst-lsavoie:
Is this still WIP?

tst-basilhuffman:
It is not, title changed.

tst-lsavoie:
Is this associated with a github issue? If not, it's a good idea to write the issue while you're fixing the bug.

tst-lsavoie:
I tested this on RHEL 6 and it works fine. If we address the other xpath uses this should be good.

tst-basilhuffman:
Modified src/installer/uninstall_fusion.sh to fall in line with extracting xpath

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -225,13 +225,25 @@ show_fusion_running_message()
 	echo -e "To use this installer to upgrade, you must stop all fusion services.\n"	
 }
 
+xml_file_get_xpath()
```
tst-ppenev: The `common.sh` file is sourced by the various install scripts under `src/installer`.  I think that's a better place for the `xml_file_get_xpath()` function than copying it in each script.

However, I think we should do away with having code replicated between `installer` scripts, and RPM install scripts.  Each time you copy code from file to another, you tend to multiply the test steps you have to run by the number of times you've copied the code, and the same applies to every fix other modification to that code, as we have in this example.

You can install Fusion and Server using the same scripts that have been tested as part of RPM installations.  What you'd have to add is copying files and links that would be done by the package manager otherwise.

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -229,9 +229,9 @@ load_systemrc_config()
 {
 	if [ -f "$SYSTEMRC" ]; then
 		# read existing config information
-		ASSET_ROOT=$(xmllint --xpath '//Systemrc/assetroot/text()' $SYSTEMRC)
-		GEFUSIONUSER_NAME=$(xmllint --xpath '//Systemrc/fusionUsername/text()' $SYSTEMRC)
-		GROUPNAME=$(xmllint --xpath '//Systemrc/userGroupname/text()' $SYSTEMRC)		
+		ASSET_ROOT=$(xml_file_get_xpath "$SYSTEMRC" "//Systemrc/assetroot/text()")
```
tst-lsavoie: There's still one more use of xmllint in this file on line 745.

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -742,8 +742,9 @@ check_asset_root_volume_size()
 check_fusion_master_or_slave()
 {
     if [ -f "$ASSET_ROOT/.config/volumes.xml" ]; then
-        EXISTING_HOST=$(xmllint --xpath "//VolumeDefList/volumedefs/item[1]/host/text()" $ASSET_ROOT/.config/volumes.xml | $NEWLINECLEANER)
-
+        #EXISTING_HOST=$(xmllint --xpath "//VolumeDefList/volumedefs/item[1]/host/text()" $ASSET_ROOT/.config/volumes.xml | $NEWLINECLEANER)
```
tst-lsavoie: Minor: probably ought to remove the commented out line as well.

earth_enterprise/src/installer/uninstall_fusion.sh
```diff
@@ -384,11 +384,11 @@ change_volume_ownership()
 
             local volume_name="test"
             local index=1
-            local max_index=$(expr "$(xmllint --xpath 'count(//VolumeDefList/volumedefs/item/localpath)' $CONFIG_VOLUME)")
-
+            local xpath_output=$(xml_file_get_xpath "$CONFIG_VOLUME" "//VolumeDefList/volumedefs/item/localpath")
+	    local max_index=`echo "$xpath_output" | grep localpath | wc -l`
```
tst-lsavoie: Minor: is this intended to be permanent? It looks like debug output.

tst-basilhuffman: The original max index takes a count of the number of elements.  The best way I could think to do this, given the limitations was to take the output, filter on localpath, and count the number of lines i.e. number of localpaths

tst-lsavoie: Got it. Makes sense.

PR #695, opened 2018-02-07
Issue #678: Performance logging tests

tst-lsavoie:
This PR adds unit tests for the performance logger added in PR #683. It also fixes bugs found through unit testing. Most notably it removes the PERF_LOG_BLOCK macro which often didn't profile an entire block.

### DIFF
--- a/earth_enterprise/src/common/performancelogger_unittest.cpp
+++ b/earth_enterprise/src/common/performancelogger_unittest.cpp
@@ -1,0 +1,108 @@
+// Copyright 2018 The Open GEE Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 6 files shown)


earth_enterprise/src/common/performancelogger_unittest.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <time.h>
+
+#include "performancelogger.h"
+
+using namespace std;
+using namespace testing;
+using namespace getime;
+
+// Mock of the PerformanceLogger class
+class MockPerformanceLogger {
```
tst-dsheremata: Not a blocker for this review - but is there a reason we don't subclass PerformanceLogger for the mock test, or have any explicit type safety?   Perhaps the output in PerformanceLogger (i.e., notify(NFY_NOTICE)  could be a separate method and overridden)

tst-lsavoie: Here's my thoughts on that. I feel that you get pretty good type safety with templates because the code won't compile if the class you provide to the template doesn't have the right functions. I could have created an abstract class to be the PerformanceLoggerInterface or something like that and coded to the interface, but that introduces another level of redirection and could be a performance problem on code like this that's called repeatedly.

Also, I intentionally am not testing the PerformanceLogger class in this PR because it only includes the log function and that is expected to change dramatically in follow up tickets, so I felt that those tests would be wasted.

tst-basilhuffman: I came to a similar conclusion from looking at the previous code, that notify(...) could possibly be moved, it would require a number of changes, and perhaps is not the right path to look down, but it is one thing I looked into. This was on previous code though

PR #697, opened 2018-02-07
test #668: Add unit tests for changes done for this fix

tst-rwildes:
Unit tests for changes done for fix of #668

### DIFF
--- a/earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
+++ b/earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
@@ -38,6 +38,10 @@ struct FakeGetIndextManifest : public DbManifest::IndextManifest {
                   }
 };
 
+bool LessThan(const ManifestEntry& lhs_entry, const ManifestEntry& rhs_entry) {
+  return (lhs_entry.orig_path < rhs_entry.orig_path);
+}
+
 class DbManifestTest : public testing::Test {
 
   protected:
@@ -88,8 +92,7 @@ class DbManifestTest : public testing::Test {
     std::vector<std::string> poi_data_files;
   };
 
-  struct FakeDbCreationParameters
-  {
+  struct FakeDbCreationParameters {
     std::string prefix_loc;
     std::vector<PoiEntry> poi_entries;
     std::vector<std::string> icon_files;
@@ -98,8 +101,7 @@ class DbManifestTest : public testing::Test {
   // nice for debugging when things go wrong
   // NOTE: none of this shows up unless you do `export KH_NFY_LEVEL=6`
   // before running unit test
-  void DumpManifest(const std::vector<ManifestEntry>& manifest)
-  {
+  void DumpManifest(const std::vector<ManifestEntry>& manifest) {
     for(size_t idx=0; idx < manifest.size(); ++idx) {
       const ManifestEntry& entry = manifest[idx];
       notify(NFY_WARN, "org=%s:cur=%s;data_size=%ld",
@@ -214,6 +216,32 @@ class DbManifestTest : public testing::Test {
     khEnsureParentDir(actual_toc_path);
     khMakeEmptyFile(actual_toc_path);
   }
+
+  bool IsEquivalent(std::vector<ManifestEntry>* manifest1,
+                    std::vector<ManifestEntry>* manifest2) {
+
+    // sort no matter what.  Makes things easier to figure out
+    // what is missing when we dump the manifests
+    std::sort(manifest1->begin(), manifest1->end(), LessThan);
+    std::sort(manifest2->begin(), manifest2->end(), LessThan);
+    if (manifest1->size() != manifest2->size()) {
+      return false;
+    }
+
+    for (size_t idx=0; idx < manifest1->size(); ++idx) {
+      ManifestEntry& entry1 = manifest1->at(idx);
+      ManifestEntry& entry2 = manifest2->at(idx);
+      if (IsEquivalent(&(entry1.dependents), &(entry2.dependents)) == false ||
+          entry1.orig_path != entry2.orig_path ||
+          entry1.current_path != entry2.current_path ||
+          entry1.data_size != entry2.data_size) {
+            return false;
+      }
+    }
+
+    return true;
+  }
+  
 };
 
 // Basic constructor tests
@@ -264,14 +292,15 @@ TEST_F(DbManifestTest, AssetRootGetPushManifestNoPoiNoIconsSingleManifest) {
 
   db_manifest->GetPushManifest(file_pool_, &single_manifest, &single_manifest, "", "");
 
-  if (expected_manifest != single_manifest) {
+  bool is_equivalent = IsEquivalent(&expected_manifest, &single_manifest);
+  if (!is_equivalent) {
     notify(NFY_WARN, "Dumping expected manifest:");
     DumpManifest(expected_manifest);
     notify(NFY_WARN, "Dumping actual manifest:");
     DumpManifest(single_manifest);
   }
 
-  EXPECT_TRUE(expected_manifest == single_manifest);
+  EXPECT_TRUE(is_equivalent);
 }
 
 TEST_F(DbManifestTest, AssetRootGetPushManifestSingleManifest) {
@@ -328,14 +357,15 @@ TEST_F(DbManifestTest, AssetRootGetPushManifestSingleManifest) {
 
   db_manifest->GetPushManifest(file_pool_, &single_manifest, &single_manifest, "", "");
 
-  if (expected_manifest != single_manifest) {
+  bool is_equivalent = IsEquivalent(&expected_manifest, &single_manifest);
+  if (!is_equivalent) {
     notify(NFY_WARN, "Dumping expected manifest:");
     DumpManifest(expected_manifest);
     notify(NFY_WARN, "Dumping actual manifest:");
     DumpManifest(single_manifest);
   }
 
-  EXPECT_TRUE(expected_manifest == single_manifest);
+  EXPECT_TRUE(is_equivalent);
 }
 
 TEST_F(DbManifestTest, AssetRootGetPushManifest) {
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -26,6 +26,22 @@
 #include "common/khxml/khdom.h"
 #include "fusion/dbmanifest/dbmanifest.h"
 
+struct FakeGetIndextManifest : public DbManifest::IndextManifest {
```
tst-nfarah: may be rename this struct to 'MockGetIndexManifest'.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -42,61 +58,73 @@ class DbManifestTest : public testing::Test {
...
     khPruneDir(test_folder_);
   }
 
-  void CreateFakeDbFolderWithPoiData(const std::string& file_loc = std::string()) {
+  struct PoiEntry {
+    std::string poi_file;
+    std::vector<std::string> poi_data_files;
+  };
+
+  struct FakeDbCreationParameters {
```
tst-nfarah: same comment as above: use 'Mock' instead of 'Fake'.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -42,61 +58,73 @@ class DbManifestTest : public testing::Test {
...
+        entry.orig_path.c_str(), entry.current_path.c_str(), entry.data_size);
+      for(size_t jdx=0; jdx < entry.dependents.size(); ++jdx) {
+        const ManifestEntry& dep_entry = entry.dependents[jdx];
+        notify(NFY_WARN, "  dep_org=%s:dep_cur=%s;dep_data_size=%ld",
+          dep_entry.orig_path.c_str(), dep_entry.current_path.c_str(), dep_entry.data_size);
+      }
+    }
+  }
+
+  void CreateFakeDbFolderWithPoiData(const FakeDbCreationParameters& params = FakeDbCreationParameters()) {
```
tst-nfarah: use 'Mock' for name

tst-rwildes:
@tst-nfarah I can change 'fake' to 'mock' there is definite gray area between these two terms...  I've seen mocks get morphed into fakes and fakes that get morphed into mocks.  I understand the official definitions of the terms but the practical application seems to have grayness.

tst-rwildes:
@tst-nfarah I'm not sure if these are to high level or not to be termed 'unit tests' or if they are really functional tests.  I do 'mock' out a dependency but I also call methods that call other methods and even functions outside the class/method under test itself.  I like to lump tests into two broad categories: tests that can run at build time without installation/setup of product code and tests that require proper install/setup of full product before they can be run.  These tests are most defiantly of the former category.  I found tests that try to isolate things too much become less meaningful.

tst-nfarah:
The tests seems to be closer to functional tests than unit tests. May be use '_tests' for file name instead of '_unittests' as post-fix.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -26,6 +26,22 @@
...
+                  std::vector<ManifestEntry> &manifest,
+                  const std::string& /*tmpdir*/,
+                  const std::string& /*disconnect_prefix*/,
+                  const std::string& /*publish_prefix*/) {
+                    manifest = manifest_; // for now we keep things simple
+                  }
+};
+
+bool LessThan(const ManifestEntry& lhs_entry, const ManifestEntry& rhs_entry) {
+  return (lhs_entry.orig_path < rhs_entry.orig_path);
```
tst-nfarah: we are comparing two paths as one is less than the other ?

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -119,24 +146,102 @@ class DbManifestTest : public testing::Test {
...
+                    std::vector<ManifestEntry>* manifest2) {
 
-    // write out the poi file
-    khEnsureParentDir(actual_poi_file_path);
-    WriteDocument(poi_file_dom, actual_poi_file_path);
+    // sort no matter what.  Makes things easier to figure out
+    // what is missing when we dump the manifests
+    std::sort(manifest1->begin(), manifest1->end(), LessThan);
+    std::sort(manifest2->begin(), manifest2->end(), LessThan);
+    if (manifest1->size() != manifest2->size()) {
```
tst-nfarah: do we need sorting before comparing the sizes ?

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -145,21 +250,310 @@ TEST_F(DbManifestTest, EnforcesAbsolutePath) {
   EXPECT_THROW(TransferOwnership(new DbManifest(&not_an_absolute_path)), khSimpleException);
 }
 
-TEST_F(DbManifestTest, AssetRootDB) {
+// asset root tests
+TEST_F(DbManifestTest, AssetRootGetPushManifestNoPoiNoIconsSingleManifest) {
```
tst-nfarah: i would a comment what this test verify

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -145,21 +250,310 @@ TEST_F(DbManifestTest, EnforcesAbsolutePath) {
...
+    notify(NFY_WARN, "Dumping expected manifest:");
+    DumpManifest(expected_manifest);
+    notify(NFY_WARN, "Dumping actual manifest:");
+    DumpManifest(single_manifest);
+  }
+
+  EXPECT_TRUE(is_equivalent);
+}
+
+TEST_F(DbManifestTest, AssetRootGetPushManifestSingleManifest) {
```
tst-nfarah: add a comment what this test case does try to verify

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -145,21 +250,310 @@ TEST_F(DbManifestTest, EnforcesAbsolutePath) {
...
+    notify(NFY_WARN, "Dumping expected manifest:");
+    DumpManifest(expected_manifest);
+    notify(NFY_WARN, "Dumping actual manifest:");
+    DumpManifest(single_manifest);
+  }
+
+  EXPECT_TRUE(is_equivalent);
+}
+
+TEST_F(DbManifestTest, AssetRootGetPushManifest) {
```
tst-nfarah: same comment about adding a comment for test case

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -145,21 +250,310 @@ TEST_F(DbManifestTest, EnforcesAbsolutePath) {
...
+  get_index_manifest = TransferOwnership(mock_get_index_manifest);
+
+  EXPECT_NO_THROW(db_manifest = TransferOwnership(new DbManifest(&db_path, get_index_manifest)));
   EXPECT_EQ(db_path_v1_, db_path); // should be unchanged
+  EXPECT_TRUE(db_manifest->Prefix().empty());
+
+  db_manifest->GetPushManifest(file_pool_, &stream_manifest, &search_manifest, "", "");
+
+  bool is_equivalent = IsEquivalent(&expected_stream_manifest, &stream_manifest);
+  if (!is_equivalent) {
```
tst-nfarah: should this fail the test execution instead of a warning ?

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
```diff
@@ -145,21 +250,310 @@ TEST_F(DbManifestTest, EnforcesAbsolutePath) {
...
+  get_index_manifest = TransferOwnership(mock_get_index_manifest);
+
+  EXPECT_NO_THROW(db_manifest = TransferOwnership(new DbManifest(&db_path, get_index_manifest)));
+  EXPECT_EQ(db_path_v1_, db_path); // should be changed now
+  EXPECT_EQ(db_manifest->Prefix(), disconnect_path);
+
+  db_manifest->GetPushManifest(file_pool_, &stream_manifest, &search_manifest, "", "");
+
+  bool is_equivalent = IsEquivalent(&expected_stream_manifest, &stream_manifest);
+  if (!is_equivalent) {
```
tst-nfarah: should this fail the test case ?

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
tst-rwildes: Yes.  I'm sorting the manifest entries based on original paths as strings.  The sort function uses less than for sorting so you have to provide a function that indicates which item is less than another item.  Based on how you want to sort things.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
tst-rwildes: We don't need it but it is helpful as the comment states.  When things are missing it is easier to see what is missing if both manifests are sorted.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
tst-rwildes: line 437 will fail the test.  To make failures easier to diagnose I added this pre-check and dump the manifests before ding the EXPECT_TRUE() below which will do the actual failing of the test.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
tst-rwildes: Same as above.  Line 538 in this case will do the failure.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
tst-nfarah: i see. its for diagnostic purposes.

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
tst-nfarah: ok

earth_enterprise/src/fusion/dbmanifest/dbmanifest_unittest.cpp
tst-nfarah: so the sorting is done based on the 'file path' length to sort them from short to long path ? just to dump the file names from short to long file names ?

tst-rwildes: No.  They are sorted alphabetically by original path name.  The point I was making is we could of checked that the two manifests had different number of entries first (before the sort) and if they where not the same numbers then just returned false right away.  *But*, the downside of doing that is that it makes looking at the failure harder to see what is missing (or is extra) because the two manifests might have things in different orders.  By taking the time to sort no matter what it makes failures easier to figure out.  Since this is just test code with relatively small numbers the trade off was worth it.

tst-nfarah: ok. yes it would be useful for debugging purpose to see the diff more easily.

PR #703, opened 2018-02-13
WIP: beginning to look at how to log messages generically

tst-basilhuffman:
examining an envelope strategy at this point. subject to change, just putting this out there to coordinate with others doing logging tasks


earth_enterprise/src/common/performancelogger.h
```diff
@@ -49,26 +72,49 @@ class BlockPerformanceLogger {
...
+        //TODO: refactor into try/catch block
+        if (type < 0 || type >= NUM_ENVELOPE_TYPE_ENUMS)
+        {
+            //print a message to stderr
+            return;
+        }
         ended = true;
+        perfEnvelope * envelope;
         const timespec endTime = getime::getMonotonicTime();
+        if (type == PERFORMANCE) {
```
tst-ppenev: Can we not just use a constructor for each class, instead of having to delegate that based on a type ID?

tst-basilhuffman: If I understand you correctly, that would defeat the point of subclassing, and b would require a lot more logic to be put into toString(), as an object (struct/class) can't self instantiate. I suppose I could break instantiation off into a separate factory function though.

tst-lsavoie: I agree with @tst-ppenev here: if we created new macros for each type of log (i.e., BEGIN_PERF_LOGGING, BEGIN_IO_LOGGING, etc), those macros could create the proper envelope and just pass it to this function. I haven't completely thought through how that would work but I think it would be cleaner.

PR #705, opened 2018-02-13
WIP: Performance logging

tst-dmendyke:
Created an added a mutex class based on the timed mutex function in pthreads.  
Simplified the formatting of the output message string as strict CVS.  That may need additional work.

The use of the notify function will have to be replaced with standard C++ file stream output calls.

### DIFF
--- a/earth_enterprise/src/common/mutex.cpp
+++ b/earth_enterprise/src/common/mutex.cpp
@@ -1,0 +1,67 @@
+// Copyright 2018 The Open GEE Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


tst-dsheremata:
@googlebot

tst-dmendyke:
@googlebot  Filled out the CLA so hopefully this allows for a successful request.

earth_enterprise/src/common/mutex.cpp
```diff
@@ -0,0 +1,67 @@
+// Copyright 2018 The Open GEE Contributors
```
tst-lsavoie: There are some existing mutex classes in khThread.cpp/.h. I think khLockGuard would do the same thing as this class. It might be simpler to use the existing code.

tst-lsavoie:
In general this looks good. Perhaps you're already working on this since this is marked WIP, but you should also add a compile flag that compiles out all of the performance logging code.

earth_enterprise/src/common/mutex.cpp
tst-ppenev: :+1:

I see that being done in #708, which, however, also includes some `pip` changes.  We just need a clean PR with the mutex.

tst-dmendyke:
Oh, right.  I failed to branch off of main.

earth_enterprise/src/common/mutex.cpp
tst-dsheremata: Any updates w/r these comments?

tst-lsavoie:
Superseded by PR #708.

PR #707, opened 2018-02-14
** WIP **: Proposed design for profiling framework

tst-lsavoie:
This PR is for reference only; it is not intended to be merged.

This is my proposed design for the profiling framework that will allow multiple people to work on it at the same time. When you fill in your function (e.g., `logIo`) you may also need to write a wrapper class that records the appropriate data and calls the function - how you do that is up to you.

### DIFF
--- a/earth_enterprise/src/common/performancelogger.h
+++ b/earth_enterprise/src/common/performancelogger.h
@@ -101,4 +101,10 @@ class BlockPerformanceLogger {
   BlockPerformanceLogger<PerformanceLogger> name(op, __VA_ARGS__)
 #define END_PERF_LOGGING(name) name.end()
 
+#define BEGIN_IO_LOGGING() // fill in as needed
+#define BEGIN_IO_LOGGING() // fill in as needed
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/common/performancelogger.h
```diff
@@ -87,4 +101,10 @@ class BlockPerformanceLogger {
   BlockPerformanceLogger<PerformanceLogger> name(op, __VA_ARGS__)
 #define END_PERF_LOGGING(name) name.end()
 
+#define BEGIN_IO_LOGGING() // fill in as needed
+#define BEGIN_IO_LOGGING() // fill in as needed
+#define BEGIN_THREAD_LOGGING() // fill in as needed
+#define BEGIN_THREAD_LOGGING() // fill in as needed
```
tst-basilhuffman: Perhaps create a third macro which creates the envelope?

Like

 `#define CREATE_IO_ENVELOPE(envlp, ...)`

tst-lsavoie:
Merged into PR #724.

PR #708, opened 2018-02-14
WIP: Development of Performance Metric

tst-dmendyke:
Basic thread safety wrapper around ostream output.

### DIFF
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -57,3 +57,12 @@ void PerformanceLogger::log(
 
   notify(NFY_NOTICE, "%s\n", message.str().c_str());
 }
+
+// Thread safety wrapper for log output
+PerformanceLogger::do_notify( const string& message, ostream& out, khMutex& mutex ) {
+
+  mutex.lock();  // wait for and then lock the mutex
+  out << message << endl;
+  mutex.unlock();  // clear the lock
+
+};  // end do_notify
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -26,6 +26,8 @@
 using namespace std;
 using namespace getime;
 
+#ifdef LOG_PERFORMANCE
+
 // Initialize static members of Profiler class
 PerformanceLogger * const PerformanceLogger::_instance = new PerformanceLogger();
 
@@ -38,22 +40,17 @@ void PerformanceLogger::log(
     const size_t size) {
 
   const timespec duration = timespecDiff(endTime, startTime);
-  const pid_t tid = syscall(SYS_gettid);
   stringstream message;
+  pthread_t tid = pthread_self();
 
   message.setf(ios_base::fixed, ios_base::floatfield);
-  message << setprecision(9)
-          << operation << " " << object << ": "
-          << "start time: " << startTime
-          << ", "
-          << "end time: " << endTime
-          << ", "
-          << "duration: " << duration
-          << ", "
-          << "thread: " << tid;
-  if (size > 0) {
-    message << ", size: " << size;
-  }
+  message << tid << ", "
+          << startTime << ", "
+          << endTime << ", "
+          << duration << ", "
+          << size << ", "
+          << operation << ", "
+          << object;
 
   notify(NFY_NOTICE, "%s\n", message.str().c_str());
 }
@@ -61,8 +58,14 @@ void PerformanceLogger::log(
 // Thread safety wrapper for log output
 PerformanceLogger::do_notify( const string& message, ostream& out, khMutex& mutex ) {
 
+  // Get the thread ID
+  pthread_t tid = pthread_self();
+
   mutex.lock();  // wait for and then lock the mutex
-  out << message << endl;
+  out << tid << ", " << message << endl;
   mutex.unlock();  // clear the lock
 
 };  // end do_notify
+
+
+#endif  // LOG_PERFORMANCE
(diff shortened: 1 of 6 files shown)


earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -57,3 +57,12 @@ void PerformanceLogger::log(
 
   notify(NFY_NOTICE, "%s\n", message.str().c_str());
 }
+
+// Thread safety wrapper for log output
+PerformanceLogger::do_notify( const string& message, ostream& out, khMutex& mutex ) {
+
+  mutex.lock();  // wait for and then lock the mutex
```
tst-ppenev: I believe the GEE way of doing this is:

```
  {
    khLockGuard lock(mutex);

    out << message << endl;
  }
```

This way, the mutex won't remain locked if there's an exception during output (e.g., disk full, etc.), which can cause threads to remain deadlocked.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -38,22 +40,32 @@ void PerformanceLogger::log(
...
-  }
+  message << tid << ", "
+          << startTime << ", "
+          << endTime << ", "
+          << duration << ", "
+          << size << ", "
+          << operation << ", "
+          << object;
 
   notify(NFY_NOTICE, "%s\n", message.str().c_str());
```
tst-lsavoie: Need to change this to call do_notify

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -38,22 +40,32 @@ void PerformanceLogger::log(
...
 }
+
+// Thread safety wrapper for log output
+PerformanceLogger::do_notify( const string& message, ostream& out, khMutex& mutex ) {
+
+  // Get the thread ID
+  pthread_t tid = pthread_self();
+
+  mutex.lock();  // wait for and then lock the mutex
+  out << tid << ", " << message << endl;
```
tst-lsavoie: I think we decided that we want to include the pid as well as the tid here.

earth_enterprise/src/common/performancelogger.cpp
tst-lsavoie: This will also involve creating a file to hold the output. We want the file name to be different for every execution, which you can do by appending a timestamp or something like that.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -38,22 +40,32 @@ void PerformanceLogger::log(
...
-          << ", "
-          << "end time: " << endTime
-          << ", "
-          << "duration: " << duration
-          << ", "
-          << "thread: " << tid;
-  if (size > 0) {
-    message << ", size: " << size;
-  }
+  message << tid << ", "
```
tst-lsavoie: We should remove the tid from here since we're adding it in do_notify

tst-lsavoie:
I think we'll need some method to pass a flag to scons that sets this compiler flag. It would be nice if we could only set it for gecombineterrain and the performance logging tests, but I'm not sure if that's possible.

earth_enterprise/src/common/performancelogger.h
```diff
@@ -83,8 +90,13 @@ class BlockPerformanceLogger {
 //
 // If you use multiple performance logging statements in the same scope you
 // must give each one a unique name.
+#ifdef LOG_PERFORMANCE
 #define BEGIN_PERF_LOGGING(name, op, ...) \
   BlockPerformanceLogger<PerformanceLogger> name(op, __VA_ARGS__)
 #define END_PERF_LOGGING(name) name.end()
+#else
+#define BEGIN_PERF_LOGGING
```
tst-lsavoie: I get compile errors if I don't specify `LOG_PERFORMANCE`. They can be fixed by adding an argument list here and on the line below - i.e., `BEGIN_PERF_LOGGING(...)`.

earth_enterprise/src/common/performancelogger.h
```diff
@@ -33,6 +37,7 @@ class PerformanceLogger {
         const timespec startTime,      // The start time of the operation
         const timespec endTime,        // The end time of the operation
         const size_t size = 0);        // The size of the object, if applicable
+    void do_notify( const string&, std::ostream&, khMutex& );  // func decloration
```
tst-lsavoie: This function should be private.

tst-dmendyke:
I had written and used code that protected against an exception being
thrown during output but you recommended that I replace it with the use of
the predefined mutex class.

earth_enterprise/src/common/performancelogger.cpp
tst-dsheremata: Hey Daniel - any updates to this comment?

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -38,22 +40,31 @@ void PerformanceLogger::log(
...
+
+// Thread safety wrapper for log output
+void PerformanceLogger::do_notify( const string& message, ostream& out, khMutex& mutex ) {
+
+  // Get the thread ID
+  pthread_t tid = pthread_self();
+
+  {  // atomic inner block
+    khLockGuard lock( mutex );
+    mutex.unlock();  // clear the lock
```
tst-lsavoie: Looks like we somehow lost the code to actually write out the data here.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -38,22 +40,32 @@ void PerformanceLogger::log(
...
 
   notify(NFY_NOTICE, "%s\n", message.str().c_str());
 }
+
+// Thread safety wrapper for log output
+void PerformanceLogger::do_notify( const string& message, ostream& out, khMutex& mutex ) {
+
+  // Get the thread ID
+  pthread_t tid = pthread_self();
+  pid_t pid = getppid();  // get the ID of the parent process
```
tst-lsavoie: I think what we want here is `getpid`, not `getppid`, though having both wouldn't hurt.

tst-lsavoie:
Merged into PR #724.

PR #709, opened 2018-02-15
Log Thread Statistics

tst-ppenev:
* Added timing of `khThread` semaphore operations, and thread starting.

### DIFF
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -16,6 +16,8 @@
 #include <fstream>
 #include <sstream>
 #include <string>
+
+#include <errno.h>
 #include <unistd.h>
 #include <sys/syscall.h>
 #include <time.h>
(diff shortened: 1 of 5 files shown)


earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -26,11 +29,42 @@
 using namespace std;
 using namespace getime;
 
-// Initialize static members of Profiler class
-PerformanceLogger * const PerformanceLogger::_instance = new PerformanceLogger();
+namespace performance_logger {
```
tst-lsavoie: I'm not super happy with the duplication of the locking code but I don't see a better way to handle this. Maybe include a comment about why it was duplicated.

tst-lsavoie: nvm, I see you added the comment in the .cpp file.

tst-lsavoie:
Merged into PR #724.

PR #717, opened 2018-02-19
WIP Gp 974 log io

tst-basilhuffman:
Not tested. Needs some further qustions answered. Pushed as WIP for initial review and expansion on doNotify, plus some comments on file name composition

### DIFF
--- a/earth_enterprise/src/common/packetfile/packetfilereaderpool.h
+++ b/earth_enterprise/src/common/packetfile/packetfilereaderpool.h
@@ -31,7 +31,6 @@
 #include <khTypes.h>
 #include <khGuard.h>
 #include <packetfile/packetfilereader.h>
-#include "common/performancelogger.h"
 
 // PacketFileReaderToken
 //
--- a/earth_enterprise/src/common/packetfile/packetfilereaderpool.h
+++ b/earth_enterprise/src/common/packetfile/packetfilereaderpool.h
@@ -31,6 +31,7 @@
 #include <khTypes.h>
 #include <khGuard.h>
 #include <packetfile/packetfilereader.h>
+#include "common/performancelogger.h"
 
 // PacketFileReaderToken
 //
@@ -65,8 +66,7 @@ class PacketFileReaderToken {
  private:
   PacketFileReaderToken(uint32 pool_hash, uint32 packet_file_index)
       : pool_hash_(pool_hash),
-        packet_file_index_(packet_file_index) {
-  }
+        packet_file_index_(packet_file_index) {}
   friend class PacketFileReaderPool;
 
   uint32 pool_hash_;
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -18,7 +18,7 @@
 #include <unistd.h>
 #include <sys/syscall.h>
 #include <time.h>
-
+#include <vector>
 #include "common/performancelogger.h"
 #include "common/notify.h"
 #include "common/timeutils.h"
@@ -27,6 +27,7 @@ using namespace std;
 using namespace getime;
 
 std::ofstream ioPrefFile;
+std::string ioFileName;
 std::ofstream timePrefFile;
 std::ofstream threadPrefFile;
 
@@ -74,9 +75,34 @@ void openFiles() {
     io << "csv";
     thread << "csv";
     time << "csv";
-    if (!ioPrefFile.is_open()) ioPrefFile.open(io.str().c_str());
-    if (!timePrefFile.is_open()) timePrefFile.open(time.str().c_str());
-    if (!threadPrefFile.is_open()) threadPrefFile.open(thread.str().c_str());
+    // if not open, open and add header
+    if (!ioPrefFile.is_open()) { 
+      ioPrefFile.open(io.str().c_str());
+      ioFileName = io.str();
+      ioPrefFile << "operation,object,startTime,endTime,bufferSize" << endl;
+    }
+    if (!timePrefFile.is_open()) {
+      timePrefFile.open(time.str().c_str());
+      timePrefFile << "operation,object,startTime,endTime,duration,threadID,objectSize" << endl;
+    }
+    if (!threadPrefFile.is_open()) {
+      threadPrefFile.open(thread.str().c_str());
+      //TODO: pavel fill in header line
+    }
+}
+
+void ioPostProcess()
+{
+    /*
+        Do IO post processing and output:
+            - total number of write requests (number of lines in file)
+            - throughput: (size/duration) * 1024 = MbPS
+
+        Also look into a simple CVS parser. I have used https://github.com/ben-strasser/fast-cpp-csv-parser
+        but not sure about licensing, as this is BSD license. I can write my own, but I'd prefer to
+        not re-invent the wheel.
+    */
+
 }
 
 void closeFiles() {
@@ -90,12 +116,12 @@ PerformanceLogger& PerformanceLogger::instance() {
     return _instance;
 }
 
-void PerformanceLogger::doNotify(string message, string fileName)
+void PerformanceLogger::doNotify(string message, ostream& fileName)
 {
     //implemented by Daniel
 }
 
-/*void PerformanceLogger::logIO(
+void PerformanceLogger::logIO(
         const string & operation,
         const string & object,
         const timespec startTime,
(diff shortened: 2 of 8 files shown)


earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,139 @@
...
+    io << val << delimiter;
+    thread << val << delimiter;
+    time << val << delimiter;
+}
+
+void openFiles() {
+    time_t t = time(0);
+    tm date = *localtime(&t);
+    std::stringstream io,time,thread;
+    io << "io.";
```
tst-ppenev: Minor: Can you use more descriptive variable names. E.g., `io_file_name`, `timing_file_name`, `threads_file_name`?

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,139 @@
...
+    time << "time.";
+    thread << "thread.";
+    checkForLeadingZero(io,thread,time,date.tm_mon);
```
tst-ppenev: I think this would be more readable, if you did something like:

```C++
char buffer[256];

// Format file names that look like <io_stast.12-25-2018-11:30:55.csv>:
strftime(buffer, sizeof(buffer), "io_stats.%m-%d-%Y-%H:%M:%S.csv", &date);
std::string io_stats_file_name(buffer);

strftime(buffer, sizeof(buffer), "timing.%m-%d-%Y-%H:%M:%S.csv", &date);
std::string timing_stats_file_name(buffer);

strftime(buffer, sizeof(buffer), "thread_stats.%m-%d-%Y-%H:%M:%S.csv", &date);
std::string thread_stats_file_name(buffer);
```

tst-ppenev: Minor: File names sort better, and dates are less ambiguous, if they are formatted from the most significant to the least significant component.  Like: `year-month-day-hour:minute:second.etc`.

I would also think that you may want to see all the files that refer to a single run next to each other in directory listing, rather than all the files with I/O stats, then all the files with timing, and then all the files with thread stats.  I.e., I suggest naming the files like: `year-month-day-hour:minute:second-io_stats.csv`, etc.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,139 @@
...
 #include "common/performancelogger.h"
 #include "common/notify.h"
 #include "common/timeutils.h"
 
 using namespace std;
 using namespace getime;
 
-// Initialize static members of Profiler class
-PerformanceLogger * const PerformanceLogger::_instance = new PerformanceLogger();
+std::ofstream ioPrefFile;
```
tst-ppenev: Suggestion: These look like they should be member variables of the `PerformanceLogger` class, rather than global variables.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,139 @@
...
+    checkForLeadingZero(io,thread,time,date.tm_min);
+    printToSStream(io,thread,time,':',date.tm_min);
+    checkForLeadingZero(io,thread,time,date.tm_sec);
+    printToSStream(io,thread,time,'.',date.tm_sec);
+    io << "csv";
+    thread << "csv";
+    time << "csv";
+    // if not open, open and add header
+    if (!ioPrefFile.is_open()) { 
+      ioPrefFile.open(io.str().c_str());
```
tst-ppenev: Minor: You don't need a C string to open a file.  A `std::string` works fine.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,139 @@
...
+    if (!threadPrefFile.is_open()) {
+      threadPrefFile.open(thread.str().c_str());
+      //TODO: pavel fill in header line
+    }
+}
+
+void ioPostProcess()
+{
+    /*
+        Do IO post processing and output:
```
tst-ppenev: I'm not sure the logger is the place for processing the log output.  It seemed to me that would be done in another utility that reads the logs, and does aggregation and other processing.  It would probably not need to be in C++, either.  E.g., there are many statistical and graphing tools in Python.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,139 @@
...
+            - throughput: (size/duration) * 1024 = MbPS
+
+        Also look into a simple CVS parser. I have used https://github.com/ben-strasser/fast-cpp-csv-parser
+        but not sure about licensing, as this is BSD license. I can write my own, but I'd prefer to
+        not re-invent the wheel.
+    */
+
+}
+
+void closeFiles() {
```
tst-ppenev: I think it may be simpler to manage file stream closing as part of the logger destruction.  In fact, that would happen automatically, if the file streams are members of the logger class.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,139 @@
...
+    return _instance;
+}
+
+void PerformanceLogger::doNotify(string message, ostream& fileName)
+{
+    //implemented by Daniel
+}
+
+void PerformanceLogger::logIO(
+        const string & operation,
```
tst-ppenev: Q: How do we differentiate between read and write operations?

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -42,18 +162,25 @@ void PerformanceLogger::log(
...
-          << "thread: " << tid;
+          << "thread: " << tid*/;
+  message << operation << ','
+          << object    << ','
+          << startTime << ','
+          << endTime   << ','
+          << duration  << ','
+          << tid       << ',';
+
   if (size > 0) {
```
tst-ppenev: This is not going to produce a valid CSV.  Every line would need to have the same number of fields.

earth_enterprise/src/common/performancelogger.h
```diff
@@ -55,16 +79,45 @@ class BlockPerformanceLogger {
       size(size),
       startTime(getime::getMonotonicTime()),
       ended(false) {}
+    ~BlockPerformanceLogger() { this->end(); }
```
tst-ppenev: Style: I think `end();` reads fine.  There isn't anything requiring an explicit `this`.

earth_enterprise/src/common/performancelogger.h
```diff
@@ -74,6 +127,11 @@ class BlockPerformanceLogger {
     bool ended;
 };
 
+// Should be called at the beginning of gecombineterrain to create files
+#define OPEN_PERF_LOG_FILES() openFiles()
```
tst-ppenev: I think it would be safer, and simpler, if you created files in the logger constructor, and let them be closed in the destructor.

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
```diff
@@ -40,9 +40,12 @@ namespace geterrain {
 // and a token is returned.  The number of packets is added to the total.
 PacketFileReaderToken CountedPacketFileReaderPool::Add(
     const std::string &packet_file_name) {
+  BEGIN_IO_LOGGING(openPacketFile,"CombintTerrain_OpenPacket","Open/add packet file to reader pool");
```
tst-ppenev: Typo: `Combint`.

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
```diff
@@ -533,9 +542,12 @@ void TerrainCombiner::PacketWriteThread() {
     PacketInfo* packet = NULL;
     while(PopWriteQueue(&packet)) {
       if (packet != NULL) {
+        BEGIN_IO_LOGGING(pwThread,"CombineTerrain_PacketWrite","Packet write thread");
         WritePacket(packet);
         delete packet;
         packets_processed++;
+        END_IO_LOGGING(pwThread);
+        // how to handle exceptions?
```
tst-ppenev: What do you need to handle during exceptions?  Is there something special we need to log about an exception?
  Currently, a block logger will have its destructor called when control flows out of the block because of an exception, which would calculate elapsed timing, etc.

earth_enterprise/src/common/packetfile/packetfilereaderpool.h
```diff
@@ -31,6 +31,7 @@
 #include <khTypes.h>
 #include <khGuard.h>
 #include <packetfile/packetfilereader.h>
+#include "common/performancelogger.h"
```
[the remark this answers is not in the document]
tst-basilhuffman: No, at one point I thought I would need to check in here but never removed it.

earth_enterprise/src/common/performancelogger.cpp
tst-basilhuffman: Yes, these will be converted away from ofstream to fstream and moved over.

earth_enterprise/src/common/performancelogger.cpp
tst-basilhuffman: will do

earth_enterprise/src/common/performancelogger.cpp
tst-basilhuffman: Fixing before merge, just had this moved over from previous version.

earth_enterprise/src/common/performancelogger.h
tst-basilhuffman: Will fix, was just always a preference of mine.

earth_enterprise/src/common/performancelogger.h
tst-basilhuffman: Will do.

earth_enterprise/src/fusion/gecombineterrain/combineterrain.cpp
tst-basilhuffman: When originally going through, I had comments scattered as notes. Forgot to remove this one.

earth_enterprise/src/common/performancelogger.cpp
tst-basilhuffman: I am not terribly certain this is a concern, as the descriptor lets us know where in code the issue occurs.

earth_enterprise/src/common/performancelogger.cpp
tst-basilhuffman: I believe that is only the case in C++11

Issue #173, opened 2017-05-17
Add status target to geserver and gefusion services


tst-ccamp:
The service files added to /etc/init.d do not have a status target. This would be helpful in diagnosing whether the services are running. 

There is example code in the installers:

1. [check_server_processes_running method](https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/installer/common.sh)
1. [check_fusion_processes_running method](https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/installer/install_fusion.sh)

As a stretch goal on this ticket:
Either implement this fix as well, or create a follow-up ticket with the info.
Improve the method of determining whether the processes are running. Using the method referred to above starting a process using `vi gehttpd`, for example, could falsely indicate  that the gehttpd process is running. A better method would be to check the pid that is written out by the services. There are currently 4 known pids for the server and fusion processes:
* /var/opt/google/run/gesystemmanager.pid
* /var/opt/google/run/geresourceprovider.pid
* /var/opt/google/pgsql/data/postmaster.pid
* /opt/google/gehttpd/logs/httpd.pid

Note on postmaster.pid: the pid is at the beginning of the file, but there is more info in the file, so you'll need to parse it out

PR #718, opened 2018-02-20
Issue #173 services need status target

jsaponara:
Implements #173 

**Verification steps:**

**for geserver:**

$ sudo /etc/init.d/geserver stop
...
$ sudo /etc/init.d/geserver status
postmaster: not running
gehttpd: not running
$ echo $?
1
$ sudo /etc/init.d/geserver start
...
$ sudo /etc/init.d/geserver status
postmaster: OK
gehttpd: OK
$ echo $?
0

**for gefusion:**

$ sudo /etc/init.d/gefusion stop
Shutting down Resource Provider... Not running
Shutting down System Manager... Not running
$ sudo /etc/init.d/gefusion status
gesystemmanager  not running
geresourceprovider  not running
$ echo $?
1
$ sudo /etc/init.d/gefusion start
Starting System Manager... Done
Starting Resource Provider... Done
$ sudo /etc/init.d/gefusion status
gesystemmanager  OK
geresourceprovider  OK
$ echo $?
0

**Comments**

wsgi:ge_* processes apparently are killed with gehttpd, thus apparently dont have or need pid files of their own--except when they arent killed.  This happened to me twice (parent process was killed but children werent) and I couldnt reproduce it.  If this is a known issue, maybe a "geserver cleanup" subcommand would be handy until it's resolved.

There is substantial repetition of paths and other info, eg in c++ and in bash.  This is clearly outside the scope of issue 173, but I tried to use already existing status-checking code to avoid worsening the DRY violations.  I think that's important enough that even where I couldnt use it, I pointed to it in a comment (implying that maybe the existing code could be improved to make it useful in the service status context).

Similarly /opt/google/gehttpd/bin/gehttpd_init has a status subcommand, but it requires lynx, which wasnt installed by following the instructions.  Thus I used ps.

Process IDs can be recycled so i grep for user or command as well.

Error codes seem to be 0 or 1, but they could be more descriptive, denoting exactly which processes are running and which arent.  Eg add 1 if proc A is not running, add 10 if proc B is missing (or similar in binary).

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/sysman/gefusion
+++ b/earth_enterprise/src/fusion/autoingest/sysman/gefusion
@@ -58,16 +58,16 @@ function check_if_master {
 }
 
 function stop_daemon {
-    echo -n "Shutting down $2... "
-    pidfile="/var/opt/google/run/$1.pid"
-    if [ -e $pidfile ] ; then
-	if gestopdaemon $1 ; then
-	    echo "Done"
-	else
-	    echo "Failed"
-	fi
+  echo -n "Shutting down $2... "
+  pidfile="/var/opt/google/run/$1.pid"
+  if [ -e $pidfile ] ; then
+    if gestopdaemon $1 ; then
+      echo "Done"
+    else
+      echo "Failed"
+    fi
   else
-  echo "Not running"
+    echo "Not running"
   fi
 }
 
@@ -81,18 +81,12 @@ function check_daemon {
     ps -ef | grep $pid | grep -q $1
     foundPid=$?
     if [ $foundPid -eq 0 ] ; then
-      message="OK"
-      returncode=0
-    else
-      message="Not running"
-      returncode=1
+      echo "$1:  Running (PID $pid)."
+      return 0
     fi
-  else
-    message="Not running"
-    returncode=1
   fi
-  echo "$1  $message"
-  return $returncode
+  echo "$1:  Not running."
+  return 1
 }
 
 start() {
@@ -130,42 +124,34 @@ stop() {
 }
 
 status() {
+  return_code=0
   if check_if_master ; then
-	check_daemon gesystemmanager
-    gesystemmanager_returncode=$?
-  else
-    gesystemmanager_returncode=0
+    check_daemon gesystemmanager || return_code=1
   fi
-  check_daemon geresourceprovider
-  geresourceprovider_returncode=$?
-  if [ $gesystemmanager_returncode -eq 0 ] && [ $geresourceprovider_returncode -eq 0 ] ; then
-    return 0
-  else
-    return 1
-  fi
+  check_daemon geresourceprovider || return_code=1
+  return "$return_code"
 }
 
 case "$1" in
     restart)
-    stop
-    sleep 1
-    start
-    ;;
+      stop
+      sleep 1
+      start
+      ;;
     start)
-    start
-    ;;
+      start
+      ;;
     stop)
-    stop
-    ;;
+      stop
+      ;;
     status)
-	status
-	exit $?
-    ;;
+      status
+      exit $?
+      ;;
     *)
-    echo "usage: $0 {start|stop|restart|status}"
-    exit 1
-    ;;
-
+      echo "usage: $0 {start|stop|restart|status}"
+      exit 1
+      ;;
 esac
 
 exit 0
--- a/earth_enterprise/src/fusion/autoingest/sysman/gefusion
+++ b/earth_enterprise/src/fusion/autoingest/sysman/gefusion
@@ -66,9 +66,33 @@ function stop_daemon {
 	else
 	    echo "Failed"
 	fi
+  else
+  echo "Not running"
+  fi
+}
+
+function check_daemon {
+  # more DRY to use "gefdaemoncheck --checkrunning"
+  #   but it seems to assume --checksysman is true
+  #   thus wont work on non-master hosts.
+  pidfile="/var/opt/google/run/$1.pid"
+  if [ -s $pidfile ] ; then
+    pid=`cat $pidfile`
+    ps -ef | grep $pid | grep -q $1
+    foundPid=$?
+    if [ $foundPid -eq 0 ] ; then
+      message="OK"
+      returncode=0
     else
-	echo "Not running"
+      message="not running"
+      returncode=1
     fi
+  else
+    message="not running"
+    returncode=1
+  fi
+  echo "$1  $message"
+  return $returncode
 }
 
 start() {
@@ -105,6 +129,22 @@ stop() {
   fi
 }
 
+status() {
+  if check_if_master ; then
+	check_daemon gesystemmanager
+    gesystemmanager_returncode=$?
+  else
+    gesystemmanager_returncode=0
+  fi
+  check_daemon geresourceprovider
+  geresourceprovider_returncode=$?
+  if [ $gesystemmanager_returncode -eq 0 ] && [ $geresourceprovider_returncode -eq 0 ] ; then
+    return 0
+  else
+    return 1
+  fi
+}
+
 case "$1" in
     restart)
     stop
@@ -117,8 +157,12 @@ case "$1" in
     stop)
     stop
     ;;
+    status)
+	status
+	exit $?
+    ;;
     *)
-    echo "usage: $0 {start|stop|restart}"
+    echo "usage: $0 {start|stop|restart|status}"
     exit 1
     ;;
 
--- a/earth_enterprise/src/fusion/autoingest/sysman/gefusion
+++ b/earth_enterprise/src/fusion/autoingest/sysman/gefusion
@@ -60,7 +60,7 @@ function check_if_master {
 function stop_daemon {
   echo -n "Shutting down $2... "
   pidfile="/var/opt/google/run/$1.pid"
-  if [ -e $pidfile ] ; then
+  if [ -e "$pidfile" ] ; then
     if gestopdaemon $1 ; then
       echo "Done"
     else
@@ -77,8 +77,8 @@ function check_daemon {
   #   thus wont work on non-master hosts.
   daemon_path=$(which "$1")
   pidfile="/var/opt/google/run/$1.pid"
-  if [ -s $pidfile ] ; then
-    pid=`cat $pidfile`
+  if [ -s "$pidfile" ] ; then
+    pid=`cat "$pidfile"`
     if [ -d "/proc/$pid" ] && [ "$(readlink "/proc/$pid/exe")" = "$daemon_path" ]; then
       echo "$1:  Running (PID $pid)."
       return 0
--- a/earth_enterprise/src/fusion/autoingest/sysman/gefusion
+++ b/earth_enterprise/src/fusion/autoingest/sysman/gefusion
@@ -75,12 +75,11 @@ function check_daemon {
   # more DRY to use "gefdaemoncheck --checkrunning"
   #   but it seems to assume --checksysman is true
   #   thus wont work on non-master hosts.
+  daemon_path=$(which "$1")
   pidfile="/var/opt/google/run/$1.pid"
   if [ -s $pidfile ] ; then
     pid=`cat $pidfile`
-    ps -ef | grep $pid | grep -q $1
-    foundPid=$?
-    if [ $foundPid -eq 0 ] ; then
+    if [ -d "/proc/$pid" ] && [ "$(readlink "/proc/$pid/exe")" = "$daemon_path" ]; then
       echo "$1:  Running (PID $pid)."
       return 0
     fi
--- a/earth_enterprise/src/fusion/autoingest/sysman/gefusion
+++ b/earth_enterprise/src/fusion/autoingest/sysman/gefusion
@@ -84,11 +84,11 @@ function check_daemon {
       message="OK"
       returncode=0
     else
-      message="not running"
+      message="Not running"
       returncode=1
     fi
   else
-    message="not running"
+    message="Not running"
     returncode=1
   fi
   echo "$1  $message"
--- a/earth_enterprise/src/server/geserver
+++ b/earth_enterprise/src/server/geserver
@@ -56,7 +56,7 @@ dashes='--------------------------------------------------------------------'
 # this script
 RETVAL=0
 
-use_su=`su $GEPGUSER -c "echo -n 1" 2>/dev/null || echo -n 0`
+use_su=`su "$GEPGUSER" -c "echo -n 1" 2>/dev/null || echo -n 0`
 
 run_as_user() {
   if [ $use_su -eq 1 ] ; then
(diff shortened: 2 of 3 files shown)


jsaponara:
I signed the CLA.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -66,9 +66,33 @@ function stop_daemon {
...
+  if [ -s $pidfile ] ; then
+    pid=`cat $pidfile`
+    ps -ef | grep $pid | grep -q $1
+    foundPid=$?
+    if [ $foundPid -eq 0 ] ; then
+      message="OK"
+      returncode=0
     else
-	echo "Not running"
+      message="not running"
```
asmitaw: "Not running"

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -66,9 +66,33 @@ function stop_daemon {
...
+      returncode=1
     fi
+  else
+    message="not running"
```
asmitaw: "Not running"

earth_enterprise/src/server/geserver
```diff
@@ -84,15 +86,62 @@ stop() {
...
+  if [ $returncode -eq 0 ]; then
+    echo "gehttpd: OK"
+  else
+    echo "gehttpd: not running"
+  fi
+  return $returncode
+}
+
+status()
+{
```
asmitaw: move the openning curly braces in previous line

jsaponara:
Regarding the orphaned wsgi processes, I noticed that any non-zero [ie error] return value of killpg [kill process group] is ignored in src/fusion/autoingest/sysman/gestopdaemon.cpp, whereas ideally the error message would be printed or logged.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -66,9 +66,33 @@ function stop_daemon {
...
+}
+
+function check_daemon {
+  # more DRY to use "gefdaemoncheck --checkrunning"
+  #   but it seems to assume --checksysman is true
+  #   thus wont work on non-master hosts.
+  pidfile="/var/opt/google/run/$1.pid"
+  if [ -s $pidfile ] ; then
+    pid=`cat $pidfile`
+    ps -ef | grep $pid | grep -q $1
```
tst-ppenev: I think it's safer to test with:

```BASH
if ps --pid "$pid" >/dev/null; then
  returncode=0
else
  returncode=1
fi
```

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -105,6 +129,22 @@ stop() {
   fi
 }
 
+status() {
+  if check_if_master ; then
+	check_daemon gesystemmanager
```
tst-ppenev: Minor: Fix indentation.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
tst-ppenev: Minor: Also, I think it's shorter to set 1 variable (e.g., `returncode`) when you test a condition, rather than 2 (`returncode` and `message`).

E.g.:
```
if [ "$returncode" = 0 ]; then
  echo "$1:   Running (PID $pid)."
  return 0
else
  echo "$1:   Not running."
  return 1
fi
```

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -66,9 +66,33 @@ function stop_daemon {
 	else
 	    echo "Failed"
 	fi
+  else
+  echo "Not running"
```
tst-ppenev: Minor: Add indentation.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -105,6 +129,22 @@ stop() {
...
+status() {
+  if check_if_master ; then
+	check_daemon gesystemmanager
+    gesystemmanager_returncode=$?
+  else
+    gesystemmanager_returncode=0
+  fi
+  check_daemon geresourceprovider
+  geresourceprovider_returncode=$?
+  if [ $gesystemmanager_returncode -eq 0 ] && [ $geresourceprovider_returncode -eq 0 ] ; then
```
tst-ppenev: Minor: Technically, the next 5 lines are equivalent to the following, if you want to be shorter:

```
[ "$gesystemmanager_returncode" = 0 ] && [ "$geresourceprovider_returncode" = 0 ]
```

Also, I think something like the following may look shorter and use less variable names:

```BASH
return_code = 0
# . . .
check_daemon gesystemmanager || return_code=1
# . . .
check_daemon geresourceprovider || return_code=1

return "$return_code"
```

earth_enterprise/src/server/geserver
```diff
@@ -84,15 +86,61 @@ stop() {
   echo
   echo $dashes
   # only stop if already running...
-  if [ -f /var/opt/google/pgsql/data/postmaster.pid ] ; then
+  if [ -f $pidfile_postmaster ] ; then
```
tst-ppenev: Lint: Please, quote variable usage. 

E.g.:

```BASH
if [ -f "$pidfile_postmaster" ]; then
# etc.
```

earth_enterprise/src/server/geserver
```diff
@@ -84,15 +86,61 @@ stop() {
...
     run_as_user $GEPGUSER "$gebin_root/pg_ctl -D /var/opt/google/pgsql/data -l /var/opt/google/pgsql/logs/pg.log stop";
-    rm -f /var/opt/google/pgsql/data/postmaster.pid
+    rm -f $pidfile_postmaster
```
tst-ppenev: Lint: Quote variable usage.

earth_enterprise/src/server/geserver
```diff
@@ -84,15 +86,61 @@ stop() {
...
+  if [ -s $pidfile_postmaster ] ; then
+    pid_postmaster=`head -1 $pidfile_postmaster`
+    # redirect to /dev/null because pg_ctl --silent option doesnt work for status subcommand
+    run_as_user $GEPGUSER "$gebin_root/pg_ctl -D /var/opt/google/pgsql/data status" >& /dev/null
+    returncode=$?
+  else
+    returncode=1
+  fi
+  if [ $returncode -eq 0 ]; then
+    echo "postmaster: OK"
```
tst-ppenev: Minor: To make 'running' and 'not running' messages more uniform, and provide more details, I'd use something like:

```BASH
  echo "postmaster:   Running (PID $pid_postmaster)."
```

earth_enterprise/src/server/geserver
```diff
@@ -84,15 +86,61 @@ stop() {
...
+  else
+    echo "postmaster: Not running"
+  fi
+  return $returncode
+}
+
+status_gehttpd() {
```
tst-ppenev: Minor: Reuse `check_daemon` from `earth_enterprise/src/fusion/autoingest/sysman/gefusion`.

We should, actually, share the code instead of rewriting the same function multiple times, but, since that may expand the scope of this ticket too much, we can use a re-implementation for now.

earth_enterprise/src/server/geserver
```diff
@@ -84,15 +86,61 @@ stop() {
...
+  if [ -s $pidfile_gehttpd ] ; then
+    pid_gehttpd=`cat $pidfile_gehttpd`
+    # not using 'gehttpd_init status' because it requires lynx install
+    ps -ef | grep $pid_gehttpd | grep -q gehttpd
```
tst-ppenev: If code isn't re-used, I suggest using `ps --pid "$pid_gehttpd"`, as mentioned above.

tst-ppenev:
@jsaponara: Looks like adding system error reporting to `src/fusion/autoingest/sysman/gestopdaemon.cpp` is something we can write an issue ticket for.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -117,8 +157,12 @@ case "$1" in
     stop)
     stop
     ;;
+    status)
+	status
```
tst-ccamp: Minor: This is really the correct indention, but it would be better to match the other cases, so I'd suggest either updating the other cases to match this one or this one to match the others

earth_enterprise/src/fusion/autoingest/sysman/gefusion
jsaponara: I usually run a couple checks (grep'ing for both pid and executable name) because pids can be reused.  When I just need to check for a live pid, I usually use [ -f /proc/$pid ], which requires no special permissions (unlike accessing /proc/$pid/*).  But of course I defer.

That does indeed shorten the code, thanks.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
jsaponara: Yes I shouldnt mix tabs and spaces, thanks.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
jsaponara: Much nicer, thanks!

earth_enterprise/src/server/geserver
jsaponara: Totally better, yes.

earth_enterprise/src/server/geserver
jsaponara: Yes, just like src/installer/common.sh, the services would benefit from some common source.  I thought about it but came to the same conclusion.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
tst-ppenev: Checking for the executable name seems like a nice idea.  You can filter `ps` results by command with the `-C` option.  (E.g., `ps -C bash`.)  However, you can easily cheat, if you're only checking the file name of the executed command.

Probably the really robust approach would be to obtain an exclusive write lock on a PID file in each daemon process, which would be released by the kernel when the process dies, and checking for the lock to tell if a daemon is running.    However, checking the absolute path of a running process should also work pretty well.  (I'm not sure how you could cheat that.)

I don't know how to check the absolute path to the executable with `ps`.  However, on Linux systems, you can use `readlink "/proc/$pid/exe"`.

So, maybe, you can use something like:

```BASH
# $1: Full path to the daemon executable.
function check_daemon {
  # . . .
  daemon_name=$(basename "$1")
  if ps --pid "$pid" >/dev/null && [ "$(readlink "/proc/$pid/exe")" = "$1" ]; then
  # . . .
  echo "$daemon_name:   Running (PID $pid)."
  # . . .
}
```

earth_enterprise/src/server/geserver
tst-ppenev: Yes, that could be a separate ticket.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
jsaponara: The readability boost is totally worth it, will do.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
jsaponara: I hadnt thought of security so much as a stale .pid file in an environment of high process turnover, so maybe the old gehttpd pid is now some unrelated process, like a grep.  But either way, I like the added specificity.  As long as we're getting into /proc, we may as well abandon ps:
if [ -f "/proc/$pid" ] && [ "$(readlink "/proc/$pid/exe")" = "$1" ]; then ...
Ah, but we dont have the full path to the daemon executable in the service init script--gestartdaemon knows that.  But I guess it's ok to repeat ourselves since this issue isnt about DRY!  Not sure I have time for another round of build/install/test tonight tho.

tst-ppenev: Yeah, I think abandoning `ps` is fine, since this init script doesn't support anything besides Linux anyway.

Actually, I don't see `gestartdaemon.cpp` implementing any path resolution.  It just relies on the daemon being in the `PATH`.

I.e., if you want to be more robust, you could, e.g.:

```BASH
# $1: Daemon executable name, or path.
function check_daemon {
  # . . .
  daemon_path=$(which "$1")
  deamon_name=$(basename "$daemon_path")
  # . . .
  if [ -d "/proc/$pid" ] && [ "$(readlink "/proc/$pid/exe")" = "$daemon_path" ]; then
    # If you want to be fancy, show the process state:
    daemon_state=$(grep -Po -e '^State:\s*\K.*' "/proc/$pid/status")
    echo "$daemon_name: $daemon_state, PID $pid."
    return 0
  fi
  # . . .
}
```

I'm not a fan of re-implementing Debian's `start-stop-daemon`, or Red Hat's `daemon` commands, but that's another discussion.

jsaponara: Thanks, I stand corrected about gestartdaemon.  Isnt daemon_name just "$1"?  And wont daemon state generally be sleeping, at least for the geserver processes that spawn workers?  I suggest that idea be considered out of scope here, maybe as a verbose option on status.

jsaponara: Maybe your check of /proc/$pid/status could be the germ of an admin dashboard (if there isnt one already).

jsaponara: In order to test my edits, I run:
sudo /etc/init.d/gefusion stop
sudo /etc/init.d/geserver stop
( cd ~/work/$opengee/earthenterprise/earth_enterprise && scons -j2 internal=1 stage_install )
( cd $opengee/earthenterprise/earth_enterprise/src/installer/ && sudo ./install_server.sh && echo $divider && sudo ./install_fusion.sh )
Maybe there's a shorter way, but if not, the install scripts might benefit from the equivalent of a "-y" option (I think I type 'n', 'c', 'c', 'c') so that this sequence can occur unattended.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
```diff
@@ -58,17 +58,34 @@ function check_if_master {
...
+  fi
+}
+
+function check_daemon {
+  # more DRY to use "gefdaemoncheck --checkrunning"
+  #   but it seems to assume --checksysman is true
+  #   thus wont work on non-master hosts.
+  daemon_path=$(which "$1")
+  pidfile="/var/opt/google/run/$1.pid"
+  if [ -s $pidfile ] ; then
```
tst-ppenev: Lint: Please, comment variable expansion on `if` line and below when using `cat`.

earth_enterprise/src/server/geserver
```diff
@@ -84,31 +86,69 @@ stop() {
...
   echo $dashes
   echo 'Done Stopping Google Earth Enterprise Server'
   echo
 }
 
+status_postmaster() {
+  if [ -s "$pidfile_postmaster" ] ; then
+    pid=`head -1 "$pidfile_postmaster"`
+    # redirect to /dev/null because pg_ctl --silent option doesnt work for status subcommand
+    run_as_user $GEPGUSER "$gebin_root/pg_ctl -D /var/opt/google/pgsql/data status" >& /dev/null
```
tst-ppenev: Lint: Unquoted variable expansion.

earth_enterprise/src/fusion/autoingest/sysman/gefusion
jsaponara: Fixed, thanks.

earth_enterprise/src/server/geserver
jsaponara: Fixed, thanks.

Issue #685, opened 2018-01-31
Trying to disassemble a portable globe or map GLC produces an empty folder


tst-nfarah:
On Red hat 6, using ''Settings' → 'Glc Disassembly' in 'GEE Server' admin web page to disassemble an existing portable globe reports 'SUCCESS' but the output folder is empty.

tst-ccamp:
Looks like one of the directories in the path (/opt/google/gehttpd/htdocs/cutter/globes) that needs to be accessed is only accessible by geapacheuser. If any of the disassebly processes is being run by the gefusion user, then the directory will not be accessible and this may cause the process to fail. However, there is also the issue that it is reporting success when obviously there is an error somewhere. 

Note: this may only occur when installing via RPM.

PR #719, opened 2018-02-21
Fixed bug #685: globe disassembly produces empty directory

tst-rwildes:
Fixed bug #685: communicate() was called multiple times for a single popen call.  Should only call this once to get the stdout and stderr in a single return.  The first call closes the pipes so the second call always fails.  Also added and cleaned up logging so finding errors is a bit easier next time.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
@@ -182,8 +182,7 @@ class GlcAssembler(object):
     """Get logger built from info.txt."""
 
     if not self.info_file:
-      # if called before setting self.info_file
-      # return None
+      # self.info_file is expected to be set before calling this method
       return None
 
     if not os.path.exists(os.path.dirname(self.info_file)):
@@ -214,7 +213,7 @@ class GlcAssembler(object):
               % (COMMAND_DIR, glc_path))
     glx_info = utils.RunCmd(os_cmd)
 
-    # the odd way errors are returned....
+    # Checking for errors and logging them if there where any
     if not glx_info[0] and len(glx_info) == 2:
       utils.PrintAndLog("ERROR: '%s' running command: '%s'" % (glx_info[1], os_cmd), logger)
       return []
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
@@ -214,7 +214,7 @@ class GlcAssembler(object):
               "--list_files"
               % (COMMAND_DIR, glc_path))
     glx_info = utils.RunCmd(os_cmd)
-    if len(glx_info) == 0:
+    if not isinstance(glx_info, list) or len(glx_info) == 0:
       utils.PrintAndLog("Unexpected empty results running command: '%s'" % os_cmd, logger)
       return []
 
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
@@ -213,7 +213,7 @@ class GlcAssembler(object):
               "--list_files"
               % (COMMAND_DIR, glc_path))
     glx_info = utils.RunCmd(os_cmd)
-    if not isinstance(glx_info, list) or len(glx_info) == 0:
+    if len(glx_info) == 0:
       utils.PrintAndLog("Unexpected empty results running command: '%s'" % os_cmd, logger)
       return []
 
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
```diff
@@ -172,12 +178,18 @@ def PrepareAssembler(self, base_path):
...
     """Get logger built from info.txt."""
-    info_file = INFO_FILE % base_path
-    if not os.path.exists(os.path.dirname(info_file)):
-      os.makedirs(os.path.dirname(info_file))
-    logger = utils.Log(info_file)
+
+    if not self.info_file:
+      # if called before setting self.info_file
+      # return a logger that does nothing
+      logger = utils.Log("/nulllogger", False)
```
tst-nfarah: why would we need this ? a logger that does nothing

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
```diff
@@ -196,11 +208,21 @@ def ExtractFileFromGlx(self, glc_path, file_in_glc, output_file):
 
   def ExtractFilelistFromGlx(self, glc_path, regex):
     """Extract given file from glc file and save it to the given path."""
+    logger = self.GetLogger()
     os_cmd = ("%s/geglxinfo "
               "--glx=\"%s\" "
               "--list_files"
               % (COMMAND_DIR, glc_path))
     glx_info = utils.RunCmd(os_cmd)
+    if len(glx_info) == 0:
```
tst-nfarah: could 'glx_info' be null triggering an exception when we try to get its length ?

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
tst-rwildes: Shouldn't happen but if it did I think a logger that does nothing is better than a run-time exception that will likely happen if I returned None.  Or if I don't return None then where do I put the file?  Again should not happen but if it does I feel a null logger is the least bad of all bad options available.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
tst-rwildes: glx_info is expected to be a list.  I'm not sure what len() does with a None object.  I *think* both an empty list and a None object evaluate to 'False' so maybe `if not glx_info:` is safer?  Or I could do a compound if statement if you feel that is more understandable.  I'll try a few things and see what works.  But if you feel a compound if statement is still the better thing to do I'll do just do: `if glx_info is None or len(glx_info) ==0:`.  Question: does Python have "short circuit if statement"?

tst-nfarah: C++ does stop if the first test fails but not sure how python does. May be try it out or google it.

tst-rwildes: According to this it does: http://bioruffo.net/lpotf/short-circuit-evaluations-in-python/188/

tst-rwildes: I verified using the interpreter that yes it does support short circuit statements.  And a None object evaluates to false in an if statement and so does an empty list.   But unlike an empty string it looks like an empty list evaluates to true in an if statement.  So I can do a compound if statement to guard against None return.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
tst-nfarah: this would not happen because we initialize the info file to an empty string ? (self.info_file = "") ? if yes, i would remove this code as it might be confusing.

tst-rwildes: I added the initialization just to be sure the attribute existed.  But it still needs to be valued with something other than empty string.

tst-ppenev: I don't see the point of the code, either.  The purpose of exceptions is to report errors.  If there's an error you should throw an exception.  All the better if the exception is already thrown for you, and you don't need to write any code for it.

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
```diff
@@ -196,11 +208,21 @@ def ExtractFileFromGlx(self, glc_path, file_in_glc, output_file):
 
   def ExtractFilelistFromGlx(self, glc_path, regex):
     """Extract given file from glc file and save it to the given path."""
+    logger = self.GetLogger()
     os_cmd = ("%s/geglxinfo "
               "--glx=\"%s\" "
               "--list_files"
               % (COMMAND_DIR, glc_path))
     glx_info = utils.RunCmd(os_cmd)
+    if not isinstance(glx_info, list) or len(glx_info) == 0:
```
tst-ppenev: I don't think there's any point in guarding against the result of `utils.RunCmd()` not being a list.  We wrote that method to return a list, if it doesn't we have a bug in our code, and should get an exception.

Giving the user a message that says "Unexpected empty results running command" is probably just going to confuse people to think they need to try executing the command in a different way, rather than telling them there's a bug in the Python code.

PR #723, opened 2018-02-23
Log io and csv

tst-basilhuffman:
Renamed branch for clarity.  IO logging and CSV file output both addressed

### DIFF
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -27,25 +27,24 @@
 using namespace std;
 using namespace getime;
 
-string ioFileName;
-string threadFileName;
-string timeFileName;
+// make sure static members get initialized
+string PerformanceLogger::ioFileName = "";
+string PerformanceLogger::threadFileName = "";
+string PerformanceLogger::timeFileName = "";
 
-void generateFileNames() {
+void PerformanceLogger::generateFileNames() {
+
     // needs mutex lock
     time_t t = time(0);
     tm date = *localtime(&t);
     char buf[256];
-    std::fill(buf,buf+256,0);
     if (!ioFileName.size()) {
         strftime(buf, sizeof(buf), "io_stats.%m-%d-%Y-%H:%M:%S.csv", &date);
         ioFileName = buf;
-        std::fill(buf,buf+256,0);
     }
     if (!timeFileName.size()) {
         strftime(buf, sizeof(buf), "time_stats.%m-%d-%Y-%H:%M:%S.csv", &date);
         timeFileName = buf;
-        std::fill(buf,buf+256,0);
     }
     if (!threadFileName.size()) {
         strftime(buf, sizeof(buf), "thread_stats.%m-%d-%Y-%H:%M:%S.csv", &date);
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -18,7 +18,7 @@
 #include <unistd.h>
 #include <sys/syscall.h>
 #include <time.h>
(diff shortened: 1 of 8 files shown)


earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,131 @@
...
+string ioFileName;
+string threadFileName;
+string timeFileName;
+
+void generateFileNames() {
+    // needs mutex lock
+    time_t t = time(0);
+    tm date = *localtime(&t);
+    char buf[256];
+    std::fill(buf,buf+256,0);
```
tst-ppenev: Q: Is this necessary?  Doesn't `strftime` already add a terminating `'\0'` character?

tst-ppenev: You can use `sizeof(buf)` instead of `256`.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,131 @@
...
 using namespace std;
 using namespace getime;
 
-// Initialize static members of Profiler class
-PerformanceLogger * const PerformanceLogger::_instance = new PerformanceLogger();
+string ioFileName;
+string threadFileName;
+string timeFileName;
+
+void generateFileNames() {
```
tst-ppenev: This looks like it may be better done in a logger constructor before any of the file names get used.

earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -18,19 +18,131 @@
...
 #include "common/notify.h"
 #include "common/timeutils.h"
+#include <cassert>
 
 using namespace std;
 using namespace getime;
 
-// Initialize static members of Profiler class
-PerformanceLogger * const PerformanceLogger::_instance = new PerformanceLogger();
+string ioFileName;
```
tst-ppenev: These look like they can be member variables, since they pertain to what the logger is doing.

earth_enterprise/src/common/performancelogger.cpp
tst-basilhuffman: Would we want to have the different files to have the same timestamp or does it matter?

earth_enterprise/src/common/performancelogger.cpp
tst-basilhuffman: I like to err on the side of caution. sizeof would be ok, but we already know the size here so I figured just use it. I suppose it comes down to standard convention.

earth_enterprise/src/common/performancelogger.cpp
tst-ppenev: Yeah, I think that's a good idea.  I don't think that would change if this function was a member function called from a constructor.  It would just insure these variables are initialized before logger methods are called.

earth_enterprise/src/common/performancelogger.cpp
tst-ppenev: This is minor, but, the documentation of `strftime` says:

> `size_t strftime( char *restrict str, size_t count, const char *restrict format, const struct tm *restrict time );`
> Converts the date and time information from a given calendar time time to a **null-terminated** multibyte character **string str** according to format string format. Up to count bytes are written.

[empasis added]

It's intended to be used with an uninitialized buffer.  I think there may be compiler flags to initialize buffers (which often already happens).

tst-basilhuffman: Fair enough. Changing.

tst-basilhuffman:
@tst-lsavoie I'm ok with combining into #724

tst-lsavoie:
Merged into PR #724.

PR #727, opened 2018-03-01
Performance improvements for profiling

tst-lsavoie:
This PR reduces the overhead of profiling by:
- Opening and closing the file once
- Batching updates and writing 64 KB to the file at a time

The resulting output file has fewer lines that an output file that is run with the old code; however, this is because there are more operations on condition variables (i.e., there are more calls to khCondVar::broadcast_all, khCondVar::signal_one, and khCondVar:wait), so the actual operations being performed are the same.

Sample command line for testing (you will have to change this based where your data is):
```
sudo /opt/google/bin/gecombineterrain --indexversion 7 --numcpus 4 --output /gevol/assets/Databases/CA_POIs.kdatabase/terrain.kta/ver002/terrain.pack/ /gevol/assets/Projects/Terrain/worldtopo.ktproject/geindex.kta/ver004/worldtopo-v4.geindex
```

### DIFF
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -110,9 +110,19 @@ void PerformanceLogger::do_notify(const string & message) {
   {  // atomic inner block
     plLockGuard lock( write_mutex );
     {
-      timeFile << pid << ',' << tid << ',' << message << endl;
+      if (buffer.size() + message.size() > BUF_SIZE) {
+        flushBuffer();
+      }
+      stringstream newLine;
+      newLine << pid << ',' << tid << ',' << message << endl;
+      buffer.append(newLine.str());
     }
   }
+}
+
+void PerformanceLogger::flushBuffer() {
+  timeFile << buffer;
+  buffer.clear();
 }
 
 } // namespace performance_logger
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -96,26 +97,34 @@ void PerformanceLogger::logTiming(
...
   pthread_t tid = pthread_self();
   pid_t pid = getpid();
 
   {  // atomic inner block
     plLockGuard lock( write_mutex );
-    { // Make sure we flush and close the output file before unlocking the mutex:
-      ofstream output_stream(fileName.c_str(), ios::app);
-      output_stream << pid << ',' << tid << ',' << message << endl;
+    {
+      if (buffer.size() + message.size() > BUF_SIZE) {
```
tst-ppenev: @tst-lsavoie: Have you compared performance when using an external `std::string` as a buffer vs. setting `std::ofstream`'s `streambuf`?  (E.g. `timeFile.rdbuf()->pubsetbuf(buffer.c_str(), buffer.capacity());`)

tst-lsavoie: Yes, I did try that. It didn't seem to have an effect. Apparently `pubsetbuf` is not guaranteed to do anything.

PR #734, opened 2018-03-12
#225: Added check to reject directory names with spaces or special characters.

### LINKED ISSUE #225: Fusion lets you create folder with space in the name
In Fusion->Asset manager , we can create a subfolder in the asset root.
It even allows you to create with folder names like "Imagery{space}resources"

when we try to save the resource to this folder, then obviously it fails because of invalid path.

bejota:
- This fix prevents the Fusion->Asset Manager->ASSET_ROOT->New Subfolder dialog from creating directories with spaces or special characters.
- To test, use New Subfolder dialog in Fusion->Asset Manager in ASSET_ROOT window.
- Tested on CentOS 7
- Fixes #225
- Note: May need to check all UI dialogs for consistency.


tst-ccamp:
We have two approvals, so merging. Thanks @bejota !

PR #736, opened 2018-03-12
#735: Add Linux thread IDs to the gecombineterrain performance log.

### LINKED ISSUE #735: Report the Linux Thread ID in `gecombineterrain` Performance Profiling Output
As a profiler, I would like the `gecombineterrain` performance log to include Linux thread IDs for all logged operations.

Thread IDs can be reported by other tools, such as `pidstat`, that take execution statistics from the Linux kernel.  Reporting the thread ID in the performance log allows relating execution statistics from such tools to the information in the performance log.

## AC:

1. The `gecombineterrain` performance log should contain Linux thread IDs for each logged operation.

tst-ppenev:
* Renamed the `tid` column to `pthread_id` in the `gecombineterrain` performance log.
* Added a `tid` column containing the Linux thread ID (as returned by the `gettid` `syscall`).

Fixes #735.

### DIFF
--- a/earth_enterprise/src/common/performancelogger.cpp
+++ b/earth_enterprise/src/common/performancelogger.cpp
@@ -30,6 +30,10 @@ using namespace getime;
 
 #ifdef LOG_PERFORMANCE
 
+#include <sys/syscall.h>
+#include <sys/types.h>
+
+
 namespace performance_logger {
 
 // make sure static members get initialized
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/common/performancelogger.cpp
```diff
@@ -30,6 +30,10 @@ using namespace getime;
 
 #ifdef LOG_PERFORMANCE
 
+#include <sys/syscall.h>
```
tst-lsavoie: Minor - this is already included above.

tst-lsavoie:
I did a quick review and test and everything looks fine. I'm going to hold off on approving for now since we want to fix the duplication of the CombineTerrain_Pop tag.

Issue #498, opened 2017-09-13
Add remove_volume functionality to geconfigureassetroot


tst-mswartz:
It would be convenient to have an option in geconfigureassetroot to remove volumes from the list, rather than editing volumes.xml manually.

tst-lsavoie:
Closed via PR #737.

PR #737, opened 2018-03-12
#498 Add 'removevolume' option to geconfigureassetroot

akrokh:
This is my first attempt at fixing bug #498 , as a coding assignment. A couple of notes:
- Documentation also needs to be updated to account for "removevolume", but when looking at  earthenterprise/docs/geedocs/5.2.0/answer/3481558.html, I found this:
============
--fixmasterhost [--assetroot path]

Optional. Change the assetroot host entry to match the current host name. (This command corrects cases where a host name is changed after installing and configuring Google Earth Enterprise Fusion.)

--addvolume volume_name:path]

Optional. Change the assetroot host entry to match the current host name. (This command corrects cases where a host name is changed after installing and configuring Google Earth Enterprise Fusion.)
============

The above two entries are identical, does it mean that "addvolume" also needs to be properly documented? Or is that document not being actively maintained?

- There is a typo that I found and fixed ("unrecognized parameters")

- With C++11, some code can be simplified--for example, you wouldn't need a "makeset" with multiple versions for different number of template arguments. However, this would require updating the build system. There are other improvements that could be made. However, in general, I tried to maintain consistency with the current style of geconfigureassetroot.cpp and only fix the issue referenced in the bug.

### DIFF
--- a/docs/geedocs/5.2.2/answer/3481558.html
+++ b/docs/geedocs/5.2.2/answer/3481558.html
@@ -90,7 +90,7 @@
   <p><i>Default</i>. Uses normal imagery for the project.</p>
   <h2><a name="63542"></a>geconfigureassetroot</h2>
   <pre>
-<code><b>geconfigureassetroot</b> {<b>--new</b> <b>--assetroot </b><i>path </i> [<b>--srcvol </b><i>path</i>] | <b>--repair</b> | <b>--editvolumes</b> | <b>--listvolumes</b> | <b>--addvolume</b> | <b>--removevolume</b> | <b>--fixmasterhost</b> | <strong>--noprompt</strong>}  [--nochown]</code></pre>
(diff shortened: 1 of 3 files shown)


akrokh:
I signed it!

tst-jhaight:
Code change looks ok.  I'd agree with the suggestions to update the documentation.  At least adding the documentation for removevolume might be within the scope of this fix.

I tested this change out on CentOS 7 to ensure the proper behavior for the following:
- removing an existing volume
- removing a non-existent volume  produces an appropriate message

akrokh:
@tst-jhaight , I updated the documentation for version 5.2.2, I believe that's the one we are targeting.

tst-ccamp:
@akrokh These changes will go into 5.2.3. 5.2.2 testing has already begun so we aren't adding any new features to that branch. Unfortunately we don't have the docs ready for 5.2.3 quite yet, I may be able to get to that tomorrow.

akrokh:
@tst-ccamp , sure--let me know which file and I will update my pull request.

tst-lsavoie:
I just wrote #770 to handle the documentation changes later.

PR #741, opened 2018-03-15
Corrected build script for RPMs

tst-dmendyke:
Altered build file 'build.gradle' in order to allow building either a DGB or a REL version of the RPMs.

### DIFF
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -31,7 +31,7 @@ EnsureSConsVersion(0, 96, 92)
 package_command_arguments = "-Pdebug"
 
 for parameter in [ 'release', 'optimize', 'profile' ] :
-  if int( ARGUMENTS.get( parameter, '0' ) ) > 0 :
+  if int( ARGUMENTS.get( parameter, '0' ) ) :
     package_command_arguments = "-P" + parameter
     break;
 
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -24,13 +24,18 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
+
+
+
 # Check for possible objects to pass on to gradle
 def gradle_options() :
   options = ""  # Default to no option
   if ARGUMENTS.get( 'release', '0' ) == '1' :
-    options = "-PRELEASE"
+    options += " -PRELEASE"
+  else :
+    options += " -PDEBUG"
   if ARGUMENTS.get( 'optimize', '0' ) == '1' :
-    options = "-POPTIMIZE"
+    options += " -POPTIMIZE"
   return options
 
 
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+# This file is 'earth_enterprise/SConstruct'
+
 import sys
 import os
 import commands
@@ -24,18 +26,24 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
+# Check for the presents of the passed argument on
+# the command line.  If exists return a formated
+# string that can be passed to gradle else return
+# an empty string
+def check_for_argument( argument ) :
+  result = ""
+  if int( ARGUMENTS.get( argument, '0' ) ) > 0 :
+    result = "-P" + argument
+  return result
 
-
-
-# Check for possible objects to pass on to gradle
+# Check for command line arguments passed that must be
+# passed on to gradle
 def gradle_options() :
-  options = ""  # Default to no option
-  if ARGUMENTS.get( 'release', '0' ) == '1' :
-    options += " -PRELEASE"
-  else :
-    options += " -PDEBUG"
-  if ARGUMENTS.get( 'optimize', '0' ) == '1' :
-    options += " -POPTIMIZE"
+  options = ""
+  options = check_for_argument( 'debug' )
+  options = check_for_argument( 'release' )
+  options = check_for_argument( 'optimize' )
+  options = check_for_argument( 'profile' )
   return options
 
 
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -31,7 +31,7 @@ EnsureSConsVersion(0, 96, 92)
 package_command_arguments = "-Pdebug"
 
 for parameter in [ 'release', 'optimize', 'profile' ] :
-  if int( ARGUMENTS.get( parameter, '0' ) ) :
+  if int( ARGUMENTS.get( parameter, '0' ) ) > 0 :
     package_command_arguments = "-P" + parameter
     break;
 
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -40,9 +40,9 @@ def check_for_argument( argument ) :
 # pass them on to Gradle.
 def gradle_options() :
   options = "-Pdebug"
-  options = check_for_argument( 'release' )
-  options = check_for_argument( 'optimize' )
   options = check_for_argument( 'profile' )
+  options = check_for_argument( 'optimize' )
+  options = check_for_argument( 'release' )
   return options
 
 
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -171,7 +171,7 @@ stage_tutorial = env.PhonyTargets(**stage_tutorial_cmds)
 env.Depends(stage_tutorial, stage_data)
 
 package_staged_cmds = {'package_install': [
-    "cd rpms; ./gradlew osPackage " + CommandLineParameter().option
+    "cd rpms; ./gradlew osPackage " + package_command_arguments
 ]}
 
 package_install = env.PhonyTargets(**package_staged_cmds)
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -39,7 +39,7 @@ def check_for_argument( argument ) :
 # Check for command line arguments, if they are found
 # pass them on to Gradle.
 def gradle_options() :
-  options = ""
+  options = "-Pdebug"
   options = check_for_argument( 'release' )
   options = check_for_argument( 'optimize' )
   options = check_for_argument( 'profile' )
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -26,24 +26,17 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
-# Check for the presence of the passed argument on
-# the command line.  If exists return a formated
-# string that can be passed to Gradle else return
-# an empty string
-def check_for_argument( argument ) :
-  result = ""
-  if int( ARGUMENTS.get( argument, '0' ) ) > 0 :
-    result = "-P" + argument
-  return result
+# An object used to check for a specific command line parameter.
+# Set the instance variable 'option' based on passed parameter.
+class CommandLineParameter( object ) :
 
-# Check for command line arguments, if they are found
-# pass them on to Gradle.
-def gradle_options() :
-  options = "-Pdebug"
-  options = check_for_argument( 'profile' )
-  options = check_for_argument( 'optimize' )
-  options = check_for_argument( 'release' )
-  return options
+  option = "-Pdebug"
+
+  def __init__( self ) :
+    for parameter in [ 'profile', 'optimize', 'release' ] :
+      if int( ARGUMENTS.get( parameter, '0' ) ) > 0 :
+        self.option = "-P" + parameter
+
 
 
 # process commandline arguments
@@ -180,7 +173,7 @@ stage_tutorial = env.PhonyTargets(**stage_tutorial_cmds)
 env.Depends(stage_tutorial, stage_data)
 
 package_staged_cmds = {'package_install': [
-    "cd rpms; ./gradlew osPackage " + gradle_options()
+    "cd rpms; ./gradlew osPackage " + CommandLineParameter().option
 ]}
 
 package_install = env.PhonyTargets(**package_staged_cmds)
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -24,6 +24,16 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
+# Check for possible objects to pass on to gradle
+def gradle_options() :
+  options = ""  # Default to no option
+  if ARGUMENTS.get( 'release', '0' ) == '1' :
+    options = "-PRELEASE"
+  if ARGUMENTS.get( 'optimize', '0' ) == '1' :
+    options = "-POPTIMIZE"
+  return options
+
+
 # process commandline arguments
 installdir = ARGUMENTS.get('installdir', '/tmp/fusion_os_install')
 with open('rpms/build.properties', 'w') as f:
@@ -158,7 +168,7 @@ stage_tutorial = env.PhonyTargets(**stage_tutorial_cmds)
 env.Depends(stage_tutorial, stage_data)
 
 package_staged_cmds = {'package_install': [
-    'cd rpms; ./gradlew osPackage'
+    "cd rpms; ./gradlew osPackage " + gradle_options()
 ]}
 
 package_install = env.PhonyTargets(**package_staged_cmds)
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -26,16 +26,14 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
-# An object used to check for a specific command line parameter.
-# Set the instance variable 'option' based on passed parameter.
-class CommandLineParameter( object ) :
+# Set 'package_command_arguments' based on a parameter
+# that is to be passed on to Gradle
+package_command_arguments = "-Pdebug"
 
-  option = "-Pdebug"
-
-  def __init__( self ) :
-    for parameter in [ 'profile', 'optimize', 'release' ] :
-      if int( ARGUMENTS.get( parameter, '0' ) ) > 0 :
-        self.option = "-P" + parameter
+for parameter in [ 'release', 'optimize', 'profile' ] :
+  if int( ARGUMENTS.get( parameter, '0' ) ) :
+    package_command_arguments = "-P" + parameter
+    break;
 
 
 
--- a/earth_enterprise/SConstruct
+++ b/earth_enterprise/SConstruct
@@ -26,9 +26,9 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
-# Check for the presents of the passed argument on
+# Check for the presence of the passed argument on
 # the command line.  If exists return a formated
-# string that can be passed to gradle else return
+# string that can be passed to Gradle else return
 # an empty string
 def check_for_argument( argument ) :
   result = ""
@@ -36,11 +36,10 @@ def check_for_argument( argument ) :
     result = "-P" + argument
   return result
 
-# Check for command line arguments passed that must be
-# passed on to gradle
+# Check for command line arguments, if they are found
+# pass them on to Gradle.
 def gradle_options() :
   options = ""
-  options = check_for_argument( 'debug' )
   options = check_for_argument( 'release' )
   options = check_for_argument( 'optimize' )
   options = check_for_argument( 'profile' )
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -230,7 +230,7 @@ task stageOpenGeeInstall(type: Exec, dependsOn: 'compileOpenGee') {
     // "installdir=${build_install_dir.getAbsolutePath()}", so we
     // overwrite whatever is in the default stage install directory:
     commandLine(
-        'scons', '-j8', scons_options, 'stage_install',
+        'scons', '-j8', scons_option, 'stage_install',
         "installdir=${stagedInstallDir}")
 }
 
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -83,14 +83,15 @@ if (new File('build.properties').exists()) {
 
 def openGeeVersionFile = new File( gee_long_version_txt )
 
-if (!openGeeVersionFile.exists()) {
+// Call scons to create the file 'gee_long_version.txt'
+// The actual parameter used by scons can remain 'release=1'
+if ( !openGeeVersionFile.exists() ) {
     GeeCommandLine.expand(
-            ["/usr/bin/scons", "release=1", gee_long_version_txt ],
-            //      ["/usr/bin/scons", "release=1", openGeeVersionFile ],
-            "Runing `scons` to produce <gee_long_version_txt> failed!",
-            null, // Use default environment variables.
-            new File( working_dir )
-        )
+      ["/usr/bin/scons", "release=1", openGeeVersionFile ],
+      "Runing `scons` to produce <gee_long_version_txt> failed!",
+      null, // Use default environment variables.
+      new File( working_dir )
+    )
 }
 
 def openGeeVersion =
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -54,13 +54,13 @@ def working_dir = "../src/"
 // Check for either "-PRELEASE" or "-POPTIMIZE" command line parameters.
 // Return the correct directory based on which paramater is used.
 def native_x86 = {
-  if ( project.hasProperty( 'RELEASE' ) ) { return "NATIVE-REL-x86_64" }
   if ( project.hasProperty( 'OPTIMIZE' ) ) { return "NATIVE-OPT-x86_64" }
-  return "NATIVE-DBG-X86_64"
+  if ( project.hasProperty( 'DEBUG' ) ) { return "NATIVE-DBG-x86_64" }
+  return "NATIVE-REL-x86_64"
 }  // end native_x86
 
 // Build the actual file to be used
-def gee_long_version_txt = working_dir + native_x86 + "gee_long_version.txt"
+def gee_long_version_txt = working_dir + native_x86() + "/gee_long_version.txt"
 
 
 def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
@@ -85,9 +85,9 @@ def openGeeVersionFile = new File( gee_long_version_txt )
 
 if (!openGeeVersionFile.exists()) {
     GeeCommandLine.expand(
-      //            ["/usr/bin/scons", "release=1", gee_long_version_txt ],
-            ["/usr/bin/scons", "release=1", openGeeVersionFile ],
-            "Runing `scons` to produce <gee_long_version.txt> failed!",
+            ["/usr/bin/scons", "release=1", gee_long_version_txt ],
+            //      ["/usr/bin/scons", "release=1", openGeeVersionFile ],
+            "Runing `scons` to produce <gee_long_version_txt> failed!",
             null, // Use default environment variables.
             new File( working_dir )
         )
@@ -438,7 +438,6 @@ task openGeeServerRpm (type: GeeRpm, dependsOn: openGeeSharedFiles) {
     installUtils = file('build/shared/install-utils.sh')
 
     // Add geresetpgdb script:
-    //    installUtils = file( new File( working_dir + "server/gepgcommon" ) )
     installUtils = file( new File( stagedInstallDir_server_opt, 'google/bin/gepgcommon' ) )
 
     preInstall = file('opengee-server/pre-install.sh')
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -152,7 +152,7 @@ def packageInstallLibDir = new File(packageInstallRootDir, 'opt/google/lib')
 
 def postGisInstallDir = new File(
         project.projectDir,
-        "${src_dir}${native_dir()}third_party/postgis/install"
+        "${src_dir}${native_dir()}/third_party/postgis/install"
     ).getAbsolutePath()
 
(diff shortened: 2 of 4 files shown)


earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +43,17 @@ apply plugin: "nebula.ospackage"
...
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def working_dir = "../src/"
+
+// Use the command line option "-PDEBUGRPMS" in order to build
+// for NATIVE-DBG-x86_64.  Default is NATIVE-REL-x86_64
+def native_x86 = ( project.hasProperty( 'DEBUGRPMS' ) ? "NATIVE-DBG-x86_64/" : "NATIVE-REL-x86_64/" )
+
+// Build the actual file to be used
+def long_txt = working_dir + native_x86 + "gee_long_version.txt"
```
tst-lsavoie: Minor: might be nice to have a more descriptive variable name than `long_txt` - even `long_version_txt` would work.

earth_enterprise/rpms/build.gradle
```diff
@@ -417,7 +429,8 @@ task openGeeServerRpm (type: GeeRpm, dependsOn: openGeeSharedFiles) {
     installUtils = file('build/shared/install-utils.sh')
 
     // Add geresetpgdb script:
-    installUtils = file(new File(stagedInstallDir_server_opt, 'google/bin/gepgcommon'))
+    installUtils = file( new File( working_dir + "server/gepgcommon" ) )
```
tst-lsavoie: I don't think this line needs to change. `stagedInstallDir_server_opt` is a directory under `/tmp/fusion_os_install`, which should be the same regardless of the version built.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +43,17 @@ apply plugin: "nebula.ospackage"
 
 import org.opengee.shell.GeeCommandLine
 
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def working_dir = "../src/"
+
+// Use the command line option "-PDEBUGRPMS" in order to build
+// for NATIVE-DBG-x86_64.  Default is NATIVE-REL-x86_64
+def native_x86 = ( project.hasProperty( 'DEBUGRPMS' ) ? "NATIVE-DBG-x86_64/" : "NATIVE-REL-x86_64/" )
```
tst-lsavoie: Minor: it would be nice to also support the "optimize" option here. In your scons command:
- if you specify `release=1` you get `NATIVE-REL-x86_64`
- if you specify `optimize=1` you get `NATIVE-OPT-x86_64`
- if you specify nothing you get `NATIVE-DBG-x86_64`
It would be nice if the script supported all 3 versions.

tst-lsavoie:
It might be nice to pass this parameter from scons automatically when it is needed. When I build RPMs, I run `scons -j8 release=1 package_install`, which calls this gradle code. I think it would be handy if scons passed the correct parameter to gradle when it is called with the `package_install` target.

earth_enterprise/rpms/build.gradle
tst-dmendyke: I can look into modifying SCons to achieve this.

earth_enterprise/rpms/build.gradle
tst-dmendyke: More information is needed on the file 'gepgcommon' and it's location.

earth_enterprise/rpms/build.gradle
tst-dmendyke: Fair point.  I agree.

earth_enterprise/rpms/build.gradle
tst-ppenev: I think you should pass the build name to Gradle (`REL`, `OPT`, `DBG`, etc.).  E.g., `project.buildName`.

I propose one of the following ways to easily implement that:

1. Copy the PostGIS build directory to the 'staged install' directory.

1. Create a `.properties` file in the 'staged install' directory, which stores a property value, such as `buildName`.

This way there would be less danger of specifying a `src/NATIVE-${buildName}-x86_64` directory that doesn't correspond to the staged install directory.  The `build.gradle` script already receives a staged install location, so it won't need any new options to support the different builds.

I think we would want to add the name of any non-release build to the package name, or version.

Of course, in the future, it would be best if we don't bundle PostGIS and PostgreSQL with Open GEE.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +43,17 @@ apply plugin: "nebula.ospackage"
 
 import org.opengee.shell.GeeCommandLine
 
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def working_dir = "../src/"
```
tst-ppenev: This seems like a misleading name. The Gradle working directory is usually the project directory.

earth_enterprise/rpms/build.gradle
tst-ppenev: There's already a variable that has the path to `gee_long_version.txt` as a `File` below.  It is called `openGeeVersionFile`.  I think the name should correspond.  Or, just use the `File` variable.  It converts to `String`, and supports many other path operations.

tst-ppenev: Minor: Also, I think it's more readable to use Groovy string templates, rather than string arithmetic to construct a string.

E.g.:

```Groovy
def openGeeVersionFile = new File("../src/NATIVE-${project.buildName}-x86_64/gee_long_version.txt")
```

earth_enterprise/rpms/build.gradle
tst-ppenev: @tst-dmendyke: I don't understand your comment.  What information is needed about `gepgcommon`?

earth_enterprise/rpms/build.gradle
```diff
@@ -22,6 +22,10 @@
 // This Scons target copies build artifacts, binaries and documents into the
 // /tmp/fusion_os_install path.
 //
+//     ./gradlew -PDEBUGRPMS
```
tst-lsavoie: Looks like this comment needs to be updated.

earth_enterprise/rpms/build.gradle
```diff
@@ -62,14 +81,15 @@ if (new File('build.properties').exists()) {
     }
 }
 
-def openGeeVersionFile = new File("../src/NATIVE-DBG-x86_64/gee_long_version.txt")
+def openGeeVersionFile = new File( gee_long_version_txt )
 
 if (!openGeeVersionFile.exists()) {
     GeeCommandLine.expand(
-            ["/usr/bin/scons", "release=1", "NATIVE-DBG-x86_64/gee_long_version.txt"],
+      //            ["/usr/bin/scons", "release=1", gee_long_version_txt ],
```
tst-lsavoie: Minor: delete unused code.

earth_enterprise/rpms/build.gradle
```diff
@@ -417,7 +438,8 @@ task openGeeServerRpm (type: GeeRpm, dependsOn: openGeeSharedFiles) {
     installUtils = file('build/shared/install-utils.sh')
 
     // Add geresetpgdb script:
-    installUtils = file(new File(stagedInstallDir_server_opt, 'google/bin/gepgcommon'))
+    //    installUtils = file( new File( working_dir + "server/gepgcommon" ) )
```
tst-lsavoie: Minor: remove commented out code

earth_enterprise/rpms/build.gradle
tst-ppenev: - [ ] Rename variable to, e.g. `src_dir`.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +47,21 @@ apply plugin: "nebula.ospackage"
...
 import org.opengee.shell.GeeCommandLine
 
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def working_dir = "../src/"
+
+// Check for either "-PRELEASE" or "-POPTIMIZE" command line parameters.
+// Return the correct directory based on which paramater is used.
+def native_x86 = {
+  if ( project.hasProperty( 'RELEASE' ) ) { return "NATIVE-REL-x86_64" }
```
tst-ppenev: Grumble: This looks like it should be one variable with 3 possible values, rather than 3 boolean variables that are mutually exclusive.  It also seems more usual to me to have build default to release, rather than debug, which is pretty much only useful to developers.  But this mirrors the state of affairs in the `scons` build scripts, so it's OK for now.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +47,21 @@ apply plugin: "nebula.ospackage"
 
 import org.opengee.shell.GeeCommandLine
 
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def working_dir = "../src/"
+
+// Check for either "-PRELEASE" or "-POPTIMIZE" command line parameters.
+// Return the correct directory based on which paramater is used.
+def native_x86 = {
```
tst-ppenev: Minor: Could you include something the name of this function that indicates that it is the name of directory, and it's build directory?

tst-lsavoie:
I realized that build.gradle also has an option to call scons, but it will call it with release=1 by default. Someone could end up in a case where they run scons from gradle and build the release version, then it tries to create RPMs from the debug version. Off the top of my head, here are some options to fix that:

- Make the call to scons respect the build type setting
- Remove the call to scons and force users to call scons on their own
- Make the default setting "release"

earth_enterprise/rpms/build.gradle
tst-lsavoie: I like having the same interface as scons but since the default was previously "release" we should leave that the default so that we don't surprise anyone.

tst-ccamp:
Is there a ticket for this?

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +47,21 @@ apply plugin: "nebula.ospackage"
...
+  if ( project.hasProperty( 'OPTIMIZE' ) ) { return "NATIVE-OPT-x86_64" }
+  if ( project.hasProperty( 'DEBUG' ) ) { return "NATIVE-DBG-x86_64" }
+  return "NATIVE-REL-x86_64"
+}  // end native_x86
+
+// Build the actual file to be used
+def gee_long_version_txt = working_dir + native_x86() + "/gee_long_version.txt"
```
tst-ppenev: If you're using `scons release=1` to generate `gee_long_version.txt` below, you need to use `"../src/NATIVE-REL-x86_64/gee_long_version.txt"` as the path here.

- [ ] Make path to `gee_long_version.txt` consistent with the command to generate it.

earth_enterprise/rpms/build.gradle
```diff
@@ -62,15 +81,17 @@ if (new File('build.properties').exists()) {
     }
 }
 
-def openGeeVersionFile = new File("../src/NATIVE-DBG-x86_64/gee_long_version.txt")
+def openGeeVersionFile = new File( gee_long_version_txt )
```
tst-ppenev: Minor: I don't see the need to generate a `File` containing the path to `gee_long_version.txt`, and a `string` containing the same thing.
- [ ] Consider using one variable with the path to `gee_long_version.txt`.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +47,21 @@ apply plugin: "nebula.ospackage"
...
 import org.opengee.shell.GeeCommandLine
 
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def working_dir = "../src/"
+
+// Check for either "-PRELEASE" or "-POPTIMIZE" command line parameters.
+// Return the correct directory based on which paramater is used.
+def native_x86 = {
+  if ( project.hasProperty( 'OPTIMIZE' ) ) { return "NATIVE-OPT-x86_64" }
```
tst-ppenev: Minor: We don't have an issue ticket that specifies what build targets should be supported, but what `scons` supports is:

```Python
if release:
  builddir = 'REL-'
elif optimize:
  builddir = 'OPT-'
elif gprof:
  builddir = 'PROF-'
else:
  builddir = 'DBG-'
if sandbox:
  builddir += 'SBOX-'
builddir += march

if native_cc:
  builddir = 'NATIVE-' + builddir
```

Currently, this code partially supports `release=1`, `optimize=1` and `debug=1`, but not `profile=1`, or `sandbox=1`.  (I'm not sure anyone has tested cross-compiling.)

Should we support what `scons` supports, or write another ticket to refactor the two to be more consistent?

tst-ppenev: Minor: The project variable, and the variables passed to `scons` are not in all capitals.  I think it would easier, if users didn't have to remember when to capitalize, and when not to.

earth_enterprise/SConstruct
```diff
@@ -24,6 +24,21 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
+
+
+
+# Check for possible objects to pass on to gradle
+def gradle_options() :
+  options = ""  # Default to no option
+  if ARGUMENTS.get( 'release', '0' ) == '1' :
```
tst-ppenev: I don't think `== '1'` is a good test to use.  E.g., a user passing `scons release=2` would be interpreted as false here, but as true in `earth_enterprise/src/SConstruct`.

You can make your code consistent by parsing the same way:

```Python
release = int(ARGUMENTS.get('release', 0))
optimize = int(ARGUMENTS.get('optimize', 0))
gprof = ARGUMENTS.get('profile', 0) or ARGUMENTS.get('gprof', 0)
sandbox = int(ARGUMENTS.get('sandbox', 0))
native_cc = not int(ARGUMENTS.get('cross_cc', 0))
```

- [ ] Interpret `scons` command-line options consistently across `scons` scripts.

earth_enterprise/rpms/build.gradle
```diff
@@ -62,15 +81,17 @@ if (new File('build.properties').exists()) {
...
+// Call scons to create the file 'gee_long_version.txt'
+// The actual parameter used by scons can remain 'release=1'
+if ( !openGeeVersionFile.exists() ) {
     GeeCommandLine.expand(
-            ["/usr/bin/scons", "release=1", "NATIVE-DBG-x86_64/gee_long_version.txt"],
-            "Runing `scons` to produce <gee_long_version.txt> failed!",
-            null, // Use default environment variables.
-            new File("../src") // Working directory.
-        )
+      ["/usr/bin/scons", "release=1", openGeeVersionFile ],
```
tst-lsavoie: I think the most straightforward solution to this is to change "release=1" here to match whatever is passed in.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +56,21 @@ apply plugin: "nebula.ospackage"
 
 import org.opengee.shell.GeeCommandLine
 
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def src_dir = "../src/"
+
+// Look for command line paramters that specify how the packages are
```
tst-ppenev: Very minor: Project properties can be passed in `.settings` files, or injected from other projects, in addition to being specified on the command line.  We don't distinguish among these methods.

earth_enterprise/rpms/build.gradle
```diff
@@ -62,15 +90,18 @@ if (new File('build.properties').exists()) {
     }
 }
 
-def openGeeVersionFile = new File("../src/NATIVE-DBG-x86_64/gee_long_version.txt")
+// Create a file handle for gee_long_version_txt
+def openGeeVersionFile = new File( gee_long_version_txt )
 
-if (!openGeeVersionFile.exists()) {
+// Call scons to create the file 'gee_long_version.txt'
+// The actual parameter used by scons can remain 'release=1'
```
tst-ppenev: This second comment is wrong.  More details are below.

earth_enterprise/rpms/build.gradle
```diff
@@ -62,15 +90,18 @@ if (new File('build.properties').exists()) {
...
+if ( !openGeeVersionFile.exists() ) {
     GeeCommandLine.expand(
-            ["/usr/bin/scons", "release=1", "NATIVE-DBG-x86_64/gee_long_version.txt"],
-            "Runing `scons` to produce <gee_long_version.txt> failed!",
-            null, // Use default environment variables.
-            new File("../src") // Working directory.
-        )
+      ["/usr/bin/scons", "release=1", gee_long_version_txt ],
```
tst-ppenev: This command only works for a `gee_long_version.txt` path under the `NATIVE-REL-x86-64` build directory, and will fail for all other options.

You need to pass `debug`, `release`, `optimize` and `profile` back to `scons` to produce the output.

Here's how to test you code, e.g. for building with `debug=1`:

* Remove your `src/NATIVE-DBG-x86_64/` directory (as well as the other build directories like `src/NATIVE-REL-x86_64` for the purposes of testing).
* `cd earth_enterprise/rpms`
* `./gradlew -Pdebug tasks`
* If you see the following error, this code doesn't work correctly.

```

* Where:
Build file '/home/pav/prod/bitbucket.com/t-sciences/geoplatform/earthenterprise_code/earthenterprise/earth_enterprise/rpms/build.gradle' line: 103

* What went wrong:
>                 Runing `scons` to produce <../src/NATIVE-DBG-x86_64/gee_long_version.txt> failed!
                  Command: [/usr/bin/scons, release=1, ../src/NATIVE-DBG-x86_64/gee_long_version.txt]
                  Exit code: 2.
                  Error output:
                  
  scons: warning: The build_dir keyword has been deprecated; use the variant_dir keyword instead.
  File "/home/pav/prod/bitbucket.com/t-sciences/geoplatform/earthenterprise_code/earthenterprise/earth_enterprise/src/SConstruct", line 519, in <module>
  scons: *** Do not know how to make File target `NATIVE-DBG-x86_64/gee_long_version.txt' (/home/pav/prod/bitbucket.com/t-sciences/geoplatform/earthenterprise_code/earthenterprise/earth_enterprise/src/NATIVE-DBG-x86_64/gee_long_version.txt).  Stop.
  
                  Standard output:
                  scons: Reading SConscript files ...
  Checking for C library gtest... (cached) no
  Checking for C header file libpng12/png.h... (cached) yes
  scons: done reading SConscript files.
  scons: Building targets ...
```

earth_enterprise/rpms/build.gradle
```diff
@@ -113,7 +144,7 @@ def packageInstallLibDir = new File(packageInstallRootDir, 'opt/google/lib')
 
 def postGisInstallDir = new File(
         project.projectDir,
-        '../src/NATIVE-DBG-x86_64/third_party/postgis/install'
+          "${src_dir}${native_dir()}third_party/postgis/install"
```
tst-ppenev: Code format: This line seems to have an unequal indentation to the previous one.

earth_enterprise/rpms/build.gradle
```diff
@@ -178,6 +209,7 @@ ospackage {
 // Build packages for all platforms by default:
 defaultTasks 'osPackage'
 
+
```
tst-ppenev: General comment: If we're adding adding options like `debug` and `optimize` to `build.gradle`, we should pass them on to all `scons` commands, rather than just for outputting `gee_long_version.txt`.  Otherwise, it may be unexpected for a user to have these options work for only some tasks, and be ignored for others.

It's a small change, but we can create a separate ticket for it, if it would help get tasks completed.

earth_enterprise/rpms/build.gradle
```diff
@@ -22,6 +25,16 @@
 // This Scons target copies build artifacts, binaries and documents into the
 // /tmp/fusion_os_install path.
 //
+//     ./gradlew -Prelease
+// Build the RPMs using the files in 'NATIVE-REL-x86_64'
```
tst-ppenev: I don't see any code to process a `release` property.  I think this comment about `-Prelease` should be removed.

earth_enterprise/SConstruct
```diff
@@ -24,6 +26,27 @@ from khEnvironment import khEnvironment
...
+  result = ""
+  if int( ARGUMENTS.get( argument, '0' ) ) > 0 :
+    result = "-P" + argument
+  return result
+
+# Check for command line arguments passed that must be
+# passed on to gradle
+def gradle_options() :
+  options = ""
+  options = check_for_argument( 'debug' )
```
tst-ppenev: I'm not sure how you expect this to work.  There's currently no well-defined `debug` for the `scons` build, and, in fact, `src/SConstruct` would refuse to run with such an unrecognized option.

So, typically, there would never be a `debug` option passed to `scons`, and this line would have no effect.

earth_enterprise/SConstruct
```diff
@@ -24,6 +26,27 @@ from khEnvironment import khEnvironment
...
+  options = check_for_argument( 'release' )
```
tst-ppenev: On the other hand, `src/SConstruct` would make a debug build when `release=0`, but your code would currently not pass a `release` property to Gradle in that case.

earth_enterprise/rpms/build.gradle
```diff
@@ -190,7 +230,7 @@ task stageOpenGeeInstall(type: Exec, dependsOn: 'compileOpenGee') {
     // "installdir=${build_install_dir.getAbsolutePath()}", so we
     // overwrite whatever is in the default stage install directory:
     commandLine(
-        'scons', '-j8', 'release=1', 'stage_install',
+        'scons', '-j8', scons_options, 'stage_install',
```
tst-ppenev: Typo: `scons_options` is spelled `scons_option` in earlier locations.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +56,29 @@ apply plugin: "nebula.ospackage"
...
+// Look for command line paramters that specify how the packages are
+// to be built.  This can be either 'optimize' or 'debug'.
+// Option 'release' is the default.
+def native_dir = {
+  if ( project.hasProperty( 'optimize' ) ) {
+    scons_option = "optimize=1"
+    return "NATIVE-OPT-x86_64"
+  }  // end check for option 'optimze'
+  if ( project.hasProperty( 'debug' ) ) {
+    scons_option = ""
```
tst-ppenev: This doesn't work.  Passing an empty command-line argument to `scons` causes a failure:

```

* Where:
Build file '/home/pav/prod/bitbucket.com/t-sciences/geoplatform/earthenterprise_code/earthenterprise/earth_enterprise/rpms/build.gradle' line: 111

* What went wrong:
>                 Runing `scons` to produce <../src/NATIVE-DBG-x86_64/gee_long_version.txt> failed!
                  Command: [/usr/bin/scons, , ../src/NATIVE-DBG-x86_64/gee_long_version.txt]
                  Exit code: 2.
                  Error output:
                  
  scons: warning: The build_dir keyword has been deprecated; use the variant_dir keyword instead.
  File "/home/pav/prod/bitbucket.com/t-sciences/geoplatform/earthenterprise_code/earthenterprise/earth_enterprise/src/SConstruct", line 519, in <module>
  fusion/mtjtools/analyze/analyze.cpp:21:23: fatal error: gdal_priv.h: No such file or directory
  compilation terminated.
  scons: *** [NATIVE-DBG-x86_64/fusion/mtjtools/analyze/analyze.o] Error 1
  
                  Standard output:
                  scons: Reading SConscript files ...
  Checking for C library gtest... no
  Checking for C header file libpng12/png.h... yes
  scons: done reading SConscript files.
  scons: Building targets ...
  scons: building associated VariantDir targets: NATIVE-DBG-x86_64
  g++ -o NATIVE-DBG-x86_64/fusion/mtjtools/analyze/analyze.o -c -g -pipe -pthread -Wall -Werror -Wno-parentheses -Wno-error=unused-result -Wno-error=unused-local-typedefs -Wno-error=unused-but-set-variable -I/usr/include/libpng12 -D_LARGEFILE64_SOURCE -DQT_THREAD_SUPPORT -DQT_NO_ASCII_CAST -DQT_CLEAN_NAMESPACE -D_REENTRANT -DGUNIT_NO_GOOGLE3 -DKH_MAX_NUM_JOBS_COEFF=16 -INATIVE-DBG-x86_64/include -INATIVE-DBG-x86_64 -I. -INATIVE-DBG-x86_64/common -Icommon -Icommon/khmisc -Ikeyhole -INATIVE-DBG-x86_64/fusion -Ifusion -Ifusion/mtjtools/histogram fusion/mtjtools/analyze/analyze.cpp
```

You can set `scons_option` to `release=0`, or make this variable an array that gets combined with the other elements to construct the `scons` command line.

earth_enterprise/rpms/build.gradle
tst-ppenev: @tst-dmendyke: This minor comment still stands.  There are only `optimize` and `debug` project properties now.

earth_enterprise/rpms/build.gradle
tst-ppenev: @tst-dmendyke: This minor comment also stands; `release=1` isn't being passed to `scons` any more.

earth_enterprise/rpms/build.gradle
```diff
@@ -113,7 +152,7 @@ def packageInstallLibDir = new File(packageInstallRootDir, 'opt/google/lib')
 
 def postGisInstallDir = new File(
         project.projectDir,
-        '../src/NATIVE-DBG-x86_64/third_party/postgis/install'
+        "${src_dir}${native_dir()}third_party/postgis/install"
```
tst-lsavoie: If I'm following everything right there's a missing slash here after `${native_dir()}`.

earth_enterprise/SConstruct
```diff
@@ -24,6 +26,26 @@ from khEnvironment import khEnvironment
...
+def check_for_argument( argument ) :
+  result = ""
+  if int( ARGUMENTS.get( argument, '0' ) ) > 0 :
+    result = "-P" + argument
+  return result
+
+# Check for command line arguments, if they are found
+# pass them on to Gradle.
+def gradle_options() :
+  options = "-Pdebug"
```
tst-lsavoie: Ran into a bug here. `options` is completely overwritten on every call, so you only get the result from `check_for_argument('release')`. This needs to be updated so that it will only change `options` if a non-empty string is returned from `check_for_argument`.

tst-lsavoie: I rewrote the function like this and it seemed to work:
```
def gradle_options() :
  options = "-Pdebug"
  for opt in ['profile', 'optimize', 'release']:
    gradleOpt = check_for_argument(opt)
    if gradleOpt:
      options = gradleOpt
  return options
```

earth_enterprise/SConstruct
```diff
@@ -24,6 +26,19 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
+# An object used to check for a specific command line parameter.
+# Set the instance variable 'option' based on passed parameter.
+class CommandLineParameter( object ) :
```
tst-ppenev: Style: I don't see this class as checking for a specific command line parameter.  It checks for any parameter that changes the build directory, and which is understood by the `build.gradle` for RPM packaging.

I don't quite understand the need for a class which only has a constructor in it.  In fact, you don't even need a function, you can just evaluate all of this code as part of the `SConstruct` to produce a variable, e.g., something like:

```Python
package_command_arguments = "-Pdebug"

for parameter in [ 'release', 'optimize', 'profile' ]:
    if int( ARGUMENTS.get( parameter, '0' ) ):
        package_command_arguments = "-P" + parameter
        break
```

tst-ppenev: Style: Also the name of this class may have a high likelihood of colliding with the name of anything that processes command-line parameters in general.

earth_enterprise/SConstruct
```diff
@@ -158,7 +171,7 @@ stage_tutorial = env.PhonyTargets(**stage_tutorial_cmds)
 env.Depends(stage_tutorial, stage_data)
 
 package_staged_cmds = {'package_install': [
-    'cd rpms; ./gradlew osPackage'
+    "cd rpms; ./gradlew osPackage " + CommandLineParameter().option
```
tst-ppenev: That was minor comment, but you'd need to replace `CommandLineParameter().option` with `package_command_arguments` here.

earth_enterprise/SConstruct
```diff
@@ -26,16 +26,14 @@ from khEnvironment import khEnvironment
...
-  def __init__( self ) :
-    for parameter in [ 'profile', 'optimize', 'release' ] :
-      if int( ARGUMENTS.get( parameter, '0' ) ) > 0 :
-        self.option = "-P" + parameter
+# Set 'package_command_arguments' based on a parameter
+# that is to be passed on to Gradle
+package_command_arguments = "-Pdebug"
+
+for parameter in [ 'release', 'optimize', 'profile' ] :
+  if int( ARGUMENTS.get( parameter, '0' ) ) :
```
tst-lsavoie: Doesn't this need to be `int( ARGUMENTS.get( argument, '0' ) ) > 0` to handle the case where the user passes `release=0`?

earth_enterprise/SConstruct
```diff
@@ -24,6 +26,17 @@ from khEnvironment import khEnvironment
 
 EnsureSConsVersion(0, 96, 92)
 
+# Set 'package_command_arguments' based on a parameter
+# that is to be passed on to Gradle
+package_command_arguments = "-Pdebug"
+
+for parameter in [ 'release', 'optimize', 'profile' ] :
+  if int( ARGUMENTS.get( parameter, '0' ) ) > 0 :
```
tst-ppenev: This doesn't need to be `> 0` to handle cases when a parameter is `0`.  This implementation, actually, makes it inconsistent with how it's parsed in `src/SConstruct`.  Negative values are interpreted as `True` there and `False` here.

earth_enterprise/rpms/build.gradle
```diff
@@ -43,6 +53,29 @@ apply plugin: "nebula.ospackage"
 
 import org.opengee.shell.GeeCommandLine
 
+// Varables used for specifying location of two files
+// 'gee_long_version.txt' and 'gepgcommon'
+def src_dir = "../src/"
+def scons_option = "release=1"
+
+// Look for command line paramters that specify how the packages are
```
tst-dsheremata: parameters

Issue #252, opened 2017-06-19
Google Earth Server fails to load Admin console when there is an invalid link to a published portable file


avnishb:
As reported:

_We recently had to move to a new location and one of the NAS devices is not yet online. One of the portable maps (.glm) was quite large and was moved to the new NAS. In its place a sym-link was created. However, since the target file is not there, the admin console does not finish loading._

## Reproduction steps
1. Install GEE Server
1. Optionally push and publish a database (so it's easier to see when the page load fails)
1. Load the admin console `http://localhost/admin` (user/password: `geapacheuser`/`geeadmin`) - Page should load correctly
1. Click on Manage Portables button - portables popup window appears
1. Create a symlink in the globes directory under `/opt/google/gehttpd/htdocs/cutter/globes` to a portable globe file that doesn't exist
    ```
    sudo su - geapacheuser
    cd /opt/google/gehttpd/htdocs/cutter/globes
    ln -s /some_path_that_doesn't_exist/myglobe.glb .
    ```
1. Load or reload the the admin console `http://localhost/admin`

Expected:
1. Page loads correctly and displays available databases
1. Clicking on Manage Portables button displays portables popup window

Actual:
1. Page does not display any databases if they were previously present
1. Clicking on Manage Portables button displays the error message " There are currently no globes."
1. The web console displays a Javascript error similar to:
    ```
    TypeError: div is null[Learn More]  gees_utilities.js:83:7
        gees<.dom.show http://localhost/js/gees_utilities.js:83:7
        gees<.tools.notificationBar http://localhost/js/gees_utilities.js:469:7
        gees<.tools.errorBar http://localhost/js/gees_utilities.js:483:7
        gees.initialize.databases http://localhost/js/gees_initialize.js:109:7
        <anonymous> http://localhost/js/gees_initialize.js:236:1
    ```
1. No other errors appear in the error_log

tst-ccamp:
I've done some preliminary investigation and it looks like the place where this is occurring is earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py in the _GetPortableGlobesList function. I'm sure there is an error occurring somewhere when it's trying to get info about the file and it's being swallowed up and doesn't get logged.

PR #773, opened 2018-03-24
Fixes #765 and #252 - Fix no portables and bad links

### LINKED ISSUE #765: "Manage Portables" produces the error "There are currently no globes" incorrectly
In GEE Cutter, after a .glm is cut, it appears to work, i.e. .glm files appear in  `/opt/google/gehttpd/htdocs/cutter/globes/`, but clicking on "Manage Portables" yields the error: "There are currently no globes".

This may be related to the fix made for #252. In fixing this issue, also make sure that the original issue does not occur.

tst-jhaight:
Needed to check that portable files actually exist using the full path instead of just the filename.  Should really still add a log message for files that don't exist on Monday.

Tested on #252 and #765 on CentOS 7
Test steps:
1) Create, push, and publish a database using fusion and geserver (to see that links don't break display)
2) Create a symlink in /etc/google/gehttpd/htdocs/cutter/globes
3) Place a portable globe in /etc/google/gehttpd/htdocs/cutter/globes
4) Log in to the server admin page
5) Pushed DB from step 1 is visible
6) Clicking manage portables brings up a window containing the portable globe
7) Add another file into the portable globes file
8) Refresh the portable globes display and ensure the new globe appears


earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
```diff
@@ -1213,7 +1213,7 @@ def _GetPortableGlobesList(self, registered_portable_set):
     root = constants.CUTTER_GLOBES_PATH
     for name in os.listdir(root):
       # Ignore globes that are registered.
-      if (name not in registered_portable_set) and (os.path.isfile(name)):
+      if name not in registered_portable_set and os.path.isfile(os.path.join(root,name)):
```
tst-nfarah: so was the cause of that error "no portable..." the old code that use short file name instead of the correct file name (full path) ? Basically the globe was there but we did not find because of it's short name ?

tst-nfarah:
Since this PR address 2 issues that are targeting 5.2.2 release, the target merge branch needs to be 'release_5.2.2'.

earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
tst-jhaight: Yeah.  The previous check fixed the symlink problem because almost everything would fail the isfile check but that started preventing valid portables from being built.  Not sure what directory the server is executing in but it's apparently not /opt/google/gehttpd/htdocs/cutter/globes

tst-nfarah: Thanks. This is an introduced regression issue when fixing #252 ? (https://github.com/google/earthenterprise/commit/26081a0f8ebabaeb6343bafbeb404c6268c98e0e). Wanted to know so we would not document it an release notes for 5.2.2 if it was a introduced then fixed issue.

tst-jhaight: I believe that is correct.  Before the fix for 252 managing portables looks like it worked fine.  As long as this is fixed there should be no change in behavior.

tst-jhaight:
When I changed target branches, it looks like it's trying to merge more than just my fix.  I think I need to just create a new pull request for the release_5.2.2 target.

tst-jhaight:
Closing this PR since it was initially based off of master instead of release_5.2.2

Issue #764, opened 2018-03-21
Compile open GEE C++ code using C++11 standards


tst-rwildes:
Currently open GEE compiles using C++98 (the default) standards.  It would be good to move to a more modern standard and add the `std=C++11` or `std=gnu++11`.

tst-ppenev:
I'd vote for C++11. I doubt we'll be trying to compile with something other than GCC any time soon, but I'm also not sure how many GNU extensions we'd want to use anyway.

tst-rwildes:
GEE code seems to compile just fine with C++11 (with some changes I made) but some of the third party code used by GEE doesn't compile without using gnu++11.  Checkout: https://github.com/tst-rwildes/earthenterprise/tree/gee-c++11 if you are interested.  Will be doing a WIP soon based on that fork/branch.

tst-rwildes:
It was surprisingly few changes to get GEE to compile with C++11.

tst-rwildes:
The ABI is comparable but you do have some incompatibilities in definitions of some objects.  But if you are careful on what you pass between binaries that are compiled differently they can co-exist. 

https://gcc.gnu.org/wiki/Cxx11AbiCompatibility

PR #774, opened 2018-03-24
GEE c++11

tst-rwildes:
Addresses issue #764 to move GEE C++ code to C++11 standards.


earth_enterprise/src/SConstruct
```diff
@@ -97,6 +98,7 @@ coverage = ARGUMENTS.get('coverage', 0)
 bundle_3rd_libs = int(ARGUMENTS.get('bundle_3rd_libs', 0))
 cc_dir = ARGUMENTS.get('cc_dir', None)
 log_performance = int( ARGUMENTS.get( 'log_performance', 0 ) )
+cpp_standard = ARGUMENTS.get('cpp_standard', 'gnu++11')
```
tst-dsugar: Well, on very new distros with gcc 6 or later, c++11 is the default, not gnu++11.  Is it possible to test gcc versions from scons?

earth_enterprise/src/common/khCppStd.h
```diff
@@ -0,0 +1,31 @@
...
+
+
+#ifndef _khcpp11_h_
+#define _khcpp11_h_
+
+// This will help with transitioning from C++98 to C++11.  Once fully transitioned over
+// we should stop using this file and delete this file
+
+#if __cplusplus <= 199711L    // C++98
+#  define nullptr NULL
```
tst-dsugar: Maybe ((void *)NULL) would be a bit better....?

earth_enterprise/src/SConstruct
tst-rwildes: Yes we need to move away from gnu extensions.  The good news is I believe GEE code doesn't rely on any extensions directly but some of the 3rd party libraries do.  I'm hoping the effort to update to newer 3rd party libraries will allow us to then switch to c++11.

tst-rwildes: BTW we don't compile with c++98 either...  We need use gnu++98.  I think the dependency on gnu extensions is a different issue that should be raised and addressed outside this issue.

tst-rwildes: Created #775

earth_enterprise/src/common/khCppStd.h
tst-rwildes: Unfortunately that doesn't work because with C++ (unlike C) you can not implicitly convert a `void*` to typed pointer.  However, this made me look around a bit and see if there was something we could do better than using `NULL` as a stand in for `nullptr` when building c++98 and I found something in a book by Scott Meyers.

earth_enterprise/src/common/khRefCounter.h
```diff
@@ -175,12 +190,30 @@ class khRefCounterImpl : public ThreadPolicy::MutexHolder {
...
+  khRefCounterImpl(const khRefCounterImpl &) : refcount_(1) {
+    // intentionally left empty.  The right thing to do is nothing.
+  }
+  khRefCounterImpl & operator=(const khRefCounterImpl&) {
+    // intentionally left empty.  The right thing to do is nothing.
+    return *this;
+  }
+
+#if GEE_HAS_MOVE
+  khRefCounterImpl(khRefCounterImpl &&) noexcept : refcount_(1) {
```
tst-ppenev: Q: I'm not sure how this class is being used, but aren't you supposed to copy the reference count when you're moving an object?

earth_enterprise/src/common/verref_storage_io.h
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2018 Google Inc.
```
tst-ppenev: Was this file really written by Google?

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -63,10 +65,43 @@ class DerivedAssetHandleD_ : public virtual BaseD_, public ROBase_
  public:
   DerivedAssetHandleD_(void) : BBase(), BaseD(), ROBase() { }
   DerivedAssetHandleD_(const std::string &ref_) :
+      // Only call the common (virtually inherited) base class with the initializtion state.
+      // Its the only one that has state anyway.  Also explicitly calling the virtual base
```
tst-ppenev: Typo: `Its` -> `It's`.

tst-ppenev: Consider a comma after `Also`.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -39,6 +40,7 @@ class DerivedAssetHandleD_ : public virtual BaseD_, public ROBase_
   typedef ROBase_ ROBase;
```
tst-ppenev: Q: I know this is not your code, but what's the purpose of defining type names ending in `_`, especially, if we're aliasing them to types without the underscore afterwards?

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -235,27 +283,61 @@ class MutableAssetHandleD_ : public virtual Base_ {
   }
 
 
-  // the compiler generated assignment and copy constructor are fine for us
-  // ref & handle have stable copy semantics and we don't have to worry about
-  // adding to the dirtyMap because the src object will already have
-  // done that
+  // We need to more explicit about copy/move constructor and assignment as we are using virtual
```
tst-ppenev: Typo: `need to more explicit`: missing preposition?

earth_enterprise/src/common/khRefCounter.h
tst-rwildes: No.  And the reason is the reference count (i.e. the base classes state) has nothing to do with the derived class (or its state).  This is one of those odd cases where the base class actually should support a do nothing move/copy.  The job of the base class has nothing to do with of whatever the derived class is trying to implement.  Instead the base class has the job of managing the one and only memory block it is assigned when first constructed and nothing more and that assignment never changes.  So as  the object state of the derived class changes because of move/copy (or anything else) this base class doesn't care because that doesn't have any influence on weather or not that block of memory can be cleaned up or not (i.e. it doesn't effect the number of external pointers pointing to this memory block).  The memory is cleaned up only when external intrusive pointers have zero references to that memory.  The fact that the state of the memory changed doesn't at all change how many external pointers are referencing this memory block (i.e. the following operation doesn't at all change the reference count: `*p1 = *p2`).  The base class has a completely different mission from the derived class: memory management.  So it can't and shouldn't participate in any assignment/move operations that the derived class is doing as that doesn't effect the job of memory management (for this memory block).  This is probably why the C++ committee instead added the make_shared<> as it gives you most of the same benefits of intrusive shared objects without the weirdness of things like this.

earth_enterprise/src/common/khCppStd.h
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2018 Google Inc.
```
tst-ppenev: This should say `Copyright 2018 the OpenGEE Contributors`.  I don't think Google is contributing to this project any longer.

tst-rwildes: Good catch!  And thanks for all the feedback!  I'm going fix these things up and do another push.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
tst-rwildes: It is good practice to typedef template parameters.  The standard library pretty much follows this practice too.  There is currently no way in C++ to extract this information external to the template class (when they add reflection then that might change).  Many times when using a template class externally this comes in handy

earth_enterprise/src/common/khRefCounter.h
tst-ppenev: I see.  I think this class just needs its documentation to say it just provides storage for a reference counter, but does nothing, and the derived classes are expected to make any use of the counter.

earth_enterprise/src/fusion/autoingest/AssetHandle.h
```diff
@@ -108,12 +109,17 @@ class AssetHandle_  {
...
+  // adding to the cache because the src object will already have done that.
+  // Same goes for move constructor and assignment.
 
   virtual ~AssetHandle_(void) { }
-  std::string Ref(void) const { return ref; }
+  const std::string& Ref(void) const { return ref; }
   bool Valid(void) const;
-  operator bool(void) const { return Valid(); }
+  // This is better then overloading the bool operator as it
+  // more closely emulates what a pointers boolean operations does
```
tst-ppenev: Typo: `pointers` -> `pointer's`.

tst-ppenev: Also, you can put punctuation at the end of the sentence.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
tst-ppenev: Ah, yes, the C++ compiler needs a lot of explicit help.

tst-dsugar: Well, actually there is a preference to using "using" in place of typedef in c++11 and later...

thisisbasil:
@tst-rwildes I will review... Just because

tst-basilhuffman:
@tst-rwildes I will review. Just because

PR #776, opened 2018-03-26
Fix no portables 765

tst-jhaight:
Fixes issue #765.  Needed to check that portable files actually exist using the full path instead of just the filename.

Tested that #252 is still also fixed as well. Tested on CentOS 7
Test steps:

1. Create, push, and publish a database using fusion and geserver (to see that links don't break display)
2. Create an invalid symlink in /etc/google/gehttpd/htdocs/cutter/globes
3. Place a portable globe in /etc/google/gehttpd/htdocs/cutter/globes
4. Log in to the server admin page
5. Pushed DB from step 1 is visible
6. Clicking manage portables brings up a window containing the portable globe
7. Add another file into the portable globes file
8. Refresh the portable globes display and ensure the new globe appears
9. Opened /opt/google/gehttpd/logs/gestream_publisher.out and verified there is a warning related to the invalid symlink created in step 2.


tst-mswartz:
I've now tested it and it works correctly, I now get a list of portable globes from the admin console when i click "Manage portables"

PR #779, opened 2018-03-29
properly allocating CPUs for gecombineterrain

tst-basilhuffman:
seems to be better than just counting the number of processors total. although
this functionality can be provided by "nproc --all"

not tested, compiles ok


earth_enterprise/src/common/config/geConfigUtil.cpp
```diff
@@ -42,15 +44,28 @@ bool IsRedHat(void) {
 
 uint GetNumCPUs(void) {
   uint numcpus = 0;
-  std::ifstream in("/proc/cpuinfo");
+  /*std::ifstream in("/proc/cpuinfo");
```
tst-ppenev: I think we should write in C++ instead of pretending that C++ is BASH.

In C++11 you can use http://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency.

Otherwise, POSIX defines `sysconf`(3) for that purpose, e.g.:

```C++
#include <unistd.h>

long cpu_count = sysconf(_SC_NPROCESSORS_ONLN);
```

tst-basilhuffman: It might be better to use sysconf until we are actually using 11 (it is currently still in review??)

Either way, this function must change, as well as GetPhysicalMemorySize( ) below because, as they stand, they are nothing more than re-interpretations of `grep | wc -l`

tst-ppenev:
Is there a ticket for this pull request?  Otherwise, you could refactor `GetPhysicalMemorySize`, too.

earth_enterprise/src/common/config/geConfigUtil.cpp
```diff
@@ -41,41 +41,14 @@ bool IsRedHat(void) {
...
-        ++numcpus;
-      }
-    }
-  }
-
-  return std::max(uint(1), numcpus);
+  return std::max(uint(1), uint(sysconf(_SC_NPROCESSORS_ONLN)));
 }
 
+// want memory size in Kb
```
tst-ppenev: I'm not sure what this comment means.  It looks like this function is trying to return the memory size in bytes.

earth_enterprise/src/common/config/geConfigUtil.cpp
```diff
@@ -41,41 +41,14 @@ bool IsRedHat(void) {
...
-        return 0;  // parse failure...too many tokens or not "kB"
-      }
-    }
-    return 0;  // parse failure
-  }
-
-  return 0;  // parse failure
+    uint64 pages = sysconf(_SC_PHYS_PAGES),
+           page_size = sysconf(_SC_PAGE_SIZE)/1024;
+    return (pages * page_size * 1000);
```
tst-ppenev: Do we really need to return an incorrect memory size?

I think this should return the memory size in bytes, and any functions that use the memory size should make the proper calculations.

tst-basilhuffman: This would necessitate a review to see where this function is called

tst-ppenev: Yeah, this is the only place it is used:

```
src/fusion/gecombineterrain/gecombineterrain.cpp:290
```

tst-basilhuffman: Do we want to go ahead and fix this here or open a new issue?

tst-ppenev: I don't think we need a new issue.  I think `gecombineterrain` will work just fine using the actual amount of memory.

earth_enterprise/src/common/config/geConfigUtil.cpp
```diff
@@ -41,41 +41,14 @@ bool IsRedHat(void) {
...
-    std::string line;
-    while (std::getline(in, line)) {
-      if (StartsWith(line, "processor")) {
-        ++numcpus;
-      }
-    }
-  }
-
-  return std::max(uint(1), numcpus);
+  return std::max(uint(1), uint(sysconf(_SC_NPROCESSORS_ONLN)));
```
tst-lsavoie: I'm a bit concerned about this since `_SC_NPROCESSORS_ONLN` is a non-standard GNU extension. Where possible I'd prefer to write code that works on all systems. As messy as it is, that might be the reason for the original code that reads through /proc.

Same comment applies to the other `sysconf` calls.

tst-lsavoie:
In our discussion about this ticket we decided it would be best to change `numcpus` to `numCompressThreads` or something like that since it more closely describes how the parameter is used. We should leave `numcpus` as well but make it an alias for `numCompressThreads`.

Also, the documentation needs to be updated to make sure the description of `numcpus` is correct, to add `numCompressThreads`, and to mark `numcpus` as deprecated.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,46 @@ int main(int argc, char **argv) {
...
+
     if (argn == argc) {
-      usage(progname, "No input indexes specified");
+      usage(progname, "No input indices specified");
     }
+    uint cmdDefaultCPUs = CommandlineNumCPUsDefault();
+    uint numavailable = GetMaxNumJobs();
+
+    /*
+ 	Choosing number of CPUs:
```
tst-lsavoie: Minor: change tabs to spaces.

earth_enterprise/src/common/config/geConfigUtil.cpp
tst-basilhuffman: Understood. There is also pulling in the results from a call to `nproc`. `std::thread::hardware_concurrency()` is another option, but it would necessitate waiting on c++11 migration to be complete

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,46 @@ int main(int argc, char **argv) {
...
+	The methodology	first checks if the --numcpus flag was present, if
+	it was, it will go with what was passed, no matter the number. If
+	it was not, it will take the minimum value of what is defined in the
+	systemrc file and what is currently available (values described 
+	above). If the maxjobs tag in systemrc is invalid (i.e. less than
+	1, it will default to the available CPUs.
+    */
+    if (!numcpus) {
+	if (cmdDefaultCPUs <= 1)
+		cmdDefaultCPUs = numavailable;
```
tst-lsavoie: Minor: it would be nice to have brackets around this if statement.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,46 @@ int main(int argc, char **argv) {
...
+	  4. the number of currently available cpus, obtained through sysconf
+
+	The methodology	first checks if the --numcpus flag was present, if
+	it was, it will go with what was passed, no matter the number. If
+	it was not, it will take the minimum value of what is defined in the
+	systemrc file and what is currently available (values described 
+	above). If the maxjobs tag in systemrc is invalid (i.e. less than
+	1, it will default to the available CPUs.
+    */
+    if (!numcpus) {
```
tst-lsavoie: I'm a bit confused here - your comment above says numcpus defaults to 1 if it's not present, but this seems to indicate that numcpus will be 0 if it is not provided. It looks like you may have set it to be 0 if it's not present and the comment above means that you will set it to 1. If that's the case I think the comments could use some cleanup.

tst-lsavoie:
One other thing I'd like to see - this code is from combineterrain.cpp around line 300:
```
  if (num_cpus_ >= 4) {
    compress_cpus--; // If we have more than 4 cpus save one for reads/writes.
  }
```
So in some cases it subtracts one from the number of threads it starts. I'd like to remove this as well so that there is a strong connection between the number passed in the command line and the number of compressor threads are always the same.

tst-lsavoie:
Sorry, closed accidentally.

earth_enterprise/src/version.txt
```diff
@@ -1 +1 @@
-5.2.2
\ No newline at end of file
+5.2.2
```
tst-lsavoie: Not related to this PR, but I wonder if this was checked in by accident.

tst-basilhuffman: Yeah, I originally deleted it by accident iirc

earth_enterprise/src/fusion/config/gefConfigUtil.cpp
```diff
@@ -153,16 +153,16 @@ uint32 GetMaxNumJobs() {
...
   // with maximum number of concurrent jobs equals 0 (internal usage).
   if (max_num_jobs == 0) {
     max_num_jobs = GetDefaultMaxNumJobs();
   }
 
   // Set the max_num_jobs to the min of the values: number of CPUs or max
   // allowable number of jobs.
-  max_num_jobs = std::min(max_num_jobs, GetNumCPUs());
+  uint numCPUs = GetNumCPUs();
+  max_num_jobs = std::min(max_num_jobs, numCPUs);
```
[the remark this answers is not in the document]
tst-basilhuffman: Was using it when debugging, will remove

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,47 @@ int main(int argc, char **argv) {
...
     }
+    uint cmdDefaultCPUs = CommandlineNumCPUsDefault();
+    uint numavailable = GetMaxNumJobs();
+
+    /* 
+        Choosing number of CPUs:
+
+        numcpus: initially, the command line parameter --numcpus
+	
+        CommandLineNumCPUsDefault(): defined in /etc/opt/google/systemrc via maxjobs tag
```
tst-ppenev: @tst-basilhuffman: Yes, I think, ideally, we should change this to a better name.  However, that's a minor improvement, so I'll approve even without it.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,47 @@ int main(int argc, char **argv) {
...
+
+        GetMaxNumJobs(): the number of available jobs, defined as the minimum value of either:
+         1. the currently available jobs or the default number of jobs i.e. kMaxNumJobs (is 8) multiplied by KH_MAX_NUM_JOBS_COEFF, which is a compiler option
+	 2. kMaxNumJobsDefault = 8
```
tst-ppenev: Minor: This line seems to have an inconsistent indentation.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,47 @@ int main(int argc, char **argv) {
...
+
+    /* 
+        Choosing number of CPUs:
+
+        numcpus: initially, the command line parameter --numcpus
+	
+        CommandLineNumCPUsDefault(): defined in /etc/opt/google/systemrc via maxjobs tag
+
+        GetMaxNumJobs(): the number of available jobs, defined as the minimum value of either:
+         1. the currently available jobs or the default number of jobs i.e. kMaxNumJobs (is 8) multiplied by KH_MAX_NUM_JOBS_COEFF, which is a compiler option
```
tst-ppenev: Minor: This line seems rather long.  Perhaps, it could be split.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,47 @@ int main(int argc, char **argv) {
...
+	 2. kMaxNumJobsDefault = 8
+         3. environment variable, KH_GOOGLE_MAX_NUM_JOBS, if defined
+         4. the number of currently available cpus, obtained through sysconf
+
+         The methodology	first checks if the --numcpus flag was present, if
```
tst-ppenev: Minor: there seems to be an interesting amount of space between words on this line.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -232,28 +234,47 @@ int main(int argc, char **argv) {
...
+         it was, it will go with what was passed, no matter the number. If
+         it was not, it will take the minimum value of what is defined in the
+         systemrc file and what is currently available (values described 
+         above). If the maxjobs tag in systemrc is invalid (i.e. less than
+         1, it will default to the available CPUs.
```
tst-ppenev: Minor: Close brackets.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -225,35 +230,79 @@ int main(int argc, char **argv) {
...
+        Choosing number of CPUs:
+
+        numcpus: initially, the command line parameter --numcpus
+	
+        CommandLineNumCPUsDefault(): defined in /etc/opt/google/systemrc via maxjobs tag
+
+        GetMaxNumJobs(): the number of available jobs, defined as the minimum value of either:
+         1. the currently available jobs or the default number of jobs i.e. kMaxNumJobs 
+            (is 8) multiplied by KH_MAX_NUM_JOBS_COEFF, which is a compiler option
+	 2. kMaxNumJobsDefault = 8
```
tst-ppenev: Lint: Uneven indentation.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -225,35 +230,79 @@ int main(int argc, char **argv) {
...
+    */
+ 
+    if (numCompressThread)
+    {
+       if (numcpus) {
+         usage(progname,"--numcpus deprecated in favor of --numCompressedThreads");
+       } else {
+         // if numcpus is not present, but numCompressThreads is, set
+         // numcpus equal to numCompressThreads
+         numcpus = numCompressThreads;
```
tst-ppenev: Syntax: Missing closing `}`.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -225,35 +230,79 @@ int main(int argc, char **argv) {
...
-
-    notify(NFY_WARN, "gecombineterrain numcpus: %llu ",
-               static_cast<long long unsigned int>(numcpus));
-    notify(NFY_WARN, "gecombineterrain CommandlineNumCPUsDefault(): %llu ",
-               static_cast<long long unsigned int>(CommandlineNumCPUsDefault()));
-    
-    numcpus = std::min(numcpus, CommandlineNumCPUsDefault());
     
+    if (!numcpus) {
+	if (cmdDefaultCPUs <= 1) {
```
tst-ppenev: Lint: Convert tabs to spaces.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -88,7 +88,9 @@ void usage(const std::string &progn, const char *msg = 0, ...) {
       "      --help | -?:  Display this usage message\n"
       "      --sortbuf <buffer_size>: size of packet index sort buf in MB\n"
       "        (default %u)\n"
-      "      --numcpus:    Number of CPUs to use (default %u)\n"
+      "      --numCompressThreads:    Number of CPUs/threads to use \n("
```
tst-ppenev: Minor: The command-line option sets the number of threads that will be used.  We don't know how many CPUs the underlying schedulers will use.

earth_enterprise/src/fusion/gecombineterrain/gecombineterrain.cpp
```diff
@@ -225,35 +230,80 @@ int main(int argc, char **argv) {
...
+        systemrc file and what is currently available (values described 
+        above). If the maxjobs tag in systemrc is invalid (i.e. less than
+        1) it will default to the available CPUs.
+
+         
+        --numcpus is deprecated in favor of --numCompressThreads. One
+        or the other may be passed in, but not both.
+    */
+ 
+    if (numCompressThread)
```
tst-lsavoie: This is a compile error - should be `numCompressThreads` (with an 's' on the end).

tst-lsavoie: I can fix this quickly online if you like.

tst-ccamp:
@tst-basilhuffman Please make sure there is a GitHub Issue created for this PR

PR #795, opened 2018-04-05
Updated earth_enterprise/rpms/build.gradle to output Debian package for common libraries

### LINKED ISSUE #762: Package Libraries and Utilities Common to Fusion and Server for Ubuntu 16
As a Google Earth user, I want to be able to install Fusion and Server from a Debian package manager on Ubuntu 16.  There are libraries and utilities common to both Fusion and Server.

There are 3 tickets (sub-tasks) for packaging Earth Server:

* **Package libraries and utilities common to Fusion and Server. [This ticket.]**
* Package PostGIS. (#761)
* Package the rest of Earth Server. (#760)

And there are 2 tickets (sub-tasks) for packaging Fusion:

* **Package libraries and utilities common to Fusion and Server. [This ticket.]**
* Package the rest of Fusion. (#763)

## AC

* Update the `earth_enterprise/rpms/build.gradle` script to output a Debian package for the common libraries and utilities used by Google Earth Server and Fusion.
* The package should install on Ubuntu 16.
* The package should have the correct dependencies, so it can be installed by a single dpkg -i <package-name>.deb command on a minimal Ubuntu system (e.g., Docker `ubuntu` image). [Can be split into a sub-task.]
* Rename the `rpms` directory to, e.g. `packages`, if its name is still RPM-specific. [This name can change to gain consensus.]

## Guidelines

* There is a stub `openGeeCommonDeb` task in `rpms/build.gradle`. You can use the `openGeeCommonRpm` task as an example.
* Share as much install script code as possible between RPM and Debian packages.  There are `pre-instal.sh`, `post-install.sh`, etc. scripts under the `opengee-common`, `opengee-fusion` and `opengee-server` that are used in the RPM packages.  Examples of techniques already used:
    * You can write shell script that detects platform-specific tools at run-time, and uses what's available.
    * You can use Groovy templates that expand variables, conditionally include or remove code blocks, or conditionally include other shell script files.
* There are Groovy classes under `rpms/buildSrc/src` that call `/usr/lib/rpm/find-provides` and `/usr/lib/rpm/find-requires`. Those scripts are included in the `rpm` Debian package, and do run on Debian systems like Ubuntu.  It may be possible to process their output on Ubuntu to auto-generate package dependencies.  However, that should be left as a separate task.  As a first iteration you can hard-code the Debian package dependencies.

mrstephen:
Updated the openGeeCommonDeb task in build.gradle to install common files and specify dependencies.

Tested the build and installation of the resulting opengee-common***.deb package on Ubuntu 16.04.
Installed using sudo dpkg -i opengee-common***.deb. Installed dependencies with sudo apt-get -f install.

For me, the install of dependencies was hit-and-miss, but I believe that this was due to some issues I was having with my virtual machine. In the instances where it did not install the dependencies, apt-get would not allow me to install anything at all (using sudo apt-get install <package>).

The list of dependencies was compiled from the list in BUILD_Ubuntu.md and needs a follow-on task to clean up the list.

Fixes #762

### DIFF
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -403,35 +403,19 @@ if (project.hasProperty('buildOpenGee')) {
 }
 
 task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
-    packageName = 'opengee-common'
-    release = "1.${rpmPlatformString}"
+    packageName = openGeeCommonRpm.packageName
+    release = openGeeCommonRpm.release
     version = openGeeVersion
-    user = 'root'
-    permissionGroup = 'root'
-    packageGroup = 'Application/Productivity'
-    summary = 'Third-party libraries bundled with Open GEE'
+    user = openGeeCommonRpm.user
+    permissionGroup = openGeeCommonRpm.permissionGroup
+    packageGroup = 'misc'
+    summary = openGeeCommonRpm.summary
     // The package build failed when packageDescription was copied here from openGeeCommonRpm, so it is omitted for now
-    license =
-        'ASL 2.0 and ' +
-        'BSD and ' +
-        'GPLv2+ and ' +
-        'Psycopg2 GPL 3+ and ' +
-        'LGPLv2 and ' +
-        'MIT and ' +
-        'MPLv1.1 and ' +
-        'PSF 2 and ' +
-        'Python Imaging Library 1.1.7 and ' +
-        'mgrs 1.1.0 and mm 1.4.2 and ' +
-        'OpenLDAP and ' +
-        'OpenSSL and ' +
-        'Skia 5.1.2'
+    license = openGeeCommonRpm.license
+    priority = 'optional'
 
     setArch('amd64')	// if this is not specified, the architecture is x86_64, and dpkg won't install without the --force-architecture flag
 
-    requires('alien')
-    requires('autoconf')
-    requires('automake')
-    requires('doxygen')
     requires('flex')
     requires('freeglut3-dev')
     requires('gettext')
@@ -447,9 +431,6 @@ task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
     requires('python-psycopg2')
     requires('python-setuptools')
     requires('python2.7')
-    requires('python-git')
-    requires('shunit2')
-    requires('swig')
 
     // Prefix variable definitions to all install scripts:
     installUtils = file('build/shared/install-utils.sh')
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -79,16 +79,6 @@ def gee_long_version_txt = "${src_dir}${native_dir()}/gee_long_version.txt"
 
 def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
 
-def osId = org.opengee.os.Platform.osId
-
-def packageDep
-
-if (osId == "ubuntu") {
-    packageDep = "openGeeDebs"
-} else {
-    packageDep = "openGeeRpms"
-}
-
 def stagedInstall_path =
     project.hasProperty('stageInstallPath') ?
         project.stageInstallPath :
@@ -712,7 +702,7 @@ task openGeeRpms(
 task openGeeDebs(
     dependsOn: ['openGeeServerDeb', 'openGeeFusionDeb', 'openGeeCommonDeb'])
 
-task osPackage(dependsOn: packageDep)
+task osPackage(dependsOn: org.opengee.os.Platform.osId == 'ubuntu' ? openGeeDebs : openGeeRpms)
 
 // Mostly for documentation purposes.  You need to comment out the rest of the
 // Gradle file to run this task:
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -409,8 +409,7 @@ task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
     user = openGeeCommonRpm.user
     permissionGroup = openGeeCommonRpm.permissionGroup
     packageGroup = 'misc'
-    summary = openGeeCommonRpm.summary
-    // The package build failed when packageDescription was copied here from openGeeCommonRpm, so it is omitted for now
+    formatPackageDescription(openGeeCommonRpm.summary + '\n' + openGeeCommonRpm.packageDescription)
     license = openGeeCommonRpm.license
     priority = 'optional'
 
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -1,4 +1,4 @@
-// Copyright 2017 the Open GEE Contributors
+// Copyright 2017, 2018 the Open GEE Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -416,7 +416,7 @@ task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
     setArch('amd64')	// if this is not specified, the architecture is x86_64, and dpkg won't install without the --force-architecture flag
 
     requires('flex')
-    requires('freeglut3-dev')
+    requires('freeglut3')
     requires('gettext')
     requires('libc6')
     requires('libfreetype6')
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -79,6 +79,16 @@ def gee_long_version_txt = "${src_dir}${native_dir()}/gee_long_version.txt"
 
 def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
 
+def osId = org.opengee.os.Platform.osId
+
+def packageDep
+
+if (osId == "ubuntu") {
+    packageDep = "openGeeDebs"
+} else {
+    packageDep = "openGeeRpms"
+}
+
 def stagedInstall_path =
     project.hasProperty('stageInstallPath') ?
         project.stageInstallPath :
@@ -702,7 +712,7 @@ task openGeeRpms(
 task openGeeDebs(
     dependsOn: ['openGeeServerDeb', 'openGeeFusionDeb', 'openGeeCommonDeb'])
 
-task osPackage(dependsOn: ['openGeeRpms', 'openGeeDebs'])
+task osPackage(dependsOn: packageDep)
 
 // Mostly for documentation purposes.  You need to comment out the rest of the
 // Gradle file to run this task:
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -402,16 +402,84 @@ if (project.hasProperty('buildOpenGee')) {
     openGeeCommonRpm.dependsOn stageOpenGeeInstall
 }
 
-
-task openGeeCommonDeb (type: Deb) {
+task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
     packageName = 'opengee-common'
+    release = "1.${rpmPlatformString}"
+    version = openGeeVersion
+    user = 'root'
+    permissionGroup = 'root'
+    packageGroup = 'Application/Productivity'
+    summary = 'Third-party libraries bundled with Open GEE'
+    // The package build failed when packageDescription was copied here from openGeeCommonRpm, so it is omitted for now
+    license =
+        'ASL 2.0 and ' +
+        'BSD and ' +
+        'GPLv2+ and ' +
+        'Psycopg2 GPL 3+ and ' +
+        'LGPLv2 and ' +
+        'MIT and ' +
+        'MPLv1.1 and ' +
+        'PSF 2 and ' +
+        'Python Imaging Library 1.1.7 and ' +
+        'mgrs 1.1.0 and mm 1.4.2 and ' +
+        'OpenLDAP and ' +
+        'OpenSSL and ' +
+        'Skia 5.1.2'
 
-    from(stagedInstallDir_common) {
+    setArch('amd64')	// if this is not specified, the architecture is x86_64, and dpkg won't install without the --force-architecture flag
+
+    requires('alien')
+    requires('autoconf')
+    requires('automake')
+    requires('doxygen')
+    requires('flex')
+    requires('freeglut3-dev')
+    requires('gettext')
+    requires('libc6')
+    requires('libfreetype6')
+    requires('libperl4-corelibs-perl')
+    requires('libpng12-0')
+    requires('libstdc++6')
+    requires('libtool')
+    requires('libxml2-utils')
+    requires('openssl')
+    requires('python-imaging')
+    requires('python-psycopg2')
+    requires('python-setuptools')
+    requires('python2.7')
+    requires('python-git')
+    requires('shunit2')
+    requires('swig')
+
+    // Prefix variable definitions to all install scripts:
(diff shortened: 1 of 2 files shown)


earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,84 @@ if (project.hasProperty('buildOpenGee')) {
...
-task openGeeCommonDeb (type: Deb) {
+task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
     packageName = 'opengee-common'
+    release = "1.${rpmPlatformString}"
+    version = openGeeVersion
+    user = 'root'
+    permissionGroup = 'root'
+    packageGroup = 'Application/Productivity'
+    summary = 'Third-party libraries bundled with Open GEE'
+    // The package build failed when packageDescription was copied here from openGeeCommonRpm, so it is omitted for now
```
tst-ppenev: @mrstephen: Perhaps, the description may not have been formatted properly.  What error did you get?

earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,84 @@ if (project.hasProperty('buildOpenGee')) {
...
+        'Python Imaging Library 1.1.7 and ' +
+        'mgrs 1.1.0 and mm 1.4.2 and ' +
+        'OpenLDAP and ' +
+        'OpenSSL and ' +
+        'Skia 5.1.2'
 
-    from(stagedInstallDir_common) {
+    setArch('amd64')	// if this is not specified, the architecture is x86_64, and dpkg won't install without the --force-architecture flag
+
+    requires('alien')
```
tst-ppenev: The following packages contain build tools, and should not be necessary at runtime: `alien`, `autoconf`, `automake`, `doxygen`, `python-git`, `shunit2`, `swig`.

earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,84 @@ if (project.hasProperty('buildOpenGee')) {
...
+task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
     packageName = 'opengee-common'
+    release = "1.${rpmPlatformString}"
+    version = openGeeVersion
+    user = 'root'
+    permissionGroup = 'root'
+    packageGroup = 'Application/Productivity'
+    summary = 'Third-party libraries bundled with Open GEE'
+    // The package build failed when packageDescription was copied here from openGeeCommonRpm, so it is omitted for now
+    license =
```
tst-ppenev: Can you please share values between RPM and Deb packages?  You could access the RPM task's properties, e.g. `openGeeCommonRpm.license`, or, if you prefer, you can make a class to hold the various properties for a package, so it can be shared between RPM and Deb packages.

earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,84 @@ if (project.hasProperty('buildOpenGee')) {
...
 
-
-task openGeeCommonDeb (type: Deb) {
+task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
     packageName = 'opengee-common'
+    release = "1.${rpmPlatformString}"
+    version = openGeeVersion
+    user = 'root'
+    permissionGroup = 'root'
+    packageGroup = 'Application/Productivity'
```
tst-ppenev: If this sets the `section` field of the Debian package, it should probably have a value of `misc`.  I don't think there's a section called `Application/Productivity` in Debian.  You can have a look at the list of defined sections at https://www.debian.org/doc/debian-policy/#sections.

earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,84 @@ if (project.hasProperty('buildOpenGee')) {
     openGeeCommonRpm.dependsOn stageOpenGeeInstall
 }
 
-
-task openGeeCommonDeb (type: Deb) {
+task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
```
tst-ppenev: I think you should also add `priority = 'optional'` to the package fields.

earth_enterprise/rpms/build.gradle
mrstephen: build failure

This is the error I see. I had to track it to the packageDescription property by trial and error. 
As for the format, I copied and pasted it directly from the `openGeeCommonRpm` task, so I was a little surprised it didn't build, because `openGeeCommonRpm` builds with no problem. 

I tried formatting it more like the the `license` property (each line enclosed in single quotes and joined by a +), but that didn't help. I also tried just setting it to `openGeeCommonRpm.packageDescription`, but that similarly failed. 

I built it with `scons -j8 release=1 package_install`. Is there a flag I can pass into that call to produce more detailed error messages?

earth_enterprise/rpms/build.gradle
mrstephen: Thanks. I'll remove those dependencies.

mrstephen:
@tst-ppenev I made the edits that you recommended, but I'm still stymied by the issue with the `packageDescription` property.

I tried two additonal formats for the string, but neither would work:
1) I formatted it similar to how the `license` property is formatted (each line surrounded by quotes, concatenated by a '+')
2) Instead of having it as a multi-line string, I made it a single line with '\r\n' inserted where the line breaks should be. That didn't work either.

I was able to set `packageDescription` to 'test' with no issues, so it will accept a value. I'm just not sure what the issue is with a longer string.

earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,65 @@ if (project.hasProperty('buildOpenGee')) {
...
+    version = openGeeVersion
+    user = openGeeCommonRpm.user
+    permissionGroup = openGeeCommonRpm.permissionGroup
+    packageGroup = 'misc'
+    summary = openGeeCommonRpm.summary
+    // The package build failed when packageDescription was copied here from openGeeCommonRpm, so it is omitted for now
+    license = openGeeCommonRpm.license
+    priority = 'optional'
+
+    setArch('amd64')	// if this is not specified, the architecture is x86_64, and dpkg won't install without the --force-architecture flag
```
tst-ppenev: Yes, the correct value is `amd64`.

There's no `x86_64` architecture string in Debian packages.  According to the Debian Policy Manual (https://www.debian.org/doc/debian-policy/#architecture-specification-strings), you can get the list of recognized architecture strings with `dpkg-architecture -L`.

tst-ppenev:
@mrstephen: The required format for the `Description` field is described at https://www.debian.org/doc/debian-policy/#description.

I presume that the OS Package plug-in prepends the `summary` field to the `packageDescription` field.  Although, that may not be the case.  You can see what ends up in the package meta-data after building.

Basically, you want a description that has a first line containing a summary of up to, maybe, 60-something to 70 characters.  All following lines need to be indented by a space, and blank-lines should be represented as an indented `.`.

If that fixes the problem, I suggest adding a field to the `GeeDeb` class that would take a string, process it to conform to the required `Description` field formatting and set `packageDescription` to the result.

mrstephen:
@tst-ppenev Please see the updated code. I added a method to GeeDeb.groovy called `formatPackageDescription` that takes a String, formats it properly, and assigns it to `packageDescription`.
I haven't written groovy code before, so if that method could be more efficient, let me know.

This builds and installs properly, and the Description seems to display properly in the package meta-data.

Thanks for all of your input, by the way. You've been extremely helpful :)

earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,64 @@ if (project.hasProperty('buildOpenGee')) {
     openGeeCommonRpm.dependsOn stageOpenGeeInstall
 }
 
+task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeSharedFiles) {
```
tst-ccamp: Can you update the copyright year in the header to 2017, 2018?

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeDeb.groovy
```diff
@@ -82,4 +82,19 @@ class GeeDeb extends com.netflix.gradle.plugins.deb.Deb {
 
         findRequires(packageInputFiles).each { requires(it) }
     }
-}
\ No newline at end of file
+
```
tst-ccamp: Please add a copyright header (copied from build.gradle should be fine)

mrstephen: Sure thing. Just the part from the first line down to the line "limitations under the License"?

earth_enterprise/rpms/build.gradle
mrstephen: Sure. Should it be "2017, 2018" or "2017-2018"? Or does it matter either way?

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeDeb.groovy
tst-ccamp: Yep, you got it!

mrstephen: Good, because I went ahead and committed it earlier ;)

earth_enterprise/rpms/build.gradle
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 the Open GEE Contributors
+// Copyright 2017, 2018 the Open GEE Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
```
tst-ccamp: I ran into an issue where it was trying to run rpmrebuild on the ubuntu platform. But, I got around it by only having it build deb packages on Ubuntu. I already had this code ready, but not checked in yet. I think it would be a good idea to go ahead and add that also. 

This would go right after `def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString`
```
def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
 
def osId = org.opengee.os.Platform.osId

def packageDep

if (osId == "ubuntu") {
    packageDep = "openGeeDebs"
} else {
    packageDep = "openGeeRpms"
}
```

And replace the line `task osPackage(dependsOn: ['openGeeRpms', 'openGeeDebs'])` with `task osPackage(dependsOn: packageDep)`

earth_enterprise/rpms/build.gradle
```diff
@@ -79,6 +79,16 @@ def gee_long_version_txt = "${src_dir}${native_dir()}/gee_long_version.txt"
 
 def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
 
+def osId = org.opengee.os.Platform.osId
+
+def packageDep
```
tst-ppenev: Minor: I don't think we need a new variable just to change one dependency.  It could just be:

```Groovy
task osPackage(dependsOn: org.opengee.os.Platform.osId == "ubuntu" ?  "openGeeDebs" : "openGeeRpms")
```

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeDeb.groovy
```diff
@@ -82,4 +96,19 @@ class GeeDeb extends com.netflix.gradle.plugins.deb.Deb {
...
+    protected void formatPackageDescription(String descrip){
```
tst-ppenev: Minor: I was thinking of using a field, but this is acceptable.  E.g., you could define:

```Groovy
void setFreeformDescription(String value)
```

so you can use it like this:

```Groovy
task(type: GeeDeb) {
    freeformDescription = 'Some\nMultiline description.'
}
```

tst-ppenev: Minor: I don't think this should be `protected` it's intended for public use.

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeDeb.groovy
```diff
@@ -82,4 +96,19 @@ class GeeDeb extends com.netflix.gradle.plugins.deb.Deb {
...
+        packageDescription = ''
+        List lines = descrip.split( '\n' )
+        lines.eachWithIndex{ item, index ->
```
tst-ppenev: Minor: I think you could do with:

```Groovy
descrip.eachLine { line ->
    // . . .
}
```

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeDeb.groovy
mrstephen: Interesting. I didn't know fields were assigned in that manner. So it would automatically invoke `setFreeformDescription` when it sees `freeformDescription = 'Some\nMultiline description.'`? 
The only counter-point I would make is that making it a property seems like it involves an extra step later (`packageDescription = freeformDescription`), unless I'm missing something. But maybe using fields is the more Groovy-esque way to do things. Would you prefer that I make it a field? I'll defer to your judgement if you have a preference.

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeDeb.groovy
mrstephen: Very handy. I didn't know about eachLine. Thanks.

earth_enterprise/rpms/buildSrc/src/main/groovy/GeeDeb.groovy
```diff
@@ -82,4 +96,24 @@ class GeeDeb extends com.netflix.gradle.plugins.deb.Deb {
 
         findRequires(packageInputFiles).each { requires(it) }
     }
-}
\ No newline at end of file
+
+    // Formats a multi-line description to be compatible with a Debian build
+    void formatPackageDescription(String descrip){
```
tst-ppenev: @mrstephen: I'm commenting on the latest pull request commit, because the earlier comments became outdated, are would be harder to find.

Yes, in Groovy, basically, all expressions to access an object's member variable are implicit calls to getters and setters.  You can have a look at http://groovy-lang.org/style-guide.html#_getters_and_setters.  It uses the same accessor convention used in Java for accessing Java bean properties.  The getter for `propertyName` is the method called `getPropertyName` and the setter is called `setPropertyName`.  (If the property is a boolean, you can use `isProperty` instead of `getProperty`.)

I was thinking that it would be more in line with configuring Gradle tasks to use a property, like having another version of `packageDescription`.  However, this is a stylistic choice, so I'll approve even if it's implemented with a method that's not an setter.

earth_enterprise/rpms/build.gradle
```diff
@@ -402,16 +402,64 @@ if (project.hasProperty('buildOpenGee')) {
...
+    permissionGroup = openGeeCommonRpm.permissionGroup
+    packageGroup = 'misc'
+    formatPackageDescription(openGeeCommonRpm.summary + '\n' + openGeeCommonRpm.packageDescription)
+    license = openGeeCommonRpm.license
+    priority = 'optional'
+
+    setArch('amd64')	// if this is not specified, the architecture is x86_64, and dpkg won't install without the --force-architecture flag
+
+    requires('flex')
+    requires('freeglut3-dev')
```
tst-ccamp: Do you know if this is a runtime requirement? `freeglut3` without "-dev" might work fine. We can either test that now or add it to a follow-up ticket

mrstephen: By "test that now", do you mean just the build and install? I assume there's nothing in opengee-common that we can test by executing, correct?

mrstephen: It builds and installs fine with `freeglut3` instead of `freeglut3-dev`. I guess that means it'll be fine, but I don't know. Do you want me to add that to this pull request?

tst-ccamp: That's true, we won't be able to truly test until we do the follow-up. For now, I say let's go with the non-dev version. Fewer packages is better, and `freeglut3-dev` installs about 5 more packages than just plain `freeglut3`.

mrstephen: Sounds like a plan. I'll get it added shortly.

PR #796, opened 2018-04-05
Remove field from query

tst-eclamar:


earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
```diff
@@ -653,8 +653,7 @@ def _QueryPublishContextByTargetPath(self, target_path):
     query_string = ("""SELECT publish_context_table.snippets_set_name,
            publish_context_table.search_def_names,
            publish_context_table.supplemental_search_def_names,
-           publish_context_table.poifederated, 
-           publish_context_table.ec_default_db
```
tst-nfarah: just to understand the cause-result of that line: so passing `publish_context_table.ec_default_db` into the query was causing the query to fail and later the re-publishing to fail ? or how that line caused the republishing to fail ?

PR #805, opened 2018-04-12
Changed SConscript in gtest to not build gtest if in release. 

calhoojp:
This appears to remove the dependency from the rpms created. Only tested on CentOS 7. I was also able to build internal/release and build the rpms. rpms were tested on a different CentOS VM without gtest installed. Both /etc/init.d/geserver restart and /opt/google/bin/geserveradmin ran without failing to link gtest. I tried other approaches of removing gtest from libs in different ways but this seemed to work best.

### DIFF
--- a/earth_enterprise/src/third_party/gtest/SConscript
+++ b/earth_enterprise/src/third_party/gtest/SConscript
@@ -22,10 +22,12 @@ buildGtest = False
 
 # Check environment for libgtest.so
 conf = Configure(env)
-if not conf.CheckLib('gtest'):
-    buildGtest = True
-
 env = conf.Finish()
+
+# If building release, don't include gtest so scons won't add it as a dependency.
+if not env['release']:
+    if not conf.CheckLib('gtest'):
+        buildGtest = True
 
 # Build libgtest as a shared library. If we make gtest static it forces all
 # executables linked against it to link statically and that messes up the


calhoojp:
I signed it!

earth_enterprise/src/third_party/gtest/SConscript
```diff
@@ -22,11 +22,13 @@ buildGtest = False
...
 # Check environment for libgtest.so
 conf = Configure(env)
-if not conf.CheckLib('gtest'):
-    buildGtest = True
-
 env = conf.Finish()
 
+# If building release, don't include gtest so scons won't add it as a dependency.
+if not env['release']:
+    if not conf.CheckLib('gtest'):
```
tst-ccamp: I'll check it tomorrow, but this solution may only work for RHEL/CentOS. In Ubuntu, it never builds GTest, but still links it.

earth_enterprise/src/third_party/gtest/SConscript
```diff
@@ -22,11 +22,13 @@ buildGtest = False
 
 # Check environment for libgtest.so
 conf = Configure(env)
-if not conf.CheckLib('gtest'):
-    buildGtest = True
-
 env = conf.Finish()
 
+# If building release, don't include gtest so scons won't add it as a dependency.
+if not env['release']:
```
tst-ppenev: Style: This code could be just:

```Python
buildGtest = not env['release'] and not conf.CheckLib('gtest')
```

earth_enterprise/src/third_party/gtest/SConscript
```diff
@@ -22,11 +22,13 @@ buildGtest = False
...
+    if not conf.CheckLib('gtest'):
+        buildGtest = True
+
 # Build libgtest as a shared library. If we make gtest static it forces all
```
tst-ppenev: I don't think you should need to build a library that's installed by a package.  The `gtest` package on Red Hat platforms already contains the built shared library.  Unfortunately, I think there isn't such a package on Ubuntu.

I'd, personally, rather build a Ubuntu package with the shared library like we did for RHEL 6, rather than have this SCons script.

Q: Are we packaging binaries in the RPMs that link against GoogleTest?  Do you know where the dependency comes from?

tst-ppenev: Thanks, @calhoojp. I was wondering which binaries those were.  I could only imagine unit test executables being linked against `gtest`.

calhoojp: That is all I can imagine, also. Which obviously doesn't make sense and isn't what we would want to be happening. Scons is completely new to me, so I'm not sure what exactly is causing this to happen. But I can do some more investigating and try to figure out what, exactly, is being linked.

tst-ppenev:
I'll add executables I found linking against GTest to the issue #651.

PR #811, opened 2018-04-16
Resolve host name from ServerName or FQDN instead of using Listen command in gehttpd.con

### LINKED ISSUE #810: When GEE Server listen to specific ip address the returned default local virtual host is an ip address instead of the hostname
- Using sudo, edit `opt/google/gehttpd/conf/gehttpd.conf` (ex. `sudo nano /opt/google/gehttpd/conf/gehttpd.conf`) and change `Listen 80` to `Listen 127.0.0.1:80`
- Restart GEE Server: `sudo /etc/init.d/geserver restart`
- List Virtual hosts: `/opt/google/bin/geserveradmin --listvhs`
This returns:
1. local, http://localhost/local_host
2. public, http://127.0.0.1/public_host
3. secure, http://127.0.0.1/secure_host

- Expected returned hosts should be:
1. local, http://localhost/local_host
2. public, http://MyHostName/public_host
3. secure, http://MyHostName/secure_host

where `MyHostName` is the host name where GEE Server is installed and running.

tst-nfarah:
Fixes #810

When specific ip address is being listened to current logic of getting host name of the local virtual host use s that ip address instead of resolving from the machine. 

Verification Steps:
- Using sudo, edit opt/google/gehttpd/conf/gehttpd.conf (ex. `sudo nano /opt/google/gehttpd/conf/gehttpd.conf`) and change `Listen 80` to `Listen 127.0.0.1:80`

- Restart GEE Server: `sudo /etc/init.d/geserver restart`

- List  Virtual hosts: `/opt/google/bin/geserveradmin --listvhs`

- Verify the returned list
local, http://localhost/local_host
public, http://MyHostName/public_host
secure, http://MyHostName/secure_host

where MyHostName is the host name where GEE Server is installed and running.

### DIFF
--- a/earth_enterprise/src/server/wsgi/common/utils.py
+++ b/earth_enterprise/src/server/wsgi/common/utils.py
@@ -155,11 +155,11 @@ def BuildServerUrl(host_):
     server URL in format scheme://host_[:port]
   """
   assert host_
-  scheme_host_port = GetApacheSchemeHostPortFromListen()
-  if not scheme_host_port:
+  scheme_port = GetApacheSchemePortFromListen()
+  if not scheme_port:
     return None
 
-  (scheme, port) = scheme_host_port[0], scheme_host_port[2]
+  (scheme, port) = scheme_port[0], scheme_port[1]
   assert scheme
   assert port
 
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/server/wsgi/common/utils.py
```diff
@@ -216,6 +216,7 @@ def GetApacheSchemeHostPortFromListen():
     assert port
     if not scheme:
       scheme = "https" if port == "443" else "http"
+    host = None
```
tst-dsugar: This is about the fact that utils.py parses the apache config file itself directly, and will use the ip address of an apache bind as the "host name" for other opengee operations if an ip address is used to bind the apache listen to a specific interface.  If something like "Listen 80" is used, then it just returns the default system hostname.  If you have a bind like "Listen 127.0.0.1:80", then it actually makes 127.0.0.1 a hostname in other operations, such as the list of hosts returned by geserveradmin --listvhs.  My own impression is that this is not desired behavior, and this change makes it always fall thru and use the system hostname regardless of what was in the apache Listen directive.

tst-rwildes: It seems this function is doing exactly what it was documented to do and what its name implies that it will do.  Is the bug with this function or is the bug really with the calling code for incorrectly using the host returned from this function?  If we are always going to return `None` for host then maybe rename this function and change the documentation to state it doesn't return a host and just have it return a tuple with two values (scheme and port) instead of three.

tst-rwildes:
Code changes all look good.  Building/testing code now.

tst-twilk:
Looks good. Thanks!

PR #815, opened 2018-04-18
Issue 791 clean up javascript

tst-kfreiheit:
Removing obsolete files from maps api and updating example code as needed.


tst-lsavoie:
I'm OK with my commits being included on this PR.

earth_enterprise/src/maps/bootstrap.js
```diff
@@ -1,32 +0,0 @@
-/*
```
tst-lsavoie: I've seen web apps that use these two boostrap files. Are they causing problems? If not, it might be wise to keep them.

tst-kfreiheit: I see the bigger picture now: these are easy entry points for users who want to override the api version with a specific build's subfolder.  I'll add them back.

tst-kfreiheit: ...and after discussing it with you a bit more, we concluded that it's best to check this PR in as-is, and potentially re-add the versioned API bootstrapper(s) when the alternate versions are re-introduced to the codebase.

PR #818, opened 2018-04-19
update travis to add cpp98 build

tst-rwildes:
Just testing adding a C++98 build check to travis builder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,30 +15,25 @@ cache:
   directories:
     - $HOME/cache
 jobs:
+  env:
+    - CPP_STD=98
+    - CPP_STD=11
   include:
-    - stage: build third party
-      script:
-        - cd earth_enterprise/src
-        - if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then tar xf $HOME/cache/third_party$CPP_STD.tgz; fi
-        - scons -j3 release=1 cpp_standard=gnu++$CPP_STD third_party > build.log
-        - mkdir -p $HOME/cache
-        - tar cfz $HOME/cache/third_party$CPP_STD.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
-      - env:
-        - CPP_STD=98
-      - env:
-        - CPP_STD=11
-    - stage: build fusion and test
-      script:
-        - cd earth_enterprise/src
-        - if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then tar xf $HOME/cache/third_party$CPP_STD.tgz; fi
-        - cd ..
-        - scons -j3 release=1 cpp_standard=gnu++$CPP_STD build > build.log
-        - cd src/NATIVE-REL-x86_64/bin/tests
-        - travis_wait 20 ./RunAllTests.pl
-      - env:
-        - CPP_STD=98
-      - env:
-        - CPP_STD=11
+  - stage: build third party
+    script:
+      - cd earth_enterprise/src
+      - if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then tar xf $HOME/cache/third_party$CPP_STD.tgz; fi
+      - scons -j3 release=1 cpp_standard=gnu++$CPP_STD third_party > build.log
+      - mkdir -p $HOME/cache
+      - tar cfz $HOME/cache/third_party$CPP_STD.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
+  - stage: build fusion and test
+    script:
+      - cd earth_enterprise/src
+      - if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then tar xf $HOME/cache/third_party$CPP_STD.tgz; fi
+      - cd ..
+      - scons -j3 release=1 cpp_standard=gnu++$CPP_STD build > build.log
+      - cd src/NATIVE-REL-x86_64/bin/tests
+      - travis_wait 20 ./RunAllTests.pl
 
 notifications:
   slack: opengee:YjzzzTnc7pgITPsrQPATIGwc
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ jobs:
       - ls $HOME/btemp
   - stage: build fusion and test
     script:
-      - mkdir -p $HOME/btemp/bfat98
+      - mkdir -p $HOME/btemp/bfat11
       - cd earth_enterprise/src
       - if [ -f $HOME/cache/third_party11.tgz ]; then tar xf $HOME/cache/third_party11.tgz; fi
       - cd ..
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,37 +14,24 @@ cache:
   timeout: 600
   directories:
     - $HOME/cache
-
+env:
+  - CPP_STD=98
+  - CPP_STD=11
 jobs:
   include:
     - stage: build third party
       script:
         - cd earth_enterprise/src
-        - if [ -f $HOME/cache/third_party.tgz ]; then tar xf $HOME/cache/third_party.tgz; fi
-        - scons -j3 release=1 third_party > build.log
+        - if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then tar xf $HOME/cache/third_party$CPP_STD.tgz; fi
+        - scons -j3 release=1 cpp_standard=gnu++$CPP_STD third_party > build.log
         - mkdir -p $HOME/cache
-        - tar cfz $HOME/cache/third_party.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
+        - tar cfz $HOME/cache/third_party$CPP_STD.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
     - stage: build fusion and test
       script:
         - cd earth_enterprise/src
-        - if [ -f $HOME/cache/third_party.tgz ]; then tar xf $HOME/cache/third_party.tgz; fi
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -14,24 +14,29 @@ cache:
...
         - cd earth_enterprise/src
-        - if [ -f $HOME/cache/third_party.tgz ]; then tar xf $HOME/cache/third_party.tgz; fi
-        - scons -j3 release=1 third_party > build.log
+        - if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then tar xf $HOME/cache/third_party$CPP_STD.tgz; fi
+        - scons -j3 release=1 cpp_standard=gnu++$CPP_STD third_party > build.log
         - mkdir -p $HOME/cache
-        - tar cfz $HOME/cache/third_party.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
+        - tar cfz $HOME/cache/third_party$CPP_STD.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
+      env:
+        - CPP_STD=98
```
tst-ccamp: This seems to just be setting the CPP_STD variable twice in the same job. Maybe try adding a matrix section above jobs:

```
matrix:
  include:
    -env:
         - CPP_STD=98
    -env:
         - CPP_STD=11
```
Example: https://docs.travis-ci.com/user/languages/cpp/#GCC-on-Linux

tst-ccamp: Another option might be to have 2 env sections under the stage:

```
jobs:
   include:
     - stage: build third party
       script:
         ...
       - env:
         CPP_STD=98
       - env:
         CPP_STD=11
```

tst-rwildes: I'll give that a try.  The documentation doesn't help much with what I want to do.  I think I can get it working the way I want if you duplicate the script code twice and name the stages the same.  I duplicated the stages but called them something different but that made the run take twice as long....  I saw an example where they named the stage the same and liked named stages run in parallel (at least that is what the example said).  But I didn't like that because I wanted to re-use the code in the stages....

.travis.yml
```diff
@@ -14,24 +14,31 @@ cache:
...
       script:
         - cd earth_enterprise/src
-        - if [ -f $HOME/cache/third_party.tgz ]; then tar xf $HOME/cache/third_party.tgz; fi
-        - scons -j3 release=1 third_party > build.log
+        - if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then tar xf $HOME/cache/third_party$CPP_STD.tgz; fi
+        - scons -j3 release=1 cpp_standard=gnu++$CPP_STD third_party > build.log
         - mkdir -p $HOME/cache
-        - tar cfz $HOME/cache/third_party.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
+        - tar cfz $HOME/cache/third_party$CPP_STD.tgz NATIVE-REL-x86_64 .sconsign.dblite .sconf_temp
+      env:
```
tst-ccamp: This is the problem, anything in a list has to have `- ` in front of it. Otherwise, whatever is last simply replaces what was there before.

tst-ccamp: Try updating it to exactly what I posted: 

```
       - env:
         CPP_STD=98
       - env:
         CPP_STD=11
```

tst-rwildes: I personally like the simplicity of JSON.  XML use to be my favorite generic and loosely coupled data exchange format but when I started using JSON I switched to liking JSON better.  Yea I'm not a fan at all of yaml...

.travis.yml
tst-rwildes: Turns out that isn't valid yaml syntax.  I found https://www.json2yaml.com/ so I can do things in JSON and see what the equivalent yaml is.  I think what I want to do can not be done with the new stage syntax.  I basically want to matrix the job object to run with different environments but it looks like that isn't supported...  Maybe if I move the env array up one more level...  If that doesn't work I'm going to just go with repeating the stages...

tst-rwildes:
I don't think Travis builder will let me do what I want to do...  I want two groups of stages running in two different parallel jobs (one for c++98 and one for c++11).  I can get different stages to run in parallel but I think they are running in the same `job`.  This means they are sharing the same virtual environment/workspace/code.  That isn't going to be good because the build output and intermittent files will be stomping on each other.   I could mangle the output directory to include c++ standard.  Change `NATIVE-REL-x86_64` to `NATIVE-REL-x86_64-std98` for example.  Then maybe we could have parallel builds from the same source code building two different ways.  However, I'm not 100% sure our build scripts handle this correctly.  It would be a nice feature of our build system to allow this and also through in platform to the mix.  Then we can have one source tree building different permutations of things in parallel.

tst-rwildes:
Good news!  I was mis-understanding the documentation and it turns out each stage is its own independent environment.  So I can just create a two shell scripts: 1) to build and cache third party and 2) a script to use the cached third party and do the build.  The scripts can take a parameter to know what standard we want to build.  Then create two stages called `build third party` and call the first script with c++98 for one of the stages and c++11 for the other stage.  And then create two more stages both called `build fusion and test` and do the same thing of passing one c++98 and the other C++11.  Then C++11 and C++98 will both build in parallel so our build times will not double.

Now that I understand what needs to be done I think I'll throw this away and do another branch/pr.

tst-rwildes:
Doing one more negative test to be sure travis build fails with C++11 syntax.

tst-ccamp:
I think it's good to keep the PR intact for the history. If you just want to replace the previous commits, you can just do a --force push to this branch with just your new commit

tst-rwildes:
Looks like my test worked as expected and the C++11 build succeeded but the C++98 build failed.  I'll clean things up and push.

scripts/ci/build_fusion_and_test.sh
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the license.
+#
+# This script is used by Travis build stages (see travis.yml at root) that build
+# fuison and run unit tests. Assumes the following environment variables are set:
+# CPP_STD    - Should be set to `98` or `11` depending on the standard to build with
+# BUILD_TYPE - The type of build `release`, `optimize`, or `internal`
+
+case "$CPP_STD" in
+    98|11)
```
tst-dsugar: we may add 14 or 17 one day...this does make it clean and simple to specify which version(s) we will build with.

scripts/ci/build_fusion_and_test.sh
```diff
@@ -0,0 +1,51 @@
...
+
+set -x
+cd earth_enterprise/src
+
+if [ -f $HOME/cache/third_party$CPP_STD.tgz ]; then
+  tar xf $HOME/cache/third_party$CPP_STD.tgz;
+fi
+
+cd ..
+scons -j3 $BUILD_TYPE=1 cpp_standard=gnu++$CPP_STD build > build.log
```
tst-dsugar: How does this effect us if we later migrate to just c++11 vs gnu++11?  Will we always use gnu++11 rather than c++11, even on newest Ubuntu (18.04) and (future) rhel 8?

scripts/ci/build_fusion_and_test.sh
tst-rwildes: Yes we can just update this as we add/remove support for different standards.

scripts/ci/build_fusion_and_test.sh
tst-rwildes: There is a separate issue opened to move us off of gnu.  I figured that effort can then change this.

.travis.yml
```diff
@@ -14,24 +14,28 @@ cache:
...
+      - CPP_STD=98
+      - BUILD_TYPE=release
+    script: scripts/ci/build_third_party.sh
+  - stage: build fusion and test
+    env:
+      - CPP_STD=98
+      - BUILD_TYPE=release
+    script: travis_wait 45 scripts/ci/build_fusion_and_test.sh
+  - stage: build third party
+    env:
```
tst-nfarah: it would be great if we can separate the two build into two ci ... Would that be doable ?
That way if C++98 failed only we would know about it, instead of not knowing why the build failed ...

scripts/ci/build_third_party.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+#
+# Copyright 2017 Open GEE Contributors
```
tst-nfarah: update the copyright year to 2017-2018 ... ?

scripts/ci/build_fusion_and_test.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+#
+# Copyright 2017 Open GEE Contributors
```
tst-nfarah: update copyright year to 2017-2018 ?

.travis.yml
tst-rwildes: This does that.  The C++11 and C++98 run independently and in parallel of each other.  You can tell if it is the C++11 or the C++98 build that failed.  See:

tst-rwildes: The one thing it doesn't do is give you a build matrix in side of the MD file which is what I think you wanted.  The unfortunate thing is with the newer Travis syntax we are using this feature is not yet supported.  We are not using the `matrix` syntax but instead are using the `stage` syntax and I can't find anything that lets you create something like what the `matrix` has for reporting results in a cool looking matrix...  Maybe they will add something where the individual stages can report results.  

Also not that the link above doesn't show the environment variables as I didn't add them until later.  With the environment variables you can more clearly see which stage is running what build.  See later runs for what I'm talking about.

tst-nfarah: Looking at this recent build https://travis-ci.org/google/earthenterprise/builds/369736104 ... the C++ version is listed there.
Ok about not having the `matrix` setup. We can revisit it later.

Issue #817, opened 2018-04-19
Temporary change to work around python certificate validation issues


tst-jhaight:
Some supported platforms utilize python versions which default to enable CA certificate verification ([additional info](https://access.redhat.com/articles/2039753)).  Because of this and the fact that OpenGEE utilizes its own packaged version of openssl, some geserver administrative commands via https console will not work due to certificate verification failure (#731).  Until a better solution is available that doesn't involve multiple trust stores, we will turn off python certificate verification for the ge publish server.

tst-ccamp:
Looks like this one can be closed. There is no additional work planned for this branch

PR #819, opened 2018-04-19
Disabling CA certificate authentication for the ge publish server to …

tst-jhaight:
Addresses #817.  Temporary work around to avoid issues publishing via https admin console.

Tested on ubuntu 16, centos7, and RHEL7 by attempting to publish a database to any virtual host via the https admin console.


earth_enterprise/src/server/wsgi/serve/publish/publish_app.wsgi
```diff
@@ -1 +1,4 @@
+import os
+os.environ['PYTHONHTTPSVERIFY'] = '0'
```
tst-nfarah: may be add a comment explaining this change... ex: this is a temp fix for #issueNumber to disable ...

tst-ccamp:
Tested on Ubuntu 16.04 and it works!

PR #830, opened 2018-04-27
Issue #821: Fusion Maps API infinite loop

### LINKED ISSUE #821: Calling geeInitMap in the Fusion Maps API results in an infinite recursive loop
A simple web page that calls geeInitMap ends up in an infinite loop of recursion, eventually crashing the web site and preventing the map from loading. This issue does not occur if you replace the js and maps folders under /opt/google/gehttpd/htdocs with the ones from 4.3. We need to track down and fix the source of the issue so that clients can reliably use the fusion maps API.

tst-lsavoie:
Fixes #821

Fixes an infinite loop in the fusion maps API when calling geeInitMap. The function tries to load a polygon that does not exist, resulting in a NaN that causes an infinite loop.

Also updates some of the bootstrap files to load jQuery, which is required for the latest version of the Google Maps API.

### DIFF
--- a/earth_enterprise/src/maps/api/bootstrap.js
+++ b/earth_enterprise/src/maps/api/bootstrap.js
@@ -32,6 +32,7 @@ google.maps = google.maps || {};
 
   var loadScriptTime = (new Date).getTime();
 
+  getScript(GEE_BASE_URL + '/js/jquery.js');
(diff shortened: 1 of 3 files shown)


tst-kfreiheit:
Pulled code to my local server to test.  All 4 example_local*.html pages pass -- no longer seeing the recursion issue.

Tested with the tutorial data: SF_2d_Merc map.

PR #832, opened 2018-04-28
Issue 337: Checks GCC version and exits build if version < 4.8.0

erikmillergalow:
Makes sure that GCC is at least version 4.8.0.  If the script cannot determine the version it emits a warning but continues.

### DIFF
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -479,21 +479,22 @@ env = khEnvironment(
     python_major_version=python_major_version
 )
 
-# Check for GCC from an alternative toolchain installation on RHEL 6:
-if cc_dir is None and os.path.isfile('/opt/rh/devtoolset-2/root/usr/bin/g++'):
-    cc_dir = '/opt/rh/devtoolset-2/root/usr/bin'
-    rpathlink_dirs += [
+# Check if the version of the default compiler is at least 4.8:
+version = get_cc_version(env['CXX'])
+if not is_version_ge(version, [4, 8]):
+    # Check for GCC 4.8 from an alternative toolchain installation on RHEL 6:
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/SConstruct
```diff
@@ -479,15 +479,22 @@ env = khEnvironment(
...
 )
 
-# Check for GCC 4.8 from an alternative toolchain installation on RHEL 6:
+# Check for GCC from an alternative toolchain installation on RHEL 6:
 if cc_dir is None and os.path.isfile('/opt/rh/devtoolset-2/root/usr/bin/g++'):
-    # Check if the version of the default compiler is, at least 4.8:
-    if not is_version_ge(get_cc_version(env['CXX']), [4, 8]):
-        cc_dir = '/opt/rh/devtoolset-2/root/usr/bin'
-        rpathlink_dirs += [
+    cc_dir = '/opt/rh/devtoolset-2/root/usr/bin'
```
tst-ppenev: I think we should prefer using the `g++` installation in the platform default location, instead of the `devtoolset` `g++` in the case where the one at the default location is sufficiently new.

erikmillergalow: I'd mistakenly come to the conclusion that the devtoolset compiler was required to build on CentOS6 but looking at it again, I agree. Changed so that it first checks for version 4.8 or greater, then if an outdated compiler is detected (or no version is detected) it looks for the devtoolset version.

tst-ppenev: You do need the `devtoolset` compiler on Cent OS 6, or some other compiler that supports `g++ 4.8` features.  However, this code runs on all platforms, on which the platform compiler can be sufficiently new, but you could also have a `devtoolset` compiler.

PR #834, opened 2018-05-03
PR - #835: Fix for server_url issue in GEE web services

### LINKED ISSUE #835: GEE Server Publish Service gets the Server URL from the client instead of locally
In publish_manager.py, in several places the "request_url" or "server_url" is retrieved from the Request object, which comes from the client. The consequence of this is that the server will try to connect to basically anything that the client tells it to. This can be used in an attack vector to direct the GEE Server to connect to an invalid or even dangerous server.

Example:publish_manager.py line 227:

request_url = request.GetParameter(constants.ORIGIN_REQUEST_HOST)

This is one example. The same error happens throughout publish_manager.py. This ticket is to fix all instances of the problem.

## Acceptance Criteria:

1. Throughout publish_manager.py, the server URL the known correct URL (loaded from a configuration). The URL should not be taken from the client request.

1. If possible, do not add a new configuration value - use the existing Apache configuration or similar.

    * If a new configuration value is added, the installers and the documentation on changing the hostname need to be updated.

1. Survey the code to make sure no other code pulls the server URL from the client request. Write follow up tickets as necessary.

tst-hrajamoney:
Fixes #835 

server_url that was initialized by init is stored in
 os.environ["GESERVER_URL"] = server_url
Used the above, whenever we need to send message back instead of using
request_url from request message.

### DIFF
--- a/earth_enterprise/src/server/wsgi/serve/publish/publish_manager.py
+++ b/earth_enterprise/src/server/wsgi/serve/publish/publish_manager.py
@@ -114,6 +114,7 @@ class PublishManager(object):
       raise exceptions.PublishServeException(
           "Couldn't get self-referential server URL.")
     logger.debug("PublishManager init: server URL %s", server_url)
+    os.environ["GESERVER_URL"] = server_url
     self.__ResetPublishManager(server_url)
 
   def HandlePingRequest(self, request, response):
@@ -154,9 +155,8 @@ class PublishManager(object):
       response: response object.
     """
     logger.debug("HandleResetRequest...")
-    request_url = request.GetParameter(constants.ORIGIN_REQUEST_HOST)
-    assert request_url
-    self.__ResetPublishManager(request_url)
+    geserver_url = os.environ.get("GESERVER_URL")
+    self.__ResetPublishManager(geserver_url)
     http_io.ResponseWriter.AddBodyElement(
         response, constants.HDR_STATUS_CODE, constants.STATUS_SUCCESS)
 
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/server/wsgi/serve/publish/publish_manager.py
```diff
@@ -114,6 +114,7 @@ def __init__(self):
       raise exceptions.PublishServeException(
           "Couldn't get self-referential server URL.")
     logger.debug("PublishManager init: server URL %s", server_url)
+    os.environ["GESERVER_URL"] = server_url
```
tst-ccamp: Is there a way to do this without modifying the os.environ? There should be a better way to share info within the class.

earth_enterprise/src/server/wsgi/serve/publish/publish_manager.py
```diff
@@ -154,9 +155,8 @@ def HandleResetRequest(self, request, response):
       response: response object.
     """
     logger.debug("HandleResetRequest...")
-    request_url = request.GetParameter(constants.ORIGIN_REQUEST_HOST)
-    assert request_url
```
tst-ccamp: We could probably keep the assert commands, just use it for geserver_url instead of request_url

tst-hrajamoney: ok, Will do that.

earth_enterprise/src/server/wsgi/serve/publish/publish_manager.py
tst-hrajamoney: I  thought if we need this information later, outside the class as well , this will be centralized place

tst-ccamp: I don't think it's a good idea to access the OS environment from within a web service. They are meant to be self-contained. We have no real control over how our environment is setup and there's no guarantee for repeatable behavior because we're at the mercy of how the web server wants to behave. 

Looks like this class is already storing member variables. Just a few lines up you have `self._publish_helper = publish_manager_helper.PublishManagerHelper()`. I suggest following that same pattern here.

tst-hrajamoney: ok, I will add a member variable.

PR #853, opened 2018-05-18
fix(build): Create settings.gradle for RPM project

tst-twilk:
GEE uses gradle to build rpms. Mark the root of the project tree so gradle does not traverse up the dir structure in search of settings.gradle.


earth_enterprise/settings.gradle
```diff
@@ -0,0 +1 @@
+
```
tst-rwildes: If we add the file here we should add an include line to point to the RPM folder so that becomes the child folder.

tst-ppenev:
@tst-twilk: Is there an issue for this pull request? I can't tell what the problem is, and this makes me suspect it may not be the best solution.

tst-ppenev:
@tst-twilk: You don't have to create a new pull request, just because you made an incorrect commit.  You can do an interactive rebase, and amend the author.

PR #854, opened 2018-05-18
Update master with latest changes from release_5.2.2

tst-lsavoie:


tst-lsavoie:
All commits have been previously added to this repo, so authors have previously consented.

PR #855, opened 2018-05-18
add settings.gradle to mark the root of the project

tst-twilk:
Add earth_enterprise/rpms/settings.gradle so gradle does not traverse dir structure up looking for settings.gradle.


tst-rwildes:
Looks good to me

PR #862, opened 2018-05-25
refactor(build): support RPM linting during build

tst-dsugar:


tst-ppenev:
Is there a ticket with a description for this pull request?

earth_enterprise/rpmlintrc
```diff
@@ -0,0 +1,41 @@
...
+
+# Disable to review/remediate later...
+addFilter("E: non-standard-executable-perm")
+
+# Disable to ignore for now, often issues from third party code, some is valid though.
+addFilter("^.* /opt/")
+addFilter("^.* /var/opt/")
+
+# Because this is in init.d it gets LOTS of error hits...
+addFilter("E: .* /etc/init.d/gevars.sh")
```
tst-ppenev: Shouldn't we fix this by moving the file, instead of ignoring it?

earth_enterprise/rpmlintrc
```diff
@@ -0,0 +1,41 @@
...
+
+# We are not currently renaming the logrotate file...
+addFilter("E: incoherent-logrotate-file")
+
+# We do not currently do changelogs in rpm specfiles...
+addFilter("E: no-changelogname-tag")
+
+# We should fix this...
+addFilter("E: devel-dependency proj-devel")
```
tst-ppenev: Yeah, why add a filter, if we need to fix it?

earth_enterprise/rpms/build.gradle
```diff
@@ -520,6 +520,20 @@ task openGeeServerRpm (type: GeeRpm, dependsOn: openGeeSharedFiles) {
     directory('/var/opt/google/pgsql', 0700)
     directory('/var/opt/google/pgsql/logs')
 
+    filesMatching(
+        ['/etc/init.d/geserver']
```
tst-ppenev: Can't we fix these permissions in the repository, instead of adding packaging code?

tychosoft: I actually fixed the perms in the repo, but wasn't sure why they were getting the wrong perms/needed to have it being set here in the first place.

earth_enterprise/rpmlintrc
tychosoft: I had wanted to speak to you about this first.  But yes, we should...

tst-dsugar: I wanted to speak with you about this first before moving it.

earth_enterprise/rpmlintrc
tst-dsugar: I know we had problems removing this before....

earth_enterprise/rpms/build.gradle
tst-dsugar: I was not sure why we were setting these perms in the gradle build in the first place, so I wanted to make sure they were at least correct there, and I did already change the perms in the repo.

earth_enterprise/rpmlintrc
tst-ppenev: As far as I'm concerned, I think we should get this file out of `/etc/init.d`.

Actually, it seems that this file currently contains variables used in Fusion, as well as Server.  We could split it, and package each part with the service it's part of. If we don't want to fix everything it once, perhaps, under `/etc/opt/google/opengee-common` would be good place.

earth_enterprise/rpmlintrc
tst-ppenev: OK.  We could use a ticket for fixing that, then.

It may still be useful to see the errors that we need to fix.

earth_enterprise/rpms/build.gradle
tst-ppenev: OK.  I'm not sure what the scope of this pull request is.  If verifying that this code is no longer needed, and removing it is not in scope, it could be done on another ticket.

earth_enterprise/rpmlintrc
tst-dsugar: The purpose was to be able to use lint to verify no "new" errors are being introduced with future builds, hence why they are filtered, and any error halts the target.  The important thing to note, yes, is that those issues that are not being remediated now are documented and should be addressed in separate issues.  Some potential remediations I was cautious about because I did not want to break builds, and I thought they would be better addressed as stand-alone issues on their own.

tst-eclamar:
Had to add one filter to get the linting to complete.

On a side note, the ownerships for /opt/google/gehttpd/logs/{geresourceprovider.log,gesystemmanger.log} are incorrect: they where 'geapacheuser:490' upon initial inspection. Changing them to 'geapacheuser:gegroup' allowed /etc/init.d/gefusion to start.

earth_enterprise/rpmlintrc
tst-ppenev: That makes perfect sense.  Writing tickets to fix these is fine.

earth_enterprise/rpmlintrc
tst-ppenev: As you explain below, this would be outside the scope of this pull request, so fixing this would be in a separate ticket.

tst-dsugar: We can introduce a separate issue for that.  I did briefly outline the goals and purpose of this branch in a github ticket, #866

PR #863, opened 2018-05-28
fix(fusion): Check for NULL return from NewPYRTexture in NewTextureFromPath (#342)

### LINKED ISSUE #342: gefusion crash when opening non supported type of file
screenshot from 2017-07-14 14-15-56

BCLaird:
Check for NULL return from NewPYRTexture in NewTextureFromPath.
Reproduction steps:  The tutorial fies can be used to reproduce the crash.  In the Fusion UI, click the "Open File in Preview" button on the toolbar.  Open the file from the tutorial at /gevol/assets/Resources/Terrain/SFTerrain.ktasset/product.kta/ver001/preview.png.aux.xml.  Before the fix, gefusion would crash.  After the fix, the UI reports "File has unknown problems" and gefusion does not crash.
Tested fix on Ubuntu 16.04.
Fixes #342

### DIFF
--- a/earth_enterprise/src/fusion/gst/gstTextureManager.cpp
+++ b/earth_enterprise/src/fusion/gst/gstTextureManager.cpp
@@ -261,11 +261,15 @@ gstTextureGuard gstTextureManager::NewTextureFromPath(
       std::string str = path;
       str += "/" + kHeaderXmlFile;
       gstTextureGuard texture = NewPYRTexture(str, std::string());
-      *is_mercator_imagery = texture->IsMercatorImagery();
+      if (texture) {
+        *is_mercator_imagery = texture->IsMercatorImagery();
+      }
       return texture;
     } else if (base.endsWith(".xml")) {
       gstTextureGuard texture = NewPYRTexture(path, std::string());
-      *is_mercator_imagery = texture->IsMercatorImagery();
+      if (texture) {
+        *is_mercator_imagery = texture->IsMercatorImagery();
+      }
       return texture;
     }
   }


tst-rwildes:
Testing with centos 7 STIG machine.

tst-rwildes:
I tested this  with centos 7 STIG machine and it does fix the issue.

tst-ccamp:
The code looks fine. I'm just wondering if you checked out why NewPYRTexture might be returning NULL?

BCLaird:
Regarding "why NewPYRTexture might be returning NULL":

NewPYRTexture tries to construct a gstPYRTexture for the file being opened (preview.png.aux.xml).  The constructor calls khRasterProduct::Open() which looks for a header.xml file in the same dir as preview.png.aux.xml.  The header.xml file does not exist.  That results in the gstPYRTexture constructor throwing khException and ultimately results in the NULL texture in NewTextureFromPath().

Should there be a header.xml file in that directory?

tst-ccamp:
I looked at some of the assets I have and they do have a header.xml file, but I'm not sure which directory it's trying to open: `./Blue_Marble2.kiasset/product.kia/ver001/raster.kip/header.xml`. Maybe that's worth taking a second look at.

BCLaird:
Ok.  Continuing investigation.

BCLaird:
The file I used to reproduce the issue was at "./SFTerrain.ktasset/product.kta/ver001/preview.png.aux.xml".  The khRasterProduct::Open() code is looking for a header.xml file in the same dir as preview.png.aux.xml.  There actually is a header.xml one dir down in the raster.ktp dir.  Looking at all the header.xml files in my assets they are all under a dir named raster.k?p, product.k?p, or mask.k?p.

When trying to preview a file in Fusion, by default it lists all files.  When you drop down the list of file types, XML files are not listed which I think reinforces this issue's original title that said the crash happened when trying to preview an unsupported type of file.

Note that there is debug output when fusion is unable to open the file.  It reads, "Fusion Warning: Unable to open "/gevol/assets/Resources/Terrain/SFTerrain.ktasset/product.kta/ver001/preview.png.aux.xml".

I don't think there is anything additional needed for this issue, unless we want additional debug output when it doesn't find the header.xml file.  If so, maybe that should be a separate issue.

tst-ccamp:
OK, I looked at it again and it looks like it's able to successfully open the file if it's a header.xml file with corresponding *.pyr files, so I think this works. Thanks for taking a second look.

Issue #871, opened 2018-05-31
Switch code to connect to PostgreSQL via unix domain sockets 


tst-rwildes:
This is needed to allow for tighter security more easily.  Doing this change the GEE admin can change pg_hba.conf to disallow all connections but local connections and can then change local connection to from `trust` to `peer` to ensure underprivileged users can not connect to PostgreSQL.

PR #872, opened 2018-05-31
switch to using unix domain sockets for PostgreSQL connections

tst-rwildes:
fixes issue #871 

To verify all connections are no using Unix domain sockets connect to postgresql and run this query:

`SELECT client_addr, client_port, query FROM pg_stat_activity;`

All client_addr should be blank and client_port should be -1.  The connection you are using will also show with the above query.  If you didn't connect without using a host parameter then your connection might not show this.  This is okay.

To connect to postgresql:
`sudo su - gepguser`
`psql -U gepguser -d postgres -h /tmp`


tst-nfarah:
i'm testing this on Ubuntu. Let's not merge it yet.

Issue #869, opened 2018-05-30
Support WMS in portable


tst-lsavoie:
2D maps can be published with WMS support in GEE server. We should try to use the same code, if possible, to provide WMS support in portable.

PR #876, opened 2018-06-01
WIP: Initial code for portable WMS

### LINKED ISSUE #118: Remove obsolete pywms
From discussion at #115, we should remove the old pywms scripts.

> The src/server/pywms is obsolete. The wms-implementation is in the src/server/wsgi/wms.
> 
> The src/server/pywms dir could be deleted.

tst-lsavoie:
This adds initial support for WMS to portable. The changes include the following:

- The existing WMS code is included in the portable bundle
- A WMS endpoint is added to the portable server
- The existing WMS code is called from the WMS endpoint
- Removes the unused `pywms` folder

Note that this PR only includes initial support for WMS - the portable server will respond to WMS requests but it may only return errors. Future PRs will fix the WMS code to work fully with portable.

To test these changes:

1. Build, install, and run portable using the instructions here: https://github.com/google/earthenterprise/wiki/Portable-Server
2. In a browser, go to http://localhost:9335/wms/anything. It should return an error message indicating that an invalid WMS request was received.
3. If you submit a valid WMS query like http://localhost:9335/wms?version=1.3.0&request=GetCapabilities&service=WMS you should get a valid WMS response but it will likely indicate there was an error.
4. Since changes were made to the WMS code, you should also run a basic WMS test using the standard server.

This code partially completes issue #869
Fixes #118


tst-lsavoie:
I've added the ticket. As for verification, this PR is not expected to fully support WMS - it sets up the infrastructure but some additional changes to the code are probably required for the WMS code to work with portable. Thus, at this point we don't expect this code to work with QGIS. We'll want to test with something like QGIS when the feature is complete, of course.

tst-nfarah:
Ah yes, the ticket was there, just i did not notice it. Ok for this being initial building block for later full WMS support.

tst-ccamp:
Due to no activity on the PR, I'm going to close it. You are welcome to open a new one when it becomes a higher priority

Issue #897, opened 2018-06-23
C++98 should no longer be supported for 5.2.4


tst-rwildes:
After 5.2.3 C++98 will no longer be supported.  Need to update the Travis builder to remove building C++ code using C++98 flag.

PR #898, opened 2018-06-23
remove support for C++98

tst-rwildes:
Address issue #897 as of 5.2.4 C++98 will no longer be supported.


.travis.yml
```diff
@@ -16,16 +16,6 @@ cache:
     - $HOME/cache
 jobs:
   include:
-  - stage: build third party
```
tst-dsugar: indeed once we actually start actively using c++11 we cannot go back...

scripts/ci/build_fusion_and_test.sh
```diff
@@ -21,7 +21,7 @@
 # BUILD_TYPE - The type of build `release`, `optimize`, or `internal`
 
 case "$CPP_STD" in
-    98|11)
+    11)
```
tst-dsugar: I wonder if this should eventually include 14/17 too...

tst-rwildes: Yes that would be a good idea.  It would make moving to those feature standards easier.  Would also want to add stages to make sure new code stays comparable with those standards.

tst-rwildes: May also want to add stages for debug builds too.  Right now we only build release.

tst-rwildes:
I will need to wait for the PR that has the draft 5.2.4 review notes to be merged before I can add this to those notes.

scripts/ci/build_fusion_and_test.sh
tst-rwildes: @tst-nfarah I updated the release notes

tst-rwildes: you can ignore the two 7160003.html files.  They will go away as soon as master is updated with two other PRs.  I'll make sure that those PRs are merged first.

Issue #902, opened 2018-06-25
POI search is not working in 5.2.3 if PostgreSQL tcp listener is disabled


tst-twilk:
It appears that the following PR https://github.com/google/earthenterprise/pull/872 changed how we connect to PostgreSQL but there are still a few places in the code that still use loopback address instead of UNIX domain sockets. Here are a few examples:

- earthenterprise/earth_enterprise/src/server/wsgi/search/conf/PoiSearch.conf
- earthenterprise/earth_enterprise/src/server/wsgi/search/conf/GEPlacesSearch.conf
- earthenterprise/earth_enterprise/src/server/wsgi/search/conf/ExampleSearch.conf

It would be nice to update these files as well for consistency.

PR #905, opened 2018-06-26
Gee 902 consistently use Unix sockets

tst-rwildes:
fixes issue #902


docs/geedocs/5.2.3/answer/7160003.html
```diff
@@ -42,12 +42,19 @@ <h5>
...
   </h5>
     <p><strong>C++11 support</strong>. With release 5.2.3 Open GEE C++ code now supports and defaults to compiling using the <code>-std=gnu++11</code> compile flag.  This default can be overridden and the C++ code can be compiled using the <code>-std=gnu++98</code> flag instead.  However, Open GEE 5.2.3 will be the last release to support compiling C++ code using the <code>-std=gnu++98</code> flag.  The subsequent releases after Open GEE 5.2.3 will require the use of the <code>-std=gnu++11</code> flag when compiling C++ code.</p>
-    <p><strong>Remove TCP/IP connections to PostgreSQL</strong>. With release 5.2.3 Open GEE code will now connect to PostgreSQL exclusively using Unix domain sockets. This will make it easier to secure PostgreSQL for security sensitive installations.</p>
+    <p><strong>Remove TCP/IP connections to PostgreSQL</strong>. With release 5.2.3 Open GEE code will now, by default, connect to PostgreSQL exclusively using Unix domain sockets. This will make it easier to secure PostgreSQL for security sensitive installations. For those upgrading from a previous release the following files will need to be changed before disabling PostgreSQL tcp listener:
```
tst-hrajamoney: Does the user have to do this or It is done in code :  disable PostgreSQL tcp listener 
If the user has to do do we need to specify  how to do that ?

tst-rwildes: The user has to do this and we can assume if they are going to do it they know enough about PostgreSQL to do this.  We could create a page in our wiki to guide them on possible ways to be more secure with open GEE PostgreSQL instance (too much to do in release notes here) but that should be a new issue.

Issue #622, opened 2017-11-28
Queries for Map Definitions should return proper JSON in GEE Server (Apache)


tst-ahernandez:
When a GEE server is queried for Server Definitions the response is not valid JSON. It would be better if the returned values were valid JSON so that they can be read by parsers.

To see an example run a query against a server:
 `http://<server>/<map>/query?request=Json`

Edit: desired behavior and versioning

The plan is to phase the new behavior into GEE. 
1. Add a v= parameter for the version of "Json" to return
    1. "v=1" will be the same behavior as current
    1. "v=2" will return properly formatted Json
1. For the first release (5.2.5), the default behavior will be "v=1", the same as it is now
1. For subsequent releases (5.2.6+), the default behavior will be "v=2"

Note a function now exists that can convert js to json: https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/fusion/autoingest/JsUtils.h#L34
This should be able to be used to help solve this issue.

tst-ccamp:
In order to maintain backward compatibility with clients, we'll need to add a new parameter to the query string

tst-ccamp:
This will also need to be implemented separately for Server and for Portable

tst-ppenev:
I'm not sure backward compatibility would be an issue.  I think it would be better to have the data in standard minified JSON when building a database, and not have to have it in two different formats.

Standard JSON is valid Java Script, so any client, such as a browser, that supports Java Script would be compatible with JSON.  Do we have clients that we're supporting that would not support JSON?

macornwell:
@tst-ppenev I definitely think it would make sense to convert the format to JSON. At a minimum for clarity, given much of the codebase uses the word JSON when its actually JavaScript.

tst-ppenev:
This discussion involves some decisions that are beyond just this issue.

I would say, if we want to support custom clients, we should define an API, and, version it, and specify how long versions would be supported, and how transitions would occur.  We would also want to keep some track of what versions of the API are used, and how often.

I don't think we can try to support something unknown for very long. It adds overhead to all of the development that deals with this part of the code, and we have no way of verifying that anything we're supposedly supporting actually works.  The only thing we can do is not change, which is not a very good way to develop, and adds an increasing amount of overhead with time.

It would help to have an idea of what real use problems something is solving to decide what effort is worth spending on it.

PR #907, opened 2018-06-26
Fixes issue #622 - Returns proper JSON for Map Definitions

macornwell:
This pull request fixes issue #622. It fixes it by adding a new method that can return JSON with a class intended on converting JavaScript dictionary objects into valid JSON.

The existing code located in serverdbReader.h and .cpp were refactored so that all references that discussed "json" were changed to JavaScript (since that is what they were referring to). The one exception to this was changing line 398 which if effectively the handler for the REST endpoint.

The suggestion by @tst-ccamp was used for the naming of the new REST endpoint (JsonOnly). 
Example: http://localhost/testdb/query?request=JsonOnly

**Recommendations For Further Fixes:**
- There is quite a bit of mixup in language in this section of the code on JSON and JavaScript. You can see it in  serverdb.cpp, serverdbReader.cpp, etc.
- Either all terminology should be changed to use the word javascript, or the file on the disk "earth.json.DEFAULT" and all associated code should be converted to JSON instead of JavaScript.
- It may also make sense to change how the underlying file 

**Notes of Caution**
While this fixes the issue, it is fairly brittle. If the earth.json.DEFAULT format changes slightly, then the regular expression used to clean up the result will fail to function. The best approach is likely to write out valid JSON in the first place because converting JavaScript into JSON is complicated, at best, without the use of JavaScript, which can do it quite well.


tst-rwildes:
Quick note: your Travis builder failed because you have some C++11 code.  The good news is there is a PR that is going in soon that will allow the use of C++11 #898 .  As soon as that is merged you could update your branch and that should take are of your build issues.  The bad news is I see you are using `<regex>`...  the unfortunate thing is we still support GCC 4.8 and its implementation of `<regex>` isn't very good: https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions

earth_enterprise/src/common/serverdb/serverdbUtilities_unittest.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+class ServerdbUtilitiesTest : public testing::Test {
+protected:
+    virtual void SetUp() {}
+
+    virtual void TearDown() {
+    }
+
+    std::string GetJavaScriptString() {
+        std::string test_string =
```
tst-rwildes: The use of a raw string might make this more readable.  See: https://en.cppreference.com/w/cpp/language/string_literal

macornwell: Good idea about the readability of the string. I wasn't crazy about what I had but now I've got it formatted and looking good.

I am looking at alternatives to using <regex>. It looks like for *nix <regex.h> can work. I am testing out a rough solution now. Having said that it won't work for windows.

Another option is bringing in boost regex or pcre.

tst-rwildes: Yes I was going to sugest <regex.h> too.  But that is a good point about Windows.  We currently don't have boost as one of our third party libraries.  I would like to add it but that might increase this scope to add it now.   There are other libraries that would be nice to have access to in boost.

macornwell: What about using "#ifdef COMPILER_MSVC" and using <regex> with msvc, and then <regex.h> for every other situation?

This should cover all *nix and Windows.

tst-rwildes: I think you have the a good idea there.  The one twist is Open GEE uses GCC for windows too *but* I believe it is a later version that has proper support for <regex>.  We could drive the decision to use <regex> vs <regex.h> based on compiler version.  CentOS 6 is the only supported OS that uses the older compiler when it builds.

earth_enterprise/src/common/serverdb/map_tile_utils.cpp
```diff
@@ -171,6 +171,7 @@ const MimeType MapTileUtils::kOctetMimeType("application/octet-stream");
 const MimeType MapTileUtils::kJpegMimeType("image/jpeg");
 const MimeType MapTileUtils::kPngMimeType("image/png");
 const MimeType MapTileUtils::kJavascriptMimeType("text/javascript");
+const MimeType MapTileUtils::kJSONMimeType("application/json");
```
tst-ppenev: Style: To be consistent with other variable names, this would be `kJsonMimeType`.

earth_enterprise/src/common/serverdb/serverdbReader.cpp
```diff
@@ -396,8 +398,11 @@ const MimeType ServerdbReader::GetData(
     } else if (request == "Json") {
       std::string json_variable_name = arg_map["var"];
       // CheckJavascriptName(json_variable_name);
-      GetJson(json_variable_name, language, region, buf);
+      GetJavaScript(json_variable_name, language, region, buf);
       return MapTileUtils::kJavascriptMimeType;
+    } else if (request == "JsonOnly") {
+      GetJSON(language, region, buf);
```
tst-ppenev: Style: To be consistent with other names, this function name would be `GetJson`.

earth_enterprise/src/common/serverdb/serverdbReader.cpp
```diff
@@ -576,21 +581,30 @@ const MimeType ServerdbReader::GetResampledImageryMapsTile(
...
   } else {
-    // Return a Javascript version of the JSON with the variable name declared.
-    std::string javascript = "var " + json_variable_name + " = " + json + ";\n";
+    // Return a version of the JavaScript with the variable name declared.
+    std::string javascript = "var " + javascript_variable_name + " = " + javascript + ";\n";
     buf.SetValue(javascript);
   }
 }
 
+void ServerdbReader::GetJSON(const std::string& language,
```
tst-ppenev: Style: Camel case.

earth_enterprise/src/common/serverdb/serverdbUtilities.cpp
```diff
@@ -0,0 +1,10 @@
+// Copyright 2018 Open GEE Contributors
+#include "serverdbUtilities.h"
+#include <string>
+#include <regex>
+
+std::string ServerdbUtilities::ParseValidJSON(const std::string& javascript_dict) {
```
tst-ppenev: Naming: This function doesn't parse valid JSON.  It applies a string replace to put quotes around object property names.

Correctness: This function would only work for special cases of Java Script code, e.g. where there are no values containing a string of letters followed by a space and a colon.

This is an implementation that requires verifying that no such values may occur in the Java Script code generated by Fusion.  It is also potential source of hard-to-diagnose problems in the future, in case such values come into use.  E.g., how is a developer supposed to know that a Map client will break if a certain sequence of characters occurs in Java Script code?

earth_enterprise/src/common/serverdb/serverdbUtilities_unittest.cpp
```diff
@@ -0,0 +1,78 @@
...
+        ","
+        "serverUrl : \"http://localhost/testdb-v001\""
+        "}";
+        return test_string;
+    }
+
+    std::string GetJSONString() {
+        std::string json_string =
+        "{"
+        "\"dbType\" : \"gedb\","
```
tst-ppenev: Although this is valid JSON, some parsers don't handle white-space in JSON properly.  I don't know what clients we're trying to support, but I think we'd get wider compatibility, if we use minified JSON, i.e. with no white-space.

tst-rwildes:
@macornwell #898 has been merged to master.  If you update your fork to sync with the downstream master and then sync your branch with with your forked master the C++11 build issues will go away.  I do see there was also a unit test failure so that would need addressing too.

earth_enterprise/src/common/serverdb/serverdbUtilities_unittest.cpp
tst-ppenev: I think using regular expressions for parsing, or modifying a data structure is not a good approach in general. Regular expression don't offer a way to represent a complex data structure, such as a heterogenous map, or an array.  It is usually very easy to introduce a hard-to-spot bug in a regular expression, which only shows up on some occasions.

macornwell: @tst-dsugar I am in agreement that I think the better answer is refactoring and modifying the code to do what it says it does (read/write JSON).

Having said that, if the original idea is to keep backwards compatibility (returning JavaScript) then that constricts the fix a bit. If regex isn't a good approach, then either another third party library needs to be brought in that handles this situation, or a customized JSON parser has to be built for this specific problem set.

macornwell:
@tst-ppenev thanks for providing useful comments.
@

PR #911, opened 2018-06-26
WIP - Fixes issue #622 - Returns proper JSON for Map 

macornwell:
This pull request fixes issue #622. It fixes it by simply quoting the Key when serializing the data.

This approach still allows the returned value (JSON) to be used as valid JavaScript.

**Notes**

- In order to activate this change, the map has to be unpublished, then republished which triggers the writing and reading of the JSON object. 
- Previous discussion of another approach can be found at #907 

@tst-ppenev, @tst-rwildes, @tst-dsugar, @tst-ccamp any thoughts about possible issues to test for with this? Unit Tests are passing, and the basic functionality is working on my end.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.files
 /*.includes
 
+# CLion project files.
+.idea/
+cmake-build*/
--- a/earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
+++ b/earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
@@ -316,7 +316,7 @@ std::string JsonUtils::JsonObject(
     if (iter != field_map.begin()) {
       buffer << ",\n";  // Note: IE6 doesn't like trailing commas.
     }
-    buffer << iter->first << " : " << iter->second;
+    buffer << Quoted(iter->first) << " : " << iter->second;
   }
   buffer << "\n}\n";
   return buffer.str();


tst-rwildes:
I agree with @tst-ppenev  comments.  I'd like to run some tests using this change just to be be sure the EC still like things.  Then I'll approve.

tst-ccamp:
I'm concerned that we are not allowing any way to get the legacy behavior. This could have significant impact on existing client implementations. I can be convinced to change the default, but we at least need a parameter that will cause it to use the old behavior.

.gitignore
```diff
@@ -21,3 +21,6 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.files
 /*.includes
 
+# CLion project files.
+.idea/
```
tst-ccamp: Consider ignoring these directories at all directory levels using the "/**" syntax

tst-rwildes: I also recently read (can't remember where) that because editors are a user preference and editors come and go as preferences change it is a good idea to use `~/.gitignore_global` for those things.  Unless it is a tool the project itself requires the use of.  BTW I really like the `CLion` editor too but I wasn't sure how well it would work without `cmake`.

https://davidwalsh.name/global-gitignore

tst-rwildes:
For what it is worth I did build and test these changes (did some small testing using tutorial testing scripts).  and both the google earth client (EC) with 3D globe and the browser with 2D map seem to work still.

I will test again once the requested changes are made.

.gitignore
macornwell: @tst-ccamp good point. I'll add that in.

@tst-rwildes  I have found the Jet brains IDE to be just one of the bests hands down. I was using it for the debugger on this. I was a bit surprised to see that I could not build in a custom build/run functionality. I also find the vim implementation to be great in Clion/Pycharm. As of right now, I found that the most effective way to build/run was just making a small script that I keep in a console window. And then attaching to the process.

tst-ccamp:
@tst-rwildes My concern isn't for standard viewing through the EC or Google Maps API, but for custom integrations that may expect the output in a certain format. Since we have no way of knowing all the different ways users have done this over the years, I propose to make as few changes to the way it works as possible unless the integrator specifies they want something different. We really have no tests for this at the moment. Even if we did, we don't know whether it would be representative of the existing integrations. It's this unknown element that makes me reluctant to change the default behavior.

tst-rwildes:
@tst-ccamp Understood.  I think we can still keep a simple approach of just adding or not adding the quotes based on the request parameter and forgo the regex stuff.

macornwell:
@tst-ccamp  you're certainly making a good point. To put it in perspective here are a few options.

1. We make the change as shown now. If a client is taking the string and bringing it directly into JavaScript, it _should_ work. I went to a few JavaScript validation services to confirm its valid JavaScript.
- The downside here, is if its being processed in some sort of other way, (such as someone else adding quotes) this could cause problems that are unexpected. 

2. We revert this back, and go with the other approach which that creates a new interface. Behind the scenes the application will still be using relying upon JavaScript (rather than JSON). In order to make this interface, a new external library dependency may be needed to be brought in. Having said that @tst-ppenev  has made some good arguments regarding a regex approach (the most straight forward), and dealing with the cases that have been mentioned, make a solution requiring a custom build parser. The potential for bugs will remain in the new interface.

3. We keep this change, add in a new restpoint that delivers the unchanged result (with quotes), and then change how the request=Json currently gets its value, by effectively doing the same thing as the 2nd option, but in reverse. (Convert JSON into a JavaScript dictionary). Both option 2 and 3 will require either a dependency added for regex/json parser, or a custom parser that can intelligently navigate the structure. The downside to this is this has the potential to add bugs into the old interface.

macornwell:
@tst-rwildes  unfortunately its not that simple of a fix. The document is written to the disk either quoted or not quoted. This happens when the globe is published. When the API is requested it goes through a few classes and methods to get to it.

tst-ccamp:
I think there may be some info missing on the original issue. I found that if you don't specify the "var" paramenter, then it just returns JSON without the "var =" part. Example: http://earth.t-sciences.com/Blue_Marble_Globe-v001/query?request=Json

One thought I had is adding some kind of version command, like:
v=legacy : return current behavior
v=2 : return correctly formatted JSON along with correct mimetype for JSON

The 2 would correspond with the edition of the standard, as shown here: http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf

We'd then just need to decide what  the default should be. There are pros and cons for both and we can take a gradual approach like we did with the C++11 migration. For 5.2.4 we add the option for the new version, deprecate the legacy behavior. In 5.2.5 we change the default to v=2, but by this time they've hopefully had advanced warning that it's coming and in the meantime have made any necessary changes to their legacy apps.

tst-rwildes:
Yes! I really like the idea if a new 'version' and introducing the new version in 5.2.4 but not changing the default version to that new version until 5.2.5 and maybe depreciate the first version in 5.2.6.  It bothers me that we call something JSON when it really is JavaScript.  Versioning the API lets you transition to the correct thing without having to break people and then give them a path forward.  We could also add a JavaScript type if there is a need to return JavaScript and those clients can transition as needed.

tst-rwildes:
@macornwell I missed your point about when the JSON/JavaScript is created.  So your change effects how things are created when the DB is pushed to server and does not effect anything at request time.  Maybe @tst-ppenev suggestion of doing both formats at publishing time and using the correct file at request time will help work around this.  That does mean for requests of things we don't have a format for because of older publishes we will have to maybe return 404 error.

tst-rwildes:
BTW I think we should (another issue) bring in a real JSON parser into the project for other reasons beyond this issue.  This one looks really interesting: https://github.com/nlohmann/json

tst-ccamp:
@tst-ppenev I'd like to use a number for a version, but what the heck would you call GEEs non-json json format verion number? 0? -1? NAN? That's why I called it legacy. What would you suggest?

tst-ppenev:
It would be a version of the HTTP API. I'd call it `1`.

tst-ccamp:
I haven't seen any activity on this PR for a while. I'm going to update the original ticket with the latest decision on versioning and close this PR.

Issue #361, opened 2017-07-20
Fusion UI : Provide "Save and Build" as an additional file menu item


tst-awankhede:
we have save and we have build. most of the times we save and immediately build from same file menu. It would save time if we can "save and build" at the same time  and it can be exposed as a separate menu option.

tst-ccamp:
Closing with fix in #914

PR #914, opened 2018-06-29
Issue #361 Fusion UI: "Save and Build" as an additional file menu item

jnk7667:
Implements #361


tst-rwildes:
Looks like the tests might be timing out.  I do see failures but that might be from things being ripped out from under the tests because of timeouts?  @jnk7667 (Joseph) can you try adding a longer timeout to:

tst-rwildes:
Looks like you did another push.  We can see how this one goes.  Maybe you won't need to change the timeout

earth_enterprise/src/fusion/fusionui/AssetBase.cpp
```diff
@@ -327,11 +332,19 @@ void AssetBase::Build(void) {
 }
 
 
+void AssetBase::SaveAndBuild(void) {
+  if (Save) {
```
tst-dsugar: Should this not be if (Save()) { ??

jnk7667: Thanks for the save there!

earth_enterprise/src/fusion/fusionui/AssetBase.h
```diff
@@ -95,6 +96,7 @@ class AssetBase : public QMainWindow {
   virtual void SetHidden(bool on);
   virtual void AboutToShowFileMenu();
   virtual void AboutToHideFileMenu();
+  virtual void SaveAndBuild();
```
tst-dsugar: This may not need to be a virtual, as the Save() and Build() methods already are, and this is a convenience slot to invoke them chained together.

jnk7667: I wanted to make sure it passed all of the tests first. I'll go ahead and make this change now.

tst-rwildes:
The code changes look good to me.  I'm going to grab the code and build/test things.

tst-rwildes:
During my testing things don't seem to work.  I see the new menu item but it is never grayed out like the save button is when there is nothing to save.  Also the new menu seems to do nothing when clicked even when there is something to save.

jnk7667:
When I was working on this I could not actually find the UI it appeared in when using Qt Designer. Is there a way I can build a debug version of Fusion to test it on my machine? I wasn't sure if that was possible, but working on it the way I have been has been hit or miss.

tst-rwildes:
Going to pull the recent changes and test them out today.

tst-ccamp:
I haven't had the chance to review the code yet, but there's one last little chore to finish this up. The release notes need to be updated: https://github.com/google/earthenterprise/blob/master/docs/geedocs/5.2.4/answer/7160004.html

jnk7667:
In this case it would just be updating the version number, correct? I looked at the page for semantic versioning to determine what to update it to. Not sure what else needs to be updated in the release notes.

tst-ccamp:
@jnk7667 we try to keep the release notes up to date as development goes on for each new feature or bug fix that occurs. You can take a look at this PR for an example in 5.2.3 development https://github.com/google/earthenterprise/pull/872/files

So, add a new entry under "New Features" that describes this change

earth_enterprise/src/version.txt
```diff
@@ -1 +1,5 @@
+
 5.2.4
+
+5.2.4
```
tst-ccamp: Can you remove the changes to version.txt? That must have been a mistake

tst-ccamp: This is actually pretty weird. It doesn't say that any of your commits added these lines. I think you may just need to merge master into the feature branch to clear this up.

earth_enterprise/src/fusion/fusionui/AssetBase.cpp
```diff
@@ -63,12 +63,14 @@ AssetBase::AssetBase(QWidget* parent)
...
   build_action_ = new QAction(this);
   //  build_action_->setIconSet(QIconSet(QPixmap::fromMimeSource("notes.png")));
+  savebuild_action_ = new QAction(this);
   close_action_ = new QAction(this);
   close_action_->setIconSet(QIconSet(QPixmap::fromMimeSource("fileclose.png")));
   hidden_action_ = new QAction(this);
   hidden_action_->setToggleAction(true);
   notes_action_ = new QAction(this);
   notes_action_->setIconSet(QIconSet(QPixmap::fromMimeSource("notes.png")));
+  savebuild_action_ = new QAction(this);
```
tst-ccamp: This is instantiated twice. I'm assuming it's because of a merge conflict. Please remove the redundant instantiation

tst-rwildes:
I re-tested with the new change and everything still working as expected.

tst-ccamp:
Since the changes still work, I'm going to go ahead and approve and merge

Issue #918, opened 2018-07-10
geedocs folder has bad links for answer, art, and css


tst-rwildes:
The links have invalid charter at the end of them.  This is causing the wiki to not update correctly

PR #919, opened 2018-07-10
fix bad links

tst-rwildes:
fixes issue #918


tst-rwildes:
The diff doesn't look like much but there was some garbage at the end of those link files.  I think that is why the wiki update was failing and we don't have a 5.2.4 section showing up.

PR #937, opened 2018-07-26
#936: Add progress log messages to `gesystemmanager`.

### LINKED ISSUE #936: Provide progress indicators during gesystemmanager processing
As a GEE user I want gesystemmanager to print out information about the progress of planning a build so that I can have an idea of how long it will take.

Most fusion executables print some kind of progress indicator showing how much work has been done and how much remains. However, gesystemmanager doesn't provide such indicators, even though the process of planning a build can take a long time. We should provide some kind of indication of what fusion is doing while it is planning a build so that users can determine if it is making progress and how much work remains. These messages do not need to indicate percentages of work remaining (e.g., "5% done"). Something as simple as an indication of the number of items processed would suffice. For example, here is the function AssetImplD::UpdateInputs from AssetD.cpp.

```C++
void
AssetImplD::UpdateInputs(std::vector<AssetVersion> &inputvers) const
{
  inputvers.reserve(inputs.size());
  for (std::vector<std::string>::const_iterator i = inputs.begin();
       i != inputs.end(); ++i) {
    AssetVersionRef verref(*i);
    if (verref.Version() == "current") {
      // we only need to update our input if it's an asset ref
      // or a version ref with "current"
      // if it points to a specific version there's nothing to do
      AssetD asset(verref.AssetRef());
      bool needed = false;
      inputvers.push_back(asset->Update(needed));
    } else {
      inputvers.push_back(AssetVersion(verref));
    }
  }
}
```

This function could include a logging statement that says something like "Processing input %x of %y for asset %z". We may be able to provide something similar for building the children of each asset as well.

Note that the above code is an example of how to accomplish this goal. It is likely that we will need to modify several functions that iterate over lists of inputs/insets/children/etc. The main goal is that we need to print periodic progress updates so that the user knows that something is going on.

## AC:

* gesystemmanager prints periodic messages to the logs indicating its progress while it determines what tasks it needs to run to complete a BuildAsset command.

Reported by @tst-lsavoie.

tst-ppenev:
* Added progress log messages to the loop that updates input resources.

* Added progress log messages to the loop that iterates over imagery inset
  assets.

Fixes #936.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
@@ -270,6 +270,10 @@ void RasterProjectAssetImplD::AddInsets(
   for (std::vector<RasterProjectModifyRequest::Item>::const_iterator item =
          items.begin();
        item != items.end(); ++item) {
+    
+    notify(NFY_PROGRESS, "Raster project::AddInsets: Finding position to add inset %lu of %lu.",
+      item - items.begin(), items.size());
+
     // error if there's anything I already have
     if (have.find(item->dataAsset) != have.end()) {
       throw khException(kh::tr("Project already has inset '%1'")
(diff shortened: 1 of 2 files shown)


tst-dsheremata:
Approved!

PR #940, opened 2018-07-30
dynamic logging levels

tst-basilhuffman:
initial WIP PR, added enum to string function for syntactic sugar, 
added method to khSystemManager to change level, this needs work - could
possiby put in ::Run() but looks shoehorned, maybe a blocker here?
noticed logic error in gevectorpackupgrade.cpp vis a vis setting logging levels


earth_enterprise/src/common/notify.cpp
```diff
@@ -68,6 +69,52 @@ khNotifyLevel getNotifyLevel() {
...
+    case NFY_INFO2: retval = "NFY_INFO2"; break;
+    case NFY_DEBUG: retval = "NFY_DEBUG"; break;
+    case NFY_VERBOSE: retval = "NFY_VERBOSE"; break;
+    case NFY_NOTICE:
+    default: retval = "NFY_NOTICE";
+    };
+    return retval;
+}
+
+khNotifyLevel stringTokhNotifyLevel(std::string level)
```
tst-rwildes: Consider using `const std::string &` for the parameter.

tst-basilhuffman: Agreed. Changing.

earth_enterprise/src/common/notify.cpp
```diff
@@ -68,6 +69,52 @@ khNotifyLevel getNotifyLevel() {
...
+    case NFY_INFO2: retval = "NFY_INFO2"; break;
+    case NFY_DEBUG: retval = "NFY_DEBUG"; break;
+    case NFY_VERBOSE: retval = "NFY_VERBOSE"; break;
+    case NFY_NOTICE:
+    default: retval = "NFY_NOTICE";
+    };
+    return retval;
+}
+
+khNotifyLevel stringTokhNotifyLevel(const std::string&s level)
```
tst-rwildes: You have an extra `s`.  That is why build failed.  should be `const std::string& level`

earth_enterprise/src/fusion/autoingest/sysman/khSystemManager.cpp
```diff
@@ -72,20 +73,40 @@ khSystemManager::SetWantExit(void)
   }
 }
 
+// when SIGHUP is given, reload systemrc
+void systemrc_reload()
+{
+    Systemrc systemrc;
+    LoadSystemrc(systemrc);
```
tst-basilhuffman: compilation error here

earth_enterprise/src/fusion/autoingest/sysman/khSystemManager.cpp
```diff
@@ -72,20 +73,43 @@ khSystemManager::SetWantExit(void)
...
+    uint32 logLevel = systemrc.logLevel;
+    assert(logLevel >= 0 && logLevel <= 7);
```
tst-lsavoie: I don't think an assert is the right way to handle cases where a user's input might be wrong. I think a better option would be to check the log level in LoadSystemrc. If the level is invalid you can set it to a valid level and print a warning.

earth_enterprise/src/fusion/autoingest/sysman/khResourceProvider.cpp
```diff
@@ -136,6 +135,13 @@ khResourceProvider::Run(void)
 
   Systemrc systemrc;
   LoadSystemrc(systemrc);
+
+  uint32 logLevel = systemrc.logLevel;
+  assert(logLevel >=0 && logLevel <= 7);
```
tst-lsavoie: Same: I'm not fond of the idea of checking user input with asserts.

earth_enterprise/src/fusion/autoingest/sysman/SConscript
```diff
@@ -108,7 +108,7 @@ gesystemmanager = env.executable('gesystemmanager',
       env['FUSIONVER'],
       ] +
     env.ObjFromOtherDir(extraObjs),
-    LIBS=['gegdal', 'geraster', 'gesearchtabs', 'genet', 'gepublish',
+    LIBS=['gegdal', 'geraster', 'gesearchtabs', 'genet', 'gepublish', 'geconfigutil','gefconfigutil',
```
tst-lsavoie: If possible, remove `geconfigutil` from here - I think you only need `gefconfigutil`.

tst-lsavoie:
We should have a documentation update about this as well.

tst-basilhuffman:
@tst-lsavoie I started to update the confluence page for khthread but I decided to hold back until implementation is approved

tst-lsavoie:
I was thinking more along the lines of adding a blurb to the html docs explaining how to change the log level while gesystemmanager is running.

earth_enterprise/src/common/notify.cpp
```diff
@@ -68,6 +69,53 @@ khNotifyLevel getNotifyLevel() {
...
+    else if (level == "NFY_INFO")
+        retval = NFY_INFO;
+    else if (level == "NFY_INFO2")
+        retval = NFY_INFO2;
+    else if (level == "NFY_DEBUG")
+        retval = NFY_DEBUG;
+    else if (level == "NFY_VERBOSE")
+        retval = NFY_VERBOSE;
+    else
+        retval = NFY_NOTICE;          // NFY_NOTICE is default, either do this or handle error
```
tst-lsavoie: Minor: you could use `NFY_DEFAULT_LEVEL` here.

tst-basilhuffman:
@tst-lsavoie what behavior are you seeing?

earth_enterprise/src/fusion/autoingest/sysman/khSystemManager.cpp
```diff
@@ -72,20 +73,43 @@ khSystemManager::SetWantExit(void)
...
 {
   sigset_t waitset;
   sigemptyset(&waitset);
   sigaddset(&waitset, SIGINT);
   sigaddset(&waitset, SIGTERM);
-
+  sigaddset(&waitset, SIGHUP);
 
   int sig;
   sigwait(&waitset, &sig);
```
tst-lsavoie: It appears that this call never returns, even if I send a signal to the gesystemmanager. Somehow it's not picking up signals. It's not caused by your changes, but your changes won't work if it's not fixed.

earth_enterprise/src/fusion/autoingest/sysman/SConscript
tst-basilhuffman: done

earth_enterprise/src/fusion/autoingest/sysman/khSystemManager.cpp
```diff
@@ -72,23 +73,34 @@ khSystemManager::SetWantExit(void)
...
-    notify(NFY_NOTICE, "Received signal %d. Ignoring ...", sig);
-  }
+
+void
+khSystemManager::SignalLoop(void)
+{
+  signal(SIGINT,handleExitSignals);
+  signal(SIGTERM,handleExitSignals);
+  signal(SIGHUP,systemrc_reload);
+  while(1);
```
tst-lsavoie: I don't think you need this while here. Now that you've changed the way the system handler works this can just exit, I think. Signals will still be sent to the appropriate handlers.

earth_enterprise/src/common/notify.cpp
```diff
@@ -68,6 +69,55 @@ khNotifyLevel getNotifyLevel() {
...
+        retval = NFY_INFO;
+    else if (level == "NFY_INFO2")
+        retval = NFY_INFO2;
+    else if (level == "NFY_DEBUG")
+        retval = NFY_DEBUG;
+    else if (level == "NFY_VERBOSE")
+        retval = NFY_VERBOSE;
+    else if (level == "NFY_NOTICE")
+        retval = NFY_NOTICE;
+    else if (level == "NFY_DEFAULT_LEVEL")
```
tst-lsavoie: I'm getting a compile error on CentOS 6 here because retval may be uninitialized. If you remove the "if" part of this "if else" statement (so that it returns NFY_DEFAULT_LEVEL for any string that doesn't match one of the others), that fixes the build error.

docs/geedocs/5.2.2/answer/176676.html
```diff
@@ -6,6 +6,7 @@
   <div class="sidebar1">
 <p class="sidebar-toc"><a href="../answer/3481464.html">Before you configure Google Earth Enterprise Fusion</a></p>
 <p class="sidebar-toc"><a href="../answer/6008655.html">Running Fusion on a Machine with Multiple CPUs</a></p>
+<p class="sidebar-toc"><a href="../answer/7160003.html">Dynamically adjusting logging levels</a></p>
```
tst-lsavoie: These changes should be in the 5.2.4 documentation folder, since that's the version they'll be introduced in. We should also add a note to the release notes about the new feature.

docs/geedocs/5.2.2/answer/7160003.html
```diff
@@ -0,0 +1,36 @@
...
+<p class="sidebar-toc"><a href="../answer/6086672.html">Troubleshoot push/publishing issues 5.x</a></p>
+<p class="sidebar-toc"><a href="../answer/3481558.html">Command reference</a></p>
+  </div>
+  <div class="content">
+<a name="top_of_file"></a>
+<p><img src="../art/common/googlelogo_color_260x88dp.png" width="130" height="44" alt="Google logo" /></p>
+<h1><a href="../index.html">Google Earth Enterprise Documentation Home</a> | Fusion administration</h1>
+<h2>Dynamically adjusting logging levels</h2>
+<p>In some instances, it may be desirable to change the logging level during execution.  The log level can be set in <code>/etc/opt/google/systemrc</code> via the <code>&lt;logLevel&gt;</code> tag, and this file may be reloaded at any time by executing: <code>/etc/init.d/gefusion reload</code>. Acceptable values range from 0 to 7, with 7 being the highest level. </p>
+<!--
```
tst-lsavoie: Minor: remove commented out code

tst-lsavoie:
There's apparently a caching mechanism on the systemrc that prevents it from being loaded multiple times.

docs/geedocs/5.2.2/answer/7160003.html
```diff
@@ -0,0 +1,36 @@
...
+<p class="sidebar-toc"><a href="../answer/6051700.html">Publish databases using disconnected publishing</a></p>
+<p class="sidebar-toc"><a href="../answer/6086672.html">Troubleshoot push/publishing issues 5.x</a></p>
+<p class="sidebar-toc"><a href="../answer/3481558.html">Command reference</a></p>
+  </div>
+  <div class="content">
+<a name="top_of_file"></a>
+<p><img src="../art/common/googlelogo_color_260x88dp.png" width="130" height="44" alt="Google logo" /></p>
+<h1><a href="../index.html">Google Earth Enterprise Documentation Home</a> | Fusion administration</h1>
+<h2>Dynamically adjusting logging levels</h2>
+<p>In some instances, it may be desirable to change the logging level during execution.  The log level can be set in <code>/etc/opt/google/systemrc</code> via the <code>&lt;logLevel&gt;</code> tag, and this file may be reloaded at any time by executing: <code>/etc/init.d/gefusion reload</code>. Acceptable values range from 0 to 7, with 7 being the highest level. </p>
```
tst-lsavoie: It may be good to mention that you can only change the logging level of gesystemmanager dynamically. For example, "...it may be desirable to change the gesystemmanager logging level..."

earth_enterprise/src/common/notify.cpp
```diff
@@ -44,6 +44,7 @@ static void* notifyCallData = NULL;
 khNotifyLevel getDefaultNotifyLevel() {
   char* variable;
   khNotifyLevel lev;
+  notify(NFY_WARN,"in getDefaultNotifyLevel");
```
tst-lsavoie: This looks like temporary code and should probably be removed.

tst-mreinert:
Reviewed changes requested by @tst-dsheremata and approved

PR #942, opened 2018-08-04
#882: Use `g++` from `devtoolset-2-toolchain` automatically on CentOS 6, if installed

### LINKED ISSUE #882: GEE Portable Build script fails on CentOS 6
Attempting to start portable server on CentOS6 results in the following error "ImportError: No module named _glc_unpacker".  Was attempting to run portable for 'portable server scenarios' (2.8.3).  Portable server can be built manually if the developer build instructions for Open GEE are followed and then 'source /opt/rh/devtoolset-2/enable' is performed to make the newer g++ compiler available.

Seems likely due to the build script attempting to use the default g++ compiler, which doesn't support the C++11 compile flag. The build script is supposed to automatically select the devtools-2 compiler, but seems to fail. The work-around is to manually source the devtoolset, as stated above.

## Additional Acceptance Criteria:
1. The build script should fail and report an error if any of the compile commands returns an error

tst-ppenev:
Fixes #882 
* Added code to check if the g++ compiler version is high enough to
  `build_and_test.py`.

* Added code to use g++ from the devtoolset2 installation on Cent OS 6, if the
  default g++ compiler version is too low.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
@@ -34,7 +34,7 @@ import util
 
 def get_cc_version(cc_path):
   """Extracts the version number of a C/C++ compiler (like GCC) as a list of
-  strings."""
+  strings.  (Like ['5', '4', '0'].)"""
 
   process = subprocess.Popen(
     [cc_path, '--version'], stdout=subprocess.PIPE)
@@ -73,9 +73,10 @@ def path_prepend(value, new_path, if_present='skip', separator=':'):
   """Adds a component to a PATH variable, where paths are separated by
   `separator`.
 
-  if_present = What to do, if PATH already contains `new_path`.
-    'skip': leave PATH as is.
-    'move': move `new_path` to the beginning of PATH.
+    if_present = What to do, if PATH already contains `new_path`.
+      'skip': leave PATH as is.
+      'move': move `new_path` to the beginning of PATH.
+    separator = String that separates path components in the variable value.
   """
 
   # Check if the PATH variable already contains `new_path`:
@@ -111,6 +112,18 @@ def path_prepend(value, new_path, if_present='skip', separator=':'):
 def environ_var_path_prepend(
   var_name, new_path, if_present='skip', separator=':'
 ):
+  """Prepend a path component to a PATH-like environment variable (i.e., one
+  which contains paths separated by `separator`).
+
+    var_name = Name of the environment variable to add a path component to.
+  The variable is created, if it doesn't exist.
+    new_path = Path to add to the variable value.
+    if_present = What to do, if the variable already contains `new_path`.
+      'skip': Leave the variable value as is.
+      'move': Move `new_path` to the beginning of the variable value.
+    separator = String that separates path components in the variable value.
+  """
+
   try:
     value = os.environ[var_name]
   except KeyError:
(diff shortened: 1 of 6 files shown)


earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
```diff
@@ -25,18 +25,145 @@
...
 import distutils.sysconfig
 import os
+import re
 import shutil
+import subprocess
 import sys
 import util
 
 
+def get_cc_version(cc_path):
```
tst-lsavoie: I'm sure you've already thought of this, but is this code duplicated from somewhere? Is there any way to eliminate the duplication?

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
```diff
@@ -25,18 +25,145 @@
...
+  if not version:
+    raise ValueError('Unable to determine g++ version!')
+  if not is_version_ge(version, [4, 8]):
+    # Check for GCC 4.8 from the devtoolset-2-toolchain package on Red Hat 6:
+    cc_dir = '/opt/rh/devtoolset-2/root/usr/bin'
+    if os.path.isfile('{0}/g++'.format(cc_dir)):
+      environ_var_path_prepend('PATH', cc_dir, if_present='move')
+      environ_var_path_prepend(
+        'LIBRARY_PATH',
+        '/opt/rh/devtoolset-2/root/usr/lib',
```
tst-lsavoie: Minor: could move most of this path into a shared variable so that it's not repeated.

tst-lsavoie:
Fixes #882

tst-lsavoie:
Issue #882 also says that we should fix the script so that it fails and returns non-zero when any of the compile commands fails.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
tst-ppenev: Yes, I was looking at that.  There is some utility code, but it's specific to things like WSGI scripts, or Portable Globe CGI scripts.

I could create a Python module near the top-level, and add it to `PYTHONPATH` for scons, and Pontable Server build scripts.

Maybe we can get some votes, or other proposals? (@tst-lsavoie, @tst-dsheremata, @tst-basilhuffman, @tst-ccamp)

tst-ppenev:
Yes, the build script terminates, but I forgot to check the exit code.

tst-ppenev:
@tst-lsavoie: I can't reproduce what you describe in your comment.

Here's what I get when I run on Cent OS 6 with the check for a GCC minimum version removed:

```
[root portableserver]# cat /etc/centos-release 
CentOS release 6.10 (Final)
[root portableserver]# ./build.py 
Build and test file unpacker library for Linux Portable Server.
Executing: cp ../shared/* .

Executing: swig -python -c++ glc_unpacker.i

Executing: g++ -std=gnu++11 -O2 -fPIC -I/usr/include/python2.6 -c glc_unpacker_wrap.cxx

cc1plus: error: unrecognized command line option "-std=gnu++11"
FAIL: Unable to execute g++ -std=gnu++11 -O2 -fPIC -I/usr/include/python2.6 -c glc_unpacker_wrap.cxx

[root portableserver]# echo $?
1
```

tst-lsavoie:
@tst-ppenev I get the same thing. I must have been using an old version. Approving.

PR #944, opened 2018-08-08
GP-1547 Update/fix imagery_viewer.js

tst-eclamar:
Fix bug in imagery_viewer.js - specifically, lines 27013 and 27045: "b=a(b)" => "b=a.Qh(b)".


tst-eclamar:
This fixes bug in GP-1547.

PR #951, opened 2018-08-22
Added js to json funciton to JsonUtils.  Also added single unit test for new method.

tst-rwildes:
This will be used to help address issue #622


tst-rwildes:
I'm going to have to re-do this PR...  I accidentally commit from a box that didn't have git fully configured so the name is wrong on the commit record.

tst-ppenev:
@tst-rwildes: You should be able to just rebase, instead of creating a new pull request.

PR #963, opened 2018-08-29
extended existing gtest to cover a wide variety of javascript syntax …

justrandyc:
35 additional cases of JavaScript and expected Json conversion strings, to be fed into the JsonUtility.  Examples only of JavaScript Objects, but including arrays and subobjects.  Boolean, Numeric, and String instances in various combinations.

This gtest can be run with the other gtests as JsonUtils_unittest and was tested on Cent OS 7.
It is an expansion of the fix for issue #622

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
+++ b/earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
@@ -23,13 +23,342 @@ namespace {
 
 class JsonUnitTest : public testing::Test { };
 
-TEST_F(JsonUnitTest, singleFieldNospaces) {
+// Testing when the Javascript is simply null
+// The conversion method should handle it correctly
+TEST_F(JsonUnitTest, nullFields) {
+    const std::string js = R"js_()js_";
+    const std::string expectedJson = R"js_()js_";
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// Testing the Javascript could be simply empty
+TEST_F(JsonUnitTest, emptyFields) {
+    const std::string js = R"js_({})js_";
+    const std::string expectedJson = R"js_({})js_";
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// Testing the value can be empty
+TEST_F(JsonUnitTest, unsetValue) {
+    const std::string js = R"js_({name:""})js_";
+    const std::string expectedJson = R"js_({"name":""})js_";
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// this checks the quotes are added to the key when string
+TEST_F(JsonUnitTest, noKeyQuotes) {
     const std::string js = R"js_({name:"fred"})js_";
     const std::string expectedJson = R"js_({"name":"fred"})js_";
     std::string result = JsonUtils::JsToJson(js);
 
     EXPECT_EQ(expectedJson, result);
 }
+
+// this checks the conversion detects they are strings and quotes them
+TEST_F(JsonUnitTest, noValueQuotes) {
+    const std::string js = R"js_({name:fred})js_";
+    const std::string expectedJson = R"js_({"name":"fred"})js_";
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// this is the most "normal" string case
+TEST_F(JsonUnitTest, doubleKeyQuotes) {
+    const std::string js = R"js_({"name":"fred"})js_";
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -23,14 +23,343 @@ namespace {
...
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// Ok so apparently using single quotes is somewhat common
+// Surely the conversion utility shouldn't just "accept" them
+// and put double quotes INSIDE of them!
+//TEST_F(JsonUnitTest, singleQuotes) {
+//    const std::string js = R"js_({'name':'fred'})js_";
```
tst-rwildes: Yes that is a good catch

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -23,14 +23,343 @@ namespace {
...
+    const std::string expectedJson = R"js_({"fred's":"name"})js_";
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// While unlikely and dangerous, this seems legal in Javascript
+// Perhaps a bogus test/unlikely edge case
+//TEST_F(JsonUnitTest, doubleQuoteInKey) {
+//    const std::string js = R"js_({"freds\"car\"":"christine"})js_";
```
tst-rwildes: I think you are correct here.  This is legal JS.

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -23,14 +23,343 @@ namespace {
...
+    const std::string js = R"js_({"car":"fred's"})js_";
+    const std::string expectedJson = R"js_({"car":"fred's"})js_";
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// Can we escape quotes?  This is legit for Javascript
+//TEST_F(JsonUnitTest, doubleQuoteInValue) {
+//    const std::string js = R"js_({"carname":"\"christine\""})js_";
```
tst-rwildes: Yes this is a good test.

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -23,14 +23,343 @@ namespace {
...
+}
+
+// Should it eat white space? I would think the conversion would collapse it.
+// This may not be a bug.  Yes the test is ugly on purpose to contrast output
+//TEST_F(JsonUnitTest, spansMultipleLines) {
+//    const std::string js = R"js_({name:"fred", 
+//         car:"christine"
+//        ,color:"red"})js_";
+//    const std::string expectedJson = 
+//        R"js_({"name":"fred","car":"christine","color":"red"})js_";
```
tst-rwildes: Yea I wasn't clear on that.  I think currently it preserves white space.  Maybe we can add an option to create compact JSON.  @tst-jhaight can confirm behavior.

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -23,14 +23,343 @@ namespace {
...
+//    std::string result = JsonUtils::JsToJson(js);
+//
+//    EXPECT_EQ(expectedJson, result);
+//}
+
+// Return/NewLine should be... eaten? Replaced with a space?
+// Surely not just have extra escape chars thrown at it
+//TEST_F(JsonUnitTest, newlineInValue) {
+//    const std::string js = R"js_({"name":"fred\r\nis\r\ncool\r\n"})js_";
```
tst-rwildes: In this case you are stating you want new lines in your data.  This is perfectly allowed and should be transformed to JSON...  I'm wondering with the raw string you could just add literal new lines.  But in any case it should be preserved in the JSON too.

Curious what this would do:

```
R"js_({"name":"fred
is
cool
"})js_"
```

It should produce the same string for JSON (with the literal new lines).  I'm not sure how to do return carriages in a raw string as all characters are taken literally.

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -23,14 +23,343 @@ namespace {
...
+    const std::string js = 
+        R"js_({names:{first:fred,new:false,house:10}})js_";
+    const std::string expectedJson = 
+        R"js_({"names":{"first":"fred","new":false,"house":10}})js_";
+    std::string result = JsonUtils::JsToJson(js);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+// possible additional tests
```
tst-rwildes: Yes multi-line tests would be good adds!

PR #964, opened 2018-08-29
add logging information to find what may be causing .new.new creation during cache purge

tst-basilhuffman:
I am unable to reproduce the error. I will manually add files and check.


tst-ccamp:
Can we get a better title on this PR?

tst-basilhuffman:
Sorry about that, more descriptive now

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -189,7 +189,13 @@ class MutableAssetHandleD_ : public virtual Base_ {
       for (std::vector<std::string>::iterator it = toDelete.begin();
            it != toDelete.end(); ++it) {
         if (dirtyMap.find(*it) != dirtyMap.end()) {
-          std::string filename = dirtyMap[*it]->XMLFilename() + ".new";
+          std::string filename = dirtyMap[*it]->XMLFilename();
```
tst-lsavoie: I'm not sure if this will completely fix the problem. I think we want to avoid getting files that end in .new into the dirtyMap in the first place. This will accomplish the goal of preventing .new.new files from appearing, but the underlying issue that .new files are being processed as if they were .xml files will still exist.

Thoughts?

tst-basilhuffman: Were we not wanting to overwrite the files that are currently there?

tst-ppenev: This will probably create bugs.  You'll be writing an XML file that doesn't correspond to a valid build of an asset.  E.g., until an asset build succeeds, there will be files files that are partially built, or metadata that has been updated, but the files it refers to won't be there, yet,or they won't have the correct contents.

If Fusion services are terminated before an asset build completes, you'll end up with an XML describing an invalid build that looks just like a successfully built asset.  If you kill Fusion services at the right time, it's probably also possible to end up with an incompletely written XML and loose data.

You also need to know that there's no other code that can request information about a building asset that has been purged from the cache.  Otherwise, that code will receive asset metadata that doesn't correspond to built files on disk.

If you've already read all the code, and know there are no such possible problems, it may be OK to overwrite existing XML files.  Otherwise, I think the first step is to add documentation about how asset matadata is updated during a build, and how the asset cache handles updates, especially during a cache purge.

tst-basilhuffman: This gets back to a previous question I had about .new remnants being fully formed or not.

tst-ccamp:
@tst-basilhuffman Please update the GitHub issue number in the title and description

earth_enterprise/src/common/generic_utils.h
```diff
@@ -16,7 +16,8 @@
 
 #ifndef GEO_EARTH_ENTERPRISE_SRC_COMMON_GENERIC_UTILS_H_
 #define GEO_EARTH_ENTERPRISE_SRC_COMMON_GENERIC_UTILS_H_
-
+#include <string>
```
tst-lsavoie: Are these necessary? I don't see any other changes in this file.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
tst-lsavoie: I think we decided to back out the change to avoid creating a filename ending in .new.new. If I remember the conversation correctly, we thought it was safest to continue allowing .new.new files to be created and just log when they are being created.

earth_enterprise/src/fusion/autoingest/Asset.h
```diff
@@ -97,10 +97,13 @@ Asset::DoBind(const std::string &ref, bool checkFileExistenceFirst) const
   HandleType entry = CacheFind(ref);
   bool addToCache = false;
 
+  notify(NFY_NOTICE,"CheckFileExistence: %d",checkFileExistenceFirst);
```
tst-lsavoie: I would make this NFY_VERBOSE.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
tst-lsavoie: Note: you can still keep the notify statement for when we create a .new file.

tst-basilhuffman:
Addresses #1002

earth_enterprise/src/fusion/autoingest/Asset.h
tst-basilhuffman: Addressed

earth_enterprise/src/common/generic_utils.h
tst-basilhuffman: Holdover from a utility function I added but then deleted. Addressed

earth_enterprise/src/common/generic_utils.h
```diff
@@ -16,6 +16,8 @@
 
 #ifndef GEO_EARTH_ENTERPRISE_SRC_COMMON_GENERIC_UTILS_H_
 #define GEO_EARTH_ENTERPRISE_SRC_COMMON_GENERIC_UTILS_H_
+#include <string>
```
tst-lsavoie: These are still showing up for me. Is there a missing commit?

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -186,16 +186,28 @@ class MutableAssetHandleD_ : public virtual Base_ {
 
       // Save mutable items.
       khFilesTransaction filetrans(".new");
+      uint8 numDotNew=0;
```
tst-mreinert: I might consider making this larger than uint8

earth_enterprise/src/common/generic_utils.h
```diff
@@ -17,6 +17,9 @@
 #ifndef GEO_EARTH_ENTERPRISE_SRC_COMMON_GENERIC_UTILS_H_
 #define GEO_EARTH_ENTERPRISE_SRC_COMMON_GENERIC_UTILS_H_
 
+#include <string>
```
tst-lsavoie: These lines don't seem to want to go away.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -21,7 +21,7 @@
 #include <notify.h>
 #include <khThread.h>
 #include <khFileUtils.h>
-
+#include <cstdlib>
```
tst-lsavoie: This is probably unneeded as well.

earth_enterprise/src/common/generic_utils.h
tst-basilhuffman: Sorry about that, didn't pay close enough attention. Taken care of.

PR #966, opened 2018-08-29
Moved Js utils in to a separate class

tst-hrajamoney:


earth_enterprise/src/fusion/autoingest/JsUtils.h
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef GEO_EARTH_ENTERPRISE_SRC_FUSION_AUTOINGEST_JSUTILS_H_
+#define GEO_EARTH_ENTERPRISE_SRC_FUSION_AUTOINGEST_JSUTILS_H_
+
+#include <string>
+#include <vector>
```
tst-jhaight: we don't really need the include of vector or map here since they aren't used

earth_enterprise/src/fusion/autoingest/JsUtils.cpp
```diff
@@ -0,0 +1,71 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+#include "fusion/autoingest/JsUtils.h"
+#include <math.h>
```
tst-jhaight: I think we really just need the includes for JsUtils.h and array here.

tst-rwildes: Yes keep headers as light as you can.

earth_enterprise/src/fusion/autoingest/JsUtils.h
```diff
@@ -0,0 +1,40 @@
...
+#include <map>
+
+
+// JsUtils is a static utility class for JS
+class JsUtils {
+ private:
+  JsUtils() {}  // Static class. Do not allow construction.
```
tst-rwildes: Just use the new C++ delete key word instead of creating private constructors.  The error message you get is more descriptive.

`JsUtils() = delete;`
`JsUtils(const JsUtils&) = delete;`

PR #971, opened 2018-08-31
SConstruct: replace deprecated build_dir with variant_dir

tst-twilk:
Renaming deprecated build_dir to variant_dir in earthenterprise/earth_enterprise/src/SConstruct


tst-rwildes:
Should we put the `.sconsign.dblite` file in the build dir using the `SConsignFile` call?

PR #974, opened 2018-09-04
Increased the AssetCacheSize and VersionCacheSize to 128k

arjntc1:
Currently a static value (128k) -- would be great if this were configurable based on some calculated analysis of current system resources.


arjntc1:
Specified the wrong email address ... I updated to match the email used in GitHub.

arjntc1:
Update related to #978

Issue #976, opened 2018-09-06
Portable server windows build fails


tst-jhaight:
Using the mingw g++ 8.1.0 (latest available [here](https://sourceforge.net/projects/mingw-w64/files/latest/download)) attempting to compile portable server fails for both 32 and 64bit with the following error message:
`cmath:1121:11: error: '::hypot' has not been declared`

It appears the root cause of this issue is [a cpython bug](https://bugs.python.org/issue11566) which has not been fixed yet.  To work around this issue `-D_hypot=hypot` must be added to the Windows g++ command to compile `glc_unpacker_wrap.cxx`.

Additionally `-DMS_WIN64` needs to be added to g++ commands in order to link successfully when using mingw64.

tst-ppenev:
@tst-jhaight, please specify what installation of Python you're using.  This bug is logged for Python 3.3.  Portable Server runs with Python 2.7.

tst-jhaight:
@tst-ppenev 
Sorry I forgot to specify but I'm using python 2.7.15 and seeing the same behavior described by that bug.

`MS_WIN64` fixes a problem that seems to be caused by a few factors.  The swig generated implementation of `glc_unpacker_wrap.cxx` is using `Py_InitModule`.  Python's `modsupport.h` is supposed to rename `Py_InitModule` for 64bit systems.  `MS_WIN64` is used to help determine 64 vs 32 on windows but for some reason it isn't set with mingw64.

tst-ppenev:
@tst-jhaight, thanks.

So, you think the problem is `swig` only supports 32-bit Python extensions?  I wonder why I didn't see this problem on 64-bit systems before?

tst-jhaight:
@tst-ppenev I really still haven't figured out exactly what is the absolute root cause.  I don't think it is swig at fault unless they should include or define something in the auto-generated code.  It at least looks to me like including python.h should be all that is needed and they do that.

It could be caused by where we are pulling include files from.  I see a lot of the .h files for python are in both mingw's `opt/include/python2.7` and the python installation's `include` directory but they are not the same file content.  Maybe each one could expect some behavior from the other which isn't met?  It at least seems limited to only showing up for 64bit windows builds.

tst-ppenev:
Oh, you have a MinGW Python installation?

You could try specifying an include directory explicitly when compiling to override the default path.  I think we'd want to make sure GCC is using the header files corresponding to the Python interpreter that's running the build script.

[...]

Issue #973, opened 2018-09-04
Portable server builds require file generated by geserver build


tst-jhaight:
The portable server build is looking for `earth_enterprise/src/version.txt` and fails because it is not generated without the full earthenterprise build.  As a consequence of this, portable server cannot be built for windows without some manual steps to create that file and put it in the correct location.

AC:

1. Portable server builds are independent of regular geserver/gefusion builds.

tst-ppenev:
SCons should run on Windows?  Did you have trouble generating `version.txt`?

We would need to add more installation instructions.

tst-jhaight:
I didn't attempt to run a scons build since I didn't expect it to succeed and didn't want to install a bunch of tools I didn't really need.  Is there a scons target that would only generate version.txt?

I filed this since building the release_5.2.3 branch I didn't have problems with version.txt.

tst-ppenev:
The target is the path to the file you want to generate.  E.g., something like:

```bash
scons src/NATIVE-REL-x86_64/gee_long_version.txt
```

tst-rwildes:
Running `scons` from `earth_enterprise` folder with no parameters does generate the `version.txt` file without having to do a full open build but not in Windows.  The scons script fails under Windows which is another issue IMO.  A work around under Windows (version 10) is to run the `scons` command from within a `bash` window with no parameters.  This will then create the file and you can switch back to a windows command prompt to do the rest.  If you are in a Windows command prompt and have bash for Windows installed you can just execute `bash` to jump into a bash shell then execute `scons` then `exit` to get back to Windows command prompt.

tst-rwildes:
I want to correct myself.  Both I and @tst-ppenev made incorrect statements on how to get scons to generate the version.txt file...  As is right now to get the file to generate you have to cd into `earth_enterprise/src` and from within that directory you can run `scons NATIVE-DBG-x86_64/gee_version.txt` to get scons to generate the needed version.txt file. 

This has a couple of issues: 
- this requires scons and if we are going to require scons to build portable we should open another issue to just add the portable builds to the scons build system.  I do feel that should be a different issue and not addressed with this issue but I thought I'd mention it.
- the bigger issue is that running that command is unintuitive and we should have an easier command to run.  This may be addressed with the fix for #984. 
- we need to update the documentation or maybe the current build scripts (assuming the first point isn't addressed right away) to run the scons command to generate the version file.

tst-ppenev:
[...]

PR #977, opened 2018-09-07
Correct two issues blocking windows portable builds

tst-jhaight:
Fixes issue #976 

The definition of the hypot macro is a workaround for an upstream python bug which has not been resolved yet.  The other mingw define fixes a linkage issue in the swig c++ wrapper when compiling under mingw64.  Verified portable compile works as expected for win64, win32 and centos7.

Note for testing: currently need to obtain a copy of version.txt to work around #973

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
@@ -25,6 +25,7 @@ maps.
 
 import distutils.sysconfig
 import os
+import platform
 import re
 import shutil
 import subprocess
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
```diff
@@ -78,13 +79,24 @@ def BuildLibrary(os_dir, ignore_results):
 
   configure_c_compiler(os_dir)
 
+  specialDefs = ''
+  if os_dir == "Windows":
+    # The logic below fixes a method used by the swig c++ wrapper. Mingw python headers
+    # should detect and fix this but for some reason they aren't working with mingw64
+    pythonCLib = "libpython" + sys.version + ".a"
```
tst-blaird: Python's sys.version returns something like '2.7.15 (v2.7.15:ca079a3ea3, Apr 30 2018, 16:30:26) [MSC v.1500 64 bit (AMD64)]' which probably is not what was intended here.

PR #983, opened 2018-09-11
getop timeout

tst-jkent:
Adds a timeout option for getop in order to avoid freezing when working on large projects. The default timeout is 60 seconds, but can be overridden by the user by using the timeout flag paired with the desired time, ex: (getop --timeout 100).


earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -78,16 +79,20 @@ main(int argc, char *argv[])
...
     options.opt("delay", delay);
+    options.opt("timeout", timeout);
     if (!options.processAll(argc, argv, argn))
       usage(progname);
     if (help)
       usage(progname);
     if (delay <= 0)
       usage(progname, "--delay must be positive");
+    if (timeout < 0)
+      usage(progname, "--timeout must be zero or greater");
```
tst-basilhuffman: Minor: timeout must not be less than zero

Issue #987, opened 2018-09-18
Installdir does not handle absolute paths


tst-twilk:
Scons scripts do not handle absolute paths correctly. Paths that point to outside of the source tree are handled correctly (ex: /tmp/fusion_os_server) while other paths are converted to relative paths (ex: /home./centos/earthenterprise/earth_enterprise/src/build/stage -> build/stage). Converting absolute path to relative may lead to multiple installdir being created when running  `scons installdir=build/stage install`.

PR #988, opened 2018-09-18
Installdir bug fix + add ability to build third_party only

tst-twilk:
#987 

- Fix for installdir not handling absolute paths correctly
- Add ability to build third_party code only


tst-blaird:
What are the verification steps?

Issue #881, opened 2018-06-14
save layer error


AsdrtChen:
Dears , 

I build a Map Layer . 
And I change the icon .
but I can't save if I change the icon 
![image](https://user-images.githubusercontent.com/30334676/41396798-9e351024-6fe4-11e8-8f87-4d39704f3d57.png)

And I already chmod + chown those file .
it has the same error message 
how do I fix this problem 

thanks

ericwhiteau:
I have also just run into this issue,
Note that the 3D  earth build was successful with the same Icon file names.

Happens in 2D with both standard and custom icons.  Definitly not a permissions issue,  I have set read all  and as stated it works in 3d build with the same icons.

tst-ccamp:
I missed this bug before. We'll start taking a look

ericwhiteau:
With this issue in 2D maps,  I need to understand my options.
I was migrating an Earth project to a new infrastructure.   Currently runs 5.1.3  the last google release.
I did testing with OpenGee and built full 3D solution (after full test suite against 3d globe).
It was only then that I addressed the 2d maps side.  which is failing due to able issue.

Is there a way to roll back Open GEE to 5.1.3?
I get the message. 
/gevol/assets  is configured for "5.2.1"  you have 5.1.3  need to upgrade.....

Is there a valid way to downgrade to 5.1.3?    changing the version in .config file doesnt work as I tested that.
Attempting to avoid build from scratch as there is 30days of build time.

not sure what sort of timeline is on bug fixes against 5.2.4?
thanks in advance for any comments.

ericwhiteau:
PS   I note that   the library   "SkImageDecoder::DecodeFile"

is used in 2d maps and not in 3d client.

this seems to have a valid path passed to it but returns error.

tst-sbayard:
@tst-mbiancalana and I started looking at this last week, and we believe we've identified the issue. @ericwhiteau, good detective work tracking it down to SkImageDecoder::DecodeFile". The underlying problem is that, because of some fancy linking trickery in combination with how SCons executes the SConscript files, the codecs for the icons don't get linked in, so they don't even exist in the Fusion install ☹️ 
Hence, Fusion doesn't know how to decode them for 2d map layers.

 We're currently working to determine the best fix for this.

@AsdrtChen and @ericwhiteau thanks for reporting this and doing some digging on it.

tst-lsavoie:
Closed via PR #989.

ericwhiteau:
[...]

PR #989, opened 2018-09-18
#881 Allow icons on map layers

tst-mbiancalana:
This commit addresses multiple related things
1) Issue #881, which had errors returning to fusion users when saving a
map layer with an icon chosen for shields or points.  This was caused by
having no image decoders registered in the skia third-party library.

2) The skia library was not being built correctly by SCons. The source
for the skia library exists in a tarball, which is extracted as a step
in the build process, but SCons was trying to gather list of cpp files
up front before the files were extracted. Now the list of files is
gathered from the contents of the tarball.

3) To work around the issue above, the third-party libraries were being
built twice. This worked because on the second round, the cpp files
existed on disk at the right time. This second round is no longer
necessary.


earth_enterprise/src/third_party/skia/SConscript
```diff
@@ -147,9 +175,6 @@ sgl_opts_ssse3_objects = map(
 sgl_opts_files = map(lambda f: os.path.join(sgl_src_abs_dir, f), sgl_opts_cpps)
 sgl_opts_ssse3_files = map(lambda f: os.path.join(sgl_src_abs_dir, f), sgl_opts_ssse3_cpps)
 
-skia_source = File('#/../../earth_enterprise/third_party/skia/%s.tar.gz'
-                   % skia_version).abspath
-
 skia_source_files = sgl_opts_files + sgl_opts_ssse3_files + sgl_cpps
 skia_target = '%s/.extract_skia' % current_dir
```
tst-rwildes: wonder if we need this anymore...  might not hurt but might not need it too.

tst-rwildes (changes requested):
Over all looks really good.  Have some suggestions that I'd like to be addressed.  Also should work with @tst-twilk as there might be some conflict here.

tst-eclamar:
I'm attempting to compile the code and, after 30 second of successful compiling, I get the following error:
```
...
creating pr/Makefile

sent 569,803 bytes  received 675 bytes  1,140,956.00 bytes/sec
total size is 567,247  speedup is 0.99
+ touch /home/eric/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/google/protobuf-py/install/lib/python2.7/site-packages/google/__init__.py
+ touch /home/eric/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/google/protobuf-py/.install
creating pr/include/Makefile
creating pr/include/md/Makefile
creating pr/include/obsolete/Makefile
creating pr/include/private/Makefile
creating pr/src/Makefile
creating pr/src/io/Makefile
creating pr/src/linking/Makefile
creating pr/src/malloc/Makefile
creating pr/src/md/Makefile
creating pr/src/md/unix/Makefile
creating pr/src/memory/Makefile
creating pr/src/misc/Makefile
creating pr/src/threads/Makefile
creating pr/tests/Makefile
creating pr/tests/dll/Makefile
creating pr/src/pthreads/Makefile
+ touch /home/eric/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party/libjs/.configure_libjs
scons: *** [build] Error 2
[eric@centos7b earth_enterprise]$ 

```
I've cleaned and rebuilt twice, each with the same error.  (Note: what I switch back to 'master', everything compiles just fine).

earth_enterprise/src/third_party/skia/SConscript
```diff
@@ -18,29 +18,57 @@
...
+# This regexp selects the desired groups of cpp files out from everything else.
+# Strings are assumed to begin with './' as they are in the skia-5.1.2 tarball,
+# and may have trailing whitespace (e.g. newlines).  The path of the cpp file
+# is captured.
+cpp_select_re = re.compile(r'^\./(src/(?:core|effects|effects/gradients|gpu|gpu/gl|gpu/effects|images|image|pathops|utils)/[^/]+\.cpp)\s*$')
+# This regexp identifies certain cpp files to be removed from the list selected by cpp_select_re
+cpp_downselect_re = re.compile(r'^src/utils/[^/]*(?:SkDeferredCanvas|SkLua|SkThreadUtils_pthread_mach|SkThreadUtils_win|SkDumpCanvas|SkCamera)[^/]*\.cpp$')
+
+# Get a listing of the contents of the skia tarball
+tar_p = subprocess.Popen(['tar', 'tzf', skia_source], stdout=subprocess.PIPE)
```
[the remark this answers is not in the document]
tst-mbiancalana: In looking through some more of the third-party libraries, it looks like there aren't too many places where surveying the contents of the tarball beforehand will help us. We'll keep the technique in our pocket, though.

earth_enterprise/src/third_party/skia/SConscript
tst-mbiancalana: I think this is still important in order to keep the ordering of the build steps correct. The library build still must wait for the tarball to be extracted.

tst-rwildes (approved):

earth_enterprise/src/third_party/skia/SConscript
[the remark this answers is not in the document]
tst-rwildes: Because we are using fake output files (the dot files) to represent the output of the tar operation all over third party scons can not properly really know if the tar file needs to be re-extracted.  If some of all of the source is deleted for whatever reason but scons sees that dot file it will think executing the un-tar command is not needed because all if the output exists.  The more you tell scons the better and more accurate job it can do.

earth_enterprise/src/third_party/skia/SConscript
tst-rwildes: Scons will wait because it now knows that the source files to the build command is created by this untar command.  If you give scons all the correct information it does the right thing.  When you take short cuts and hide information from it then it will sometimes do the wrong thing because it doesn't know the full details.  But leaving the dot file for now isn't hurting anything.

tst-emillergalow (approved):
Code looks good and I was able to reproduce the bug in master, and verify that the issue is resolved in this branch.  I successfully saved a 2D Vector layer with the icon changed and didn't see the error message.

PR #992, opened 2018-09-18
991 dns test failure

### LINKED ISSUE #991: DNS diagnostics test fails when there is more than one hostname present
Steps to reproduce:

- Install latest Open GEE 5.2.4
- Ensure there are 2 hostnames present next to 127.0.1.1 in /etc/hosts
- Open admin page and navigate to http://<hostname>/admin/geecheck.html
- Notice that DNS test fails

Example:
No error
127.0.1.1 myhost
Error
127.0.1.1 myhost myhost.vm

tst-sbayard:
Fixes #991 

The dns test will now look at aliases in addition to the FQDN.


tst-blaird:
I'll test this against Ubuntu-14.

PR #994, opened 2018-09-19
Add ability to use non-standard build folder name and location.

tst-twilk:
Changes:
- add ability to build third_party code only
- fix `scons install` not handling installdir absolute paths correctly
- remove hardcoded build path and allow tests to be run from a non-standard build folder location
- remove hardcoded build folder name from `rpms/build.gradle`
- do not stage install postgis with server

Repro steps:
- build third_party code only
`cd earthenterprise/earth_enterprise/src
scons -j4 build_folder=/tmp/build release=1 third_party_only=1`
- build everything
`cd earthenterprise/earth_enterprise
scons -j4 build_folder=/tmp/build release=1 installdir=/tmp/build/stage`
- verify that stage folder contains separate postgis subfolder
- run unit tests
`cd /tmp/build/bin/tests
perl RunAllTests.pl -o test.out`
- build RPMs
`cd earthenterprise/earth_enterprise/rpms
./gradlew openGeeRpms`
- verify that postgis RPM located in build/distributions subfolder contains postgis files
- install/uninstall RPMs 
- run non-RPM installer
`cd earthenterprise/earth_enterprise/src/installer
sudo ./install_fusion.sh -dir /tmp/build/stage
sudo ./install_server.sh -dir /tmp/build/stage`
- stop gefusion and geserver
`sudo /etc/init.d/gefusion stop
sudo /etc/init.d/geserver stop'
- uninstall gefusion and geserver
`sudo ./uninstall_fusion.sh
sudo ./uninstall_server.sh`

### DIFF
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -58,33 +58,9 @@ import org.opengee.shell.GeeCommandLine
 def src_dir = "../src/"
 def scons_option = "release=1"
 
-// Look for command line parameters that specify how the packages are
-// to be built.  This can be either 'optimize' or 'debug'.
-// Option 'release' is the default.
-def native_dir = {
-  if ( project.hasProperty( 'optimize' ) ) {
-    scons_option = "optimize=1"
-    return "NATIVE-OPT-x86_64"
-  }  // end check for option 'optimze'
-  if ( project.hasProperty( 'debug' ) ) {
-    scons_option = "release=0"
-    return "NATIVE-DBG-x86_64"
-  }
-  return "NATIVE-REL-x86_64"
-}  // end native_dir
-
-// Build the actual file to be used
-def gee_long_version_txt = "${src_dir}${native_dir()}/gee_long_version.txt"
-
-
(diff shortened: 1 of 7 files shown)


earth_enterprise/rpms/build.gradle
```diff
@@ -95,6 +71,21 @@ if (new File('build.properties').exists()) {
...
 
+// Look for command line parameters that specify how the packages are
+// to be built.  This can be either 'optimize' or 'debug'.
+// Option 'release' is the default.
+if ( project.hasProperty( 'optimize' ) ) {
+    scons_option = "optimize=1"
+}  // end check for option 'optimze'
+else if ( project.hasProperty( 'debug' ) ) {
+    scons_option = "release=0"
+}
```
tst-rwildes: I think the original code was incorrect here...  We can keep `debug`  but we should map it to `scons_option = internal=1`.  The `release=0` is implied.  But we should also add we should add `internal` too because that is what scons really uses and have that map to the same thing of `scons_option = internal=1`.

tst-rwildes: And we should have a check for `release` property and have that map to `scons_option = "release=1"`

Issue #984, opened 2018-09-11
Build fails from 5.2.4 beta tarball due to missing version.txt


tst-blaird:
1. Install all prerequisites to build OpenGEE.
2. Download the 5.2.4 beta tarball from https://github.com/google/earthenterprise/releases/tag/5.2.4-1.beta
3. Extract the tarball.
4. cd into the extracted directory and then the earth_enterprise subdirectory.
5. Run "scons -j8 release=1 build" to build OpenGEE.
6. Note that the build immediately fails with "IOError: [Errno 2] No such file or directory: 'version.txt':"

This happens because the version.txt file is now a generated file.  The process to generate it runs git commands to retrieve tag information.  Since the extracted tarball is not a git repository these commands fail (apparently silently) and the version.txt file is not generated.

Before creating the tarball run "scons version.txt" and then create it with the generated version.txt file.

tst-rwildes:
This issue was fixed with PR #998

PR #998, opened 2018-09-19
I #984 Build fails from 5.2.4 beta tarball due to missing version.txt

tst-eclamar:
Issue #984 - Build fails from 5.2.4 beta tarball due to missing version.txt

Note: it was necessary to do step 11, 'Copy docs for next release', in order to make space for the new and updated webpages that point to 5.2.5.  Maybe I'm confused, but it looks like that this wasn't done for 5.2.4 ('Copy docs for next release').

### DIFF
--- a/scripts/build_release_archive.sh
+++ b/scripts/build_release_archive.sh
@@ -37,7 +37,18 @@ CLONE_DIR=`mktemp -d`
 
 # Download the repository
 cd "${CLONE_DIR}"
-git clone -b "${1}" --depth 1 "https://github.com/google/${REPO_NAME}.git" || exit
+git clone -b "${1}" "https://github.com/tst-eclamar/${REPO_NAME}.git" || exit # TEMPORARY: tst-eclamar => google
+
+cd "${CLONE_DIR}"/earthenterprise/earth_enterprise/src
+git remote add upstream git://github.com/google/earthenterprise.git # TEMPORARY
+git fetch upstream # TEMPORARY
+git pull upstream master # TEMPORARY
+# generate version files based on tags within git
+scons version_files
+cd ../..
+# remove all un-tracked files except for version.txt
+git clean -f -d -x -e version.txt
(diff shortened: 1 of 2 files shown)


scripts/build_release_archive.sh
```diff
@@ -37,7 +37,18 @@ CLONE_DIR=`mktemp -d`
 
 # Download the repository
 cd "${CLONE_DIR}"
-git clone -b "${1}" --depth 1 "https://github.com/google/${REPO_NAME}.git" || exit
+git clone -b "${1}" "https://github.com/tst-eclamar/${REPO_NAME}.git" || exit # TEMPORARY: tst-eclamar => google
+
+cd "${CLONE_DIR}"/earthenterprise/earth_enterprise/src
+git remote add upstream git://github.com/google/earthenterprise.git # TEMPORARY
+git fetch upstream # TEMPORARY
+git pull upstream master # TEMPORARY
```
tst-mbiancalana: When are these lines marked TEMPORARY going to be removed or changed?

scripts/build_release_archive.sh
```diff
@@ -37,7 +37,18 @@ CLONE_DIR=`mktemp -d`
...
+# generate version files based on tags within git
+scons version_files
+cd ../..
```
tst-ccamp: From a code safety perspective using relative paths in cd commands inside of scripts can be dangerous, especially if the script uses default settings and does not stop executing when an error is encountered. Basically, you don't really know that you're in the correct directory before executing a relative path. Also, if you get into the habit of using cd, you may get into the worse habit of following it up with a command that uses wildcards, such as `rm *`. These are just some things to be mindful of.

I suggest getting into the habit of using the style above `cd "${CLONE_DIR}"/earthenterprise` rather than `cd ../..`. Also, I recommend enabling stop execute on error by adding the `set -e` command after the file header comment.

PR #1003, opened 2018-09-20
SetState progress logging

tst-jkent:
Progress logging has been added in the SetState code to help address the issue of setting the state taking longer for large builds.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
@@ -344,7 +344,7 @@ void
 AssetVersionImplD::HandleChildStateChange(const std::string &) const
 {
   // NoOp in base since leaves don't need to do anything
-  notify(NFY_VERBOSE, "HandleChildStateChange: %s", GetRef().c_str());
+  notify(NFY_VERBOSE, "AssetVersionImplD::HandleChildStateChange: %s", GetRef().c_str());
 }
 
 void
@@ -968,7 +968,7 @@ LeafAssetVersionImplD::DoClean(void)
 void
 CompositeAssetVersionImplD::HandleChildStateChange(const std::string &) const
 {
-  notify(NFY_VERBOSE, "HandleChildStateChange: %s", GetRef().c_str());
+  notify(NFY_VERBOSE, "CompositeAssetVersionImplD::HandleChildStateChange: %s", GetRef().c_str());
   SyncState();
 }
 
--- a/earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
@@ -245,10 +245,10 @@ AssetVersionImplD::PropagateStateChange(void)
     notify(NFY_PROGRESS, "Iteration: %d | Total Iterations: %s | Parent: %s",
            i,
            ToString(parents.size()).c_str(),
-           ToString(parents.at(1)).c_str());
+           p->c_str());
     if (parent) {
       notify(NFY_VERBOSE, "parent: %s exists", 
-             ToString(parent).c_str());
+             p->c_str());
       notify(NFY_VERBOSE, "Calling parent->HandleChildStateChange(%s)", 
              GetRef().c_str());
       parent->HandleChildStateChange(GetRef());
@@ -274,10 +274,10 @@ AssetVersionImplD::PropagateStateChange(void)
     notify(NFY_PROGRESS, "Iteration: %d | Total Iterations: %s | Listener: %s",
            i,
            ToString(toNotify.size()).c_str(),
-           ToString(listener).c_str());
+           l->c_str());
     if (listener) {
       notify(NFY_VERBOSE, "listener: %s exists",
-             ToString(listener).c_str());
+             l->c_str());
       notify(NFY_VERBOSE, "Calling listener->HandleChildStateChange(%s, %s)", 
              GetRef().c_str(), 
              ToString(state).c_str());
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -171,14 +171,27 @@ AssetVersionImplD::SetState(AssetDefs::State newstate, bool propagate)
     notify(NFY_DEBUG, "SetState: %s %s",
```
tst-mreinert: Would it be better to just update this notify to include the `state` variable rather than the 3 notifys that you are adding?  Something like:
notify(NFY_DEBUF, "Set State to %s from %s for asset %s", ToString(newstate).c_str(),  ToString(state).c_str(), GetRef().c_str());

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -171,14 +171,27 @@ AssetVersionImplD::SetState(AssetDefs::State newstate, bool propagate)
...
+           ToString(oldstate).c_str(), 
+           ToString(state).c_str());
     state = newstate;
+    notify(NFY_VERBOSE, "Set current state: %s to newstate: %s", 
+           ToString(state).c_str(), 
+           ToString(newstate).c_str());
     try {
       // NOTE: This can end up calling back here to switch us to
       // another state (usually Failed or Succeded)
+      notify(NFY_VERBOSE, "Trying OnStateChange(%s, %s)",
```
tst-mreinert: I would consider removing this notify, it isn't really adding any information we don't already have

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -187,7 +200,14 @@ AssetVersionImplD::SetState(AssetDefs::State newstate, bool propagate)
     // set it to above. OnStateChange can call SetState recursively. We
     // don't want to notify/propagate an old state.
     if (propagate && (state == newstate)) {
+      notify(NFY_VERBOSE, "We wish to propagate and state: %s is equal to newstate: %s",
```
tst-mreinert: This notify contains a subset of the information that the next notify has, consider removing it

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -187,7 +200,14 @@ AssetVersionImplD::SetState(AssetDefs::State newstate, bool propagate)
...
     // don't want to notify/propagate an old state.
     if (propagate && (state == newstate)) {
+      notify(NFY_VERBOSE, "We wish to propagate and state: %s is equal to newstate: %s", 
+             ToString(state).c_str(), 
+             ToString(newstate).c_str());
+      notify(NFY_VERBOSE, "Calling theAssetManager.NotifyVersionStateChange(%s, %s)", 
+             GetRef().c_str(), 
+             ToString(newstate).c_str());
       theAssetManager.NotifyVersionStateChange(GetRef(), newstate);
+      notify(NFY_VERBOSE, "Calling PropagateStateChange()");
```
tst-mreinert: PropagateStateChange() has an entrance notification.  This notification is probably unnecessary

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -230,16 +250,28 @@ void
 AssetVersionImplD::PropagateStateChange(void)
 {
   notify(NFY_VERBOSE, "PropagateStateChange(%s): %s",
```
tst-mreinert: I would probably change this notify level to NFY_PROGRESS so that when we are logging at just the NFY_PROGRESS level we get the context of which asset is propagating its state change

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -300,6 +344,7 @@ void
 AssetVersionImplD::HandleChildStateChange(const std::string &) const
 {
   // NoOp in base since leaves don't need to do anything
+  notify(NFY_VERBOSE, "AssetVersionImplD::HandleChildStateChange: %s", GetRef().c_str());
```
tst-mreinert: Probably don't need to log here since this is a NoOp function

earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp
```diff
@@ -574,8 +574,12 @@ khAssetManager::NotifyVersionStateChange(const std::string &ref,
 {
   // assert that we're already locked
   assert(!mutex.TryLock());
+  notify(NFY_VERBOSE, "Asserted that mutex is already locked");
```
tst-mreinert: Assert is noisy enough that you probably don't need to log that the assert succeeded

earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp
```diff
@@ -574,7 +574,10 @@ khAssetManager::NotifyVersionStateChange(const std::string &ref,
 {
   // assert that we're already locked
   assert(!mutex.TryLock());
-
+  
+  notify(NFY_VERBOSE, "Set pendingStateChanges[ref]: %s to state: %s", 
+         ToString(pendingStateChanges[ref]).c_str(),
```
tst-mreinert: ToString(pendingStateChanges[ref]).c_str() is printing as a blank when I test this.  

> [2018-09-24 11:08:10] Verbose: Calling theAssetManager.NotifyVersionStateChange(Tutorial/Resources/Imagery/other.kiasset/product.kia?version=1, Queued)
> [2018-09-24 11:08:10] Verbose: Set pendingStateChanges[ref]:  to state: Queued

I would also probably update this notify so it says the actual ref name instead of ref

tst-jkent: So I see the blank values, but sometimes it will return Blocked. Not really sure what causes this, but I have had it return an expected value.

PR #1005, opened 2018-09-23
Fixes #1004: don't read version file for tasks that don't need to read it

### LINKED ISSUE #1004: Running simple/common gradle tasks under earth_enterprise/rpms fails
On a clean system (one that never did a `scons stage_install` simple commands like `gradle clean` or even `gradle tasks` fail to run with the following below error output.  These commands should succeeded and to the correct thing without having to run `scons stage_install` or even a `scons build` or any scons command.  This was somewhat of an issue prior to 5.2.4 but during the 5.2.4 release we inadvertently made this issue much worse via changes done for #987

Error output:
```

* Where:
Build file '/mnt/c/Users/rwildes/projects/geoplatform/earthenterprise/earth_enterprise/rpms/build.gradle' line: 100

* What went wrong:
>                 Runing `scons` to produce </tmp/fusion_os_install/gee_long_version.txt> failed!
                  Command: [/usr/bin/scons, release=1, /tmp/fusion_os_install/gee_long_version.txt]
                  Exit code: 2.
                  Error output:
                  scons: *** Do not know how to make File target `/tmp/fusion_os_install/gee_long_version.txt' (/tmp/fusion_os_install/gee_long_version.txt).  Stop.

                  Standard output:
                  scons: Reading SConscript files ...
  Checking for C library gtest... (cached) no
  Checking for C header file libpng12/png.h... (cached) yes
  scons: done reading SConscript files.
  scons: Building targets ...


* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

BUILD FAILED in 5s
```

tst-rwildes:
Fixes issue #1004 

Basically do not read from version file during the configuration phase.  Moved this to a task and placed appropriate dependencies on this new task only for the tasks that need to read this file.

### DIFF
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -89,34 +89,6 @@ def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
 // Create a file handle for gee_long_version_txt
 def openGeeVersionFile = new File( gee_long_version_txt )
 
-// Call scons to create the file 'gee_long_version.txt'
-// The actual parameter used by scons is based on the parameter
-// passing into Gradlew
-if ( !openGeeVersionFile.exists() ) {
-    GeeCommandLine.expand(
-      ["/usr/bin/scons", scons_option, gee_long_version_txt ],
-      "Runing `scons` to produce <${gee_long_version_txt}> failed!",
-      null, // Use default environment variables.
-      new File( src_dir )
-    )
-}
-
-def openGeeVersion =
-    openGeeVersionFile.readLines().
-    collect {
-        // Strip comments:
-        def commentIndex = it.indexOf('#')
-
-        commentIndex < 0 ? it : it.substring(0, commentIndex)
-    }.
-    collect { it.trim() }. // Trim whitespace.
-    findAll { !it.isEmpty() }. // Skip empty lines.
-    findResult {
-        it // The first string that remains is the version string.
-    }.
-    replaceAll(/-.*release/, '').
-    replaceAll('-', '.')
-
 def stagedInstallDir = new File(stagedInstall_path)
 def stagedInstallDir_common = new File(stagedInstallDir, 'common')
 def stagedInstallDir_common_opt = new File(stagedInstallDir, 'common/opt')
@@ -186,10 +158,8 @@ def rpmCommandFilter = {
     return rpmCapabilityMap.containsKey(it) ? rpmCapabilityMap[it] : it
 }
 
-
-
 ospackage {
-    version = openGeeVersion
+    version = "<version unset>"
     release = "1.${rpmPlatformString}"
     distribution = rpmPlatformString
     packager = 'gee-oss@googlegroups.com'
@@ -198,6 +168,29 @@ ospackage {
     os = LINUX
 }
 
+task setGeeVersion {
+    doLast {
+        // if the version file does not exist raise an error
+        if ( !openGeeVersionFile.exists() ) {
+            throw new GradleException("Missing version file '${openGeeVersionFile}'.")
+        }
+        ospackage.version =
+            openGeeVersionFile.readLines().
+            collect {
+                // Strip comments:
+                def commentIndex = it.indexOf('#')
+
+                commentIndex < 0 ? it : it.substring(0, commentIndex)
+            }.
+            collect { it.trim() }. // Trim whitespace.
+            findAll { !it.isEmpty() }. // Skip empty lines.
+            findResult {
+                it // The first string that remains is the version string.
+            }.
+            replaceAll(/-.*release/, '').
+            replaceAll('-', '.')
+    }
+}
 
 // Build packages for all platforms by default:
 defaultTasks 'osPackage'
@@ -219,7 +212,14 @@ task stageOpenGeeInstall(type: Exec, dependsOn: 'compileOpenGee') {
         "installdir=${stagedInstallDir}")
 }
 
-task openGeePostGisRpm(type: GeeRpm) {
+if (project.hasProperty('buildOpenGee')) {
+    setGeeVersion.dependsOn stageOpenGeeInstall
+}
+
+task openGeePostGisRpm(type: GeeRpm, dependsOn: [setGeeVersion]) {
+    doFirst {
+        requires('opengee-common', ospackage.version, GREATER | EQUAL)
+    }
     packageName = 'opengee-postgis'
     release = "1.${rpmPlatformString}"
     version = '2.3.4'
@@ -233,7 +233,6 @@ task openGeePostGisRpm(type: GeeRpm) {
     type = BINARY
     autoFindProvides = true
     autoFindRequires = true
-    requires('opengee-common', openGeeVersion, GREATER | EQUAL)
 
     from (stagedInstallDir_postgis_opt) {
         into new File(packageInstallRootDir, 'opt')
@@ -247,14 +246,10 @@ task openGeePostGisRpm(type: GeeRpm) {
     }
 }
 
-if (project.hasProperty('buildOpenGee')) {
-    openGeePostGisRpm.dependsOn stageOpenGeeInstall
-}
-
 // Expands the `gevars.sh.template` and any other templates so they can be
 // used in the `common` package, and prefixed to install scripts for all the
 // various packages.
-task openGeeSharedFiles(type: Copy) {
+task openGeeSharedFiles(type: Copy, dependsOn: [setGeeVersion]) {
     from file('shared')
     into file('build/shared')
 
@@ -271,7 +266,7 @@ task openGeeSharedFiles(type: Copy) {
             def templateDir = it.file.parent
 
             def templateVariables = [
-                'openGeeVersion': openGeeVersion
+                'openGeeVersion': ospackage.version
             ]
 
             def expandTemplate = { text_or_file ->
@@ -293,10 +288,12 @@ task openGeeSharedFiles(type: Copy) {
     }
 }
 
-task openGeeCommonRpm(type: GeeRpm, dependsOn: openGeeSharedFiles) {
+task openGeeCommonRpm(type: GeeRpm, dependsOn: [openGeeSharedFiles, setGeeVersion]) {
+    doFirst{
+        version = ospackage.version
+    }
     packageName = 'opengee-common'
     release = "1.${rpmPlatformString}"
-    version = openGeeVersion
     user = 'root'
     permissionGroup = 'root'
     packageGroup = 'Application/Productivity'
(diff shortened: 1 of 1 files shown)


earth_enterprise/rpms/build.gradle
```diff
@@ -186,10 +158,8 @@ def rpmCommandFilter = {
     return rpmCapabilityMap.containsKey(it) ? rpmCapabilityMap[it] : it
 }
 
-
-
 ospackage {
-    version = openGeeVersion
+    version = "<version unset>"
```
tst-rwildes: This will be set within a task if any of the package tasks need to be executed.   If none of the package tasks need to be executed then this doesn't need to be set.

earth_enterprise/rpms/build.gradle
```diff
@@ -219,7 +212,14 @@ task stageOpenGeeInstall(type: Exec, dependsOn: 'compileOpenGee') {
         "installdir=${stagedInstallDir}")
 }
 
-task openGeePostGisRpm(type: GeeRpm) {
+if (project.hasProperty('buildOpenGee')) {
+    setGeeVersion.dependsOn stageOpenGeeInstall
+}
```
tst-rwildes: This is kind of weird...  But it was what the old code was doing.  Really we should not have a task triggered like this based off of a property.  It makes the gradle command awkward...  A better way might be to have `setGeeVersion` as a `mustRunAfter` to `stageOpenGeeInstall`.  Then if the user wants to have gradle build, stage and package as oppposed to just package (with the assumption that build and stage has already been done) all the user does is run: `gradle stageOpenGeeInstall osPackage` and the right thing would be done.  This is cleaner then having to do `gradle -pbuildOpenGee osPackage`.  But changing the behavior now probably isn't a good idea.

earth_enterprise/rpms/build.gradle
```diff
@@ -219,7 +212,14 @@ task stageOpenGeeInstall(type: Exec, dependsOn: 'compileOpenGee') {
...
+
+task openGeePostGisRpm(type: GeeRpm, dependsOn: [setGeeVersion]) {
+    doFirst {
+        requires('opengee-common', ospackage.version, GREATER | EQUAL)
+    }
```
tst-rwildes: Things that we can't set during the config phase but need to be set before the task is run we can just create an action closure to the front of the action queue for that task.  As long as these things are done before the default actions are run for that task type the default actions don't really care if those things where done in the config phase or an action at the head of the tasks' action queue.  Ether way is good.

earth_enterprise/rpms/build.gradle
```diff
@@ -233,7 +233,6 @@ task openGeePostGisRpm(type: GeeRpm) {
     type = BINARY
     autoFindProvides = true
     autoFindRequires = true
-    requires('opengee-common', openGeeVersion, GREATER | EQUAL)
```
tst-rwildes: Moved from config phase to early action in task because during the config phase open GEE version isn't known.  It isn't known until after the `setGeeVersion` task is run during the execution phase.

earth_enterprise/rpms/build.gradle
```diff
@@ -247,14 +246,10 @@ task openGeePostGisRpm(type: GeeRpm) {
     }
 }
 
-if (project.hasProperty('buildOpenGee')) {
-    openGeePostGisRpm.dependsOn stageOpenGeeInstall
-}
```
tst-rwildes: `setGeeVersion` task now takes care of this dependency.

earth_enterprise/rpms/build.gradle
```diff
@@ -89,34 +89,6 @@ def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
...
-// The actual parameter used by scons is based on the parameter
-// passing into Gradlew
-if ( !openGeeVersionFile.exists() ) {
-    GeeCommandLine.expand(
-      ["/usr/bin/scons", scons_option, gee_long_version_txt ],
-      "Runing `scons` to produce <${gee_long_version_txt}> failed!",
-      null, // Use default environment variables.
-      new File( src_dir )
-    )
-}
```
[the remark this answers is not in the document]
tst-ppenev: Most of the time this is just a test if a file exists.  I don't think that's too expensive.

The `scons` code to generate `version.txt` is also not that expensive.  It reads Git tags and loops through them, but I doubt it would take a second.

earth_enterprise/rpms/build.gradle
```diff
@@ -89,34 +89,6 @@ def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
...
-
-        commentIndex < 0 ? it : it.substring(0, commentIndex)
-    }.
-    collect { it.trim() }. // Trim whitespace.
-    findAll { !it.isEmpty() }. // Skip empty lines.
-    findResult {
-        it // The first string that remains is the version string.
-    }.
-    replaceAll(/-.*release/, '').
-    replaceAll('-', '.')
```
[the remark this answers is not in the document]
tst-ppenev: You're not running the code to stage an install, or build Open GEE.  It's just the code to output the version file.

earth_enterprise/rpms/build.gradle
tst-ppenev: I don't think setting the version should depend on staging Open GEE.  Staging has a lot more side effects than generating a version file, if it doesn't exist.  The old code just generated the version file without staging Open GEE.

earth_enterprise/rpms/build.gradle
tst-ppenev: I think that's confusing.  I wouldn't expect that setting a version would do all the build work necessary for packaging.

earth_enterprise/rpms/build.gradle
tst-ppenev: This will create issues you may not be expecting.  The value of the `version` property is used to calculate the name of output files for the tasks that generate RPMs.  If those values are miscalculated during the configuration phase the incremental building of these tasks will be broken.

For example, your repository is tagged as GEE version 5.2.3, you make a build of the RPMs, then you tag your repository as 5.2.4, and try to make another build.  The output files of the RPM build task for 5.2.4 may be the same as the ones for 5.2.3, because only the version changed, but, now that's not considered as part of the output, because the version is always set to `<version unset>` when calculating output files.  So, Gradle would tell you there's nothing to build, and you may be scratching your head wondering why it's not producing packages for 5.2.4.

tst-rwildes: This is set by the `setGeeVersion` task and that task has not declared output so gradle will always run it even in incremental builds if it is called directly or indirectly because of a dependency.  All the tasks that use this value now have a dependency set on the `setGeeVersion` task so calling them even during an incremental run will trigger `setGeeVersion` to execute and update that property to its new value (if it changed) which should then trigger the other tasks to now run because that value changed.

To be sure I'll test the above senior to ensure it works as expected.

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-rwildes: True if you already have done a build and stage this is really fast...  But should I have to do a build and stage just to run `gradle tasks`? or `gradle clean`?  *That* seems like an excessive thing to have to do for those two operations.

earth_enterprise/rpms/build.gradle
tst-rwildes: True but there is no task that does anything meaningful with the version that doesn't also need the output of a build and stage available so we might as well (for now) use the build and stage as a means to get the version file.

earth_enterprise/rpms/build.gradle
tst-rwildes: If you find it confusing I could put these back but they will have no effect based on current code.

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-ppenev: As I mentioned, the old code ran neither a build, nor a stage to get a version.  It just ran the `scons` code to generate the version file.  I haven't timed it, but I don't think that would take even a second to run.

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-ppenev: I think it would be normal for a `clean` task to rely on calculating the output files from all other tasks, which would require having the version.  It's true that the `clean` operations that apply to most plugins operations are not that fine-grained, and would just delete the `build` directory without having to know the output files from each task.  Obviously, that wouldn't work in general, since there isn't anything restricting you from generating files in any arbitrary location from a task.  So, in this case, running a `clean` without version would happen to work, but in general, that wouldn't be the usual way it operates.

Task creation is also dynamic in Gradle (which is why the configuration stage needs to complete to know what tasks exist).  For the sake of argument, you could, for example, have some tasks that only existed for GEE version 5.x.x, and not for GEE version 6.x.x.  Again, we're not doing anything like that, so we can get away with using a dummy value for the GEE version, but, in general, I can make sense that you need such information just to run `gradle tasks`.

Yes, I think staging would normally depend on generating version information, so if would be normal for staging to call the tasks that output that information.  I also don't see anything wrong with getting that information before staging.  To build an RPM you need both version information and all the files that would be installed.  But when you need version information, you should perform the operations to get version information, not some other operation that also happens to generate version information.

earth_enterprise/rpms/build.gradle
tst-ppenev: Yes, I can see that's the code you've added, but that's not what I was describing.

When running a task is requested, such as, for example, `openGeeFusionRpm`, the build system needs to decide whether executing the operations defined for that task need to be run in the first place, or not.  That's typically done by checking whether any of the task's input files have changed since the task's output files were generated.  If inputs have not changed since the output was generated, the task won't run at all.  So, in this case, `openGeeFusionRpm` won't run.

There are more details about task execution.  I was describing a scenario for illustration, not to say that's the only problem that can result from this change.  I'm saying that there's a reason why some values are set during configuration, and there are consequences that may not be immediately apparent to consider.

Besides that, I think this adding more indirection for developers reading the code to follow, and adding more code to tasks.  And, I see little benefit resulting from these changes.

earth_enterprise/rpms/build.gradle
tst-ppenev: That could be argued.

But, it could also be argued that the only useful tasks in this build file are for generating packages, so it's fine for all of them to depend on building and staging Open GEE.

tst-ppenev: I think using `mustRunAfter` is a good idea, or even detecting if staged install files exist.  I think this option is mostly to make enabling a build from Gradle explicit.  It's also settable from a `.settings` file.

earth_enterprise/rpms/build.gradle
tst-rwildes: I will re-do this and do another PR where the version information is generated during the configuration phase.  BTW I was playing with things and it appears that the package tasks always rebuild the rpm files.  Even when on files (input or output) have changed.  This might be a bug with the Netflix RPM plugin we are using...

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-rwildes: The generic clean task in gradle is really just a place holder tasks for others to extend.  If you create a task that creates output you *should* extend the clean task to do proper clean up.  Many don't and many just extend clean to delete the build folder because of that.  

The reality is currently there is nothing that uses the version information until execution phase.  But yes, that could in theory change later.

earth_enterprise/rpms/build.gradle
tst-ppenev: :laughing: There you go, one mess solved by another.

I have lost count of the work-arounds I've had to do for OS Package.

Thanks for verifying.  Then, I'd be OK with these changes.  It would be great if we have a cleaner version later.

tst-ppenev: :laughing: There you go, one mess solved by another.

I have lost count of the work-arounds I've had to do for OS Package.

Thanks for verifying.  Then, I'd be OK with these changes.  It would be great if we have a cleaner version later.

earth_enterprise/rpms/build.gradle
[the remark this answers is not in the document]
tst-ppenev: Oh, yes, I have a clean task:

under the rug

PR #1010, opened 2018-09-25
#788 Add flags for poi search, update documentation.

### LINKED ISSUE #788: Allow users to enable POI search when publishing from the command line
As a GEE administrator I want to be able to enable POI search when publishing databases from the command line so that I can administer my databases without using the UI.

### AC:
1. Add a command line parameters that allow the user to enable POI search and enhanced search when running `geserveradmin --publishdb`
1. Add error checking so that the new command line parameters are only valid when the user specifies `--publishdb` and the enhanced search parameter is only valid if the user enables POI search
1. Update the documentation to describe the new parameters
1. Add new tests (automated if possible) to verify the new parameters

CTJeffries:
Added flags to geserveradmin to allow publishing databases with POI search information. Updated the documentation of geserveradmin to reflect these new flags.

Automatic testing these changes would require rewriting PublisherClient.cpp, and testing the command line utility itself would require an additional testing framework outside of gtest. Because of the lack of automatic testing, manual test instructions are included below.

To test POI search flag for geserveradmin:

Create a test database, my test database used the 4km bluemarble imagery data and the California population vector data. Add POI information making sure it is searchable, I added the names of the cities.

Test 1) Publish the database using:
    `geserveradmin --publishdb <db_name> --targetpath <publish_name>`
Upon command success, open up the earth client. Try to search something, you should get no results as POI search is not enabled. Unpublish the database using either console commands or the geserver webpage.

Test 2) Publish the database using:
    `geserveradmin --publishdb <db_name> --targetpath <publish_name> --enable_poisearch`
Upon command success, open up the earth client. Search for something that is in your POI data (In my case I search "san", as there are many cities in California with "san" in their name). You should get results that are only in your database. Unpublish the database using either console commands or the geserver webpage.

Test 3) Publish the database using:
    `geserveradmin --publishdb <db_name> --targetpath <publish_name> --enable_poisearch --enable_enhancedsearch`
Open up the earth client. Search for something that is not in your POI data, perhaps a city in the US (In my case I search "Wooster", as there is no city in California called Wooster, but there are a few scattered throughout the rest of the US). You should get results for data outside your database using the GeocodingFederated plug-in. Unpublish the database using either console commands or the geserver webpage.

Test 4) Attempt to publish the database using:
    `geserveradmin --publishdb <db_name> --targetpath <publish_name> --enable_enhancedsearch`
You should get a fusion fatal error, indicating that you cannot enable enhanced search without POI search being enabled.

Fixes #788

### DIFF
--- a/docs/geedocs/5.2.5/answer/3481558.html
+++ b/docs/geedocs/5.2.5/answer/3481558.html
@@ -75,7 +75,7 @@
 --<strong>no_historical_imagery</strong>] -o <em>projectname</em>{[--<strong>maxlevel</strong> <em>level</em>] <em>insetresource</em>}</code>...</pre>
   <h3>Purpose</h3>
   <p>Creates a new imagery project. This tool is capable of building Mercator imagery projects for 2D databases, or Flat (Plate Carrée) imagery projects with or without Historic Imagery Support for 3D databases.
-  
+
   <h3>Commands</h3>
   <pre>
 <code><b>--mercator </b></code></pre>
@@ -442,6 +442,16 @@ geserveradmin --garbagecollect
       <td><pre><code><b>--setecdefault</b> </code></pre></td>
       <td>Optional</td>
       <td>Publish this database as the default one for the Earth Client to connect to if no database or virtual host is specified upon initial connection.</td>
+    </tr>
+    <tr>
+      <td><pre><code><b>--enable_poisearch</b> </code></pre></td>
+      <td>Optional</td>
+      <td>Enable Point of Interest search if database contains POI data.</td>
+    </tr>
+    <tr>
+      <td><pre><code><b>--enable_enhancedsearch</b> </code></pre></td>
+      <td>Optional</td>
+      <td>If POI search is enabled, enable enhanced search.</td>
     </tr>
   </tbody>
 </table>
--- a/docs/geedocs/5.2.5/answer/3481558.html
+++ b/docs/geedocs/5.2.5/answer/3481558.html
@@ -417,7 +417,7 @@ geserveradmin --garbagecollect
 </table>
 <h4><a name="geserveradmin_pdb"></a>Publish database</h4>
 <pre>
-<code><b>--publishdb</b> <i>db_name</i> <b>--targetpath</b> <i>target_path</i> [<b>--vhname</b> <i>vh_name</i>] [<b>--setecdefault</b>]</code></pre>
+<code><b>--publishdb</b> <i>db_name</i> <b>--targetpath</b> <i>target_path</i> [<b>--vhname</b> <i>vh_name</i>] [<b>--setecdefault</b>] [<b>--enable_poisearch</b> [<b>--enable_enhancedsearch</b>]]</code></pre>
 <p>Publish the specified database on the specified target path. If the virtual host name is omitted, it publishes to the default virtual host: "public".</p>
 <table class="nice-table">
   <tbody>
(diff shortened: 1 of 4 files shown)


earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
```diff
@@ -623,6 +630,10 @@ int main(int argc, char* argv[]) {
         throw khException("Target path is not specified.\n");
       }
 
+      if (enable_enhancedsearch && !enable_poisearch) {
```
tst-lsavoie: I'd like to see one other error check here - poi search and enhanced search shouldn't be specified unless publishdb is set.

tst-lsavoie:
@CTJeffries the docs that need to be updated are located at https://github.com/google/earthenterprise/blob/master/docs/geedocs/5.2.5/answer/3481558.html. There's a section on geserveradmin that needs to have the new parameters added.

You may need to rebase your branch on the latest master to get access to the 5.2.5 version of the docs. Ping me if you'd like help with that.

earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
CTJeffries: Check has been added!

CTJeffries:
Changes to the documentation have been added, along with the addition requested check.

docs/geedocs/5.2.5/answer/3481558.html
```diff
@@ -443,6 +443,16 @@ <h4><a name="geserveradmin_pdb"></a>Publish database</h4>
       <td>Optional</td>
       <td>Publish this database as the default one for the Earth Client to connect to if no database or virtual host is specified upon initial connection.</td>
     </tr>
+    <tr>
```
tst-lsavoie: I'd like to also see these options in the publishdb line above. It should look something like
```
--publishdb db_name --targetpath target_path [--vhname vh_name] [--setecdefault] [--enable_poisearch [--enable_enhancedsearch]]
```

earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
```diff
@@ -623,6 +631,14 @@ int main(int argc, char* argv[]) {
         throw khException("Target path is not specified.\n");
       }
 
+      if ((enable_poiseach || enable_enhancedsearch) && !publishedb) {
```
tst-lsavoie: There's a build error right here - missing 'r' in enable_poisearch

docs/geedocs/5.2.5/answer/3481558.html
CTJeffries: Added in the most recent commit.

earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
```diff
@@ -623,6 +631,14 @@ int main(int argc, char* argv[]) {
         throw khException("Target path is not specified.\n");
       }
 
+      if ((enable_poisearch || enable_enhancedsearch) && !publishdb.size()) {
```
tst-lsavoie: OK, I ran my tests, and everything looks good, but I'm going to ask for one last thing. Can we move these checks closer to the beginning of main. Right after where the command line arguments are parsed (around line 396) seems like a good place to me. I'd like to have these checks run before (almost) anything else so we can catch error conditions as quickly as possible.

PR #1013, opened 2018-09-25
997 git based alpha builds

tst-kfreiheit:
Parallel development of a release branch and a future release are now possible on the earthenterprise repo.

Steps to enable alpha release development:
1. Create a branch for the current release (i.e. release_5.2.4):

```
git checkout master
git checkout -b release_5.2.4

```

2. On the master branch, add an empty commit and tag it using Major.Minor.Release-Patch format (i.e. 5.2.5-0):

```
git checkout master
git commit --allow-empty -m"5.2.5-0 commit for tag"
```

That's it.  The next commit to master will produce the version 5.2.5-0.alpha, and commits on the release_5.2.4 branch will continue on the 5.2.4 version sequence.

The script earthenterprise/earth_enterprise/src/scons/getversion.py generates the version string, and can be run independently:

# long version (5.2.5-0.alpha.timestamp.hash)
earthenterprise/earth_enterprise/src/scons/getversion.py --long

# short version (5.2.5)
earthenterprise/earth_enterprise/src/scons/getversion.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.files
 /*.includes
 
+.vscode/
--- a/earth_enterprise/src/scons/getversion.py
+++ b/earth_enterprise/src/scons/getversion.py
@@ -32,10 +32,6 @@ def GetLongVersion(backupFile, label=''):
     return ret
 
 
-def _GetDateString():
-    """Returns formatted date string representing current UTC time"""
-    return datetime.utcnow().strftime("%Y%m%d%H%M")
-
 def _GitGeneratedLongVersion():
     """Take the raw information parsed by git, and use it to
        generate an appropriate version string for GEE."""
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -22,3 +22,4 @@ earth_enterprise/tutorial/FusionTutorial/KML
 /*.files
 /*.includes
 
+.vscode/
```
tst-dsugar: Shouldn't this be in the top level .gitignore like before??

tst-kfreiheit: I made the change so regardless of which directory you load as your root in vscode, the project will ignore its contents.

I must have opened the source at earthenterprise/earth_enterprise, and saw the extra folder in my untracked files.

This change avoids the issue in all scenarios (including root).

earth_enterprise/src/scons/khEnvironment.py
```diff
@@ -236,149 +236,6 @@ def EmitLongVersionStrfunc(target, backupFile, label):
   return 'EmitLongVersion(%s, %s, %s)' % (target, backupFile, label)
```
tst-kfreiheit: Moved all of the version-related code to its own file, and export only the functions used by khEnvironment (GetVersion, GetLongVersion).

earth_enterprise/src/scons/getversion.py
```diff
@@ -0,0 +1,202 @@
...
+        base = _ReadBackupVersionFile(backupFile)
+        ret = '-'.join([base, _GetDateString()])
+
+  # Append the label, if there is one.
+    if len(label):
+        ret = '.'.join([ret, label])
+
+    return ret
+
+
```
tst-kfreiheit: Everything below is 'not public' implementation detail.  Public interface is above.

earth_enterprise/src/scons/getversion.py
```diff
@@ -0,0 +1,202 @@
...
+        return _VersionForTaggedHead()
+    else:
+        return _VersionFromTagHistory()
+
+
+def _IsCurrentCommitTagged():
+    """True if the current commit is tagged, otherwise False"""
+    raw = _GetCommitRawDescription()
+    # If this condition hits, then we are currently on a tagged commit.
+    return (len(raw.split("-")) < 4)
```
tst-jhaight: An alternative way of doing this which would allow a little more flexibility in tag naming: `git describe --exact-match HEAD` will return the tag name if the current commit is tagged and will fail otherwise.  There's nothing wrong with what is here and other places also have the tag scheme built in.

tst-kfreiheit: Thanks -- will leave code as-is, but that's a great tip on git describe.

Issue #1011, opened 2018-09-25
Defined but not used variable error when building with gcc 5.5.0


tst-twilk:
When building Open with gcc 5.5.0 the following warning causes build to fail.

`/home/tomek/geoplatform/earthenterprise/earth_enterprise/src/gee_build/REL-linux-4.4.0-135-generic-gcc5.5.0-x86_64/third_party/skia/include/images/SkForceLinking.h:20:12: error: 'linking_forced' defined but not used [-Werror=unused-variable]`
` static int linking_forced = SkForceLinking(false)`
`            ^`
`fusion/gst/maprender/SGLHelps.cpp:27:1: note: in expansion of macro `
`__SK_FORCE_IMAGE_DECODER_LINKING'`
` __SK_FORCE_IMAGE_DECODER_LINKING;`
 `^`
`cc1plus: all warnings being treated as errors`

PR #1015, opened 2018-09-25
1011 unused var

tst-twilk:
#1011 
Silence error coming from SGLHelps.cpp. `linking_forced` variable included with `#include <SKForceLinking.h>` in SGLHelps.cpp is not referenced  causing an unused variable warning. The warning is treated as en error when building on Ubuntu 16.04 with gcc 5.5.0.

To verify: 
- build open on Ubuntu using gcc 5.5.0 and make sure that build succeeds.


earth_enterprise/src/fusion/gst/maprender/SGLHelps.cpp
```diff
@@ -15,7 +15,11 @@
 
 #include "fusion/gst/maprender/SGLHelps.h"
 
+#pragma GCC diagnostic ignored "-Wunused-variable"
```
tst-dsugar: 10 commits one file change??
Should not the -Wunused-variable be inside the push/pop??

Issue #938, opened 2018-07-27
The `opengee-fusion` RPM is Missing a Dependency on `which`


tst-ppenev:
## Steps to Reproduce

1. Build Open GEE RPMs.
1. Have a Cent OS 7 instance that doesn't have the `which` command installed, such as Docker's `centos:7` image.
1. Install the Open GEE RPMs.
1. Run `/etc/init.d/gefusion status`.

## Observed Behavior

```
/etc/init.d/gefusion: line 78: which: command not found
gesystemmanager:  Not running.
/etc/init.d/gefusion: line 78: which: command not found
geresourceprovider:  Not running.
```

## Expected Behavior

```
gesystemmanager:  Running (PID 504).
geresourceprovider:  Running (PID 508).
```

## Notes

This issue can be resolved by adding a dependency on the `which` command, or by refactoring the `init.d` script to use the `type` shell build-in.  It should make little difference, but get some agreement before implementing.

tst-ccamp:
I vote for switching to the `type` command. The fewer the dependencies, the better.

tst-ccamp:
This was previously fixed in #1019, so closing

Issue #761, opened 2018-03-21
Package Google Earth's Bundled Version of PostGIS for Ubuntu 16


tst-ppenev:
As a Google Earth user, I want to be able to install Server from a Debian package manager on Ubuntu 16.  Currently, Earth Server bundles PostGIS.  However, PostGIS is distributed under a more restrictive license than OpenGEE, and, therefore, should be packaged separately.  There are 3 tickets (sub-tasks) for packaging Earth Server:

* Package libraries and utilities common to Fusion and Server.
* **Package PostGIS. [This ticket.]**
* Package the rest of Earth Server. (#760)

## AC

* Update the `earth_enterprise/rpms/build.gradle` script to output a Debian package for Google Earth Server's version of PostGIS.
* The package should install on Ubuntu 16.
* The package should have the correct dependencies, so it can be installed by a single dpkg -i <package-name>.deb command on a minimal Ubuntu system (e.g., Docker ubuntu image). [Can be split into a sub-task.]
* Rename the `rpms` directory to, e.g. `packages`, if its name is still RPM-specific. [This name can change to gain consensus.]

## Guidelines

* Share as much install script code as possible between RPM and Debian packages.  There are `pre-instal.sh`, `post-install.sh`, etc. scripts under the `opengee-common`, `opengee-fusion` and `opengee-server` that are used in the RPM packages.  Examples of techniques already used:
    * You can write shell script that detects platform-specific tools at run-time, and uses what's available.
    * You can use Groovy templates that expand variables, conditionally include or remove code blocks, or conditionally include other shell script files.
* There are Groovy classes under `rpms/buildSrc/src` that call `/usr/lib/rpm/find-provides` and `/usr/lib/rpm/find-requires`. Those scripts are included in the `rpm` Debian package, and do run on Debian systems like Ubuntu.  It may be possible to process their output on Ubuntu to auto-generate package dependencies.  However, that should be left as a separate task.  As a first iteration you can hard-code the Debian package dependencies.

PR #1019, opened 2018-09-26
#968: Fix Fusion Service Scripts on Red Hat and Debian Platforms

### LINKED ISSUE #968: `/etc/init.d/gefusion stop` Does Not Ensure Daemons Stop
If `gesystemmanager` is running a long process (such as planning a project build) and a user runs `/etc/init.d/gefusion stop`, `gesystemmanager` will not stop until it finishes its queued tasks; however, the `gefusion` script will return a status code indicating that `gesystemmanager` has finished running (even though it hasn't) and delete the PID file.

This may have been introduced when the signal handling was fixed in 5.2.4.

## Notes:

* I think the bug existed before 5.2.4.
* Before 5.2.4, trapping the SIGTERM signal was buggy, so `gesystemmanager` was killed instead of shutting down.  That made this bug much less obvious, since the service terminated quickly.

##### `gestopdaemon`

The current `init.d` scripts use `gestartdaemon` and `gestopdaemon`.  The `gestopdaemon` utility does the following:

```C++
  if (khExists(pidfile)) {
    FILE *pidFILE = fopen(pidfile.c_str(), "r");
    if (pidFILE) {
      int pid;
      fscanf(pidFILE, "%d\n", &pid);
      fclose(pidFILE);

      pid_t pgid = getpgid(pid);
      if ((pgid > 0) && (killpg(pgid, SIGTERM) == 0)) {
        khUnlink(pidfile);
        return 0;
      }
    }
    khUnlink(pidfile);
  }
```

As you can see, there's no error handling.
* If opening the PID file fails, the PID file is deleted without any signals being sent, as if the service is no longer running.  And no error is reported.
* There's no error handling for having invalid contents in the PID file.
* A signal is sent to the service process group, and no waiting or verification is done to see if the service processes terminate, or not.  The PID file is immediately removed and the utility returns a success exit code, as if the service is no longer running.

##### `gesystemmanager` and `geresourceprovider`

The `gesystemmanager` and `geresourceprovider` commands are not written as daemons.  They're written as command line applications, and rely on an external utility to turn them into a daemon (fork them in the background, detach them from the console, write PIDs to a file).

#### Possible Switch to LSB Functions

The Linux Standard Base (http://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/iniscrptfunc.html) has functions for starting and stopping services that should work on all Linux platforms.  Unfortunately, these functions have rather few features, so using them still requires re-implementing the functionality in shell script, or using another library.

`start_daemon` doesn't support the following adjustments which are done by `gestartdaemon`:

* Specifying a user and a group to run the service as, instead of the default `root`.
* Specifying a `umask` for the daemon process.
* Detaching a process from the console, and running it in the background.
* Writing the PID of a detached process to the PID file.

#### Possible Switch to `libdaemon`

There is more than one `libdaemon`:

* http://git.0pointer.net/libdaemon.git
* https://github.com/kisom/libdaemon

Both of the ones above have similar functionality to `gestartdaemon` and `gestopdaemon`, and neither is fully configurable.

The first one doesn't seem to support specifying a user and a group to run a daemon under.  The `umask` seems to be hardcoded to 0022 when executing a command as daemon, and to 0077 when forking the current process as a daemon. The working directory is hard coded to `/`.

The second one doesn't seem to support configuring the path to the PID file, executing a command as a daemon, or redirecting `stdout` and `stderr` to a log.  It also seems to delete the PID file before the daemon terminates.

##### Red Hat `initscripts`

Red Hat systems have shell functions in `/etc/rc.d/init.d/functions`.

The Red Hat `initscripts` `daemon` function also lacks a few features on Cent OS 6:

* Setting a group to run a deamon process under.
* Specifying a working directory for the daemon process.
* Specifying a `umask` for the daemon process.
* Forking a command in the background.
* Writing a PID file.

##### `daemonize` (Packaged for Red Hat)

The [daemonize command](https://github.com/bmc/daemonize) is packaged for Red Hat 6 and later distributions, and has a good number of features.  What it lacks is:

* Setting a group to run a daemon process under.
* Specifying a `umask` for the daemon process. (`umask(0)` is hard-coded.)

##### Debian `start-stop-daemon`

Debian-based systems use `start-stop-daemon`.  That utility supports all of features used by the Fusion services, except:

* Redirecting the daemon process's standard output and standard error to a log. (But it's trivial to add to a daemon start command.)

## AC:

* Ensure service scripts wait for `gesystemmanger` and `geresourcepvorider` terminate, or time out before returning a correct status code.
* Provide a mechanism to allow users to stop gesystemmanager immediately (or at least within a few seconds) if desired

tst-ppenev:
* Switched to using `daemonize` on Red Hat, and `start-stop-daemon` on Debian.
* Removed `gestartdaemon` and `gestopdaemon`.
* Fixed Fusion and Server init script default run levels, service dependencies and descriptions.
* Fixed package init scripts, so they detect the installation stage on both RPM and Deb systems.
* Replaced some BASH-specific commands with POSIX ones.
* Consolidated template file expansion in copy tasks.
* Switched to expanding install scripts to single files to avoid OS Package issues. (Deb packaging adds extra code around install commands which are not specified as a single file, which makes them run only during a specific install stage.  RPM packaging doesn't work with specifying a single file as an install script.)
* Added a temporary fix for running Gradle package tasks before building and staging OpenGEE with other commands.
* Added a fix for binary directories being symlinks (on CentOS 7) when adding a command as an RPM dependency.  (Now a path is converted to a path that is installed by an RPM.)
* Added fixes to `GeeDeb` tasks, so Deb tasks can use the same CPU architecture constant, `description` value, and `requiresCommands` as RPMs.

Fixes #968, #938, #761.


earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -115,26 +115,30 @@ prerequisites will be missing and the build will fail.
...
-  libXinerama-devel libXft-devel libXrandr-devel libXcursor-devel gdbm-devel \
-  libmng-devel libcap-devel libpng12-devel libXmu-devel freeglut-devel \
-  zlib-devel libX11-devel bison-devel openjpeg-devel openjpeg2-devel \
-  geos-devel proj-devel ogdi-devel giflib-devel xerces-c xerces-c-devel \
-  cmake GitPython rpm-build rpmrebuild rsync
+sudo yum install -y \
+  bison-devel cmake daemonize freeglut-devel \
+  gdbm-devel geos-devel giflib-devel GitPython \
+  libcap-devel libmng-devel libpng12-devel libX11-devel libXcursor-devel \
+    libXft-devel libXinerama-devel libxml2-devel libXmu-devel libXrandr-devel \
```
tst-lsavoie: Minor: inconsistent indentation

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -115,26 +115,30 @@ prerequisites will be missing and the build will fail.
...
-  libXinerama-devel libXft-devel libXrandr-devel libXcursor-devel gdbm-devel \
-  libmng-devel libcap-devel libpng-devel libXmu-devel freeglut-devel \
-  zlib-devel libX11-devel bison-devel openjpeg-devel openjpeg2-devel \
-  geos-devel proj-devel ogdi-devel giflib-devel xerces-c xerces-c-devel \
-  cmake GitPython rpm-build rpmrebuild rsync shunit2 python-unittest2
+sudo yum install -y \
+  bison-devel  cmake daemonize freeglut-devel \
+  gdbm-devel geos-devel giflib-devel GitPython \
+  libcap-devel libmng-devel libpng-devel libX11-devel libXcursor-devel \
+    libXft-devel libXinerama-devel libxml2-devel libXmu-devel libXrandr-devel \
```
tst-lsavoie: Minor: inconsistent indentation

tst-lsavoie:
> Added a temporary fix for running Gradle package tasks before building and staging OpenGEE with other commands.

@tst-ppenev is this temporary fix still in the code? Should it be removed, or is it meant to be permanent?

tst-lsavoie:
@tst-ppenev there are a lot of commits that aren't from you. Is it possible some commits from master will be merged into release_5.2.4 from this PR? In any case, it may be wise to rebase on the latest 5.2.4.

tst-ppenev:
> > Added a temporary fix for running Gradle package tasks before building and staging OpenGEE with other commands.
> 
> @tst-ppenev is this temporary fix still in the code? Should it be removed, or is it meant to be permanent?

Yes, it is.  There's another temporary fix that I think was merged to `master`.  There may be a better fix coming later, but since I don't see another fix on this branch, I think it's OK to keep it.  We'll have to resolve fixes with `master` later.

tst-ppenev:
> @tst-ppenev there are a lot of commits that aren't from you. Is it possible some commits from master will be merged into release_5.2.4 from this PR? In any case, it may be wise to rebase on the latest 5.2.4.

@tst-lsavoie, I'm not sure what you're asking.  This was rebased on `release_5.2.4`.  A lot of the commits got modified, because of resolving merge conflicts.  That's why they appear on the pull request.

earth_enterprise/BUILD_RHEL_CentOS.md
tst-lsavoie: There's no clear reason (at least to me) why this one line should be a sub-level when nothing else is.

tst-lsavoie:
I always forget this part - we need to add something to the release notes for this fix.

earth_enterprise/BUILD_RHEL_CentOS.md
tst-ppenev: They're ordered alphabetically.  And, this continues the same letter, rather than starting a new one.

tst-ppenev:
I see a problem with expanding `openGeeVersion`.  It's expanding to `null` in `gevars.sh`.

tst-ppenev:
> I always forget this part - we need to add something to the release notes for this fix.

Does there need to be a new copy of the documentation for the next release to add release notes?

tst-lsavoie:
A new copy has already been made on the master branch. If it's cleaner to do that in a separate PR rather than merge master into this branch, that's fine with me.

tst-ppenev:
Merged with `master` and updated release notes for 5.2.5.

PR #1022, opened 2018-09-26
Release 5.2.4

tst-twilk:
Merge latest release_5.2.4 changes into master.


tst-twilk:
These commits were already approved in previous pull requests and are already included within the repo, so additional authorization from authors is not required.

PR #1024, opened 2018-09-27
Reverting to git describe (from git tag) to determine most recent version tag

tst-kfreiheit:
Using --first-parent parameter of git describe to accurately detect latest reachable tag on branch, removing call to git tags --merged (requires git 2.7.4)

### DIFF
--- a/earth_enterprise/src/scons/getversion.py
+++ b/earth_enterprise/src/scons/getversion.py
@@ -36,49 +36,47 @@ def _GitGeneratedLongVersion():
     """Take the raw information parsed by git, and use it to
        generate an appropriate version string for GEE."""
 
+    raw = _GetCommitRawDescription()
 
     # For tagged commits, return the tag itself
-    if _IsCurrentCommitTagged():
-        return _VersionForTaggedHead()
+    if _IsCurrentCommitTagged(raw):
+        return _VersionForTaggedHead(raw)
     else:
-        return _VersionFromTagHistory()
-
-
-def _IsCurrentCommitTagged():
-    """True if the current commit is tagged, otherwise False"""
-    raw = _GetCommitRawDescription()
-    # If this condition hits, then we are currently on a tagged commit.
-    return (len(raw.split("-")) < 4)
+        return _VersionFromTagHistory(raw)
 
 
 def _GetCommitRawDescription():
     """Returns description of current commit"""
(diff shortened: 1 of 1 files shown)


tst-ccamp:
Did you try the tag format `5.2.5-0.alpha`? That would be more consistent with the beta and final formats.

tst-kfreiheit:
Yes, the 5.2.5-0.alpha tag also works, but the next commit produces 5.2.5-1.alpha version.

Thinking some more, the 5.2.5-0 tag would produce a build with 5.2.5-0 as its version...introducing a truly one-off inconsistently named build.

I've changed the tag to be 5.2.5-0.alpha.  All builds will be named with the consistent format that way.  Code tweaked a bit as well.

tst-dsugar:
I recall I used to use "git describe --abbrev=0 --tags" to get the last tag in the current branch.

tst-rwildes:
Does this need to target release 5.2.4?

tst-kfreiheit:
Here's how --first-parent alters git describe:

> Created two branches: "fake_master" and "fake_release_5.2.6".
> Starting here on "fake_master", after creating a tag named "5.2.6-0.alpha",
> and committing some fake work on it:

# git describe --tags --match '[0-9]*\.[0-9]*\.[0-9]*\-*'
5.2.6-0.alpha-1-g2e28fde

> The fake_release_5.2.5 branch has a few commits on it, the last is tagged "5.2.5-1.final".

# git merge fake_release_5.2.5
Already up-to-date! (note: all commits were empty commits to fake work being done on both branches)
Merge made by the 'recursive' strategy.


> THE PROBLEM:
> we're on the fake_master branch, which is 5.2.6 development.
> All we just did was merge 5.2.5-1.final into the ongoing 5.2.6 work...
> so we should still be on version 5.2.6-*
> .. but git describe says 5.2.5-1.final:

# git describe --tags --match '[0-9]*\.[0-9]*\.[0-9]*\-*'
5.2.5-1.final-3-gd37b1eb

> THE SOLUTION:
> adding --first-parent pins git describe to stay on the 'native' branch we're on, and
> disregard tags from merges:

# git describe --first-parent --tags --match '[0-9]*\.[0-9]*\.[0-9]*\-*'
5.2.6-0.alpha-2-gd37b1eb

PR #1025, opened 2018-09-27
Reverting change to version string logic to resolve git version issues

tst-kfreiheit:


tst-kfreiheit:
Temporarily reverting version string generation logic on master to resolve environmental issues with git versioning.

tst-kfreiheit:
Upgraded build environments with more recent git, keeping change.

PR #1030, opened 2018-09-28
#968 TEST-PR to fix Jenkins CI builds with PR-1019 

tst-dsheremata:
The gradle openGeeRpms fails on Jenkins CI environments, complaining that it cannot find the "daemonize" command.  This is likely because of the way that we are trying to locate executables as part of the packaging process - so this experiment is to see if changing the GeeCommandLine's "resolveCommandPath(command_name" function will locate "daemonize" by using a bash shell with the "login" shell option. 

> diff --git c/earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy w/earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy
index def43c6..d57cb22 100644
> --- c/earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy
> +++ w/earth_enterprise/rpms/buildSrc/src/main/groovy/org/opengee/shell/GeeCommandLine.groovy
> @@ -68,7 +68,7 @@ class GeeCommandLine {
>      // Returns the path to a given command found in a standard system PATH:
>     static def resolveCommandPath(command_name) {
>         return expand(
> -                ["bash", "-c", "type -P -p ${command_name}"],
> +                ["bash", "-l", "-c", "type -P -p ${command_name}"],
>                 "Failed to find command ${command_name} in current PATH!"
>             ).readLines()[0]
>     }


tst-ppenev:
Thanks for testing that, @tst-dsheremata.  It's almost funny how much work gets done just to reset the `PATH` variable to a system default, but, at least, it's not much code.

Can you target this pull request to my branch `tst-ppenev:I968-switch-fusion-daemon-start-and-stop-to-existing-commands`?  Otherwise, I can try to merge the last commit on my branch.

tst-dsheremata:
Also, the gefusion service no longer respects the `reload` option, as documented [here](http://www.opengee.org/geedocs/answer/7160005.html).  Suggest that this PR either restores this option to the script or alternatively update the documentation to use an alternate method of reloading the logLevel thresholds from systemrc.

tst-dsheremata:
Retargeted this fix to issue #968 the ongoing PR from @tst-ppenev  - closing this one.

PR #1034, opened 2018-10-01
Issue #1033: Don't handle SIGTERM/SIGINT

### LINKED ISSUE #1033: Kill gesystemmanager when user runs "/etc/init.d/gefusion stop"
Prior to GEE 5.2.4, gesystemmanager had signal handling code that looks like it caught SIGTERM/SIGINT signals, but it actually didn't work. Thus, when "/etc/init.d/gefusion stop" sent a sigterm to gesystemmanager it died immediately. In 5.2.4 we fixed that code, so now gesystemmanager finishes what it is doing before stopping. However "finishing what it is doing" can take days for large builds. As a short term fix for 5.2.4 we should stop handling SIGTERM/SIGINT signals so that gesystemmanager stops immediately when the user tells it to. We will implement a more elegant fix in 5.2.5 via issue #968.

The PR that fixed the signal handler is https://github.com/google/earthenterprise/pull/940/files. However, we cannot revert this entire PR because we still want the dynamic logging level capability, which uses the SIGHUP signal.

AC:
- gesystemmanager does not handle SIGTERM or SIGINT. It should, however, continue to handle SIGHUP.

tst-lsavoie:
Fixes #1033. For details, see that ticket.

To test:
1. Add a long sleep command to BuildAsset in earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp.
1. Start building an asset.
1. Run `/etc/init.d/gefusion stop` before the sleep command in step 1 finishes.
1. Run `ps aux | grep gesystemmanager` and verify that gesystemmanager is not running.
1. Run `/etc/init.d/gefusion start` and `/etc/init.d/gefusion reload`. Check the bottom of /opt/google/log/gesystemmanager.log and verify that a message similar to the one below appears at the bottom of the log:
```
[2018-10-01 10:25:50] Warning: Recieved SIGHUP, Reloading systemrc...
[2018-10-01 10:25:50] Warning: system log level changed to: NFY_<some notification level>
```

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/sysman/khSystemManager.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/khSystemManager.cpp
@@ -95,8 +95,18 @@ void handleExitSignals(int signum)
 void
 khSystemManager::SignalLoop(void)
 {
-  signal(SIGINT,handleExitSignals);
-  signal(SIGTERM,handleExitSignals);
+  // In Fusion versions 5.2.3 and earlier there was code to handle SIGTERM and
+  // SIGINT, but this code was not called when the relevant signals were sent
+  // to system manager. We fixed this in 5.2.4 with the unexpected side effect
+  // that gesystemmanager may not stop when the user runs
+  // "/etc/init.d/gefusion stop". Instead, gesystemmanager will catch the
+  // SIGTERM and wait until it finishes the current operation before stopping,
+  // which could take days. As a stopgap, we are reverting to the behavior
+  // from 5.2.3 and not handling SIGINT and SIGTERM. In 5.2.5 we will revisit
+  // this and implement better methods of starting and stopping system manager.
+  // At that point, the two lines below can be uncommented.
+  //signal(SIGINT,handleExitSignals);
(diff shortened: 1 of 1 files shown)


tst-basilhuffman:
@tst-lsavoie would way to exit somewhat gracefully i.e. through SetWantExit( ), but short-circuiting somewhat be better? If not, I approve this. Has the same functionality as before.

tst-blaird:
Verified successfully on Ubuntu 14.

tst-lsavoie:
@tst-nfarah I've updated the PR to fix the spelling errors.

tst-lsavoie:
@tst-nfarah the pop up is expected. I should have added that to the repro steps.

tst-lsavoie:
@tst-nfarah I'm going to argue that this doesn't need an entry in the release notes because it's restoring the same behavior as previous versions (i.e., it's fixing a bug that was introduced in this version).

tst-nfarah:
Ok @tst-lsavoie

tst-lsavoie:
@tst-basilhuffman: @tst-ppenev is working on a more graceful method in #968. This is just a quick-and-dirty fix for 5.2.4.

PR #1040, opened 2018-10-02
Serve Wms From gepublish

hepek:
I've added an option to gepublish to pass ServeWms flag from the command line.

Don't know if this'll be useful to somebody or there's a better way to do it programmatically (without having to visit the admin panel).

### DIFF
--- a/earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
+++ b/earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
@@ -405,7 +405,7 @@ int main(int argc, char* argv[]) {
       throw khException("Enhanced search cannot be enabled without POI search.\n");
     }
 
-    if (serve_wms && !publishdb) {
+    if (serve_wms && publishdb.empty()) {
       throw khException("serve_wms can only be used with --publishdb.\n");
     }
 
--- a/earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
+++ b/earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
@@ -101,6 +101,7 @@ void usage(const std::string &progn, const char *msg = 0, ...) {
     "                            contains POI data.\n"
     "   [--enable_enhancedsearch]If POI search is enabled, enable enhanced\n"
     "                            search.\n"
+    "   [--serve_wms]            Serve content via WMS.\n"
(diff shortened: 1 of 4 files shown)


hepek:
I signed it!

earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
```diff
@@ -350,6 +352,7 @@ int main(int argc, char* argv[]) {
     options.flagOpt("setecdefault", ec_default_db);
     options.flagOpt("enable_poisearch", enable_poisearch);
     options.flagOpt("enable_enhancedsearch", enable_enhancedsearch);
+    options.flagOpt("serve_wms", serve_wms);
```
tst-lsavoie: We probably ought to have a check at some point to verify that serve_wms is not set unless publishdb is set.

tst-lsavoie:
@hepek thanks for the contribution. From a quick glance it looks good. We'll work on getting it reviewed and tested more thoroughly.

earth_enterprise/src/fusion/gepublish/geserveradmin.cpp
hepek: This option is only used in PublishDatabase call. See line 654.

Did you mean to check this and report to the user that an invalid option combination was set?

tst-lsavoie: Yes, that's the idea. I like having a check at the beginning that verifies that invalid option combinations are not used.

hepek: Added code at line 408.

tst-basilhuffman:
Will test as well.

tst-lsavoie:
@hepek I always forget this part - we like to update the release notes with every PR. The file is here: https://github.com/google/earthenterprise/blob/master/docs/geedocs/5.2.5/answer/7160006.html. Would you add this change to the new features?

tst-basilhuffman:
Runs ok also. Good work.

tst-lsavoie:
I'm OK with including my commits in this PR.

tst-ccamp:
We need to retroactively add an issue for this for tracking

tst-ccamp:
I added and linked a ticket. I didn't fill in any details though

PR #1044, opened 2018-10-04
Issue 1020: optimize state updates

### LINKED ISSUE #1020: State updates are slow for large builds
We have found that for very large builds, updating the state of a project can take a very long time. For builds with thousands of resources, this can take days. This is particularly prevalent when running gepackgen because packgen tasks have lots of dependencies. We've seen this happen when a gepackgen task fails and a Blocked state needs to be propagated to all dependencies, and when cancelling a project.

To reproduce:
1. Create an imagery project with thousands of resources (around ~6000 seems to work best). It may help to have resources that overlap with one another.
1. Build the project. Allow the project to build until gepackgen tasks start executing.
1. Cancel the project from the Fusion UI. The UI will freeze, gesystemmanager will be running on 1 core at around 100% utilization, and no other Fusion tasks will be using much CPU. This will continue for a long time, possibly several days. Note: a similar effect happens when you use the gecancel command instead of the Fusion UI - the only difference is that the UI does not lock up.

tst-lsavoie:
Fixes #1020 

This PR improves the performance of large builds by consolidating some of the notifications that an asset's child or input has changed. By preventing (some) duplicate notifications we reduce the number of times each asset has to calculate its state, significantly improving performance.


earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -215,52 +227,156 @@ AssetVersionImplD::SyncState(void) const
...
 void
-AssetVersionImplD::PropagateStateChange(void)
-{
-  notify(NFY_VERBOSE, "PropagateStateChange(%s): %s",
-         ToString(state).c_str(), GetRef().c_str());
-  for (std::vector<std::string>::const_iterator p = parents.begin();
-       p != parents.end(); ++p) {
-    AssetVersionD parent(*p);
+AssetVersionImplD::PropagateStateChangeToParents(
+    const std::set<std::string> & toNotify,
```
tst-ppenev: Similar comment as above.  Variable names `noNotify` and `notifySet` lack descriptiveness, and it's not clear why someone trying to propagate a state change to parents needs to know about notify sets and their details.

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -215,52 +227,156 @@ AssetVersionImplD::SyncState(void) const
...
-  // HandleInputStateChange ends up calling a DelayedBuildChildren routine
-  // which build new versions with me as an input.
-  std::vector<std::string> toNotify = listeners;
-
-  for (std::vector<std::string>::const_iterator l = toNotify.begin();
-       l != toNotify.end(); ++l) {
-    AssetVersionD listener(*l);
+void
+AssetVersionImplD::PropagateStateChangeToListeners(
+    const std::set<std::string> & toNotify,
```
tst-ppenev: Similar comment.

tst-lsavoie:
I've now run a number of tests dealing with different state updates and outputs match the original code. I'm now moving this out of WIP.

tst-lsavoie:
@tst-ppenev I think I've addressed your concerns. Take another look.

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -132,11 +134,19 @@ AssetVersionImplD::StateByInputs(bool *blockersAreOffline,
                            AssetDefs::Canceled |
                            AssetDefs::Bad)) {
         ++numblocking;
+        // At this point we already know what the values of all of the outputs
```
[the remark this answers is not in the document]
tst-ppenev: Ah, yes.  You're right.

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -18,11 +18,118 @@
...
+#include "fusion/autoingest/MiscConfig.h"
+
+// ****************************************************************************
+// ***  StateUpdateNotifier
+// ****************************************************************************
+std::shared_ptr<AssetVersionImplD::StateChangeNotifier>
+AssetVersionImplD::StateChangeNotifier::GetNotifier(
+    AssetVersionImplD * const assetVersion,
+    std::shared_ptr<StateChangeNotifier> callerNotifier) {
+  if (callerNotifier) {
```
tst-ppenev: I think you could even turn this into a singleton, and use one global `StateChangeNotifier` instance for all state changes.  Then, you wouldn't need to pass a `callerNotifier` instance around, either.  But, I think this would be OK for now.

tst-lsavoie: My concern about the singleton idea is that this code is written to be multi-threaded. I know the way the implementation works right now is effectively single threaded, but I don't want to bury a nasty bug in here for someone to find later.

One other point I'd like to make - it's not possible to save up all the notifications and do them at once, at least the way the code is currently written. When something is notified, it may update its state, which may cause something else to be notified. So every notification may produce more notifications. We could fix that as well, but it would be a much bigger refactor.

tst-ppenev:
@tst-lsavoie: I assume you see this, but you also need to resolve conflicts before this request can be merged.

tst-lsavoie:
Merge conflicts have been resolved.

tst-lsavoie:
Also added release notes updates for this ticket and a previous ticket that had been missed.

PR #1049, opened 2018-10-08
Issue #622 Change default API behavior

### LINKED ISSUE #622: Queries for Map Definitions should return proper JSON in GEE Server (Apache)

williamstrong:
Fixes #622

This PR includes: changing the default behavior when requesting json data for any given map and allows users to specify "v=1" in the query string to retrieve the data in the previous format.

I choose to split the versions into different files so that it is simple to remove the older (V1) logic and easier to maintain the new logic. 

Currently the code (dbmanifest.cpp 972...) allowing users to specify which version they want to use is not working and only the new format is being served.

### DIFF
--- a/docs/geedocs/5.2.5/answer/1089795.html
+++ b/docs/geedocs/5.2.5/answer/1089795.html
@@ -56,6 +56,13 @@ var painted_map_layer = new GFusionTileLayer(geeServerDefs.serverUrl, geeServerD
 var labels_layer = new GFusionTileLayer(geeServerDefs.serverUrl, geeServerDefs.layers[2]);</code></p>
 <p>You can access the server definitions and other information about map and imagery layers directly from a web browser by using a URL in this format: <code>http://yourserver.org/virtual-server-name/query?request=Json&amp;vars=geeServerDefs</code>. Responses will look like the image below:</p>
 <p><img alt="GEE server definitions" src="../art/fusion/projects/2972270_server_definitions.jpg"/></p>
+<p>This data can also be returned by using the optional <code>v</code> parameter.o</p>
+<ul>
+<li><code>v=1</code> returns the data in the format shown in the image above.</li>
+<li><code>v=2</code> returns the data in the JSON format.</li>
+<li> A query without the <code>v</code> parameter acts as <code>v=1</code>.</li>
+</ul>
+<p>A example of a query using the <code>v</code> parameter: <code>http://&lt;GEE earth server&gt;/&lt;database_publish_point&gt;/query?request=Json&amp;v=2</code>
 </li>
 <li>Associate layers with buttons by using the <code>GMapType</code> class to define a map type for each of the <code>GTileLayer</code> objects. See the example below:
 <p><code>var painted_map_type = new GMapType([painted_map_layer],new LatLngProjection(MAX_ZOOM_LEVEL),"Maps");<br/>
--- a/docs/geedocs/5.2.5/answer/7160006.html
+++ b/docs/geedocs/5.2.5/answer/7160006.html
@@ -43,8 +43,7 @@ gtag('config', 'UA-108632131-2');
   <h5>
     <p id="overview_5.2.5">New Features</p>
   </h5>
-    <p><strong>C++98 no longer supported</strong>. As of 5.2.4 C++11 syntax can be used and building open GEE using C++98 standard will no longer be supported</p>
-
+  <p><strong>When a GEE server is queried for server definitions the response can be returned in proper JSON format.</strong> The <code>v</code> parameter can be added to query string to specify which version of the API will be returned. The valid options are <code>v=1</code> and <code>v=2</code>. Version one returns data in the current format. Version two returns data in JSON format. An example url which returns data in JSON format looks like: <code>http://&lt;GEE earth server&gt;/&lt;database_publish_point&gt;/query?request=Json&amp;v=2</code>.</p>
   <h5>
     <p id="supported_5.2.5">Supported Platforms</p>
   </h5>
--- a/earth_enterprise/src/common/serverdb/serverdbReader.cpp
+++ b/earth_enterprise/src/common/serverdb/serverdbReader.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 
+#include <array>
 #include "common/serverdb/serverdbReader.h"
 #include "keyhole/jpeg_comment_date.h"
 #include "common/quadtreepath.h"
@@ -28,6 +29,53 @@
 #include "common/proto_streaming_imagery.h"
 #include "common/geGdalUtils.h"
 #include "common/generic_utils.h"
+
+// TODO this code is from fusion/JsUtils and needs to be refactored to remove code duplication.
+namespace {
+  std::string JsToJson(const std::string& in)
+  {
+    std::ostringstream oss;
+
+    const std::string alphaNumSet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ_0123456789-abcdefghijklmnopqrstuvwxyz.";
+    const std::string valueSet = "0123456789.eE";  // .eE should only appear once but let something else enforce that
+
+    for (size_t i = 0; i < in.length(); ++i) {
+      if (isalpha(in[i])) {
+        static const std::array<std::string, 3> jsonLiterals = {"true", "false", "null"};
+        bool isLiteral = false;
+        for (const std::string& jlit : jsonLiterals) {
+          if (!jlit.compare(in.substr(i, jlit.size()))) {
+            oss << jlit;
+            i += jlit.size()-1;
+            isLiteral = true;
+          }
+        }
+
+        if (!isLiteral) {
+          size_t endEscape = in.find_first_not_of(alphaNumSet, i);
+          oss << "\"" << in.substr(i, endEscape-i) << "\"";
+          i = endEscape - 1;
+        }
+      }
+      else if (isdigit(in[i])) {
+        size_t endValue = in.find_first_not_of(valueSet, i);
+        oss << in.substr(i, endValue-i);
+        i = endValue - 1;
+      }
+      else if (in[i] == '\"') {
+        size_t escapedStrEnd = in.find("\"", i+1);
+        std::string temp = in.substr(i, escapedStrEnd-i+1);
+        oss << in.substr(i, escapedStrEnd-i+1);
+        i = escapedStrEnd;
+      }
+      else {
+        oss << in[i];
+      }
+    }
+
+    return oss.str();
+  }
+}
 
 int ServerdbReader::CheckLevel(const geindex::TypedEntry::ReadKey &read_key,
                                uint32 level, uint32 row, uint32 col) {
@@ -395,9 +443,21 @@ const MimeType ServerdbReader::GetData(
       }
     } else if (request == "Json") {
       std::string json_variable_name = arg_map["var"];
-      // CheckJavascriptName(json_variable_name);
-      GetJson(json_variable_name, language, region, buf);
-      return MapTileUtils::kJavascriptMimeType;
+      if (!arg_map["v"].empty()) {
+        // CheckJavascriptName(json_variable_name);
+        if (arg_map["v"] == "2") {
+          GetJson(json_variable_name, language, region, buf);
+        } else if (arg_map["v"] == "1") {
+          // This is added for backwards compatibility reasons.
+          GetJsonV1(json_variable_name, language, region, buf);
+        } else {
+          throw khSimpleException("Invalid argument found in v parameter: " + arg_map["v"] + ". Valid values are 1 and 2.");
+        }
+        return MapTileUtils::kJavascriptMimeType;
+      } else {
+        GetJsonV1(json_variable_name, language, region, buf);
+        return MapTileUtils::kJavascriptMimeType;
+      }
     } else if (request == "LayerDefs") {
       if (GetToc(language, region,
                  std::string() /* output= not supported for LayerDefs */,
@@ -577,6 +637,21 @@ const MimeType ServerdbReader::GetResampledImageryMapsTile(
 }
 
 void ServerdbReader::GetJson(const std::string& json_variable_name,
+                             const std::string& language,
+                             const std::string& region,
+                             ReadBuffer& buf) {
+  const std::string& json = GetLocaleKey(json_, language, region);
+  if (json_variable_name.empty()) {
+    // Return raw JSON
+    buf.SetValue(JsToJson(json));
+  } else {
+    // Return a Javascript version of the JSON with the variable name declared.
+    std::string javascript = "var " + json_variable_name + " = " + json + ";\n";
+    buf.SetValue(javascript);
+  }
+}
+
+void ServerdbReader::GetJsonV1(const std::string& json_variable_name,
                              const std::string& language,
                              const std::string& region,
                              ReadBuffer& buf) {
--- a/earth_enterprise/src/common/serverdb/serverdbReader.h
+++ b/earth_enterprise/src/common/serverdb/serverdbReader.h
@@ -24,7 +24,6 @@
 #include "common/serverdb/serverdb.h"
 #include "common/serverdb/map_tile_utils.h"
 
-
 class ServerdbReader {
  public:
   typedef geindex::UnifiedReader::ReadBuffer ReadBuffer;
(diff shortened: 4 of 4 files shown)


tst-rwildes:
I will be pulling the code today and testing it.

earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
```diff
@@ -316,7 +317,7 @@ std::string JsonUtils::JsonObject(
     if (iter != field_map.begin()) {
       buffer << ",\n";  // Note: IE6 doesn't like trailing commas.
     }
-    buffer << iter->first << " : " << iter->second;
+    buffer << JsUtils::JsToJson(iter->first) << " : " << JsUtils::JsToJson(iter->second);
```
tst-rwildes: The JsToJson should be applied to the full object string not each field.

earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
```diff
@@ -316,7 +317,7 @@ std::string JsonUtils::JsonObject(
...
   }
   buffer << "\n}\n";
   return buffer.str();
```
tst-rwildes: I think it is here you want to apply JsToJson

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -962,18 +963,40 @@ void DbManifest::GetMapsJsonFiles(
...
 
+    // Find the version parameter.
+    std::string versionParam = "v=";
+    std::string query = qurl.query();
+    int vIndex = query.find(versionParam) + 3;
+    version = std::stoi(query.substr(vIndex,1));
+
+    // Validate the version number.
+    if (version != 1 || version != 2) {
+      version = 2;
```
tst-rwildes: The issue wasn't very clear here.  I fixed it to be more clear.  But the default for the first release after this code goes in should be 1 so there is no behavior change to current usages.  Then everyone has one release to convert to the new behavior before we change the default behavior in the next release after the next release.  So I think this should be one.

earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp
```diff
@@ -962,18 +963,40 @@ void DbManifest::GetMapsJsonFiles(
...
+  int version;
   for (int i = -1; i < static_cast<int>(locale_set.supportedLocales.size());
        ++i) {
     std::string locale = i < 0 ?
         kDefaultLocaleSuffix : std::string(locale_set.supportedLocales[i]);
 
+    // Find the version parameter.
+    std::string versionParam = "v=";
+    std::string query = qurl.query();
+    int vIndex = query.find(versionParam) + 3;
```
tst-rwildes: I think you should check for the fact find might not find the string `v=` before you add anything to it.  If find returns `std::string::npos` that means the call did not include the version parameter and we should use the default (1 in the first release we have this change and 2 in the next release after).

Also, should you not be adding `2` instead of `3`?

williamstrong:
I've got the changes added @

earth_enterprise/src/fusion/autoingest/SConscript
```diff
@@ -176,7 +176,7 @@ commoncpp = [
     'LayerConfigImpl.cpp', 'MapProjectConfigImpl.cpp',
     'QueryConfigImpl.cpp', 'AssetTraverser.cpp',
     'geAssetRoot.cpp', 'SearchTabReferenceImpl.cpp',
-    'JsonUtils.cpp', 'JsUtils.cpp'
+    'JsonUtils.cpp', 'JsonUtilsV1.cpp', 'JsUtils.cpp'
```
tst-rwildes: Just for clarity I'll ask this: Doing the V1 version of `JsonUtils.cpp` isn't really needed to fix issue #622, correct?

earth_enterprise/src/common/serverdb/serverdbReader.h
```diff
@@ -18,12 +18,15 @@
...
 #define GEO_EARTH_ENTERPRISE_SRC_COMMON_SERVERDB_SERVERDBREADER_H_
 
+#include <array>
 #include <string>
 #include <map>
 #include "common/geindex/Reader.h"
 #include "common/serverdb/serverdb.h"
 #include "common/serverdb/map_tile_utils.h"
 
+std::string JsToJson(const std::string&);
```
tst-rwildes: Probably for now don't expose this in a header.

earth_enterprise/src/common/serverdb/serverdbReader.cpp
```diff
@@ -29,6 +29,55 @@
 #include "common/geGdalUtils.h"
 #include "common/generic_utils.h"
 
+
+std::string JsToJson(const std::string& in)
```
tst-rwildes: put this in an anonymous namespace to emphasize it is only used within this file.

tst-rwildes: I'll do some refactoring later and move `JsUtils::JsToJson` to common and then convert this over to use that.

earth_enterprise/src/common/serverdb/serverdbReader.cpp
```diff
@@ -29,6 +29,55 @@
...
+        oss << "\"" << in.substr(i, endEscape-i) << "\"";
+        i = endEscape - 1;
+      }
+    }
+    else if (isdigit(in[i])) {
+      size_t endValue = in.find_first_not_of(valueSet, i);
+      oss << in.substr(i, endValue-i);
+      i = endValue - 1;
+    }
+    else if (in[i] == '\n') {
```
tst-rwildes: I think we should remove this.  The output should preserve new lines of the original java script.

tst-rwildes:
Pulled the latest code as of this morning and tested it.  It does work properly in that if a `v=2` parameter is used then valid JSON is returned instead of java script.  I had a few minor comments about the code and suggested a couple of small changes.  One of the changes will make the output a little more readable.  It looks like the JS -> JSON conversion isn't preserving new lines.

earth_enterprise/src/fusion/autoingest/JsonUtilsV1.cpp
```diff
@@ -0,0 +1,337 @@
+// Copyright 2017 Google Inc. Copyright 2018 the Open GEE Contributors.
```
tst-ccamp: Why was this done as a separate copy of the utility instead of just adding a parameter to the JsonObject method inside JsonUtils.cpp/h? Can you refactor it like that?

tst-rwildes:
Since we don't really need any of the changes in:

earth_enterprise/src/fusion/autoingest/SConscript
earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
earth_enterprise/src/fusion/autoingest/JsonUtilsV1.cpp
earth_enterprise/src/fusion/autoingest/JsonUtilsV1.h
earth_enterprise/src/fusion/dbmanifest/dbmanifest.cpp

maybe just revert them.  That will get rid of the duplicate code I think @tst-ccamp is concerned about.

earth_enterprise/src/fusion/autoingest/SConscript
williamstrong: That's correct. I left it in there from my previous attempt, but I will remove it.

williamstrong:
@tst-rwildes @tst-ccamp I have made all of the requested changes.

tst-rwildes:
I will re-test with latest changes shortly.  Glanced at the changes and they look good.  But will look more closely soon.

tst-rwildes:
@williamstrong for the release notes replace this line:

tst-rwildes:
Looks like this talks about the API in the help but not really in any detail...  Not sure if it is worth updating:

https://github.com/google/earthenterprise/blob/master/docs/geedocs/5.2.5/answer/1089795.html#L57

docs/geedocs/5.2.5/answer/7160006.html
```diff
@@ -43,8 +43,7 @@ <h2>Release notes: Open GEE 5.2.5</h2>
   <h5>
     <p id="overview_5.2.5">New Features</p>
   </h5>
-    <p><strong>C++98 no longer supported</strong>. As of 5.2.4 C++11 syntax can be used and building open GEE using C++98 standard will no longer be supported</p>
-
+  <p>Server definitions and other information can now be returned in JSON format. The <code>v</code> parameter can be added to query string to specify which version of the API will be returned. The valid options are <code>v=1</code> and <code>v=2</code>. Version one returns data in the current format. Version two returns data in JSON format. An example url which returns data in JSON format looks like: <code>http://yourserver.org/virtual-server-name/query?request=Json&amp;v=2</code>.</p>
```
tst-rwildes: bold the first sentience using the `<strong>` tag like the one you replaced did.

earth_enterprise/src/common/serverdb/serverdbReader.h
```diff
@@ -18,12 +18,14 @@
 #ifndef GEO_EARTH_ENTERPRISE_SRC_COMMON_SERVERDB_SERVERDBREADER_H_
 #define GEO_EARTH_ENTERPRISE_SRC_COMMON_SERVERDB_SERVERDBREADER_H_
 
+#include <array>
```
tst-rwildes: Probably should move the `#include <array>` to the `serverdbReader.cpp` file.

docs/geedocs/5.2.5/answer/7160006.html
tst-nfarah: may be for that first phrase use: <p><strong>When a GEE server is queried for Server Definitions the returned response is a valid JSON </string> ..... </p>

earth_enterprise/src/common/serverdb/serverdbReader.cpp
```diff
@@ -29,6 +29,52 @@
 #include "common/geGdalUtils.h"
 #include "common/generic_utils.h"
 
+namespace {
+  std::string JsToJson(const std::string& in)
```
tst-nfarah: Let's add a TODO comment stating where this code was copied [from](https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/fusion/autoingest/JsUtils.cp) and refactoring is needed to remove the code duplication.

earth_enterprise/src/common/serverdb/serverdbReader.cpp
```diff
@@ -396,7 +442,12 @@ const MimeType ServerdbReader::GetData(
     } else if (request == "Json") {
       std::string json_variable_name = arg_map["var"];
       // CheckJavascriptName(json_variable_name);
-      GetJson(json_variable_name, language, region, buf);
+      if (arg_map["v"] == "2") {
+        GetJson(json_variable_name, language, region, buf);
+      } else {
```
tst-nfarah: if one pass a different version from 1 or 2 (valid options) we still want to return the v 1 result ? Let's add validation for the passed argument.

tst-rwildes: `v` parameter is optional and if not include we want default behavior to use original behavior (`v=1` behavior).  We could raise an exception if the `v` parameter present and not a value of `1` or `2`.  But if `v` is not present we should assume `v=1`.

tst-nfarah:
> Looks like this talks about the API in the help but not really in any detail... Not sure if it is worth updating:
> 
> https://github.com/google/earthenterprise/blob/master/docs/geedocs/5.2.5/answer/1089795.html#L57

It would be good if we update that document

docs/geedocs/5.2.5/answer/7160006.html
tst-nfarah: in the given example at the end of the url the database publish entry need to be appended. So the request would be:
http://<GEE earth server>/<database publish point>/query?request=Json&amp;v=2
-->
http://yourserver.org/virtual-server-name/database_publish_point/query?request=Json&amp;v=2

docs/geedocs/5.2.5/answer/1089795.html
```diff
@@ -56,6 +56,13 @@ <h3>To define map types in GEE:</h3>
 var labels_layer = new GFusionTileLayer(geeServerDefs.serverUrl, geeServerDefs.layers[2]);</code></p>
 <p>You can access the server definitions and other information about map and imagery layers directly from a web browser by using a URL in this format: <code>http://yourserver.org/virtual-server-name/query?request=Json&amp;vars=geeServerDefs</code>. Responses will look like the image below:</p>
 <p><img alt="GEE server definitions" src="../art/fusion/projects/2972270_server_definitions.jpg"/></p>
+<p>This data can also be returned by using the optional <code>v</code> parameter.o</p>
+<ul>
+<li><code>v=1</code> returns the data in the format shown in the image above.</li>
+<li><code>v=2</code> returns the data in the JSON format.</li>
+<li> A query without the <code>v</code> parameter acts as <code>v=1</code>.</li>
+</ul>
+<p>A example of a query using the <code>v</code> parameter: <code>http://yourserver.org/virtual-server-name/query?request=Json&amp;v=2</code>
```
tst-nfarah: the url would need to have the published database entry in it. See my previous comment in the release notes about the url.

earth_enterprise/src/common/serverdb/serverdbReader.cpp
```diff
@@ -395,9 +443,21 @@ const MimeType ServerdbReader::GetData(
...
-      return MapTileUtils::kJavascriptMimeType;
+      if (!arg_map["v"].empty()) {
+        // CheckJavascriptName(json_variable_name);
+        if (arg_map["v"] == "2") {
+          GetJson(json_variable_name, language, region, buf);
+        } else if (arg_map["v"] == "1") {
+          // This is added for backwards compatibility reasons.
+          GetJsonV1(json_variable_name, language, region, buf);
+        } else {
+          throw khSimpleException("Illegal argument found in v parameter: ") << arg_map["v"];
```
tst-nfarah: I would change `Illegal` to `Invalid`. May be the following error message could be an alternative: `Invalid value for the the v parameter: ... A valid value must be either 1 or 2.`

tst-nfarah: How throwing this exception affect the calling code ? will it be caught and eventually the user see it ?

tst-rwildes:
Re-tested with latest changes.  I everything works as before except now when I use anything but `1` or `2` for the `v` parameter I get a 404 error and the error log has this in it:
```
[Fri Oct 12 15:09:25 2018] [error] [client 127.0.0.1] File does not exist: /opt/google/gehttpd/htdocs/favicon.ico, referer: http://centos7builder.raw.vm/SFDb_3d-v001/query?request=Json&v=2
[Fri Oct 12 15:09:41 2018] [error] [client 127.0.0.1] Caught exception in handle request: Error:Illegal argument found in v parameter: 3
[Fri Oct 12 15:09:41 2018] [error] [client 127.0.0.1] File does not exist: /opt/google/gehttpd/htdocs/public_host
```

earth_enterprise/src/common/serverdb/serverdbReader.cpp
tst-rwildes: yes it is caught.  Looking at other code in the module it seems to be the way invalid parameters are handled.  See my testing comment below.

docs/geedocs/5.2.5/answer/1089795.html
tst-ccamp: Yeah, "virtual-server-name" should be changed to "publish-point".

tst-ccamp:
Has anyone tested this code while also setting the `var=varname` parameter?

tst-rwildes:
@tst-ccamp Just tested with and without `var=varname` in combinations of `v=1`/`v=2` and no `v` at all.  Seems to work as expected.

earth_enterprise/src/common/serverdb/serverdbReader.cpp
tst-rwildes: @williamstrong you still need to address this comment.

williamstrong: @tst-rwildes thank you for pointing this out. I didn't see it on the main page of the PR.

Issue #1012, opened 2018-09-25
Cutter produces incorrect very low resolution result from 2d mercator map


tst-blaird:
I created a 2d mercator map of San Francisco using the tutorial data.  I used cutter and manually created the polygon by outlining the high resolution area of north east SF.  I specified a region level of 18 and left the other settings at their default values.  When I register this portable map in the GEE server and then view it everything looks correct.  See image 1-ResultOfSfCut.png.

I then created a 2d mercator map of the Orlando theme park area using imagery from the USGS Earth Explorer site.  I used cutter and manually created the polygon by outlining a rectangular area that included Disney's Magic Kingdom and Epcot.  I specified a region level of 18 and left the other settings at their default values.  See image 2-CreatingOrlandoCut.png.  When I register this portable map in the GEE server and then view it the polygon contains very low resolution imagery.  See image 3-ResultIsLowRes.png.

![1-resultofsfcut](https://user-images.githubusercontent.com/40216280/46027550-20b64180-c0bc-11e8-94b8-a1b6cd34da01.png)
![2-creatingorlandocut](https://user-images.githubusercontent.com/40216280/46027551-20b64180-c0bc-11e8-9064-0534cb9e13d5.png)
![3-resultislowres](https://user-images.githubusercontent.com/40216280/46027552-20b64180-c0bc-11e8-92dd-2702e7105394.png)

One difference that might be worth noting is the Orlando imagery contains a mosaic.  If you look at image 2-CreatingOrlandoCut.png, the lighter green area outside of the polygon is from a single imagery resource created as a mosaic from nine images.

tst-ccamp:
@tst-blaird Just a guess, can you check your "Advanced Settings" and make sure the polygon level goes up to 18 as well? Also, try specify a zoom level for world also, like 8 or 10.

tst-blaird:
@tst-ccamp I checked the advanced settings and the polygon level defaulted to 18.  I generated a new cutfile after expanding the advanced settings section and the result was the same.

I then specified a world level of 8 and generated a new cutfile.  The polygon was still low resolution but it appeared the tiles outside the polygon region were one level better resolution.  There were four squares in the world region for every one square in the polygon region.  Screenshot:

![4-resultislowres-worldlevel8](https://user-images.githubusercontent.com/40216280/46036634-fd978c00-c0d3-11e8-8ddd-8f6d10ad8041.png)

PR #1071, opened 2018-10-18
#1012 Use the --ignore_imagery_depth parameter when calling geportablemapbuilder

tst-blaird:
The fix for issue #1012 is to add the existing --ignore_imagery_depth parameter to the command line for geportablemapbuilder.

The default behavior for geportablemapbuilder is to stop searching for imagery when all GEE server HTTP requests for a zoom level return no imagery.  Using the existing --ignore_imagery_depth parameter causes the code to continue sending HTTP requests if the request contains any part of the polygon defined by the cutter operation.  This continues until the code reaches the "region level" provided specified by the user.

I created a 2D mercator map with the tutorial blue marble imagery (very low res) and one very high res (level 19) image.  I then used cutter and manually drew a square around the high resolution image.  When I create the cutfile without this fix, the high resolution imagery is missing from the cutfile.  With this change the high resolution imagery is there.  For my testing this caused a 15% increase in the number of HTTP requests to the GEE server to gather the additional imagery for the cutfile.

I created another 2D mercator map using the same blue marble and very high resolution image but also with imagery at two other resolution levels that were in-between the low and high levels.  This imagery covered the same area (but larger area) as the high resolution image.  I tested both before the fix and with the fix and the resulting cutfile is unchanged.  This is the result what we want for this test.  Additionally, I found that this generated the same number of HTTP requests to the GEE server.

EDIT:  Found scenarios where geportablemapbuilder generated correct output without using the --ignore_imagery_depth option and adding that option would substantially increase the running time.  The most recent commits remove --ignore_imagery_depth from always being used and make it an option on the cutter web page.  The "Continue past empty levels" option should only display if the operation is for a 2D map, and only when the user clicks on "Advanced".  Selecting "Yes" for that option will add the --ignore_imagery_depth option to the geportablemapbuilder command line.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/portablemapbuilder.cpp
+++ b/earth_enterprise/src/fusion/portableglobe/portablemapbuilder.cpp
@@ -111,6 +111,7 @@ void PortableMapBuilder::BuildMap() {
   std::vector<uint32> level_row(24);
   std::vector<uint32> level_col(24);
   int32 level = 0;
+  int32 max_level_traversed = 0;
   std::string qt_string = "";
   qt_char[level] = '0';
   level_row[level] = 1;
@@ -135,6 +136,9 @@ void PortableMapBuilder::BuildMap() {
       // Go to next qtnode at this level.
       qt_char[level] += 1;
     } else {
+      if (level > max_level_traversed) {
+        max_level_traversed = level;
+      }
       qt_string = qt_string.substr(0, level) + qt_char[level];
       // Note that we stop descending once we reach the first
       // missing packet, which is important since this cue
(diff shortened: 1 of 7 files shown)


earth_enterprise/src/fusion/portableglobe/portablemapbuilder.cpp
```diff
@@ -360,7 +363,19 @@ bool PortableMapBuilder::WriteMapPackets(const std::string& qtpath_str,
                                            uint32 level,
                                            uint32 col,
                                            uint32 row) {
-  if (!KeepNode(qtpath_str)) {
+  /*
```
tst-jhaight: very minor - If this is useful to keep around for future metrics gathering efforts, would it be worth putting this commented out code into an `#ifdef` block?

tst-blaird: That's a good observation.  I've made that change.  Thanks!

tst-blaird:
@tst-swanson would you please review the documentation changes in earthenterprise/docs/geedocs/5.2.5/answer/3230777.html?

Issue #1072, opened 2018-10-22
Allow packet extraction from 3d portable files


tst-jhaight:
Currently `--extract_packets` in `geglxinfo` only supports extracting packets from 2d portable files.  It would be nice to also explore the content of a 3d file to see what data should be available.

AC:

1. Packet extraction from 2d portable files (both `.glm` and `.glc` containing 2d data) remains unchanged
2. Packet extraction from 3d portable files produces a dump of the imagery, terrain, vector, and metadata packets in a manner similar to the 2d output.

PR #1073, opened 2018-10-22
Allow extracting globe packets from portable file

tst-jhaight:
Addresses #1072 

When a 3d cut file is provided as input to `geglxinfo` the packets are dumped in a manner similar to 2d packet dumps.  Vector data isn't quite human readable but gives a hint at the contents of the packet.

Verification steps:
* compile or obtain `geglxinfo` for this PR as well as `geglxinfo` from master
* obtain or create the following files:
  1. 2d `.glm` file
  2. `.glc` file with 2d contents
  3. `.glc` file with 3d contents
  4. 3d `.glb` file
* Run original `geglxinfo --glx <filename> --extract_packets --layer_idx 0 &> output.txt` against file types 1 and 2 above and save the output from each
* Run PR `geglxinfo --glx <filename> --extract_packets --layer_idx 0 &> output.txt` against all portable file types and save the output
* For file types 1 and 2, ensure output of new `geglxinfo` matches the output of the old `geglxinfo` exactly.
* Ensure output of new `geglxinfo` for file types 3 and 4 produces a `maptiles` directory with all the different expected subfiles and data

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/geglxinfo.cpp
+++ b/earth_enterprise/src/fusion/portableglobe/geglxinfo.cpp
@@ -169,7 +169,7 @@ void extractAllPackets(GlcUnpacker* const unpacker,
           }
           
           if (protoPacket.HasImageData()) {
-            writePacketToFile(index_item, protoPacket.ImageData(), true);
+            writePacketToFile(index_item, protoPacket.ImageData(), true, "_img");
           }
           if (protoPacket.HasImageAlpha()) {
             writePacketToFile(index_item, protoPacket.ImageAlpha(), true, "_alpha");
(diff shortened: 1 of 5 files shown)


earth_enterprise/src/fusion/portableglobe/geglxinfo.cpp
```diff
@@ -66,65 +73,91 @@ void GetXY(int level, uint64 btree, uint64* x, uint64* y) {
...
+                    int layer_idx,
+                    bool is_composite,
+                    const std::string& index_file,
+                    const std::string& data_file) {
+  PackageFileLoc index_file_loc;
+  PackageFileLoc data_file_loc;
+
+  if (!getPackageFileLocs(unpacker, index_file, data_file, is_composite, layer_idx,
+                          index_file_loc, data_file_loc)) {
+    return;
```
tst-rwildes: Should we output something here?

earth_enterprise/src/fusion/portableglobe/geglxinfo.cpp
```diff
@@ -66,65 +73,91 @@ void GetXY(int level, uint64 btree, uint64* x, uint64* y) {
...
+  if ((buffer[6] == 'J') && (buffer[7] == 'F')
+      && (buffer[8] == 'I') && (buffer[9] == 'F')) {
+    suffix = "jpg";
+  } else if ((buffer[1] == 'P') && (buffer[2] == 'N')
+             && (buffer[3] == 'G')) {
+    suffix = "png";
+  } else {
+    suffix = "unk";
+  }
+  snprintf(path, sizeof(path), "maptiles/%d/%d/%lu/%lu%s.%s",
```
tst-rwildes: for globes can this go to `globetiles` instead of `maptiles`?

earth_enterprise/src/fusion/portableglobe/geglxinfo.cpp
tst-jhaight: I didn't output anything additional here because in getPackageFileLocs it prints a more specific message on the failure before returning false.

PR #1074, opened 2018-10-22
997 scons warning for git

tst-kfreiheit:
The opegee project will still build correctly, but may be labelled with an incorrect version string when using git versions earlier than 1.8.4.

For opengee 5.2.4+, the project allows simultaneous development of the master and release branches.  The version string is calculated via the 'git describe' command, which looks into the tag history from the current commit.  The '--first-parent' parameter is required to accurately calculate the tag from the right branch, but that option appeared in git 1.8.4.

Previous to opengee 5.2.4, the minimum supported version of git was 1.7.1.

This pull request causes a warning string to be output via the scons build, as well as via the getversion.py script when using git version less than 1.8.4.  The warning string points to a FAQ to help upgrade git to mitigate the issue.

To reproduce:
- Build the source from this PR on a machine with git version earlier than 1.8.4.  Verify the warning string is output to stderr.
- Same test with getversion.py from a command line
- Build the source with git 1.8.4+, verify the warning string does NOT appear.
- Same test with getversion.py from a command line


tst-mbiancalana:
I accept the use of my contributions here.

earth_enterprise/src/scons/getversion.py
```diff
@@ -53,10 +63,22 @@ def _GitGeneratedLongVersion():
         return _VersionFromTagHistory(raw)
 
 
+def _IsFirstParentGitDescribeSupported():
+    """Checks whether --first-parent parameter is valid for the
+    version of git available"""
+
+    return opengee.version.is_version_ge(git.Git().version_info, (1, 8, 4))
```
tst-ppenev: I think this is a (relatively) bad way of checking for a feature.  E.g., maybe it's not likely, but what would happen, if someone decided to backport `--first-parent` to version 1.8.3?  Or, if version 2.7.99 dropped `--first-parent`?  You'll have to be reading a lot of release notes, and updating this code.

I think it would be better, if you could run something like `git describe --first-parent --tags` and check for "error: unknown option \`--first-parent'".  Or, if you don't have a Git repository to run this test on, scan the output of `git describe --help` for `--first-parent`.

earth_enterprise/src/scons/getversion.py
```diff
@@ -222,17 +244,36 @@ def set_long(self, value):
 
         self.long_version_string = value
 
+    def is_warning_available(self):
```
tst-ppenev: Opinion: I'd name this something more indicative, maybe:

```suggestion
    def has_warning(self):
```

Also, I'm not sure this function is needed.  A function that returns all the warnings, and is able to return an empty result already fulfills the purpose of this function.

earth_enterprise/src/scons/getversion.py
```diff
@@ -222,17 +244,36 @@ def set_long(self, value):
...
+        """Returns whether a warning message should be shown."""
+
+        return not _IsFirstParentGitDescribeSupported()
+
+    def get_warning_message(self):
+        """Returns any issues with version calculation with mitigation steps."""
```
tst-ppenev: Minor: This docstring should specify the value returned when there are no warnings.  `None` is an alternative for indicating no warnings.

earth_enterprise/src/scons/getversion.py
```diff
@@ -222,17 +244,36 @@ def set_long(self, value):
...
+
+        return '' if not self.is_warning_available() else '''\
+WARNING: git version 1.8.4 or later is required to correctly determine the opengee version being built.
```
tst-ppenev: Style: Capitalize Git and Open GEE.

earth_enterprise/src/scons/getversion.py
```diff
@@ -1,9 +1,19 @@
...
 import argparse
 import git
+import sys
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '../lib', 'python'))
+import opengee.version
+
 from datetime import datetime
 
+def eprint(*args, **kwargs):
```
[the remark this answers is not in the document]
tst-mbiancalana: I think this function is partially re-implementing `warnings.warn()`. That one will prepend the `WARNING: ` bit on the front of your message for you as well.

earth_enterprise/src/scons/getversion.py
```diff
@@ -1,9 +1,19 @@
 #!/usr/bin/env python
+from __future__ import print_function
 import os
 import argparse
 import git
+import sys
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '../lib', 'python'))
```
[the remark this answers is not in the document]
tst-kfreiheit: Removing the import and this path addition -- using the exception-based check instead of git version number check.  No longer needed.

earth_enterprise/src/scons/getversion.py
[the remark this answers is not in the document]
tst-kfreiheit: Removing; using print with >> instead.

earth_enterprise/src/scons/getversion.py
```diff
@@ -53,10 +63,22 @@ def _GitGeneratedLongVersion():
         return _VersionFromTagHistory(raw)
 
 
+def _IsFirstParentGitDescribeSupported():
```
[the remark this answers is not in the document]
tst-kfreiheit: Agreed -- good suggestion!

earth_enterprise/src/scons/getversion.py
tst-kfreiheit: This was my first approach, but somehow I convinced myself to go the other way --- after the ugly import and extra code involved, I'm coming back to this implementation.

earth_enterprise/src/scons/getversion.py
tst-kfreiheit: Removed; returning None from get_warning_message()

earth_enterprise/src/scons/getversion.py
tst-kfreiheit: Done.

docs/_posts/2018-10-23-git-version-upgrade-for-multibranch-builds.md
```diff
@@ -0,0 +1,35 @@
...
+title:  "Git version update to support multibranch builds"
+section_id: blog
+date:   2018-10-23 13:30:00
+author: Open GEE Development Team
+---
+
+Hello Everyone!
+
+The Open GEE Development Team would like to make a special announcement about
+the version of git required to build the project. As of release `Open GEE 5.2.4`,
```
tst-ppenev: Style: Capitalize Git.

tst-kfreiheit: Thanks -- done.

earth_enterprise/src/scons/getversion.py
```diff
@@ -222,17 +242,30 @@ def set_long(self, value):
...
 
-
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("-l", "--long", action="store_true", help="Output long format of version string")
     args = parser.parse_args()
 
     print open_gee_version.get_long() if args.long else open_gee_version.get_short()
+    if open_gee_version.get_warning_message():
+        print >> sys.stderr, open_gee_version.get_warning_message()
```
tst-mbiancalana: calling `get_warning_message` twice seems undesirable to me, because there are system calls and a several-levels-deep call tree involved under the hood. And calling the same function on two lines in a row always looks odd to me.

Perhaps store the result and check it? It would  read pretty well:
```
    warning_message = open_gee_version.get_warning_message()
    if warning_message is not None:
        print >> sys.stderr, warning_message
```

tst-kfreiheit: Good callout -- fixed.

docs/_posts/2018-10-23-git-version-upgrade-for-multibranch-builds.md
tst-kfreiheit: Done.

PR #1078, opened 2018-10-24
Issue #1020: state update integration tests

tst-lsavoie:
Integration tests for verifying that states are propagated across assets correctly. These tests can be used to verify the behavior of the code updates in PR #1044.

Instructions for running these tests are listed in the provided README.md. Note that Gauge only works with Python 2.7 and up, so these tests will not run under CentOS 6 or RHEL 6.


tst-lsavoie:
@tst-dsheremata looks like you need to post in a comment that you're OK with your commits being included in this PR.

tst-dsheremata:
Hereby providing consent to include my commits!

earth_enterprise/integration_test/step_impl/assets.py
```diff
@@ -0,0 +1,420 @@
...
+def get_asset_root():
+  return os.getenv('ge_asset_root')
+
+def get_src_data_path():
+  return os.getenv('ge_src_data_path')
+
+def make_dirs(dirs):
+  try:
+    os.makedirs(dirs)
+  except OSError:
```
tst-ppenev: I don't see how this code makes sure the exception was raised because the directory exists.  To do that, you'd use code like the following: 

```suggestion
  except OSError as exc:
      if exc.errno != errno.EEXIST:
          raise
```

earth_enterprise/integration_test/test_setup.sh
```diff
@@ -0,0 +1,37 @@
...
+TUTORIAL_PATH=/opt/google/share/tutorials/fusion
+TEST_DATA_PATH=${SRC_VOLUME}/gauge_tests
+export PATH=${PATH}:/opt/google/bin
+
+# Clean up from any previous tests
+./test_teardown.sh
+
+# Script will only download if necessary
+sudo sh ${TUTORIAL_PATH}/download_tutorial.sh
+sudo rm -rf ${TEST_DATA_PATH}/*
```
tst-ppenev: Wonder: This doesn't make sense to me.  We're deleting the `${TEST_DATA_PATH}` directory, and running an `rsync` against it right after that.  How is `rsync` supposed to do anything different than a `cp`, if the `${TEST_DATA_PATH}` directory doesn't exist?

earth_enterprise/integration_test/test_teardown.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+ASSETROOT=/gevol/assets
+
+sudo service gefusion stop
```
tst-ppenev: This script doesn't ensure there are no Fusion entries referring to the files being deleted below.

For example, there are resources above that seem to imply they have Points of Interest data, which would be inserted in PostgreSQL tables.  It's true that after deleting these files there may be no resources that refer to those entries, and, therefore, this test would be creating orphaned SQL data.

If that is not the case, it should be noted here.  If it is the case, it should be fixed, if that's easy, or noted here, and another ticket should be created to fix the build up of orphaned data.

tst-lsavoie: Fusion doesn't use postgres based on the fact that postgres is not required for Fusion. Based on the section "Clean up the tutorial work space" in this: http://www.opengee.org/geedocs/answer/6028272.html, I think we can say that everything that fusion creates is stored under the asset root. Do you still think there is a need for a comment here?

tst-ppenev: Ah, yes, you're right. Those Fusion projects are never published, so there wouldn't be any of that data.

You could include a link to that document, but I'm fine even without a comment.

earth_enterprise/integration_test/specs/state_propagation.spec
```diff
@@ -0,0 +1,787 @@
...
+* Verify that the state of images for default project "StatePropagationTest_CancelProjectLate" is "Succeeded"
+
+## Clean After Cancel
+Tags: cancel, clean, rebuild
+
+Build
+* Create and build default project "StatePropagationTest_CleanAfterCancel"
+
+Cancel Resource
+* Cancel imagery resource "BlueMarble_CleanAfterCancel"
```
tst-ppenev: Minor: Those resource names could have used the same prefix as the rest of the resources.

tst-lsavoie: I probably should have done that to begin with, but at least they're all under the same parent directory. At this point I don't think it's worth the effort.

PR #1080, opened 2018-10-25
Check if OpenGEE was installed with a package manager before before running installation and uninstallation scripts.

### LINKED ISSUE #837: Manual install/uninstall scripts should be disabled when rpm installs active
## Acceptance Criteria
1. src/installer/install_server.sh, install_fusion.sh, uninstall_server.sh, and uninstall_fusion.sh refuse to run while opengee-common is installed.

jl2:
Added is_package_installed function to common.sh, refactored software_check to use it.

Made install_*.sh and uninstall_*.sh scripts check for opengee-common
installing or removing.

Fixes #837

### DIFF
--- a/docs/geedocs/5.2.5/answer/7160006.html
+++ b/docs/geedocs/5.2.5/answer/7160006.html
@@ -69,6 +69,15 @@ gtag('config', 'UA-108632131-2');
             <th>Resolution</th>
           </tr>
           <tr>
+            <td>837</td>
+            <td>Manual install/uninstall scripts should be disabled when rpm installs active</td>
+            <td>
+              The install and uninstall scripts now check if Open GEE has been installed with a package
+              manager before installation begins, and will not install or remove any files when the
+              packages are installed.
+            </td>
+          </tr>
+          <tr>
             <td>934</td>
             <td>Polygon displayed at the wrong location when viewing GLC assemly</td>
             <td>
--- a/docs/geedocs/5.2.5/answer/7160006.html
+++ b/docs/geedocs/5.2.5/answer/7160006.html
@@ -45,6 +45,7 @@ gtag('config', 'UA-108632131-2');
   </h5>
     <p><strong>When a GEE server is queried for server definitions the response can be returned in proper JSON format.</strong> The <code>v</code> parameter can be added to query string to specify which version of the API will be returned. The valid options are <code>v=1</code> and <code>v=2</code>. Version one returns data in the current format. Version two returns data in JSON format. An example url which returns data in JSON format looks like: <code>http://&lt;GEE earth server&gt;/&lt;database_publish_point&gt;/query?request=Json&amp;v=2</code>.</p>
     <p><strong>Enable WMS when publishing from the commaind line.</strong> <code>geserveradmin --publishdb</code> has a new option <code>--serve_wms</code>. Passing this flag enables WMS for the database you are publishing.</p>
+    <p><strong>Installations scripts are disabled when Open GEE packages are installed.</strong>  The install and uninstall scripts will now check if Open GEE has been installed with a package manager before installation begins, and will not make any changes when the packages are installed.
   <h5>
     <p id="supported_5.2.5">Supported Platforms</p>
   </h5>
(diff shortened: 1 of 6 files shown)


earth_enterprise/src/installer/common.sh
```diff
@@ -80,6 +80,26 @@ xml_file_get_xpath()
         tail -n +2 | head -n -1
 }
 
+is_package_installed()
+{
+    local package_installed_retval=1
```
tst-ppenev: Q: Is there a reason to use a variable, rather than just return the value?

earth_enterprise/src/installer/common.sh
```diff
@@ -80,6 +80,26 @@ xml_file_get_xpath()
...
+
+    # args: $1: Ubuntu package
+    # args: $2: RHEL package
+
+    if [ "$MACHINE_OS" == "$UBUNTUKEY" ] && [ ! -z "$1" ]; then
+        if [[ ! -z "$(dpkg --get-selections | sed s:install:: | sed -e 's:\s::g' | grep ^$1)" ]]; then
```
tst-ppenev: Minor: You could use a single `grep` pattern to tell if a package is installed, rather than piping twice through `sed`, and again through `grep`.  Perhaps, a `dpkg --status` output may be easier to parse, as well.

earth_enterprise/src/installer/common.sh
jl2: Not really. I saw a couple of nearby functions do it that way, so used that style.  Looking again I see later in the same file several functions just return.

earth_enterprise/src/installer/common.sh
jl2: I agree, that’ll simplify things. I’ll make the change when I get home this evening.

earth_enterprise/src/installer/common.sh
tst-ppenev: It's a style preference, but I think it's less code and reads better to just return, if you can.

earth_enterprise/src/installer/common.sh
```diff
@@ -148,6 +160,15 @@ determine_os()
     return $retval
 }
 
+
+show_opengee_package_installed()
+{
+    #args: $1 "install" or "uninstall"
+    echo -e "\nThe opengee-common package is installed."
```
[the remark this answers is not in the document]
tst-nfarah: The other way, we can get granular and looks for Fusion when one is trying to install/uninstall Fusion or Server. This is to handle the case of Fusion and Server being installed in different machines (Grid Setup). But keeping the generic check of opengee-common is fine as well.

earth_enterprise/src/installer/common.sh
```diff
@@ -88,20 +106,14 @@ software_check()
...
             echo -e "\nInstall $2 and restart the $1."
             software_check_retval=1
-        fi
-    elif { [ "$MACHINE_OS" == "$REDHATKEY" ] || [ "$MACHINE_OS" == "$CENTOSKEY" ]; } && [ ! -z "$3" ]; then
-        if [[ -z "$(rpm -qa | grep ^$3\$)" ]]; then
+	elif { [ "$MACHINE_OS" == "$REDHATKEY" ] || [ "$MACHINE_OS" == "$CENTOSKEY" ]; } && [ ! -z "$3" ]; then 
             echo -e "\nInstall $3 and restart the $1."
             software_check_retval=1
         fi
-    else
```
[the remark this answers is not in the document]
tst-lsavoie: I agree - it would be good to keep the else.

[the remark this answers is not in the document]
jl2: Well, the reason I took it out is that the determine_os function has to be called before the software checks (to fill in MACHINE_OS) and it already checks for an unsupported OS and prints the list of supported systems.  Since the installers exit when determine_os fails it shouldn't be possible for these checks to run on an unsupported OS.

But it's less fragile to handle it here, too, in case determine_os changes, so I'll add it back in.

docs/geedocs/5.2.5/answer/7160006.html
```diff
@@ -68,6 +68,15 @@ <h5>
             <th class="narrow">Description</th>
             <th>Resolution</th>
           </tr>
+          <tr>
+            <td>837</td>
```
tst-nfarah: Since this is an enhancement and not a bug fix per say, let's document it under **New Features** section

docs/geedocs/5.2.5/answer/7160006.html
```diff
@@ -68,6 +68,15 @@ <h5>
...
+            <td>Manual install/uninstall scripts should be disabled when rpm installs active</td>
```
tst-nfarah: `rpms installs active` --> `package manager installs is active` ?

docs/geedocs/5.2.5/answer/7160006.html
jl2: Sounds good, I’ll move it over and fix the wording this evening.

PR #1095, opened 2018-11-23
WIP: Xerces wrapper classes

tst-basilhuffman:
Added skeleton which reflects design

### DIFF
--- a/earth_enterprise/src/common/khxml/khXML.cpp
+++ b/earth_enterprise/src/common/khxml/khXML.cpp
@@ -1,0 +1,67 @@
+#include "khXML.h"
+#include <string>
+#include <thread>
+#include <mutex>
+
+static khMutexBase xmlLibLock = KH_MUTEX_BASE_INITIALIZER;
+
+bool khXML::doOp(unique_ptr<khXMLOperation> op)
+{
+  khLockGuard guard(xmlLibLock);
+  return op->op();
+}
+
+class khXMLOperation
+{
+  khXMLOperation(const khXMLOperation&) = delete;
+  khXMLOperation(khXMLOperation&&) = delete;
+  khXMLOperation& khXMLOperation(const khXMLOperation&) = delete;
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/common/khxml/khXML.cpp
```diff
@@ -0,0 +1,48 @@
+/*
```
tst-lsavoie: Minor: I'd recommend changing the names of these files so that we don't have to differentiate between khxml and khXML.

earth_enterprise/src/common/khxml/khXML.h
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  virtual ~khXMLOperation()
+  {
+    try
+    {
+      doc->release();
+    } 
+    catch (...) {}
+    if (type != khXMLOperationType::KH_XML_LOAD_FROM_STRING) khXMLOperation::Terminate();
```
tst-lsavoie: The number of Initialize calls should match the number of Terminate calls. If you Initialize unconditionally, you should Terminate unconditionally.

tst-basilhuffman: Yeah, this would be taken out. I would need to get a handle on how loads are treated.

earth_enterprise/src/common/khxml/khXML.cpp
tst-basilhuffman: Well, ideally it would replace khxml.h/.cpp. if needed, I can change the file name now though.

tst-lsavoie: No need to change the name now. Just something to take care of before it's checked in.

tst-basilhuffman:
@tst-lsavoie should i just go ahead and close this?

tst-lsavoie:
Yes, I think so. We can always reopen it later if needed.

PR #1096, opened 2018-11-23
Allow XML cache size modifications

tst-basilhuffman:
Added functionality to experiment with xerces initialize parameters. Usage is given in comments


earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -84,6 +118,7 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
                                             0);// document type object (DTD)
     return doc;
```
tst-lsavoie: If I'm reading this correctly, the mutex is never unlocked unless an exception is thrown.

tst-lsavoie: Nevermind, I see the unlock in the destructor.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -74,7 +108,7 @@ DOMDocument *
 CreateEmptyDocument(const std::string &rootTagname) throw()
 {
   InitializeXMLLibrary();
-
+  if (doTerminate) pthread_mutex_lock(&xmlDocLock);
```
tst-lsavoie: For locking an unlocking, I think declaring xmlDocLock as a khMutex: `khMutex xmlDocLoc` and then calling `lock` or `unlock` as needed will work: `xmlDocLoc.loc(); /* other stuff */ xmlDocLoc.unlock();`

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -329,12 +368,22 @@ ReadDocumentFromString(khxml::DOMLSParser *parser,
...
     doc->release();
-    return true;
+    retval = true;
   } catch (...) {
+    notify(NFY_WARN, "Unable to release document.");
   }
-  return false;
+
+  if (doTerminate) {
+    XMLPlatformUtils::Terminate();
```
tst-lsavoie: I'm not sure I like this pattern as it stands. Functionally, I think it's fine. However, it seems that it would be cleaner if we did the following:

- Don't call initialize in UsingXMLGuard
- Call initialize when creating a DOMDocument
- Call terminate when destroying a DOMDocument

However, as I said, that ends up being functionally the same as this. It's more of a style issue.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -74,7 +112,7 @@ DOMDocument *
 CreateEmptyDocument(const std::string &rootTagname) throw()
 {
   InitializeXMLLibrary();
-
+  if (doTerminate) pthread_mutex_lock(&xmlDocLock);
```
tst-ppenev: This line does not check whether the mutex was acquired successfully, or there was an error.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -84,6 +122,7 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
                                             0);// document type object (DTD)
     return doc;
    } catch (...) {
+     if (doTerminate) pthread_mutex_unlock(&xmlDocLock);
```
tst-ppenev: This also doesn't check for errors.

earth_enterprise/src/common/khxml/khxml.cpp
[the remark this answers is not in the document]
tst-ppenev: I suspect this won't work.  Does the library allow calling `XMLPlatformUtils::Terminate();`, or `XMLPlatformUtils::Initialize` for that matter, while an existing parser is in the middle of its operation?  If it doesn't you can't call any of these functions unless there are zero currently active parsers.  I.e., there can't be any open documents on any threads when you call `XMLPlatformUtils::Terminate` or `XMLPlatformUtils::Initialize`.  To ensure that, you would need a mutex, as well as, e.g., a reference counter.

tst-ppenev:
I'm not sure this implementation would work correctly in a multi-threaded environment.  It may be OK for profiling, but I think we'd want something better for production.

Since, calling `XMLPlatformUtils::Initialize` and `XMLPlatformUtils::Terminate` may be somewhat expensive, and these functions likely require having no concurrency when they're called, I think you'd get a better performance and concurrency, if the implementation has some of the following characteristics:

* Run a library re-initialization only when memory consumption goes above a given threshold, rather than each time a document is used.
* Allow an arbitrary degree of concurrency until a re-initialization is required, and after it is performed.

Here's an example of what an implementation may involve:

* Count the current number of bytes allocated by Xerces.  If they go above a configured threshold, schedule a library re-initialization.
* To schedule a library re-initialization:
    * Have an advisory lock (e.g., a POSIX read-write lock) that is acquired as a read lock (i.e., shared) before opening an XML document, and is released after closing the document.
    * When a library re-initialization is scheduled, obtain a write lock (i.e., exclusive) on the advisory lock.  Call `XMLPlatformUtils::Terminate` and `XMLPlatformUtils::Initialize`.  Have a policy of what to do, if the memory consumption didn't fall below a given amount below the threshold.  E.g., increase the threshold.  Release the lock.

earth_enterprise/src/common/khConfigFileParser.cpp
```diff
@@ -0,0 +1,137 @@
...
+        if (a == key) return true;
+    }
+    return false;
+}
+
+static khMutexBase xmlParmsLock = KH_MUTEX_BASE_INITIALIZER;
+
+void khConfigFileParser::parse(const std::string& fn)
+{
+	std::ifstream file;
```
tst-lsavoie: You've got some funny indentation here (not consistent in this function. It would be nice to clean that up.

earth_enterprise/src/common/khConfigFileParser.h
```diff
@@ -0,0 +1,72 @@
...
+    
+    // allow for range looping
+    std::map<std::string,std::string>::const_iterator begin() { return contents.begin(); }
+    std::map<std::string,std::string>::const_iterator end() { return contents.end(); }	
+    
+    // delete everything but constructor/destructor
+    khConfigFileParser()                                     = default;
+	~khConfigFileParser()									 = default;
+
+    khConfigFileParser(const khConfigFileParser&)            = delete;
```
tst-lsavoie: Minor: I'm not sure these need to be disabled. It comes down to style, though.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -37,16 +42,94 @@ ListElementTagName(const std::string &tagname)
...
+   	maxDOMHeapAllocSize < lowestBlock     ||
+	maxDOMSubAllocationSize < lowestBlock)
+    {
+	throw MinValuesNotMet();
+    }
+
+    // check to make sure that the initial size is less than the max size
+    if (maxDOMHeapAllocSize < initialDOMHeapAllocSize ||
+        maxDOMHeapAllocSize < maxDOMSubAllocationSize ||
+        initialDOMHeapAllocSize > maxDOMHeapAllocSize)
```
tst-lsavoie: I think the 3rd line in this if statement is redundant with the 1st line in the if statement.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -37,16 +42,94 @@ ListElementTagName(const std::string &tagname)
...
 
-  UsingXMLGuard(void) throw()
-  {
-   try {
-      XMLPlatformUtils::Initialize();
+  UsingXMLGuard(void) throw() {
+    // can change this to anything, just using this for convenience
+    // format is:
+    // <initialDOMHeapAllocSize> <maxDOMHeapAllocSize> <maxDOMSubAllocationSize"
+    std::string fn("/etc/opt/google/XMLparams");
```
tst-lsavoie: Looks like this is the old method of reading the config file. Shouldn't it be updated to use khConfigFileParser? Also, I think once that's done some of the new includes can be removed (i.e., fstream).

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -329,12 +416,33 @@ ReadDocumentFromString(khxml::DOMLSParser *parser,
...
+    retval = true;
   } catch (...) {
   }
-  return false;
+  if (doTerminate)
+  {
+    XMLPlatformUtils::Terminate();
+    try {
+      try {
+         validateXMLParameters();
```
tst-lsavoie: I don't see a need to validate the parameters every time we use them. As long as we validate them on input we should be fine.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -84,6 +169,8 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
                                             0);// document type object (DTD)
     return doc;
    } catch (...) {
+     // if there is an error, unlock the document
+     if (doTerminate) xmlDocLock.Unlock();
```
tst-lsavoie: Just double checking - at one point we talked about just locking around the terminate and initialize functions. Did that not work?

tst-lsavoie:
I think we can take WIP out of the title of this PR.

earth_enterprise/src/common/khConfigFileParser.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <sstream>
+
+FileNotPresentException::FileNotPresentException(const std::string& fn)
+{
+    str = fn + " is not present";
+}
+
+OptionsEmptyException::OptionsEmptyException()
+{
+	str = "no list of options present";
```
tst-lsavoie: Minor: funky indentation

tst-lsavoie:
Code looks good; still need to test before I can approve.

tst-lsavoie:
Closed this accidentally. Reopening.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -37,16 +43,116 @@ ListElementTagName(const std::string &tagname)
...
+      try {
+         validateXMLParameters();
+      } catch (const XmlParamsException& e) {
+         notify(NFY_WARN, "%s, using default xerces init values", e.what());
+         setDefaultValues();
+      }
+      XMLPlatformUtils::Initialize(initialDOMHeapAllocSize,
+                                   maxDOMHeapAllocSize,
+                                   maxDOMSubAllocationSize);
+      notify(NFY_NOTICE, "XML initialization values: %s=%zx %s=%zx %s=%zx %s=%d",
```
tst-lsavoie: This notice prints out a lot. I think it prints out for every executable that fusion starts. Can we bump it down to a lower notification level, at least when it's not called from system manager?

tst-basilhuffman: Would `NFY_DEBUG` suffice?

tst-lsavoie:
I'm running into errors when I turn on cache purging and run the gauge tests (under earth_enterprise/integration_test). If I run the test_setup.sh script it says `Error setting up tests. Please try again.` If I open the Fusion UI and cancel the asset the script created, the UI freezes. Neither action results in any error messages being printed to the log.

I'd suggest splitting this PR into two. The first would include the code to set the XML parameters, which works fine and can probably be merged quickly. The second would include the code to reinitialize Xerces periodically, which needs more testing.

tst-basilhuffman:
> I'm running into errors when I turn on cache purging and run the gauge tests (under earth_enterprise/integration_test). If I run the test_setup.sh script it says `Error setting up tests. Please try again.` If I open the Fusion UI and cancel the asset the script created, the UI freezes. Neither action results in any error messages being printed to the log.
> 
> I'd suggest splitting this PR into two. The first would include the code to set the XML parameters, which works fine and can probably be merged quickly. The second would include the code to reinitialize Xerces periodically, which needs more testing.

I will split them, which set of images were you using?

tst-lsavoie:
I'm using the tutorial images. If you run the test_setup.sh script it will download the images it needs and create the necessary resources.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -37,16 +43,109 @@ ListElementTagName(const std::string &tagname)
...
+      try {
+         validateXMLParameters();
+      } catch (const XmlParamsException& e) {
+         notify(NFY_WARN, "%s, using default xerces init values", e.what());
+         setDefaultValues();
+      }
+      XMLPlatformUtils::Initialize(initialDOMHeapAllocSize,
+                                   maxDOMHeapAllocSize,
+                                   maxDOMSubAllocationSize);
+      notify(NFY_DEBUG, "XML initialization values: %s=%zx %s=%zx %s=%zx",
```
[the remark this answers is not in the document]
tst-basilhuffman: done

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -37,16 +43,109 @@ ListElementTagName(const std::string &tagname)
...
+          // will only set the values that are present, otherwise will stay at defaults
+          if (a.first == INIT_HEAP_SIZE)
+              initialDOMHeapAllocSize = std::stol(a.second);
+          else if (a.first == MAX_HEAP_SIZE)
+              maxDOMHeapAllocSize = std::stol(a.second);
+          else if (a.first == BLOCK_SIZE)
+              maxDOMSubAllocationSize = std::stol(a.second);
+      }
+    } catch (const khConfigFileParserException& e) {
+      notify(NFY_WARN, "%s , using default xerces init values", e.what());
```
tst-lsavoie: This is another one that comes out all the time if you haven't created an XMLparams file. Perhaps we could put it at the debug level too?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -37,16 +43,109 @@ ListElementTagName(const std::string &tagname)
...
+      }
+    } catch (const khConfigFileParserException& e) {
+      notify(NFY_DEBUG, "%s , using default xerces init values", e.what());
+    }
+    
+    try {
+      try {
+         validateXMLParameters();
+      } catch (const XmlParamsException& e) {
+         notify(NFY_WARN, "%s, using default xerces init values", e.what());
```
tst-lsavoie: I missed this one - probably should make it DEBUG too.

earth_enterprise/src/common/khConfigFileParser.cpp
```diff
@@ -129,40 +129,44 @@ bool khConfigFileParser::isKeyPresent(const std::string& key)
...
+    khLockGuard guard(xmlParmsLock);
+
+    if (options.size() == 0) throw OptionsEmptyException(); // nothing to search on, exit
+
+    file.open(fn.c_str());
+    if (file.fail()) 
+    {
+        throw FileNotPresentException(fn);
+    }
+    std::stringstream fileContents;
```
tst-lsavoie: Very minor: if you had the parse function called below take a `std::istream` as a parameter instead of a `std::stringstream` you could pass the `file` directly to the function instead of reading its contents into a stringstream first.

PR #1099, opened 2018-11-30
Updating kmlrender library with latest changes.

art926:


tst-dsheremata:
As per other discussions, we agreed that the URL references to jsapi and other Google API's need to supply the FQDN to those resources, as they should reach out to the google hosts instead of relative URLs on the OpenGEE server.   @art926  - can you please amend your commit with the updated URLs? 

Thanks!

art926:
Sorry for slow response. I’ve just updated the files here:
https://github.com/art926/earthenterprise/tree/new_kmlrender/earth_enterprise/src/fusion/portableglobe/cutter/htdocs/earth/client/kmlrender
Feel free to integrate the change if it works for you.

PR #1100, opened 2018-12-01
New versioning logic using automatically generated build numbers

tst-rwildes:
Addresses issue #1101.  New versioning logic that uses branch names for version name and commit log counts for build number

**Creating a release branch** is done by branching from master or from a previous release tag and naming the branch `release_<version name>` where version name should follow GEE normal version rules as they are today (minus the old patch number and the quality tag (i.e. alpha, beta, and final)).  Also when creating a release branch you are required to create a branch 'tail tag' that should have the same name as what was used for the 'version name'.  
**Closing out a release** you have to add the final release tag to the head of the release branch (should be what is returned from the getversion.py script) and then remove the 'tail tag'

**Note:** a future PR will have scripts to automate the above steps, update the documentation, and a script to return the git hash given the build number of a release.

Things to test (tip do testing on a clone you can just through way):
1) create a release branch from point within master
  a) should generate a version like `5.2.5-707.1` where the `5.2.2` is what follows `release_` in the branch name and `707` is the commit count at the point of the branch and the the `1` is the number of commits post branching.
  b) even after merging back to master the build number should stay the same
  c) create a tag simulating the end of the release and remove the 'tail tag'.  Version number should still be the same as it should pick it up from the release tag.
  d) re-open release branch by adding the 'tail tag' back in.  Do a few commits.  build number should increase again
  e) close branch back out by adding another release tag and removing the 'tail tag'.  Version number should be the same
  f) checkout using one of the above release tags.  Version number should still be the same on disconnected head.
2) do the same thing but create the release branch from one of the previous releases (note: you need to use a new 'version name' so if your previous release had a version name of `5.2.5` you should do something like `5.2.5.1`
3) should also test running getversion on an open release but where the 'tail tag' does not exist in the clone but instead exists in a remote known to the local clone.  This tag should be automatically pulled down for the user.
4) also test building and running a few smoke tests.

### DIFF
--- a/earth_enterprise/src/scons/getversion.py
+++ b/earth_enterprise/src/scons/getversion.py
@@ -105,19 +105,35 @@ def _GitPreviousReleaseTag(versionName):
     and if it finds it then it looks for any release build tags that are also pointing to the same
     commit"""
     tags = _GetRepository().tags
-    tailTag = next((tag for tag in _GetRepository().tags if tag.name == versionName), None)
+    tailTag = next((tag for tag in tags if tag.name == versionName), None)
     if tailTag is None:
         return ''
 
     for tag in tags:
-        if tag.commit == tailTag.commit and tag.name <> tailTag.name and _IsReleseBuildTag(tag.name):
-            return tag.name
+        if tag.name != tailTag.name:
+            if  _IsReleseBuildTag(tag.name):
+                if  _GitTagRealCommitId(tag.name) == _GitTagRealCommitId(tailTag.name):
+                    return tag.name
     
     return ''
 
 
+def _GitTagRealCommitId(tagName):
+    """use shell command to retreive commit id of where the tag points to"""
+    # for some reason .hexsha was not returning the same id....
+    return os.popen("git rev-list -n 1 '{0}'".format(tagName.replace("'", "'\"'\"'"))).read().strip()
+
+
 def _IsReleseBuildTag(tagName):
-    return re.match(r'[0-9]*\.[0-9]*\.[0-9]*\-*', tagName)
+    """checks if the tag follows the pattern where if the
+    tag is split on dash and the has at least two elements
+    and the first two elements is a series of numbers delimited
+    by dot and nothing else in those first two elements"""
+    splitTag = tagName.split('-')
+    if len(splitTag) > 1:
+        return (re.match('^[0-9]+((\.[0-9]+)+)$', splitTag[0]) and re.match('^([0-9]+((\.[0-9]+)+)|[0-9]+)$', splitTag[1]))
+
+    return False
 
 
 def _IsReleaseBranch(branchName):
@@ -134,7 +150,7 @@ def _IsReleaseBranch(branchName):
             for remote in repo.remotes:
                 try:
                     remote.fetch('+refs/tags/{0}:refs/tags/{0}'.format(versionName), None, **{'no-tags':True})
-                except git.exc.GitCommandError:
+                except:
                     pass
             
             # try one more time after the fetch attempt(s)
--- a/earth_enterprise/src/scons/getversion.py
+++ b/earth_enterprise/src/scons/getversion.py
@@ -55,7 +55,7 @@ def _GitCommitCount(tagName='HEAD', baseRef=''):
     if not baseRef:
         return len(list(repo.iter_commits(tagName)))
     else:
-        return len(list(repo.iter_commits(baseRef + '.' + tagName)))
+        return len(list(repo.iter_commits(baseRef + '..' + tagName)))
 
 
 def _GitVersionNameAndBuildNumber():
--- a/earth_enterprise/src/scons/getversion.py
+++ b/earth_enterprise/src/scons/getversion.py
@@ -5,7 +5,7 @@ import git
 import sys
 
 from datetime import datetime
-
+import re
 
 def GetVersion(backupFile, label=''):
     """As getLongVersion(), but only return the leading *.*.* value."""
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/scons/getversion.py
```diff
@@ -42,16 +43,125 @@ def GetLongVersion(backupFile, label=''):
...
+                versionName = _GetReleaseVersionName(branchName)
+                return versionName, '{0}.{1}'.format(_GitBranchedBuildNumber(versionName), _GitCommitCount('HEAD', versionName))
+            else:
+                return _sanitizeBranchName(branchName),  str(_GitCommitCount())
+
+
+def _GitBranchedBuildNumber(versionName):
+    """Returns what the build number was from the branch point"""
+    prevRelTag = _GitPreviousReleaseTag(versionName)
+    prevBuildNumber = ''
```
tst-ccamp: I think something besides "BuildNumber" would be appropriate here since we're counting tags, not builds. Maybe "prevPatchNumber"

earth_enterprise/src/scons/getversion.py
```diff
@@ -42,16 +43,125 @@ def GetLongVersion(backupFile, label=''):
...
+
+    return prevBuildNumber
+
+
+def _GitPreviousReleaseTag(versionName):
+    """Looks for the tail tag for this versionName (the tag with the same name as the version name)
+    and if it finds it then it looks for any release build tags that are also pointing to the same
+    commit"""
+    tags = _GetRepository().tags
+    tailTag = next((tag for tag in _GetRepository().tags if tag.name == versionName), None)
```
tst-ccamp: The logic in this line is a little confusing. I'd suggest splitting it into multiple lines

earth_enterprise/src/scons/getversion.py
```diff
@@ -42,16 +43,125 @@ def GetLongVersion(backupFile, label=''):
...
+
+    for tag in tags:
+        if tag.commit == tailTag.commit and tag.name <> tailTag.name and _IsReleseBuildTag(tag.name):
+            return tag.name
+    
+    return ''
+
+
+def _IsReleseBuildTag(tagName):
+    return re.match(r'[0-9]*\.[0-9]*\.[0-9]*\-*', tagName)
```
tst-ccamp: I think you want to use + instead of *. We don't want to consider `..3` to be valid

earth_enterprise/src/scons/getversion.py
```diff
@@ -42,16 +43,125 @@ def GetLongVersion(backupFile, label=''):
...
+                except git.exc.GitCommandError:
+                    pass
+            
+            # try one more time after the fetch attempt(s)
+            return _gitHasTag(versionName)
+
+
+def _gitHasTag(tagName):
+    """See if a tag exists in git"""
+    return (next((tag for tag in _GetRepository().tags if tag.name == tagName), None) is not None)
```
tst-ccamp: Same comment as above. Please split these into multiple lines

earth_enterprise/src/scons/getversion.py
```diff
@@ -200,6 +234,7 @@ def _GetDateString():
     """Returns formatted date string representing current UTC time"""
     return datetime.utcnow().strftime("%Y%m%d%H%M")
 
+
```
tst-ccamp: Minor: unnecessary line

tst-rwildes:
@tst-ccamp and @tst-blaird thanks for the early feedback!  I just saw these and I'll make the suggested changes now.  I was finishing up testing/fixing things to make sure it all worked right.

earth_enterprise/src/scons/getversion.py
tst-rwildes: Okay.  How about using the name 'commitCount' as that is what we are doing, counting commits.

tst-ccamp: If I understand correctly, it's getting the number from the tag. Does the tag contain the commit count? I thought it was a patch number like we are doing now.

earth_enterprise/src/scons/getversion.py
tst-rwildes: You are correct.  There actually turns out to be a lot of things bad with this so I completely re-did it.

BTW a cool website for testing regex: https://regexr.com/

earth_enterprise/src/scons/getversion.py
tst-rwildes: Yes the commit count is known by the tag.  With the new way the commit count replaces the patch number with a build number based on commit counts.  It will be similar to how SVN does revision numbers.

earth_enterprise/src/scons/getversion.py
```diff
@@ -42,16 +43,125 @@ def GetLongVersion(backupFile, label=''):
...
+            repo = _GetRepository()
+            for remote in repo.remotes:
+                try:
+                    remote.fetch('+refs/tags/{0}:refs/tags/{0}'.format(versionName), None, **{'no-tags':True})
+                except git.exc.GitCommandError:
+                    pass
+            
+            # try one more time after the fetch attempt(s)
+            return _gitHasTag(versionName)
+
```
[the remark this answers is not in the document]
tst-rwildes: yes I changed it to `return (_gitHasTag(versionName) is not None)` which will return a Boolean now instead of an object which might be None.  This is the else case here.

earth_enterprise/src/scons/getversion.py
tst-ccamp: I think you'd still want to differentiate between `prevCommitCount` and current `commitCount` though

tst-rwildes: I do that.  This call is only concerned with getting the commit count(s) of a point from where the branch started.  The commit count for the release branch will be appended by the caller (see above).  For example the first build of `5.2.5` will have a long version of something like `5.2.5-707.0` the `707` is the commit count of master from the point the release branch was created.  The `0` is the count since the release branch was created.  If someone fixes a bug and a single commit for that fix is merged into the release branch then the long build number would be `5.2.5-707.1`.

earth_enterprise/src/scons/getversion.py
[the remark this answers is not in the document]
tst-blaird: @tst-rwildes I think there's still an implicit return None if the first conditional, if branchName[:8] == 'release_':, is false.

earth_enterprise/src/scons/getversion.py
tst-ccamp: Cool, thanks for the explanation

earth_enterprise/src/scons/getversion.py
tst-ccamp: The new code is much easier to understand

Issue #1104, opened 2018-12-05
OpenGEE Build Fails on Centos 7


tst-cjeffries:
Centos 7 (and maybe others) fails to build OpenGEE. Build seems to fail when building third party parts of Qt. This appears to be an issue with libpng.

To reproduce:
1. Follow the build [instructions](https://github.com/google/earthenterprise/blob/master/earth_enterprise/BUILD_RHEL_CentOS.md) to get dependencies.
2. `git clone http://github.com/google/earthenterprise.git`
3. `cd earthenterprise/earth_enterprise`
4. `scons -j8 release=1 build`

Example Error Messages:
```
In file included from /usr/include/libpng15/png.h:452:0,
                 from kernel/qpngio.cpp:48:
kernel/qpngio.cpp:1063:23: error: invalid use of incomplete type 'png_struct {aka struct png_struct_def}'
     if (setjmp(png_ptr->jmpbuf)) {
                       ^
In file included from kernel/qpngio.cpp:48:0:
/usr/include/libpng15/png.h:857:16: error: forward declaration of 'png_struct {aka struct png_struct_def}'
 typedef struct png_struct_def png_struct;
                ^
kernel/qpngio.cpp: In member function 'void QPNGFormat::end(png_structp, png_infop)':
kernel/qpngio.cpp:1123:27: error: invalid use of incomplete type 'png_struct {aka struct png_struct_def}'
     unused_data = (int)png->buffer_size; // Since libpng doesn't tell us
                           ^
In file included from kernel/qpngio.cpp:48:0:
/usr/include/libpng15/png.h:857:16: error: forward declaration of 'png_struct {aka struct png_struct_def}'
 typedef struct png_struct_def png_struct;
                ^
kernel/qpngio.cpp: In function 'void qt_zlib_compression_hack()':
kernel/qpngio.cpp:1238:21: error: 'compress' was not declared in this scope
     compress(0,0,0,0);
                     ^
kernel/qpngio.cpp:1239:23: error: 'uncompress' was not declared in this scope
     uncompress(0,0,0,0);
                        ^
make[1]: *** [.obj/release-shared-mt/qpngio.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory `/home/centos/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party/qt/qt-x11-free-3.3.8b/src'
make: *** [sub-src] Error 2
scons: *** Error 2
[...]

PR #1106, opened 2018-12-06
Issue #1104 - OpenGEE Build Fails on Centos 7

tst-eclamar:
This bug is due to an update of Centos 7 from "7.5" to "7.6" - libpng15 is added in "7.6", and is incompatible with libpng12.

### DIFF
--- a/earth_enterprise/src/third_party/qt/SConscript
+++ b/earth_enterprise/src/third_party/qt/SConscript
@@ -69,8 +69,6 @@ if third_party_env['is_min_ubuntu'] and not third_party_env['native_cc']:
 else:
   env_opt += ''
 
-env_opt += ' PKG_CONFIG_PATH=. '
-
 if qt_env['release'] or qt_env['optimize']:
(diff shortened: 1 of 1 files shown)


tst-nfarah:
Let's add a description for this PR and also target release_5.2.5 branch as the branch to merge to. We likely want to merge that branch into master as soon as the this PR was approved.

tst-ccamp:
OK, I have an idea that might simplify this. This is really the better solution, but we only want to do it if we can find the file. So maybe add a comment to say that this is a fix for CentOS & RHEL 7 and add a check to see if the file is there before adding the symlink and then that should be good enough.

tst-ccamp:
After the update, I tested on RHEL 7 and it worked fine. I'm going to go ahead and approve pending a successful test on RHEL 6

tst-ccamp:
This update passed on RHEL 6, CentOS 6, and Ubuntu 16.04 also. So, merging.

PR #1110, opened 2018-12-10
Issue #1104 - OpenGEE Build Fails on Centos 7 (#1106)

tst-nfarah:
* #1104 OpenGEE Build Fails on Centos 7


tst-nfarah:
updating master to get a critical bug fix

Issue #1115, opened 2018-12-11
Make sure first part of version string is numeric


tst-rwildes:
If the first part of the version string is not numeric upgrades fail.  This is only an issue with non-release builds.

tst-rwildes:
Issue fixed and merged into release 5.2.5 and master.

PR #1116, opened 2018-12-11
fix upgrading to non-releasable build

tst-rwildes:
Fixes issue #1115 Always make the first part of a version string numeric


earth_enterprise/src/scons/getversion.py
```diff
@@ -80,15 +80,17 @@ def _GitVersionNameAndBuildNumber():
         branchName = _GitBranchName()
         if not branchName:
             # we are a detached head not on a release tag so just treat
-            # the raw describe like a topic branch name
-            return _GetCommitRawDescription(), str(_GitCommitCount())
+            # the the first part of the raw describe as the release name
+            # added the b to the build number to signal this is not a releaseable build
```
tst-hrajamoney: typo: releasable

PR #1119, opened 2018-12-12
create the gevars file on fusion install too

tst-rwildes:
Fixes issue #1120

### DIFF
--- a/earth_enterprise/src/installer/install_fusion.sh
+++ b/earth_enterprise/src/installer/install_fusion.sh
@@ -31,6 +31,8 @@ DEFAULTGEFUSIONUSER_NAME="gefusionuser"
 DEFAULTGROUPNAME="gegroup"
 GEFUSIONUSER_NAME=$DEFAULTGEFUSIONUSER_NAME
 GROUPNAME=$DEFAULTGROUPNAME
+GEPGUSER_NAME="gepguser"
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -677,6 +679,9 @@ setup_fusion_daemon()
 	# setup fusion daemon
 	printf "Setting up the Fusion daemon...\n"
 
+    # Create a new file ‘/etc/init.d/gevars.sh’ and add the below lines.
```
tst-mbiancalana: Indentation mismatch here due to spaces vs hard tabs.
Here string might look prettier, but this will work.

tst-mbiancalana: So I'd say fix the indentation, but it's up to you on keeping this perfectly good command that's already been tested or switching it out and waiting for re-testing.

tst-jhaight:
Verified on CentOS 7 installing fusion only via `install_fusion.sh` installs all needed scripts and fusion can successfully start.  Also verified installing server via shell script after fusion does not cause any issues.  LGTM

tst-nfarah:
if this bug happens in previous OpenGEE version(s), we need to document it in release note as fixed.

tst-mbiancalana:
I approve.

PR #1123, opened 2018-12-13
#1122 Explicitly import Image module from PIL

### LINKED ISSUE #1122: Portable server incorrectly displays "PIL is not available"
Portable server displays a message about missing PIL library even though Python Imaging Library module is installed.

tst-twilk:
This closes #1122. Portable - test for Python Imaging Library may returns an error "PIL is not available" depending on PIL installation method.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/local_server.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/local_server.py
@@ -29,8 +29,8 @@ import sys
 
 try:
   pil_enabled = True
-  import Image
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/fusion/portableglobe/servers/local_server.py
```diff
@@ -29,8 +29,8 @@
 
 try:
   pil_enabled = True
```
tst-dsugar: of course it is amusing we set it before we do the import

Issue #1103, opened 2018-12-04
Build Instructions for Centos6 are incorrect


tst-basilhuffman:
Currently, the build instructions for CentOS 6 are incorrect.  The issues that need to be corrected in `BUILD_RHEL_CentOS.md`

- when installing gcc 4.8, `source /opt/rh/devtoolset-2/enable` should be executed in the same manner as RHEL6
- when installing Gtest, `envsubst` is not available by default and must be installed via `sudo yum -y install gettext gettext-devel gettext-libs`

The build instructions should be modified to reflect this.

tst-basilhuffman:
@justrandyc this is the issue I mentioned eariler, I provided as much information as I have

tst-rcrihfield:
The gcc 4.8 issue is simply unclear documentation, the step you refer to isn't exclusive to RHEL6 it applies to any of the 4 OS versions.

The gtest issue around envsubst I agree with, it should indeed be called out as a dependency and installed prior to building/installing the gtest package.  

I will further investigate the build issues you mention

tst-cjeffries:
Ive got a workaround to get it building:

So i followed the instructions but did two additional steps.

1st: I installed gcc-c++ (`sudo yum install gcc-c++`) (did this after the Install Additional Packages Section)

Even though it is installed in devtoolset-2, it looks for g++ in /usr/bin. And when it doesnt find it, it passes an NoneType around in python, and causes it to error out (calls the rfind method on it). (edited)
But when it is in /usr/bin, we get around that.

2nd: I installed gettext (`sudo yum install gettext`) to get the envsubst command (Did this right after my previous step). This is needed to properly install gtest.

tst-cjeffries:
I have got a workaround to get it building.

I followed the instructions (including the RHEL6 Source command)) but did two additional steps after the Installing Additional Packages step:

1st: I installed gcc-c++ (`sudo yum install gcc-c++`)

Even though it is installed in devtoolset-2, it looks for g++ in /usr/bin. And when scons doesn't find it, it passes a NoneType around in python, and causes it to error out (calls its rfind method, which doesn't exist). But when it is in /usr/bin, we get around that.

2nd: I installed gettext to get the envsubst command (`sudo yum install gettest`). 

This is needed to properly install GTest.

3: I installed doxygen (`sudo yum install doxygen').

This is needed for gdal apparently.

tst-cjeffries:
[...]

Issue #686, opened 2018-02-01
Scons fails to detect libpng library on CentOS 6


tst-ccamp:
This issue comes up a lot because a default g++ compiler is not installed when following the build setup instructions for CentOS 6. I don't believe this issue comes up in RHEL 6 because it's included with a documented package. Furthermore, even when fixing the issue it still will need to rescan for the configs (--config=force on the scons command line). This should happen by default since the time to do the scan is negligible. 

# AC
1. Update build instructions to include default g++ for CentOS 6 (gcc-g++ package) or update scons to use devtools g++ install
1. Update scons to always scan for libpng (it may be possible to disable caching)

tst-ppenev:
Have you timed how long it takes to rescan for all the libraries used in a build?

I think the proper fix is to update SCons scripts to report a proper error when they can't find a compiler, instead of making strange assumptions, continuing, and outputting hard to decipher message.

tst-ccamp:
I didn't bother timing because it's probably less time than it takes to print the message to the console.

tst-ppenev:
Scanning for libraries involves running GCC for every header, and dynamic library you're testing for.  That usually takes, at least, a few minutes for all the libraries used by GEE.

tst-ccamp:
I ran some tests. It looks like the `--config=force` flag works well for the third_party target, but I got an error for the default build target. Here are the results on the third_party target. If we just stick to that, I don't think we'll add to the total build time. And the build will stop if it can't find the library on the third_party target.

```
ccamp@Mainframe:/opt/dev/tst/earthenterprise/earth_enterprise/src$ time scons -j12 release=1 --config=force third_party
scons: Reading SConscript files ...

scons: warning: The build_dir keyword has been deprecated; use the variant_dir keyword instead.
File "/opt/dev/tst/earthenterprise/earth_enterprise/src/SConstruct", line 507, in <module>
Checking for C library gtest... yes
...
scons: done building targets.

real	0m4.143s
user	0m4.104s
sys	0m0.168s

ccamp@Mainframe:/opt/dev/tst/earthenterprise/earth_enterprise/src$ time scons -j12 release=1  third_party
scons: Reading SConscript files ...

scons: warning: The build_dir keyword has been deprecated; use the variant_dir keyword instead.
File "/opt/dev/tst/earthenterprise/earth_enterprise/src/SConstruct", line 507, in <module>
Checking for C library gtest... (cached) yes
...
scons: done building targets.

real	0m4.122s
[...]

PR #1134, opened 2018-12-17
#686 Add C++ compiler check & update build instructions

jbonhag:
`/usr/bin/g++` not being installed causes a couple of issues:

* env['CXX'] will be None, and the version check will throw an exception
* third_party g++ has /usr/bin/g++ hard-coded (e.g. in ./src/NATIVE-REL-x86_64/third_party/bin/g++)

Also, on a clean CentOS 6 install, the gtest RPM will not build because of a missing dependency.

I updated the documentation to add the default development tools, and also put a compiler check into the SConstruct file.

As long as all these requirements are met, libpng is detected and there are no problems building; I tested on both CentOS 6.8 and CentOS 6.10.

### DIFF
--- a/earth_enterprise/BUILD_RHEL_CentOS.md
+++ b/earth_enterprise/BUILD_RHEL_CentOS.md
@@ -75,12 +75,14 @@ sudo yum install git-lfs
 
 ## GCC 4.8
 
-### CentOS 7 and RHEL 7
+For all flavors, install the standard development/build tools:
 
 ```bash
 yum install ant bzip2 doxygen gcc-c++ patch python-argparse python-setuptools \
   swig tar
 ```
+
+For CentOS 6 and RHEL 6, also install the devtoolset toolchain.
 
 ### CentOS 6
 
@@ -163,6 +165,7 @@ the rest of the GEE build process.
 To clone this Git repository and build the RPM on RHEL6, execute the following:
 
 ```bash
+sudo yum install -y cmake gettext rpm-build
 mkdir -p ~/opengee/rpm-build/
 cd ~/opengee/rpm-build/
 
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -480,6 +480,10 @@ gee_version_number = env.get_open_gee_version().get_short()
 
 
 # Check if the version of the default compiler is at least 4.8:
+if not env['CXX']:
+    print "Please install the default C++ compiler for your distribution."
+    Exit(1)
+
 version = opengee.c_compiler.get_cc_version(env['CXX'])
 if not opengee.version.is_version_ge(version, [4, 8]):
     # Check for GCC 4.8 from an alternative toolchain installation on RHEL 6:


jbonhag:
I signed it!

earth_enterprise/src/SConstruct
```diff
@@ -480,6 +480,10 @@ gee_version_number = env.get_open_gee_version().get_short()
 
 
 # Check if the version of the default compiler is at least 4.8:
+if not env['CXX']:
+    print "Please install the default C++ compiler for your distribution."
```
tst-ppenev: Grammar: Please, add a comma after 'Please'.

tst-nfarah:
Does this PR fixes #1103 since #686 could not reproduced any more ?

tst-ppenev:
There need to be some documentation updates for #1103, and a re-test for a debug build.

I think #686 involves SCons caching results about what libraries have been found.  I doubt that this pull request fixes it, but, it's worth re-testing.

earth_enterprise/src/SConstruct
tst-mbiancalana: I disagree on the grammar. The sentence in question is correct without an additional comma.

tst-mbiancalana:
I was unable to reproduce the bug mentioned in #686 as well. That issue may have been fixed elsewhere.

This branch does address most, if not all, of the concerns raised in #1103, though. Maybe we should re-name this branch to reference that issue?

jbonhag:
I don't believe caching was a factor here.  The build script had no issues detecting libpng whether or not that build step was cached.

I'd be happy to rename the branch -- it does seem like there's a lot of overlap with #1103.

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -75,13 +75,15 @@ sudo yum install git-lfs
 
 ## GCC 4.8
 
-### CentOS 7 and RHEL 7
+For all flavors, install the standard development/build tools:
```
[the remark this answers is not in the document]
jbonhag: That's a good suggestion -- I'll change it to be less informal.

earth_enterprise/src/SConstruct
jbonhag: Final verdict?

tst-basilhuffman: Keep as is

tst-mbiancalana:
I'll leave the decision to rename the branch up to @tst-ccamp. Even if it doesn't happen I think we can sort out what happened.

earth_enterprise/src/SConstruct
```diff
@@ -480,6 +480,10 @@ gee_version_number = env.get_open_gee_version().get_short()
 
 
 # Check if the version of the default compiler is at least 4.8:
+if not env['CXX']:
+    print "Please, install the default C++ compiler for your distribution."
+    Exit(1)
+
 version = opengee.c_compiler.get_cc_version(env['CXX'])
```
tst-ccamp: Couldn't you also fix this by first checking to see if env['CXX'] is set? If it isn't, then you can still do the check for devtools below.

jbonhag: The system default compiler is required for building the third party modules.  Without it, the build fails here:

```
checking for c++... /home/jeff/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party/bin/g++
checking whether the C++ compiler (/home/jeff/earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party/bin/g++  ) works... no
configure: error: installation or configuration problem: C++ compiler cannot create executables.
```

This is because `third_party/bin/g++` is created with a hardcoded reference to `/usr/bin/g++`, even after running `source /opt/rh/devtoolset-2/enable`.

There might need to be a new issue to allow the third party modules to be built with another C++ compiler (e.g. the devtoolset toolchain).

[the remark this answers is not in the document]
jbonhag: Might I suggest

```
if env['CXX']:
    version = opengee.c_compiler.get_cc_version(env['CXX'])
else:
    version = []
```

since it's slightly easier to unpack?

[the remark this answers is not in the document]
tst-mbiancalana: Go for it

tst-ccamp: Ah, that makes sense. I'll write a follow-up ticket to fix that issue.

tst-ccamp: Added #1142

Issue #1137, opened 2018-12-26
Builds are failing on master branch for all OSes


tst-twilk:
Recent merge introduced memory test code that fails to build on older compilers. Added tests need to be updated to support older compilers.

```
g++ -o gee_build/REL-el7-gcc4.8.5-x86_64/fusion/autoingest/sysman/memory-tests/build-raster-project_resourcetest.o -c -O2 -DNDEBUG -pipe -pthread -Wall -Werror -Wno-parentheses -Wno-error=unused-result -Wno-error=unused-local-typedefs -Wno-error=unused-but-set-variable -std=gnu++11 -D_LARGEFILE64_SOURCE -DQT_THREAD_SUPPORT -DQT_NO_ASCII_CAST -DQT_CLEAN_NAMESPACE -D_REENTRANT -DGUNIT_NO_GOOGLE3 -DKH_MAX_NUM_JOBS_COEFF=16 "-DOPENGEE_ASSETS_DIR=get_self_dir() + \"/fusion/testdata/gevol/assets\"" -Igee_build/REL-el7-gcc4.8.5-x86_64/include -Igee_build/REL-el7-gcc4.8.5-x86_64 -I. -Igee_build/REL-el7-gcc4.8.5-x86_64/common -Icommon -Icommon/khmisc -Ikeyhole -Igee_build/REL-el7-gcc4.8.5-x86_64/fusion -Ifusion -Ifusion/asset-tests/include -Ifusion/autoingest -Ifusion/autoingest/sysman fusion/autoingest/sysman/memory-tests/build-raster-project_resourcetest.cc
fusion/autoingest/sysman/parse-raster-project-xml-no-content_unittest/parse-raster-project-xml-no-content_unittest.cc:1:35: fatal error: experimental/filesystem: No such file or directory
 #include <experimental/filesystem>
                                   ^
compilation terminated.
fusion/autoingest/sysman/memory-tests/build-raster-project_resourcetest.cc:1:35: fatal error: experimental/filesystem: No such file or directory
 #include <experimental/filesystem>
                                   ^
compilation terminated.
```

tst-lsavoie:
Closed via PR #1138.

PR #1138, opened 2018-12-26
#1137: Fix Memory Test Builds on Red Hat

tst-ppenev:
* Removed unused includes of `experimental/filesystem`.


tst-jkent:
I double checked the Travis CI logs and there appears to be another error occurring, yet it still passes.
capture

.travis.yml
```diff
@@ -9,7 +9,7 @@ before_install:
 - sudo apt-get update -qq
 - sudo apt-get install -qq
     alien bison++ bisonc++ cmake doxygen freeglut3-dev flex gettext
-    libc6 libc6-dev libcap-dev libfreetype6 libfreetype6-dev
+    libboost-dev libc6 libc6-dev libcap-dev libfreetype6 libfreetype6-dev
```
tst-rwildes: Are we including boost C++ libraries for GEE now?

tst-rwildes: Not that I'm complaining :).  Lots of good stuff in there.

tst-ppenev: I explained this in the commit message, but it seems like some versions of the `libyaml-cpp-dev` don't list `libboost-dev` as a dependency, but `libyaml` uses Boost shared pointers, so the build fails, if `libboost-dev` isn't installed.

The `libyaml-cpp-dev` package on my Ubuntu 14 installation seems to list the dependency, but it failed for Travis automated build, so I added it here to fix the build.  We may be able to remove the explicit `libboost-dev` listing in the future.

tst-ppenev:
@tst-nfarah, the Travis is build currently failing because of the script that runs the unit tests.  That's what I'm working on now.  However, there's also a new unit test for mishandling of invalid XML input.  That issues has not been fixed, yet, so the unit test will keep failing until we merge a fix.

We could either merge pull requests that only fail on that unit test, or disable it, and make sure we fix the issue and re-enable the unit test.

tst-nfarah:
@tst-ppenev i vote for disabling that unit test and enabling it back on the PR that take care of the fix.

tst-ppenev:
@tst-nfarah: OK.  I'll go with that.

tst-rwildes:
I think it is critical to get master building again as soon as we can.  Even if this means temporary removing some unit tests.  Having failed builds on master is causing a lot of confusion and disruption right now.

earth_enterprise/src/fusion/asset-tests/bin/generate_test_asset_xmls.py
```diff
@@ -9,13 +9,73 @@
...
+# Create some decorators for methods that serve as tasks to be executed.
+def execute_once(func):
+    """Make a function exectule only once.  Calls after the first one return
+    immediately.
+    """
+
+    @functools.wraps(func)
+    def wrapper_execute_once(*args, **kwargs):
+        if wrapper_execute_once.executed:
+            return
```
tst-rwildes: should you save res and return it here?

tst-ppenev: This is only used for functions that don't return a meaningful value, and this return value might tell you the function didn't execute.

There's already a library that implements memoization, which cache the return results for each set of input parameters, but I thought that was more than was needed in this case, so I just I just wrote this short decorator.

If this is confusing, I'll just switch over to the library decorator.

tst-ppenev:
@tst-nfarah, this is ready for review.

.travis.yml
```diff
@@ -7,7 +7,19 @@ compiler:  gcc
 
 before_install:
 - sudo apt-get update -qq
-- sudo apt-get install -qq libperl4-corelibs-perl libtool xorg-dev doxygen python-dev alien swig libgtest-dev libstdc++6 libxml2-dev gettext libxinerama-dev libxft-dev libxrandr-dev libxcursor-dev libgdbm-dev libc6 libc6-dev libmng-dev zlib1g-dev libcap-dev libpng12-0 libpng12-dev freeglut3-dev flex libx11-dev bison++ bisonc++ libjpeg-dev libjpeg8-dev python2.7 python2.7-dev python2.7-psycopg2 libogdi3.2-dev libgif-dev libxerces-c-dev libgeos-dev libgeos++-dev libfreetype6 libfreetype6-dev python-imaging libproj-dev python-setuptools libgif-dev libxerces-c-dev libcap-dev libpq-dev openssl libxml2-utils libxmu-dev python-git cmake
+- sudo apt-get install -qq
+    alien bison++ bisonc++ cmake doxygen freeglut3-dev flex gettext
```
tst-lsavoie: For the curious, I verified that the only meaningful changes in these lines are the addition of `libboost-dev` and `libyaml-cpp-dev`.

earth_enterprise/src/fusion/asset-tests/bin/generate_test_asset_xmls.py
```diff
@@ -9,13 +9,73 @@
...
+
+    def advance(self, progress_amount):
+        self.value += progress_amount
+        self.update_progress()
+
+    def close(self):
+        sys.stdout.write(os.linesep)
+
+    def update_progress(self):
+        if (self.value - self.start_value) % self.update_interval == 0:
```
tst-lsavoie: Minor: if `progress_amount` passed to `advance` is not 1, this may skip printing progress sometimes.

.travis.yml
tst-lsavoie: The others are needed. I just wanted to point out what the changes were so others don't have to compare the lists again.

tst-nfarah: ok @tst-lsavoie

earth_enterprise/src/fusion/asset-tests/bin/generate_test_asset_xmls.py
```diff
@@ -417,18 +480,63 @@ def generate_invalid_raster_projects(self):
...
 
+        self.progress_meter.close()
+
 
 def main():
-    generator = TestAssetGenerator()
+    parser = argparse.ArgumentParser(
+        description="Generate Fusion asset XMLs for testing.")
+
+    parser.add_argument(
```
tst-lsavoie: Minor: I prefer "positive" arguments like `--generate-valid-raster-projects`. It makes it easier to reason about the code that uses them. However, that's more of a style question.

tst-lsavoie:
It passed in Travis and I successfully reproduced the error on RHEL 6 and verified that this fixes it. Any reason not to merge?

tst-nfarah:
I don't see why not @tst-lsavoie. Since this was reviewed and tested, let's get it merged.

Issue #1140, opened 2018-12-27
Update master to use version 5.3.0


tst-rwildes:
The next version for Open GEE will be `5.3.0`.  Should update this early so master can start producing those versions.

tst-rwildes:
Master has been moved to use version 5.3.0

PR #1141, opened 2018-12-27
move master to version 5.3.0

tst-rwildes:
Addresses issue #1140 

Changes based on: https://github.com/google/earthenterprise/wiki/Development:-Release-Process (step 4)


tst-jlarocco:
This is only step 4 in the Release Process?  The documentation and release note updates will be done later?

If so, it looks good to me!

tst-rwildes:
> This is only step 4 in the Release Process? The documentation and release note updates will be done later?
> 
> If so, it looks good to me!

Yes this is only that step 4.

Issue #1063, opened 2018-10-11
Reverse the order of known issues in release notes (5.2.4 & 5.2.5)


tst-ccamp:
Make this change on the master branch

tst-nfarah:
We need to document this [issue ](https://github.com/google/earthenterprise/issues/1158)

tst-cjeffries:
This change has been made in this change has been made in the 5.2.5 release notes. Can we close this issue? @tst-ccamp

tst-nfarah:
Let's close it once this [PR ](https://github.com/google/earthenterprise/pull/1159) is reviewed and merged.

Issue #1079, opened 2018-10-24
Add ability to change the historical imagery server with gerewritedbroot


tst-emillergalow:
The gerewritedbroot utility is able to make a copy of an existing dbroot and make modifications to it but it does not support changing the server url of the location of historical imagery data. Portable globes which reference this url but are unable to access it cause Earth Client to hang. It would be useful to be able to modify or remove this from portable globes that will not have access to the server.

[Portable Server Command Line Reference](http://www.opengee.org/geedocs/5.2.4/answer/4485225.html#command_line_reference)

geglxinfo can be used to extract all files from a portable globe and gedumpdbroot will output the contents of the globe's dbroot. Somewhere in the output from gedumpdbroot there is an item named time_machine_options:

```  
geglxinfo --glx example_globe.glb --extract_all_files --output example_output_dir

gedumpdbroot example_output_dir/dbroot/dbroot_file

...
time_machine_options { 
    server_url: "http://example_server/example_globe?db=tm"
    is_timemachine: false
  }
...
```  

**Acceptance Criteria:**
- Add option to gerewritedbroot to remove references to the historical imagery server url
- Able to use the portable server command line tools to edit the dbroot of an existing globe and repackage it into a usable .glb file (geglxinfo -> gerewritedbroot ->geportableglobepacker)

tst-emillergalow:
[PR 1081](https://github.com/google/earthenterprise/pull/1081)

Issue #1068, opened 2018-10-17
Need to support WMS option when publishing databases from command line


tst-ccamp:

tst-lsavoie:
Closed vi PR #1040.

PR #1147, opened 2019-01-07
Update 5.2.5 Release Notes

tst-cjeffries:
Add fixes and known issues for 5.2.5. Also reverse the ordering for known issues per issue #1063


tst-cjeffries:
https://tst-cjeffries.github.io/earthenterprise/geedocs/5.2.5/answer/7160006.html

docs/geedocs/5.2.5/answer/7160006.html
```diff
@@ -136,6 +141,11 @@ <h5>
             <td>Symlinks are incorrectly world-writable files after installation from RPM or Deb</td>
             <td>Symlinks are now detected during the building of RPMs and Debs and are copied into the package correctly.</td>
           </tr>
+          <tr>
+             <td>1120</td>
+             <td>Installing fusion without server fails because of a missing vars file</td>
```
tst-nfarah: Since this was introduced and fixed in 5.2.5,  we don't need to document it.

tst-nfarah: These two new enhancements need to be documented:
https://github.com/google/earthenterprise/issues/1079
https://github.com/google/earthenterprise/issues/1068

See list of resolved issues for 5.2.5:
https://github.com/google/earthenterprise/issues?q=is%3Aissue+milestone%3A5.2.5+is%3Aclosed

tst-cjeffries: Add a reference to 1079, 1068 was mentioned already in the line about new publish options from the command line. Couldn't find anything else in the list of resolved issues that needed added.

tst-cjeffries: Removed it.

tst-swanson:
Few minor typos:
When a GEE server is queried for server definitions the response can be returned in proper JSON format. The v parameter can be added to **the** query string to specify which version of the API will be returned. The valid options are v=1 and v=2. Version one returns data in the current format. Version two returns data in JSON format. An example **URL that** returns data in JSON format looks like: http://<GEE earth server>/<database_publish_point>/query?request=Json&v=2.
**Installation** scripts are disabled when Open GEE packages are installed. The install and uninstall scripts will now check if Open GEE has been installed with a package manager before installation begins, and will not make any changes when the packages are installed.
Resolved Issues
1038
genewmapdatabase does not allow flat imagery to be used with mercator **databases**
Known Issues
1102
Should this definitely be REHL6, not RHEL6?
1135
**JavaScript** console error in portable server when loading build date
825
Geserver fails to **start up** fully due to conflicting protobuf library
Throughout
Cap Mercator
Use Earth Client or Earth client
Should Cutter be capped?
I think 2d and 3d should probably be 2D and 3D

tst-cjeffries:
Good catches on all of this. Though the wrong PR must be linked in the JIRA ticket. Ill make sure to fix that.

PR #1153, opened 2019-01-10
Progress messages for state updates always say they're on iteration 1

### LINKED ISSUE #1143: Progress messages for state updates always say they're on iteration 1
When updating the state of lots of asset versions, a progress message is printed out at the NFY_PROGRESS level that says "Iteration: \<x\> | Total Iterations: \<y\>". This is currently on [line 109 of AssetVersionD.cpp](https://github.com/google/earthenterprise/blob/9a16b1c99023924e5268e5a2abd38c6fcad72b71/earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp#L109). However, this message always indicates that we are on iteration 1 - it never increments (in other words, the \<x\> in the above string is always 1). We should fix this code to print the current iteration correctly.

tst-jkent:
Added a statement to increment the iteration counter when iterating through the parents and children of an asset.
To test, build an asset and monitor the /var/opt/google/log/gesystemmanager.log file to make sure the iteration count in the log message increases if there is more than one iteration.
Tested on Ubuntu 16.
Fixes #1143

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
@@ -127,6 +127,8 @@ AssetVersionImplD::StateChangeNotifier::SendNotifications(
       notify(NFY_WARN, "'%s' has broken %s '%s'",
              assetVersion->GetRef().c_str(), typeName.c_str(), ref.c_str());
     }
+
+    i++;
   }
   toNotify->clear();
 }


earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -127,6 +127,8 @@ AssetVersionImplD::StateChangeNotifier::SendNotifications(
       notify(NFY_WARN, "'%s' has broken %s '%s'",
              assetVersion->GetRef().c_str(), typeName.c_str(), ref.c_str());
     }
+
+    i++;
```
tst-ppenev: Preference: Although, it would not make difference in this case, my style preference is to rewrite the range `for` to a regular `for` with an explicit iteration expression, and add `i++` to that expression.  That way, you don't have to worry about `continue` statements making the counter out of sync.

PR #1159, opened 2019-01-16
Add 1158 to known issues.

tst-cjeffries:


docs/geedocs/5.2.5/answer/7160006.html
```diff
@@ -163,6 +163,11 @@ <h5>
           <th class="narrow">Description</th>
           <th>Workaround</th>
         </tr>
+        <tr>
+          <td>1158</td>
+          <td>Portable server loses geeServerDefs for a served map/globe as soon as it is viewed</td>
```
tst-jcupp:
```suggestion
          <td>Portable files served from Portable Server don't provide geeServerDefs to clients through the Maps API</td>
```

tst-nfarah: @tst-jcupp this issue happens only if someone is viewing that map/globe. Otherwise if no one is viewing it, the issue does not happen and server definition is returned ok.

tst-cjeffries: So how do we want to word this?

docs/geedocs/5.2.5/answer/7160006.html
```diff
@@ -163,6 +163,11 @@ <h5>
           <th class="narrow">Description</th>
           <th>Workaround</th>
         </tr>
+        <tr>
+          <td>1158</td>
+          <td>Portable files served by Portable Server don't provide geeServerDefs to clients through the Maps API if the file is being viewed</td>
```
tst-nfarah: `file` --> `map or globe`

tst-cjeffries: Can do.

PR #1162, opened 2019-01-18
Merge state-update-optimization fixes into master

tst-jkent:
Merged state-update-optimization into master and resolved all conflicts.


earth_enterprise/src/scons/getversion.py
```diff
@@ -8,6 +8,8 @@
 import re
 
 
+
```
tst-lsavoie: Very minor: there's a couple of random extra newlines here. I'm fine with keeping them or removing them.

PR #1174, opened 2019-01-23
add xml cache purge

tst-basilhuffman:
terminates when cache is full
due to pointer semantics, its best to terminate on release of documents and parsers

### DIFF
--- a/docs/geedocs/5.2.5/answer/176738.html
+++ b/docs/geedocs/5.2.5/answer/176738.html
@@ -273,6 +273,10 @@ MAX_HEAP_SIZE=&lt;value&gt;
 <br><br>
 # the size of the allocation blocks<br>
 BLOCK_SIZE=&lt;value&gt;
+<br><br>
+# whether or not purge the xml cache
+# false = 0, true = 1
+PURGE = &lt;value&gt; 
 </code>
 <h2>Learn more</h2>
 <ul>
--- a/docs/geedocs/5.2.5/answer/176738.html
+++ b/docs/geedocs/5.2.5/answer/176738.html
@@ -273,14 +273,6 @@ MAX_HEAP_SIZE=&lt;value&gt;
 <br><br>
 # the size of the allocation blocks<br>
 BLOCK_SIZE=&lt;value&gt;
-<br><br>
-# whether or not purge the xml cache
-# false = 0, true = 1
-PURGE = &lt;value&gt; 
-<br><br>
-# the level at which to purge, may be 1 (most often) to 5 (least often)
-# default is 3
-PURGE_LEVEL=3
 </code>
 <h2>Learn more</h2>
 <ul>
--- a/docs/geedocs/5.2.5/answer/176738.html
+++ b/docs/geedocs/5.2.5/answer/176738.html
@@ -277,6 +277,10 @@ BLOCK_SIZE=&lt;value&gt;
 # whether or not purge the xml cache
 # false = 0, true = 1
 PURGE = &lt;value&gt; 
+<br><br>
+# the level at which to purge, may be 1 (most often) to 5 (least often)
+# default is 3
+PURGE_LEVEL=3
 </code>
 <h2>Learn more</h2>
 <ul>
--- a/docs/geedocs/5.3.0/answer/176738.html
+++ b/docs/geedocs/5.3.0/answer/176738.html
@@ -278,7 +278,8 @@ BLOCK_SIZE=&lt;value&gt;
 PURGE=0
 <br><br>
 # the level at which to purge, may be 1 (most often) to 5 (least often)
-PURGE_LEVEL=100
+# default is 3
+PURGE_LEVEL=3
 </code>
 <h2>Learn more</h2>
 <ul>
--- a/docs/geedocs/5.3.0/answer/176738.html
+++ b/docs/geedocs/5.3.0/answer/176738.html
@@ -277,8 +277,8 @@ BLOCK_SIZE=&lt;value&gt;
 # whether or not to purge: 0 or 1, default is 0
 PURGE=0
 <br><br>
-# the percentage of the cache at which to purge, 0-100 default 100
-PERCENT=100
+# the level at which to purge, may be 1 (most often) to 5 (least often)
+PURGE_LEVEL=100
 </code>
 <h2>Learn more</h2>
 <ul>
--- a/docs/geedocs/5.3.0/answer/176738.html
+++ b/docs/geedocs/5.3.0/answer/176738.html
@@ -257,6 +257,29 @@ The asset cache size is controlled by <code>AssetCacheSize</code> and the asset 
 Note that the asset version cache is generally larger than the asset cache because it may contain multiple versions of the same asset.</p>
 
 Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
+<br><br>
+In some instances, memory utilization may become problematic. An experimental feature is included which will allow for modifying aspects of the the XML cache. This is controlled through <code>/etc/opt/google/XMLparams</code>. It takes the form of a standard configuration file, and it is in the following format.
+<br><br>
+<code>
+# comments are preceded by '#'
+<br>
+# options are case-insensitive, and all need not be present
+<br>
+# the initial size of the heap<br>
+INIT_HEAP_SIZE=&lt;value&gt;
+<br><br>
+# the max size of the heap<br>
+MAX_HEAP_SIZE=&lt;value&gt;
+<br><br>
+# the size of the allocation blocks<br>
+BLOCK_SIZE=&lt;value&gt;
+<br><br>
+# whether or not to purge: 0 or 1, default is 0
+PURGE=0
+<br><br>
+# the percentage of the cache at which to purge, 0-100 default 100
+PERCENT=100
+</code>
 <h2>Learn more</h2>
 <ul>
 <li><a href="../answer/6033487.html">Benchmarking outcomes for applying task rules</a></li>
--- a/earth_enterprise/src/common/khxml/khxml.cpp
+++ b/earth_enterprise/src/common/khxml/khxml.cpp
@@ -91,7 +91,7 @@ class PercentError : public XmlParamsException
 public:
     const char* what() const noexcept
     {
-        return "Cache purging can only be performed when capacity is between 0-100%";
+        return "Cache purging levels can only be 1-5";
     }
 };
 
@@ -114,7 +114,7 @@ void validateXMLParameters()
         throw SizeError();
     }
 
-    if (percent > 100 || percent < 0)
+    if (percent > 5 || percent < 1)
     {
         throw PercentError();
     }	
@@ -126,7 +126,7 @@ void setDefaultValues()
    maxDOMHeapAllocSize     = 0x20000;
    maxDOMSubAllocationSize = 0x1000;
    terminateCache = false;
-   percent = 1.0;
+   percent = 1.5;
 }
 
 void getInitValues()
@@ -155,7 +155,14 @@ void getInitValues()
                 terminateCache = std::stol(it.second);
             else
             {
-                percent = static_cast<float>(std::stol(it.second)/100.0);
+                switch (std::stol(it.second))
+                {
+                    case 1: percent = 0.00; break;
+                    case 2: percent = 0.75; break;
+                    case 3: percent = 1.50; break;
+                    case 4: percent = 2.25; break;
+                    case 5: percent = 3.00; break;
+                };
             }
         }
     }
@@ -224,7 +231,7 @@ static khMutexBase checkTermLock = KH_MUTEX_BASE_INITIALIZER;
 static khMutexBase cacheLock = KH_MUTEX_BASE_INITIALIZER;
 
 // only terminate when certain conditions are met
-void  readyForTerm()
+void  reInitIfReady()
 {
   khLockGuard guard(checkTermLock);
   static XMLSSize_t threashold = static_cast<XMLSSize_t>
@@ -250,7 +257,7 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
 
   if (terminateCache)
   {
-    readyForTerm();
+    reInitIfReady();
   }
   try {
     DOMImplementation* impl =
@@ -427,7 +434,7 @@ CreateDOMParser(void) throw()
   InitializeXMLLibrary();
   if (terminateCache)
   {
-    readyForTerm();
+    reInitIfReady();
   }
   class FatalErrorHandler : public DOMErrorHandler {
    public:
--- a/earth_enterprise/src/common/khxml/khxml.cpp
+++ b/earth_enterprise/src/common/khxml/khxml.cpp
@@ -89,7 +89,7 @@ public:
  * helper class for synchronizing events related to purging
  * the Xerces cache
  *
- * keeps an internal queue and a notifier as to whether or
+ * keeps an internal count and a notifier as to whether or
  * not the mutex on creating docs/parsers is currently locked
  *
  * singeton class, thread-safe
@@ -98,7 +98,7 @@ class terminateGuard
 {
 private:
     static khMutexBase qMutex, lockMutex;
-    static std::queue<char> objQ;
+    static uint32_t numDocs;
     static bool objLock;
     terminateGuard() = default;
 
@@ -109,22 +109,22 @@ public:
         return _instance;
     }
 
-    static void enqueue()
+    static void addObj()
     {
         khLockGuard guard(qMutex);
-        objQ.push('0');
+        ++numDocs;
     }
 
-    static void dequeue()
+    static void removeObj()
     {
         khLockGuard guard(qMutex);
-        objQ.pop();
+        if (numDocs > 0) --numDocs;
     }
 
     static uint32_t size()
     {
         khLockGuard guard(qMutex);
-        return objQ.size();
+        return numDocs;
     }
 
     static void lock()
@@ -148,7 +148,7 @@ public:
 
 khMutexBase terminateGuard::qMutex = KH_MUTEX_BASE_INITIALIZER; 
 khMutexBase terminateGuard::lockMutex = KH_MUTEX_BASE_INITIALIZER; 
-std::queue<char> terminateGuard::objQ;
+uint32_t terminateGuard::numDocs = 0;
 bool terminateGuard::objLock = false;
 
 void validateXMLParameters()
@@ -302,7 +302,7 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
       xmlLibLock.Lock();
       terminateGuard::instance().lock();
     }
-    terminateGuard::instance().enqueue();
+    terminateGuard::instance().addObj();
   }
   try {
     DOMImplementation* impl =
@@ -317,7 +317,7 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
      {
        xmlLibLock.Unlock();
        terminateGuard::instance().unlock();
-       terminateGuard::instance().dequeue();
+       terminateGuard::instance().removeObj();
      }
      return 0;
    }
@@ -488,7 +488,7 @@ CreateDOMParser(void) throw()
        xmlLibLock.Lock();
        terminateGuard::instance().lock();
     }
-    terminateGuard::instance().enqueue();
+    terminateGuard::instance().addObj();
   }
   class FatalErrorHandler : public DOMErrorHandler {
    public:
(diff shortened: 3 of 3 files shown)


earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -337,6 +418,16 @@ khxml::DOMLSParser*
 CreateDOMParser(void) throw()
 {
   InitializeXMLLibrary();
+  bool flag = false;
```
tst-lsavoie: This flag is never set to true.

tst-lsavoie:
I think this is getting too clever. I think the general idea is right: calling terminate on the release of documents and parsers. However, this code is too complicated for me to verify that it works as expected. Here are some of my concerns:

- I can't be sure that if multiple threads run this code at the same time that it will be correct. For one, `objCount` is read and written by multiple threads at the same time, so it should be declared `volatile`. We can fix that, but I don't know what other multi-threading bugs I've missed.
- We're doing unnecessary work. The Xerces spec says that it's fine to call initialize and terminate more than once, as long as you call them the same number of times. We don't need to track how many objects have called initialize; we can let Xerces do that and just call terminate every time an object is destroyed.

Given the above, I suggest the following:
- Call Initialize at the beginning of CreateEmptyDocument and CreateDOMParser
- Call Terminate at the end of DestroyDocument and DestroyParser
- Protect those 4 calls with a mutex so that they aren't called concurrently
- Remove all calls to InitializeXMLLibrary (or refactor it so that it calls Initialize every time and never calls terminate)

I think if we do that we will accomplish the same thing as the changes on this PR in a simpler way.

tst-lsavoie:
I tried a simple version of my changes and I got a segfault, so my suggestion is too simplistic. I have to believe there is a simpler way to accomplish this, though.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -76,6 +84,73 @@ class SizeError : public XmlParamsException
...
+    static void lock()
+    {
+        khLockGuard guard(lockMutex);
+        objLock = true;
+    }
+
+    static void unlock()
+    {
+        khLockGuard guard(lockMutex);
+        objLock = true;
```
tst-lsavoie: This should be false, correct?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -182,6 +313,12 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
                                             0);// document type object (DTD)
     return doc;
    } catch (...) {
+     if (terminateCache)
+     {
+       xmlLibLock.Unlock();
```
tst-lsavoie: I'm not sure if this is safe here - you may be calling unlock without calling lock

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -433,17 +601,42 @@ DestroyDocument(khxml::DOMDocument *doc) throw()
     retval = true;
   } catch (...) {
   }
+  if (terminateCache)
+  {
+        khLockGuard guard(checkTermLock);
+        terminateGuard::instance().removeObj();
+        if (terminateGuard::instance().isLocked() &&
+            !terminateGuard::instance().size())
```
tst-lsavoie: Minor: I prefer `terminateGuard::instance().size() == 0` - I think it's a bit clearer.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -433,17 +601,42 @@ DestroyDocument(khxml::DOMDocument *doc) throw()
     retval = true;
   } catch (...) {
   }
+  if (terminateCache)
+  {
+        khLockGuard guard(checkTermLock);
+        terminateGuard::instance().removeObj();
+        if (terminateGuard::instance().isLocked() &&
```
tst-lsavoie: Should this have a "not" in front of it - that is, do we want the instance to be locked or unlocked? Perhaps I don't understand the purpose of terminateGuard::instance().lock/unlock/isLocked.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -27,8 +27,12 @@
 #include <array>
 #include <algorithm>
 #include <exception>
+#include <queue>
```
tst-lsavoie: Minor: I don't think this include is used any more.

tst-lsavoie:
This is the one that I didn't see terminate being called.

docs/geedocs/5.2.5/answer/176738.html
```diff
@@ -273,6 +273,10 @@ <h2 id="cache_configuration">Cache configuration</h2>
 <br><br>
 # the size of the allocation blocks<br>
 BLOCK_SIZE=&lt;value&gt;
+<br><br>
+# whether or not purge the xml cache
+# false = 0, true = 1
+PURGE = &lt;value&gt; 
 </code>
```
tst-lsavoie: The PERCENT option is missing here.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -161,18 +220,38 @@ class UsingXMLGuard
...
+static khMutexBase checkTermLock = KH_MUTEX_BASE_INITIALIZER;
+static khMutexBase cacheLock = KH_MUTEX_BASE_INITIALIZER;
+
+// only terminate when certain conditions are met
+void  readyForTerm()
+{
+  khLockGuard guard(checkTermLock);
+  static XMLSSize_t threashold = static_cast<XMLSSize_t>
+         (maxDOMHeapAllocSize * percent); 
+  if (cacheCapacity >= threashold)
```
tst-lsavoie: Minor: there's no 'a' in threshold.

tst-lsavoie: Thinking out loud: it might make things simpler to remove the threshold code and simply terminate the cache every time there's an opportunity to do so. That would remove some of the code and the need to stat the file. On the other hand, it's nice to have a way to tune the terminate behavior.

docs/geedocs/5.3.0/answer/176738.html
```diff
@@ -257,6 +257,29 @@ <h2 id="cache_configuration">Cache configuration</h2>
 Note that the asset version cache is generally larger than the asset cache because it may contain multiple versions of the same asset.</p>
 
 Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
+<br><br>
```
tst-lsavoie: Good catch.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -161,18 +220,38 @@ class UsingXMLGuard
 };
 
 static khMutexBase xmlLibLock = KH_MUTEX_BASE_INITIALIZER;
+static khMutexBase checkTermLock = KH_MUTEX_BASE_INITIALIZER;
+static khMutexBase cacheLock = KH_MUTEX_BASE_INITIALIZER;
+
+// only terminate when certain conditions are met
+void  readyForTerm()
```
tst-lsavoie: Minor: The name of this function makes it sound like it doesn't actually do the terminate. Maybe change to something like `reinitIfReady`?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,55 +125,97 @@ void setDefaultValues()
...
+    catch (const XMLException& toCatch)
+    {
+        notify(NFY_FATAL, "Unable to initialize Xerces: %s",
+               FromXMLStr(toCatch.getMessage()).c_str());
+    }
+}
+
+void ReInitializeXerces()
+{
+    XMLPlatformUtils::Terminate();
```
tst-lsavoie: Shouldn't we be holding a lock during this function?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -161,18 +227,38 @@ class UsingXMLGuard
 };
 
 static khMutexBase xmlLibLock = KH_MUTEX_BASE_INITIALIZER;
```
tst-lsavoie: Minor: for simplicity you might consider using a single mutex instead of the three you have here. Splitting them up like this probably performs better, but I'm not sure the difference is enough to outweigh the simplification of only having one.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,55 +125,97 @@ void setDefaultValues()
...
+        for (const auto& it : config_parser)
+        {
+            if (it.first == INIT_HEAP_SIZE)
+                initialDOMHeapAllocSize = std::stol(it.second);
+            else if (it.first == MAX_HEAP_SIZE)
+                maxDOMHeapAllocSize = std::stol(it.second);
+            else if (it.first == BLOCK_SIZE)
+                maxDOMSubAllocationSize = std::stol(it.second);
+            else if (it.first == PURGE)
+                terminateCache = std::stol(it.second);
```
tst-jhaight: Minor: converting string to long and then implicitly to bool, maybe assign terminateCache = (std::stol(it.second) != 0)

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -284,7 +377,7 @@ bool
 WriteDocumentToString(DOMDocument *doc, std::string &buf) throw()
 {
   bool success = false;
-  InitializeXMLLibrary();
+  //InitializeXMLLibrary();
```
tst-jhaight: I'm not familiar at all with how and when xml is used but it looks possible for this to be called before any other functions which initialize the library.  I'm also assuming `getDOMImplementation` is a Xerces function.  So could this cause a call to an uninitialized library?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -191,7 +278,7 @@ bool
 WriteDocumentImpl(DOMDocument *doc, const std::string &filename) throw()
 {
 
-  InitializeXMLLibrary();
+  //InitializeXMLLibrary();
```
tst-jhaight: Is it guaranteed that the XML library will be initialized before this function is called?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -85,14 +104,19 @@ void validateXMLParameters()
...
     }
 
     // check to make sure that the initial size is less than the max size
     if (maxDOMHeapAllocSize < initialDOMHeapAllocSize ||
         maxDOMHeapAllocSize < maxDOMSubAllocationSize)
     {
         throw SizeError();
+    }
+
+    if (percent > 5 || percent < 1)
```
tst-jhaight: nitpick: comparing float against ints and magic numbers.  Could make constants and then re-use in PercentError

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -182,18 +342,23 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
...
-  InitializeXMLLibrary();
+  
+  if (terminateCache) 
+  {
+    while(purgeGuard::instance().isLocked());
+  }
+  purgeGuard::instance().increment();
   bool success = false;
-
+  notify(NFY_INFO2, "starting WriteDocumentImpl()");
```
tst-lsavoie: Need to get rid of this debugging statement.

tst-lsavoie:
I'm going to continue testing but everything seems to be working now. Two points:

- This PR could really use some comments to explain the thinking behind the changes.
- I suspect there are ways this code could be cleaned up and simplified further. The commenting effort may reveal some things, and if I find anything I'll make specific comments.

docs/geedocs/5.2.5/answer/176738.html
tst-lsavoie: Bump - still a missing option in the documentation. It was added to the 5.3.0 docs but not 5.2.5.

docs/geedocs/5.3.0/answer/176738.html
```diff
@@ -257,6 +257,29 @@ <h2 id="cache_configuration">Cache configuration</h2>
...
+MAX_HEAP_SIZE=&lt;value&gt;
+<br><br>
+# the size of the allocation blocks<br>
+BLOCK_SIZE=&lt;value&gt;
+<br><br>
+# whether or not to purge: 0 or 1, default is 0
+PURGE=0
+<br><br>
+# the level at which to purge, may be 1 (most often) to 5 (least often)
+PURGE_LEVEL=100
```
tst-lsavoie: This is no longer correct - the purge level can't be 100.

tst-lsavoie: Would be good to note the default here as well.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,195 @@ void setDefaultValues()
...
+        {
+            if (it.first == INIT_HEAP_SIZE)
+                initialDOMHeapAllocSize = std::stol(it.second);
+            else if (it.first == MAX_HEAP_SIZE)
+                maxDOMHeapAllocSize = std::stol(it.second);
+            else if (it.first == BLOCK_SIZE)
+                maxDOMSubAllocationSize = std::stol(it.second);
+            else if (it.first == PURGE)
+                terminateCache = std::stoi(it.second);
+            else
```
tst-lsavoie: I agree with @tst-jhaight here - we should check explicitly for PURGE_LEVEL and then have an else that prints out an error for invalid options.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,195 @@ void setDefaultValues()
...
+                    case  4: percent = 2.25; break;
+                    case  5: percent = 3.00; break;
+                    default: percent = 1.50;
+                };
+            }
+        }
+    }
+    catch (const khConfigFileParserException& e)
+    {
+        notify(NFY_DEBUG, "%s , using default xerces init values", e.what());
```
tst-lsavoie: Do you need to call `setDefaultValues()` here?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,195 @@ void setDefaultValues()
...
+    static uint32_t getCount()
+    {
+        khLockGuard guard(pLock);
+        return count;
+    }
+
+    static void lock()
+    {
+        khLockGuard guard(pLock);
+        _lock = true;
```
tst-lsavoie: What is the purpose of this `_lock` variable?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,195 @@ void setDefaultValues()
...
+            notify(NFY_DEBUG, "Failure to terminate in ~UsingXMLGuard");
+        }
     }
   }
 };
 
-static khMutexBase xmlLibLock = KH_MUTEX_BASE_INITIALIZER;
+void  reInitIfReady()
+{
+  khLockGuard guard(checkTermLock);
```
tst-lsavoie: I think this lock can be replaced with cacheLock. You are checking the same variable that is protected by cacheLock, so it doesn't make sense to have a separate lock.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,195 @@ void setDefaultValues()
...
+  static XMLSSize_t threashold = static_cast<XMLSSize_t>
+         (maxDOMHeapAllocSize * percent); 
+  if (cacheCapacity >= threashold)
+  {
+    purgeGuard::instance().lock();
+    while(purgeGuard::instance().getCount() != 0);
```
tst-lsavoie: Good place for a comment - something like "Wait for ongoing writes to complete"

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,195 @@ void setDefaultValues()
...
-                                   maxDOMSubAllocationSize);
-      notify(NFY_DEBUG, "XML initialization values: %s=%zu %s=%zu %s=%zu",
-             "initialDOMHeapAllocSize", initialDOMHeapAllocSize,
-             "maxDOMHeapAllocSize", maxDOMHeapAllocSize,
-             "maxDOMSubAllocationSize", maxDOMSubAllocationSize);
-    } catch(const XMLException& toCatch) {
-      notify(NFY_FATAL, "Unable to initialize Xerces: %s",
-             FromXMLStr(toCatch.getMessage()).c_str());
-    }
+    initXercesValues();
```
tst-lsavoie: For complete safety, this and the terminate call in the destructor should be called while the reinitLock is held so that we're absolutely certain multiple Initializes and Terminates can't happen at the same time.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -182,18 +341,22 @@ CreateEmptyDocument(const std::string &rootTagname) throw()
...
 WriteDocumentImpl(DOMDocument *doc, const std::string &filename) throw()
 {
-
-  InitializeXMLLibrary();
+  
+  if (terminateCache) 
+  {
+    while(purgeGuard::instance().isLocked());
+  }
+  purgeGuard::instance().increment();
```
tst-lsavoie: There is a race condition here. What happens if the purgeGuard is locked again before write calls increment?

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,195 @@ void setDefaultValues()
...
+               FromXMLStr(toCatch.getMessage()).c_str());
+    }
+}
+
+static khMutexBase xmlLibLock = KH_MUTEX_BASE_INITIALIZER;
+static khMutexBase checkTermLock = KH_MUTEX_BASE_INITIALIZER;
+static khMutexBase cacheLock = KH_MUTEX_BASE_INITIALIZER;
+static khMutexBase reinitLock = KH_MUTEX_BASE_INITIALIZER;
+
+class purgeGuard
```
tst-lsavoie: I think what you're trying to implement is a read/write lock, except that you have a write/purge lock. Writes can happen at the same time, but purges can't happen during writes or other purges. It would be safer if you used a standard read/write lock and did a write lock for purges and a read lock for writes (I know that's paradoxical, but it's the right choice in this situation).

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +125,211 @@ void setDefaultValues()
...
+                maxDOMHeapAllocSize = std::stol(it.second);
+            else if (it.first == BLOCK_SIZE)
+                maxDOMSubAllocationSize = std::stol(it.second);
+            else if (it.first == PURGE)
+                terminateCache = std::stoi(it.second);
+            else
+            {
+                switch (std::stoi(it.second))
+                {
+                    case  1: percent = 0.00; break;
```
tst-lsavoie: I actually preferred letting the user pick the percent rather than giving them a distinct set of levels. It gives them more options. That's more of a personal preference, though.

tst-lsavoie:
Overall this looks good and my tests are passing. I have a few more minor comments - rather than make the comments it was easier to make a commit with my suggested changes. You can see it here: https://github.com/tst-lsavoie/earthenterprise/commit/2a2ea86001cd85da69fa76f58e00f0fd16e4abf4. I've still got a bit more testing to do, but assuming my tests pass and you integrate my last few comments I can approve this.

docs/geedocs/5.2.5/answer/176738.html
```diff
@@ -273,6 +273,14 @@ <h2 id="cache_configuration">Cache configuration</h2>
 <br><br>
 # the size of the allocation blocks<br>
 BLOCK_SIZE=&lt;value&gt;
+<br><br>
```
tst-lsavoie: It just occurred to me that you don't need these changes here. The PURGE and PURGE_LEVEL options weren't available in 5.2.5.

earth_enterprise/src/common/khxml/khxml.cpp
```diff
@@ -101,66 +126,191 @@ void setDefaultValues()
...
-             FromXMLStr(toCatch.getMessage()).c_str());
-    }
+    initXercesValues();
   }
 
   ~UsingXMLGuard(void) throw() {
-    try {
-      XMLPlatformUtils::Terminate();
-    } catch (...) {
+    if (!terminateCache)
```
tst-lsavoie: Minor: I don't know that this check is necessary. We can call terminate when the program shuts down regardless of whether we call terminate at other times as well.

tst-lsavoie:
I'm ok with my commits being included on this PR.

earth_enterprise/src/common/khxml/khxml.cpp
tst-lsavoie: A DOMDocument is passed to this function and you can't get one of those without initializing Xerces, so this is safe.

Issue #949, opened 2018-08-08
Blog spacing is not correct


tst-ccamp:
Example: http://www.opengee.org/2018/07/18/523-release.html

It was discovered that the spacing issue is due to a missing profile image in some of the blog posts. To fix the issue, it's been decided to create a logo for the Google Earth Enterprise Software Engineers and to update the blogs which have the issue.

PR #1181, opened 2019-01-24
949 blog post spacing issue

tst-nfarah:
This address #949

For verification: https://tst-nfarah.github.io/earthenterprise/blog.html

and click on blog posts linked on the right side panel.


tst-nfarah:
> So what became of the logo mentioned in #949 ?

Good question. @tst-ccamp do we want a logo there or this is fine the way it is fixed ?

PR #1182, opened 2019-01-24
1154 add overlay cli

### LINKED ISSUE #1154: Add overlay option for genewterrainproject and gemodifyterrainproject
Currently there is no way to change the "overlay" option for a terrain project using the CLI tools. In the Fusion UI, this can be toggled via the 'Build "Overlay"' checkbox in the terrain project dialog box.

Add an optional parameter to genewterrainproject and gemodifyterrainproject to allow the user to set this flag via the command line.

### AC
- Calling genewterrainproject with the optional overlay parameter produces a terrain project with the 'Build "Overlay"' flag enabled
- Calling gemodifyterrainproject with the overlay parameter value specified enables and disables the flag as appropriate
- Release notes are updated to describe the enhanced functionality

tst-sbayard:
Fixes #1154 
Adds 4 new options to `genewterrainproject` and `gemodifyterrainproject`. The new options are as follows:

- `--terrain_overlay` (enables the "terrain overlay" flag for a terrain project)
- `--no_terrain_overlay` (disables the "terrain overlay" flag for a terrain project)
- `--start_level` (even integer between 4 and 24 inclusive)
- `--resource_min_level` (integer between 4 and 24 inclusive)

For verification, call `genewterrainproject` and `gemodifyterrainproject` with different combinations of these options (including none of them), and verify that the values appear as expected when viewing the project in question in the Fusion UI.

Verify `geaddtoterrainproject` supports the newly added options like `genewterrainproject` does.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
@@ -641,6 +641,24 @@ void RasterProjectAssetImplD::HandleAddToRequest(
     project->config.is_timemachine_project_ = is_timemachine_project;
     notify(NFY_DEBUG, "TimeMachine: %s for %s\n", req.assetname.c_str(),
            is_timemachine_project ? "on" : "off");
+
+    // By default keep is_overlay_terrain_project_ the same unless directed
+    // otherwise.
+    bool is_overlay_terrain_project = project->config.is_overlay_terrain_project_;
+    if (req.enable_terrain_overlay){
+      is_overlay_terrain_project = true;
+    }
+    else if (req.disable_terrain_overlay){
+      is_overlay_terrain_project = false;
+    }
+    if (is_overlay_terrain_project) {
+      // Set start level and/or resources min level if specified
+      if (req.overlay_terrain_start_level != 0)
+        project->config.overlay_terrain_start_level_ = req.overlay_terrain_start_level;
+      if (req.overlay_terrain_resources_min_level != 0)
+        project->config.overlay_terrain_resources_min_level_ = req.overlay_terrain_resources_min_level;
+    }
+    project->config.is_overlay_terrain_project_ = is_overlay_terrain_project;
   } else {
     throw khException(kh::tr("No %1 project named %2")
                       .arg(ToQString(req.type), ToQString(req.assetname)));
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
@@ -544,6 +544,8 @@ void RasterProjectAssetImplD::HandleNewRequest(
     notify(NFY_DEBUG, "TimeMachine: %s for %s\n", req.assetname.c_str(),
            is_timemachine_project ? "on" : "off");
     project->config.is_overlay_terrain_project_ = req.enable_terrain_overlay;
+    project->config.overlay_terrain_start_level_ = req.overlay_terrain_start_level;
+    project->config.overlay_terrain_resources_min_level_ = req.overlay_terrain_resources_min_level;
   }
 }
 
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
@@ -584,20 +584,22 @@ void RasterProjectAssetImplD::HandleModifyRequest(
     project->config.is_timemachine_project_ = is_timemachine_project;
     notify(NFY_DEBUG, "TimeMachine: %s for %s\n", req.assetname.c_str(),
            is_timemachine_project ? "on" : "off");
-           
+
     // By default keep is_overlay_terrain_project_ the same unless directed
     // otherwise.
     bool is_overlay_terrain_project = project->config.is_overlay_terrain_project_;
     if (req.enable_terrain_overlay){
       is_overlay_terrain_project = true;
+    }
+    else if (req.disable_terrain_overlay){
+      is_overlay_terrain_project = false;
+    }
+    if (is_overlay_terrain_project) {
       // Set start level and/or resources min level if specified
       if (req.overlay_terrain_start_level != 0)
         project->config.overlay_terrain_start_level_ = req.overlay_terrain_start_level;
       if (req.overlay_terrain_resources_min_level != 0)
         project->config.overlay_terrain_resources_min_level_ = req.overlay_terrain_resources_min_level;
-    }
-    else if (req.disable_terrain_overlay){
-      is_overlay_terrain_project = false;
     }
     project->config.is_overlay_terrain_project_ = is_overlay_terrain_project;
   } else {
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
@@ -543,9 +543,14 @@ void RasterProjectAssetImplD::HandleNewRequest(
     project->config.is_timemachine_project_ = is_timemachine_project;
     notify(NFY_DEBUG, "TimeMachine: %s for %s\n", req.assetname.c_str(),
            is_timemachine_project ? "on" : "off");
+
     project->config.is_overlay_terrain_project_ = req.enable_terrain_overlay;
-    project->config.overlay_terrain_start_level_ = req.overlay_terrain_start_level;
-    project->config.overlay_terrain_resources_min_level_ = req.overlay_terrain_resources_min_level;
+    if (req.enable_terrain_overlay){
+      if (req.overlay_terrain_start_level != 0)
+        project->config.overlay_terrain_start_level_ = req.overlay_terrain_start_level;
+      if (req.overlay_terrain_resources_min_level != 0)
+        project->config.overlay_terrain_resources_min_level_ = req.overlay_terrain_resources_min_level;
+    }
   }
 }
 
@@ -579,6 +584,22 @@ void RasterProjectAssetImplD::HandleModifyRequest(
     project->config.is_timemachine_project_ = is_timemachine_project;
     notify(NFY_DEBUG, "TimeMachine: %s for %s\n", req.assetname.c_str(),
            is_timemachine_project ? "on" : "off");
+           
+    // By default keep is_overlay_terrain_project_ the same unless directed
+    // otherwise.
+    bool is_overlay_terrain_project = project->config.is_overlay_terrain_project_;
+    if (req.enable_terrain_overlay){
+      is_overlay_terrain_project = true;
+      // Set start level and/or resources min level if specified
+      if (req.overlay_terrain_start_level != 0)
+        project->config.overlay_terrain_start_level_ = req.overlay_terrain_start_level;
+      if (req.overlay_terrain_resources_min_level != 0)
+        project->config.overlay_terrain_resources_min_level_ = req.overlay_terrain_resources_min_level;
+    }
+    else if (req.disable_terrain_overlay){
+      is_overlay_terrain_project = false;
+    }
+    project->config.is_overlay_terrain_project_ = is_overlay_terrain_project;
   } else {
     throw khException(kh::tr("No %1 project named %2")
                       .arg(ToQString(req.type), ToQString(req.assetname)));
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
@@ -539,9 +539,11 @@ void RasterProjectAssetImplD::HandleNewRequest(
       project->ThrowIfAnyInvalidDates();
       is_timemachine_project = true;
     }
+
     project->config.is_timemachine_project_ = is_timemachine_project;
     notify(NFY_DEBUG, "TimeMachine: %s for %s\n", req.assetname.c_str(),
            is_timemachine_project ? "on" : "off");
+    project->config.is_overlay_terrain_project_ = req.enable_terrain_overlay;
   }
 }
 
--- a/earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
+++ b/earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
@@ -139,9 +139,10 @@ main(int argc, char *argv[]) {
       usage(argv[0], "--historical_imagery is not a valid option for mercator "
           "imagery projects.");
     }
-    if ((enable_terrain_overlay || disable_terrain_overlay) && AssetDefs::Terrain != AssetType){
-      usage(argv[0], "--terrain_overlay and --no_terrain_overlay are not valid options for "
-          "imagery projects.");
+    if ((enable_terrain_overlay || disable_terrain_overlay || start_level != 0 ||
+         resource_min_level != 0) && AssetDefs::Terrain != AssetType){
+      usage(argv[0], "--terrain_overlay, --no_terrain_overlay, --start_level, and "
+          "--resource_min_level are only valid for terrain projects.");
     }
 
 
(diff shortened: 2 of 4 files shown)


earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -47,7 +47,13 @@ usage(const std::string &progn, const char *msg = 0, ...)
           "      --no_historical_imagery : make this a normal project\n"
           "                         (i.e., not a historical imagery project)\n"
           "   By default, new projects are NOT time machine projects"
-          " unless --historical_imagery is specified.\n",
+          " unless --historical_imagery is specified.\n"
+          "      --terrain_overlay : make this terrain project an overlay project\n"
+          "      --no_terrain_overlay : make this terrain project a normal project\n"
```
tst-nfarah: we may not need to add `--no_terrain-overlay` but document it as this is the default behavior. Two reasons:
1- Backward compatibility of any tool or scripts using this CLI: no need for them to change how they call the command
2- Simpler: if one need overlay terrain they add it. Otherwise nothing need to be specified.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -575,6 +584,24 @@ void RasterProjectAssetImplD::HandleModifyRequest(
...
+    bool is_overlay_terrain_project = project->config.is_overlay_terrain_project_;
+    if (req.enable_terrain_overlay){
+      is_overlay_terrain_project = true;
+    }
+    else if (req.disable_terrain_overlay){
+      is_overlay_terrain_project = false;
+    }
+    if (is_overlay_terrain_project) {
+      // Set start level and/or resources min level if specified
+      if (req.overlay_terrain_start_level != 0)
```
tst-nfarah: we probably need to do a little more validation for start level and resource level:
- start level can't be negative and may exceed 24 or so
- same for resource level
- any relationship between start level and resource level ? if any does it need to be enforced ?

tst-nfarah:
> > let's add this new capability to release notes of 5.3.0
> 
> It sounds like the additions in the release_5.2.4.1 branch will eventually be merged into release_5.3.0 branch (or perhaps a subsequent release branch). Should we hold off updating the release notes until that time?

let's create an issue for this in Github so we don't forget documenting this when we want to merge 5.2.4.1 branch into 5.3.0.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
tst-nfarah: I now see that there is validation done in ```options.opt("start_level", start_level, &khGetopt::IsEvenNumberInRange<uint, 4, 24>);``` Is that only place where arguments validation is needed ?

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -88,6 +99,11 @@ main(int argc, char *argv[]) {
     options.opt("maxlevel", overridemax);
     options.opt("historical_imagery", enable_historical_imagery);
     options.opt("no_historical_imagery", disable_historical_imagery);
+    options.opt("terrain_overlay", enable_terrain_overlay);
+    options.opt("no_terrain_overlay", disable_terrain_overlay);
+    options.opt("start_level", start_level, &khGetopt::IsEvenNumberInRange<uint, 4, 24>);
```
tst-nfarah: instead of 4 and 24 magic numbers, let's add const with descriptive names.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -151,21 +173,39 @@ main(int argc, char *argv[]) {
...
-            !(enable_historical_imagery || disable_historical_imagery)) {
-        // For adding/modifying, we always need insets except:
-        // non-mercator when historical_imagery/no_historical_imagery is
-        // specified.
+        !( (AssetDefs::Imagery == AssetType && (enable_historical_imagery || disable_historical_imagery)) ||
+           (AssetDefs::Terrain == AssetType && 
+              (enable_terrain_overlay || disable_terrain_overlay || 
+              start_level != 0 || resource_min_level != 0))
+        )){
+            // !(AssetDefs::Imagery == AssetType && (enable_historical_imagery || disable_historical_imagery)) ||
```
tst-nfarah: Is this is a dead code ? If yes let's remove it

tst-nfarah: Executing `genewimageryproject --help` output information about terrain overlay. I think even if imagery and terrain are using common code we probably should not output specific terrain info when dealing with imagery.
Here is the specific terrain overlay info i get:
--> ```--terrain_overlay : make this terrain project an overlay project
      --no_terrain_overlay : make this terrain project a normal project
      --start_level : the level from which to start building the terrain overlay project
      --resource_min_level : the threshold level that separates fill terrain from overlay terrain
   By default, new terrain projects are NOT overlay projects unless --terrain_overlay is specified.```

I get the same behavior when i run `gemodifyimageryproject --help`

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-mbiancalana: On the other hand, this mirrors `--historical-imagery`/`--no-historical-imagery` that already exist.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -151,21 +173,39 @@ main(int argc, char *argv[]) {
...
+            // !(AssetDefs::Terrain == AssetType && 
+            //   (enable_terrain_overlay || disable_terrain_overlay || 
+            //   start_level != 0 || resource_min_level != 0))) {
```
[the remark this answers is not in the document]
tst-mbiancalana: I see that @tst-nfarah commented on this already. Kindly ignore this.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -129,6 +145,12 @@ main(int argc, char *argv[]) {
       usage(argv[0], "--historical_imagery is not a valid option for mercator "
           "imagery projects.");
     }
+    if ((enable_terrain_overlay || disable_terrain_overlay || start_level != 0 ||
```
tst-nfarah: if imagery project we may need to look for those arguments as they were not valid to be used and hence we would not need this code ?

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-sbayard: Will do.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-sbayard: Yes, that's dead code. I'll remove it.

tst-sbayard: I wrestled with this a bit while writing the code. On the one hand, the original developers set a precedent of including the same usage text for all commands that are handled by geaddtoimageryproject.cpp. That means that even `ge***terrainproject` would list `--historical_imagery` and `--no_historical_imagery` in their `--help` text.

I thought about adding logic to only display appropriate usage text but decided to leave it as is for now. However, that's probably not hard to change. I'll take a stab at separating the imagery and terrain project usage text.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-sbayard: I'm not sure I understand your concern here. This is the code that is checking to make sure those options aren't being passed in when using one of the imagery commands.

tst-nfarah: we don't need to validate those arguments if they were not to be used. We can just ignore them. It's the same situation if some one pass some arguments that are not part of the command, we can't know what those are, unless if we want to exclude any extra argument and in that case we can do it for any thing not expected and not just the overlay terrain.

tst-sbayard: Okay, I see what you're saying. There's already code later in the function that prevents the terrain options from being used for a non-terrain project, so I guess there's no real need to have this here.
Especially since the `--help` text has been updated to only show applicable options.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
tst-sbayard: I think that is sufficient. That guarantees that the user can't pass in anything outside that range. Additionally, if the user doesn't specify a value, it will have the default value that was assigned earlier in the function. Also, in the `RasterProjectAssetImplD::HandleXXXXRequest` methods, the values for the different options are only used if the user specified them.

tst-sbayard:
> > > let's add this new capability to release notes of 5.3.0
> > 
> > 
> > It sounds like the additions in the release_5.2.4.1 branch will eventually be merged into release_5.3.0 branch (or perhaps a subsequent release branch). Should we hold off updating the release notes until that time?
> 
> let's create an issue for this in Github so we don't forget documenting this when we want to merge 5.2.4.1 branch into 5.3.0.

Should we make the issue just for this feature or for any/all contained in the 5.2.4.1 branch?

tst-sbayard:
> I would create an issue with generic text and list this new enhancement as an example of changes to be documented. That way if we end up doing more in 5.2.4.1 we can documented them as well.
Created issue #1184.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -35,20 +35,40 @@ usage(const std::string &progn, const char *msg = 0, ...)
...
-          "      --mercator : use mercator projection\n"
-          "      --flat : use flat(Plate Carre) projection (default)\n"
-          "      --historical_imagery : change this project to be a historical\n"
-          "                             imagery project\n"
-          "      --no_historical_imagery : make this a normal project\n"
-          "                         (i.e., not a historical imagery project)\n"
-          "   By default, new projects are NOT time machine projects"
-          " unless --historical_imagery is specified.\n",
-          progn.c_str());
+  if (AssetDefs::Terrain == AssetType){
```
tst-nfarah: I created a new terrain project using the following command `genewterrainproject -o terrainprojecttest --terrain_overlay`: no error or warning were posted about missing start level and min level resource parameters and values. When i opened that project in Fusion "Build Overlay" checkbox was not enabled. If the command did not error or post a warning, i was excepting that overlay will be used with default parameters. If not we should probably post an error or warning.

tst-nfarah: I created a new terrain project using `genewterrainproject -o terrainprojecttest2 --terrain_overlay --start_level 4 --resource_min_level 4`: no error but when i opened the terrain project in fusion "Build Overlay" checkbox was not checked and the start and min levels values were the default ones: 4 and 12. Is this is expected ?

tst-sbayard: @tst-nfarah I'm not sure what's happening there. I copied that command straight from your comment and ran it on my development machine, and when I look at the project in Fusion, the 'Build "Overlay"' checkbox is enabled, and "Start Level" and "Resources Min Level" have the default values. I'll do a pull and rebuild, then try it again to see if I see the same thing.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -35,20 +35,40 @@ usage(const std::string &progn, const char *msg = 0, ...)
...
+    fprintf(stderr,
+            "\nusage: %s [options] -o <projectname> {[--maxlevel <level>] "
```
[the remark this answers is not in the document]
tst-sbayard: Working to update the `--help` text to be accurate for each command. Changes incoming.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-sbayard: @tst-nfarah I did a complete rebuild with the latest commit, and both of these commands work for me as expected. Let me know if you're seeing something different.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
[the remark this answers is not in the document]
tst-sbayard: Each of the commands should now have help text appropriate to their functions.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-nfarah: @tst-sbayard I rebuilt then i run `genewterrainproject -o terrainproject9 --terrain` and looked at the project in Fusion UI. It did have `Build "Overlay"` checkbox enabled ! Are we not looking for the full `terrain_overlay` and only partially to decide if overlay to be on/off ?

tst-sbayard: I'm seeing the same behavior. I think this may be a bug in the option parsing code. On a hunch, I executed this command `genewimageryproject -o imageryproject --hist` and sure enough, it created an imagery project with "Support Historical Imagery" enabled. I'm going to try the same thing on a 5.2.4 release build to verify that this was present in previous versions. If it already existed, I can take a quick look, but that may need to be filed as a separate bug.

tst-sbayard:
> Built the branch and installed but Fusion UI shows that it is version 5.2.5 instead of 5.2.4
Hmm... When I click Help->About in the Fusion UI, it shows 5.2.1-4 beta, which is also incorrect, but not the same as what you're seeing. Where do you see the 5.2.5?

fusion_5241

tst-nfarah:
> > Built the branch and installed but Fusion UI shows that it is version 5.2.5 instead of 5.2.4
> > Hmm... When I click Help->About in the Fusion UI, it shows 5.2.1-4 beta, which is also incorrect, but not the same as what you're seeing. Where do you see the 5.2.5?
> 
> fusion_5241

I see it in the same place as the screen shot you shared above.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-sbayard: After further research, this seems to be by design. The option parsing code calls `getopt_long_only` which is a standard library function. The documentation states that 

> Long-option names can be abbreviated if the abbreviation is unique or an exact match for some defined option

To test, I executed `grep -r getopt_long_only .` in the opengee repository. Then I ran `grep --recurs getopt_long_only .`, using `--recurse` instead of `--recursive`. That produced the exact same results as the first call. 

I had no idea, but apparently that's just a standard feature.

tst-sbayard:
> I see it in the same place as the screen shot you shared above.

Is there possibly any cruft left over from a previous Fusion installation? I have it installed on a fresh Ubuntu 16 vm.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
tst-nfarah: After installing the new built branch, i did not have this issue any more. I had previous Fusion installed and it appears that using the commands built with this branch is not enough to get to a clean state. This is now resolved.

tst-nfarah: @tst-sbayard it looks like this was intended to work that way. I guess the idea is to accommodate with close enough strings.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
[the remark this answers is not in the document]
tst-nfarah: @tst-sbayard i verified the help output and it looks good.

earth_enterprise/src/fusion/autoingest/tools/geaddtoimageryproject.cpp
```diff
@@ -21,6 +21,54 @@
...
+            "   By default, new terrain projects are NOT overlay projects"
+            " unless --terrain_overlay is specified.\n";
+
+const char * MODIFY_TERRAIN_USAGE = "\nusage: %s [options] -o <projectname> {[--maxlevel <level>] "
+            "<insetresource>}...\n"
+            "   Supported options are:\n"
+            "      --help | -?:      Display this usage message\n"
+            "      --mercator : use mercator projection\n"
+            "      --flat : use flat(Plate Carre) projection (default)\n"
+            "      --terrain_overlay : make this terrain project an overlay project\n"
```
tst-nfarah: Let's swap the order of `terrain-overlay` and `no_terrain_overlay` so `no-terrain_overlay` is first and `terrain_overlay` is second to be close to the extra arguments that goes with it (`start_level` and `resource_min_level`).

tst-nfarah:
> > I see it in the same place as the screen shot you shared above.
> 
> Is there possibly any cruft left over from a previous Fusion installation? I have it installed on a fresh Ubuntu 16 vm.

@tst-sbayard was the version showed in the about fixed ?

Issue #127, opened 2017-05-03
Incorrect error messages from fusion installer


tst-lsavoie:
Some of the error messages printed by the fusion installer have incorrect information.  For example, if you install fusion and then try to install it again with a different username, you get an error like this:

```
It is strongly recommended that you use the default values for the fusion username and group.

Default Fusion User: 				gefusionuser
Default Fusion User Group: 			gegroup
----------------
Selected Fusion User: 				gefusionuser
Selected Fusion User Group: 			gegroup
X (Exit) the installer and change the asset root location - C (Continue) to use the asset root that you have specified.
```

There are two things to note.  First, the selected fusion user is the default user name, not the user name requested by the user.  Second, the last line refers to the asset root even though the error is about the user name.  This second error (referring to the asset root in error messages that don't involve the asset root) also appears in other error messages.

PR #1186, opened 2019-02-01
google#127: remove prompt on reinstall with -u

Calder-Ty:
Stops erroneous message from appearing when installing again
with `-u` flag. This flag is only valid for first install.
However the warning about not using the default username would
appear anytime you used it. This ensures that the message will
only show on first time installs.

Also update prompts for accuracy

see #127

### DIFF
--- a/docs/geedocs/5.3.0/answer/7160007.html
+++ b/docs/geedocs/5.3.0/answer/7160007.html
@@ -127,11 +127,6 @@ gtag('config', 'UA-108632131-2');
           <td>No current work around. The created backup can be deleted.</td>
         </tr>
         <tr>
-          <td>127</td>
-          <td>Incorrect error messages from Fusion installer</td>
-          <td>No current work around.</td>
-        </tr>
-        <tr>
           <td>190</td>
           <td>Hostname mismatch check in installers doesn't work as expected</td>
           <td>No current work around.</td>
--- a/docs/geedocs/5.3.0/answer/7160007.html
+++ b/docs/geedocs/5.3.0/answer/7160007.html
@@ -69,6 +69,13 @@ gtag('config', 'UA-108632131-2');
             <th>Resolution</th>
           </tr>
           <tr>
+            <td>127</td>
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -418,6 +418,9 @@ parse_arguments()
 				if [ $IS_NEWINSTALL == false ]; then
 					echo -e "\nYou cannot modify the fusion user name using the installer because Fusion is already installed on this server."
 					parse_arguments_retval=1
+					# Don't show the User Group dialog since it is invalid to change the fusion
+					# username once fusion is installed on the server
+					show_user_group_recommendation=false
```
tst-lsavoie: We should disable this a few lines down as well where the user has previously installed fusion and is trying to change the group.

tst-nfarah: I would update copyright notice here from `Copyright 2017 Google Inc.` to `Copyright 2017 Google Inc, 2019 Open GEE Contributors`

docs/geedocs/5.3.0/answer/7160007.html
```diff
@@ -68,6 +68,13 @@ <h5>
             <th class="narrow">Description</th>
             <th>Resolution</th>
           </tr>
+          <tr>
```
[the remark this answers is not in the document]
tst-nfarah: Let's fix the release notes page title as it indicate 5.2.5. To be changed to 5.3.0.
```<title>Release notes: Open GEE 5.2.5</title>``` --> ```<title>Release notes: Open GEE 5.3.0</title>```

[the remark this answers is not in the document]
Calder-Ty: Fixed in eeb6fb

earth_enterprise/src/installer/install_fusion.sh
Calder-Ty: bcd994d Updates the code for group.

PR #1188, opened 2019-02-01
Use POSIX file API to support large files on 32-bit Android

tst-jlarocco:
Address issue: #1190. The C library functions (fopen/ftell/fseek/...) that were used to read GLC files don't support files larger than 2 Gb on 32-bit Android.  This change uses the Posix API which does support large files on 32-bit Android.

Needs testing on Windows.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/portable_glc_reader.cpp
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/portable_glc_reader.cpp
@@ -105,7 +105,7 @@ bool PortableGlcReader::ReadData(
 
   try {
     lseek64(glc_file_, offset, SEEK_SET);
-    read(reinterpret_cast<char*>(buffer), size, 1, glc_file_);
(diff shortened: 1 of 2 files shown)


tst-blaird:
@tst-jlarocco Do defines need to be added for _FILE_OFFSET_BITS and _LARGEFILE64_SOURCE?  I found https://stackoverflow.com/questions/4772182/how-to-make-lseek64-actually-return-64-bit-offset when doing some googling.

tst-jlarocco:
I can't reproduce the behavior in the StackOverflow question.

Here's the test program I used:
https://gist.github.com/tst-jlarocco/d2cb74a82713211a26908b0a10c7da74

Adding #define _FILE_OFFSET_BITS 64 and #define _LARGEFILE64_SOURCE made a difference when I made it use lseek, but lseek64 returns the correct size with with our without them defined.

Might be able to use this same test program see what the behavior is on Windows.

tst-blaird:
Ok, thank you for checking on that.  The changes look good to me.  Approving.  I did not test as I think someone else is doing that.

PR #1198, opened 2019-02-15
Merged recent updates into patch 5.2.5

tst-sbayard:
Pulled in fixes and enhancements from the following to include in a 5.2.5 patch:

- state-update-optimization branch
- #1096 
- #1189 
- #1174

### DIFF
--- a/docs/geedocs/5.2.5/answer/176738.html
+++ b/docs/geedocs/5.2.5/answer/176738.html
@@ -257,23 +257,6 @@ The asset cache size is controlled by <code>AssetCacheSize</code> and the asset 
 Note that the asset version cache is generally larger than the asset cache because it may contain multiple versions of the same asset.</p>
 
 Another feature included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
-<br><br>
-In some instances, memory utilization may become problematic. An experimental feature is included which will allow for modifying aspects of the the allocated memory size.  This is controlled through <code>/etc/opt/google/XMLparams</code>. It takes the form of a standard configuration file, and it is in the following format.
-<br><br>
-<code>
-# comments are preceded by '#'
-<br>
-# options are case-insensitive, and all need not be present
-<br>
-# the initial size of the heap<br>
-INIT_HEAP_SIZE=&lt;value&gt;
-<br><br>
-# the max size of the heap<br>
-MAX_HEAP_SIZE=&lt;value&gt;
-<br><br>
-# the size of the allocation blocks<br>
-BLOCK_SIZE=&lt;value&gt;
-</code>
 <h2>Learn more</h2>
 <ul>
 <li><a href="../answer/6033487.html">Benchmarking outcomes for applying task rules</a></li>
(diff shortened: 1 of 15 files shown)


tst-lsavoie:
I think we're fine on the author consent question. All of these commits were already in the repo.

docs/geedocs/5.2.5/answer/176738.html
```diff
@@ -256,7 +256,24 @@ <h2 id="cache_configuration">Cache configuration</h2>
 The asset cache size is controlled by <code>AssetCacheSize</code> and the asset version cache size is controlled by <code>VersionCacheSize</code>.
 Note that the asset version cache is generally larger than the asset cache because it may contain multiple versions of the same asset.</p>
 
-Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
+Another feature included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
```
tst-lsavoie: Probably ought to remove these changes to the documentation. If these features weren't in the original 5.2.5 release we don't want them showing up in the 5.2.5 documentation when we merge this back to master.

tst-sbayard: > Probably ought to remove these changes to the documentation. If these features weren't in the original 5.2.5 release we don't want them showing up in the 5.2.5 documentation when we merge this back to master.

It looks like the cache purge disable made it into release 5.2.5 as an experimental feature. I think the change to that line is just for clarification. 

The following lines about XMLParams look like they should be moved to 5.3.0 documentation. However, it's already in the 5.2.5 docs in master, and there are no 5.3.0 docs in this branch. We may need to create a separate issue to move that to the 5.3.0 docs in master.

tst-lsavoie: Ok, sounds good.

tst-lsavoie: Did a bit of digging - looks like the part above is valid ("Another feature included within..."). However, the part below (starting "In some instances..." and going to the end of the changes) is only in 5.3.0. They're talking about two different caches.

tst-lsavoie: And we're going to have to do some documentation clean up, but that's a separate issue.

tst-lsavoie: I keep changing my mind here. I think it's best to back out all the changes in this file and leave the 5.2.5 docs exactly as they are in the 5.2.5 release. The sentence above doesn't really change the meaning of the docs, and the part below is irrelevant for 5.2.5. It's cleanest just to leave the docs as they are and handle any necessary doc updates on master.

tst-sbayard: I created issue #1200 to handle moving that section to the 5.3.0 documentation. I'll remove that section from this branch as well.

tst-sbayard: > I keep changing my mind here. I think it's best to back out all the changes in this file and leave the 5.2.5 docs exactly as they are in the 5.2.5 release. The sentence above doesn't really change the meaning of the docs, and the part below is irrelevant for 5.2.5. It's cleanest just to leave the docs as they are and handle any necessary doc updates on master.

It doesn't offend me too bad since it's just re-wording, but you're probably right. It'd be more consistent to keep the 5.2.5 docs as they were in the 5.2.5 release.

tst-sbayard: I've backed out the changes to that file.

PR #1199, opened 2019-02-17
Adding "v" parameter to portable server json requests; v=2 produces proper JSON. (#799)

### LINKED ISSUE #799: Queries for Map Definitions should return proper JSON in GEE Portable Server
When a GEE server is queried for Server Definitions the response is not valid JSON. It would be better if the returned values were valid JSON so that they can be read by parsers.

To see an example run a query against a server:
 `http://<server>/<map>/query?request=Json`

This is a follow-on for #622. Be sure to use the same parameter as used in the solution for that ticket.

ZetaZeta:
This change fixes issue #799 by adding a 'v' parameter to the GEE Portable Server's server definitions query; when it is set to 2, the result will be proper JSON.  Currently it defaults to version 1 (the previous behavior) if no version is specified.

Tested on the CentOS 7 dev environment by building the portable server and accessing tutorial_glm with both the old version and the current version, as well as with no version specified.  tutorial_glm should produce old-style output with no value or a v of 1 (starting with a variable definition, keys unquoted), and proper JSON with a v of 2.

Fixes #799

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/local_server.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/local_server.py
@@ -520,7 +520,7 @@ class LocalServer(object):
     json_text = "%s/%s%s" % (
       json_start, tornado.web.globe_.GlobeShortName(), json_end)
 
-    if json_version > 1:
+    if json_version == 2:
       json_text = self.JStoJson(json_text)
 
     handler.write(json_text)
--- a/earth_enterprise/src/fusion/portableglobe/servers/portable_server.py
+++ b/earth_enterprise/src/fusion/portableglobe/servers/portable_server.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #
-# Copyright 2017 Google Inc.
+# Copyright 2017 Google Inc, 2019 Open GEE Contributors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/docs/geedocs/5.3.0/answer/7160007.html
+++ b/docs/geedocs/5.3.0/answer/7160007.html
@@ -83,6 +83,13 @@ gtag('config', 'UA-108632131-2');
               Files are linked in <code>/tmp</code> using symbolic links back to original locations.
             </td>
           </tr>
+          <tr>
+            <td>799</td>
+            <td>The GEE Portable Server does not return valid JSON when queried with <code>http://&lt;server&gt;/&lt;map&gt;/query?request=Json</code>.</td>
+            <td>
+              Added a <code>v</code> parameter to the query; when its value is above 1, the GEE Portable Server will return proper JSON.  Currently, it defaults to 1 if not specified.
+            </td>
+          </tr> 
            <tr>
             <td>1158</td>
             <td><code>geeServerDefs</code> are not returned by Portable Server after being requested, once the globe has been viewed by a client. It is not possible to request <code>geeServerDefs</code> for globes other than the current <code>selectedGlobe</code>.</td>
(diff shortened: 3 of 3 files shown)


ZetaZeta:
I signed it!

earth_enterprise/src/fusion/portableglobe/servers/local_server.py
```diff
@@ -506,8 +517,13 @@ def LocalJsonHandler(self, handler, is_2d=False):
       json_end += "};"
 
     # Adding globe name helps ensure clearing of cache for new globes.
-    handler.write("%s/%s%s" % (
-        json_start, tornado.web.globe_.GlobeShortName(), json_end))
+    json_text = "%s/%s%s" % (
+      json_start, tornado.web.globe_.GlobeShortName(), json_end)
+
+    if json_version > 1:
```
tst-lsavoie: Minor: might be nice to check explicitly for '2' here in case we want to add other versions in the future.

earth_enterprise/src/fusion/portableglobe/servers/local_server.py
```diff
@@ -409,9 +410,19 @@ def ParseGlobeReqName(self, req):
...
-    
 
-  def LocalJsonHandler(self, handler, is_2d=False):
+  def JStoJson(self, js_string):
+    """Converts a JS server definition string to valid JSON."""
+    # Remove "var geeServerDefs = " or similar from start.
+    # Then add quotes to JSON keys that don't have them.
+    # Strip out the trailing ';'
+    # Finally, push it through json.dumps to ensure consistently-formatted output.
+    out_string = re.sub(r"^var \w+ ?= ?", "", js_string)
```
tst-lsavoie: This is fairly minor, but is this code duplicated in the regular server code? If it's not too much work it would be nice to have them shared.

tst-lsavoie:
We should add something to the release notes about this fix: https://github.com/google/earthenterprise/blob/master/docs/geedocs/5.3.0/answer/7160007.html

earth_enterprise/src/fusion/portableglobe/servers/local_server.py
ZetaZeta: I considered this!  But after looking at where it's located in the codebase, it feels like trying to access it here (or restructuring the data flow so it can be converted someplace where that code is easily accessed) would add more complication and risk of problems than it would avoid.

earth_enterprise/src/fusion/portableglobe/servers/local_server.py
ZetaZeta: I thought about this a bit!  That's what it was originally.  My reasoning for going for version > 1 instead is that, in the event that later versions are added, one of two cases are probably true:

1.  They want to do something totally different at this particular point, in which case they'll be editing it here and replacing this check.

2.  They want to do something different elsewhere based on the version number, in which case the "expected" result of incrementing the version is probably that they want to build on top of version 2 rather than regressing to version 1.

tst-lsavoie: That's sound reasoning. Works for me.

earth_enterprise/src/fusion/portableglobe/servers/local_server.py
tst-lsavoie: I'm good with that.

earth_enterprise/src/fusion/portableglobe/servers/local_server.py
tst-nfarah: if one pass v=3 (or any number > 1) they will get the new functionality while it was only intended for specific version (2). I think having specific version would resolve this issue and if future version support is needed, code would need to be changed any way.

earth_enterprise/src/fusion/portableglobe/servers/portable_server.py
```diff
@@ -292,14 +292,18 @@ class QueryHandler(portable_server_base.BaseHandler):
   def get(self):
```
tst-nfarah: Let's update the copyright notice to add `2019 Open GEE Contributors`. 
--> `Copyright 2017 Google Inc, 2019 Open GEE Contributors`

ZetaZeta: All right, done!

PR #1205, opened 2019-02-21
1166 validate imagery date

### LINKED ISSUE #1166: Fusion UI allows users to enter an invalid date/time for Imagery Resources
**Reproduction steps:**

1. Create a new imagery resource
2. For date, enter "2000-00-00"
3. Save the resource

**Expected:** An error message indicating the date is invalid
**Actual:** The resource saves, using this resource in a historical imagery project breaks the timeslider functionality

Also check to see if CLI tools are affected by this bug.

nrichman:
Fixes #1166 

This change uses the QT library to validate the date of an imagery resource. When an invalid date is entered (ex: 2019-02-29) and the user attempts to save the resource, an error message appears that says 'Invalid date.' The resource cannot be saved until a valid date is entered.


nrichman:
I signed it!

earth_enterprise/src/fusion/fusionui/QDateWrapper.cpp
```diff
@@ -130,6 +131,12 @@ void QDateWrapper::SetDate(const std::string& date) {
   }
 }
 
+bool QDateWrapper::IsValidDate() {
+  return QDate::isValid(year_edit_->text().toInt(),
```
tst-hrajamoney: We can check if time is entered, using the TimeIncluded() function. If included: Time should also be validated here, using QDateTime:isValid .

nrichman: I wasn't sure if I should validate the time as well - it didn't seem like the user was able to input a valid time. With 00:00:00 being a valid time and each field being constrained to its upper bounds. Should the function I've created still validate the time as well?

tst-hrajamoney: When I dug deeper, I found QDateWrapper.cpp already has :

void QDateWrapper::SetDateValidators(QObject* parent) {
 QIntValidator* year_validator = new QIntValidator(0, 9999, parent);
 QIntValidator* month_validator = new QIntValidator(0, 12, parent);
 QIntValidator* day_validator = new QIntValidator(0, 31, parent);

if we set the correct range here, then we may not need extra validations
this function prevents User from entering invalid values.

nrichman: I don't think that'd work for dates like it does for times since different months have a different number of days, especially with February and leap years.

tst-hrajamoney: yes, you are right. we need the function you added for additional checking. But should we  also modify the SetDateValidators so that 0 is not a valid  entry ?

earth_enterprise/src/fusion/fusionui/RasterAssetWidget.cpp
```diff
@@ -649,6 +649,10 @@ void RasterAssetWidget::AssembleEditRequest(
   } else {
     request->config.provider_id_ = 0;
   }
+
+  if (!acquisition_date_wrapper_->IsValidDate()) {
```
tst-ccamp: We have to make an exception for 0000-00-00. This should still be allowed and represents a blank date. It's just dates, such as 2000-00-00 that shouldn't be allowed.

tst-ccamp: I also tried saving with the valid date of 2000-01-01 and it still gives me the error saying I have entered an invalid date.

tst-ccamp: This was tested on CentOS 7.5

tst-ccamp:
@nrichman I'm still getting the same error when I tried saving with the default, all zeros. Are you getting a different result in Ubuntu?

image

tst-ccamp:
With the latest updates, I tested it and got the behavior I was expecting. The code changes themselves look good to me. One last thing, please add an entry for this change to the release notes here: https://github.com/google/earthenterprise/blob/master/docs/geedocs/5.3.0/answer/7160007.html

PR #1207, opened 2019-02-22
eliminate unnecessary update calls from autogen code

tst-basilhuffman:
initial commit

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
+++ b/earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
@@ -636,6 +636,7 @@ EOF
 # cachedinputs_ is empty. when this happens, 
 # UpdateInputs()is called twice: here, and in the
 # FixConfigBeforeUpdateCheck() if block, we want
+# to prevent this
 
 if ($hasinputs) {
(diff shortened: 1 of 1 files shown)


tst-lsavoie:
@tst-basilhuffman I just thought of something. It's possible that we're only seeing failures from the RasterProject code because that's the only code we're testing. It may be that my original suggestion to try this optimization is misguided and it's broken everywhere, and we just haven't noticed it because we're focused on imagery projects. Can you list all the autogenerated files that are affected by this change? We may need to come up with additional tests cases to hit all of them.

tst-basilhuffman:
@tst-lsavoie 
RasterProjectAssetD.cpp
RasterProductAssetD.cpp
MercatorRasterProductAssetD.cpp
VectorLayerXAssetD.cpp
MapLayerAssetD.cpp
MercatorRasterProjectAssetD.cpp

tst-lsavoie:
We're probably hitting RasterProject and RasterProduct but almost certainly not the others. While testing this PR we'll need to make sure we're exercising all the code that has changed.

tst-jkent:
Minor:
The indentation of the condition statements in the auto-generated code is off.
capture
capture2

PR #1209, opened 2019-02-26
Ensure the 5.3.0 copy of the 5.2.5 release notes matches the 5.2.5 copy of the release notes

tst-lsavoie:
It looks like some parts of the 5.2.5 release notes were missed in the 5.3.0 release notes. This PR fixes that.

### DIFF
--- a/docs/geedocs/5.3.0/answer/7160006.html
+++ b/docs/geedocs/5.3.0/answer/7160006.html
@@ -79,6 +79,14 @@ gtag('config', 'UA-108632131-2');
             <th>Resolution</th>
           </tr>
           <tr>
+            <td>200</td>
+            <td><code>stage_install</code> fails on the tutorial files when <code>/home</code>
+              and <code>/tmp</code> are on different file systems</td>
+            <td>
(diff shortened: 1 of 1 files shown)


docs/geedocs/5.3.0/answer/7160006.html
```diff
@@ -78,6 +78,14 @@ <h5>
             <th class="narrow">Description</th>
             <th>Resolution</th>
           </tr>
+          <tr>
```
tst-nfarah: would it be possible to push this change to github doc so the review/comparison with [5.2.5 release notes](https://www.opengee.org/geedocs/answer/7160006.html) be easier ?

tst-lsavoie:
Rendered release notes are here:
5.2.5 version of 5.2.5 release notes: https://tst-lsavoie.github.io/earthenterprise/geedocs/5.2.5/answer/7160006.html
5.3.0 version of 5.2.5 release notes: https://tst-lsavoie.github.io/earthenterprise/geedocs/5.3.0/answer/7160006.html

Also, the two files should be identical except that the 5.3.0 version has an extra link to the 5.3.0 release notes near the top.

Issue #1194, opened 2019-02-08
Sockets build up during large project builds


tst-jhaight:
It has been observed that running `getop` or the System Manager Fusion UI while gesystemmanager is mapping out very large builds can lead to an accumulation of sockets in the CLOSE_WAIT state.  For very large builds this socket buildup can result in fusion crashes due to resource exhaustion.

The half closed sockets are caused by System Manager holding a mutex for the duration of build planning which is also needed for reporting task status to both `getop` and the System Manager UI.  By default `getop` will close its side of a socket ~60 seconds after no response but System Manager waits until the mutex is available before realizing it should close its side.  Both getop and the Fusion UI will repeatedly query for task status until they are closed.

Workaround:
For very large builds close the System Manager Fusion UI and do not run getop.

Reproduction steps:

1. Set up and kick off a very large build (~6000 imagery resources should work)
1. Run `getop --timeout 1`
1. While system manager is planning the build repeatedly run `sudo lsof | grep CLOSE | wc -l` and observe the number of sockets in close wait state is increasing (sockets will be closed once system manager finishes build planning and releases the mutex)

tst-sbayard:
Fixed by PR #1204.

PR #1212, opened 2019-02-28
Updated release notes to include undocumented updates including 5.2.5.1 changes

tst-sbayard:
Included so far are:
- Controlling the XML cache size
- XML cache purging
- The addition of the `long_version` parameter to the SCons build command
- The fix for the socket build-up issue (#1194)

### DIFF
--- a/docs/geedocs/5.3.0/answer/7160007.html
+++ b/docs/geedocs/5.3.0/answer/7160007.html
@@ -44,7 +44,7 @@ gtag('config', 'UA-108632131-2');
   <h5>
     <p id="overview_5.3.0">New Features</p>
   </h5>
-    <p><strong>Added the ability to control the XML cache size . </strong>
+    <p><strong>Added the ability to control the XML cache size. </strong>
       The size can be modified by configuring the <code>INIT_HEAP_SIZE</code>, <code>MAX_HEAP_SIZE</code>, and <code>BLOCK_SIZE</code> settings in
     <code>/etc/opt/google/XMLparams</code>. This file must be created manually. View the <a href="../answer/176738.html">Configure Fusion performance</a> document for more details.</p>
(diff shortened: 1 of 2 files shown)


docs/geedocs/5.3.0/answer/7160007.html
```diff
@@ -44,7 +44,13 @@ <h2>Release notes: Open GEE 5.3.0</h2>
   <h5>
     <p id="overview_5.3.0">New Features</p>
   </h5>
-    <p><strong>Example feature</strong> replace with real feature description.</p>
+    <p><strong>Added the ability to control the XML cache size . </strong>
```
tst-jhaight: minor: space between `size` and the period

docs/geedocs/5.3.0/answer/7160007.html
```diff
@@ -111,6 +117,11 @@ <h5>
               Help has been added through the <code>--help</code> option, or if any invalid parameter is given
             </td>
           </tr>
+          <tr>
+            <td>1194</td>
+            <td>Sockets build up during large project builds</td>
+            <td>The Fusion UI and <code>getop</code> will now timeout when attempting to communicate with System Manager in order to avoid keeping sockets open indefinitely. The timeout is configurable.</td>
```
tst-jhaight: Minor: technically SystemManager is the side which will timeout so it doesn't build up a backlog of requests while another task is holding the lock.
There's also a new configurable parameter for `misc.xml` (`MutexTimedWaitSec`).  I'm not sure if that's something that goes in the release notes or not.

tst-sbayard: Agreed that release notes may not be the best place, but it does seem like it should go somewhere. Probably in 176738.html (Configure Fusion performance). I'll put a note in there and update the wording here to be more accurate.

PR #1223, opened 2019-03-14
make all method implementations in geterrain.h inline

tst-rwildes:
Fixes issue #1224


tst-blaird:
I don't think this change needs to go in the release notes as it won't affect users at all.

tst-jlarocco:
Looks good to me.

PR #1229, opened 2019-03-19
I1228 adding scroll bar to glc assembly web page

tst-eclamar:
This fix solves an issue where there is no scroll bar on the GLC Assembly webpage under Firefox, thus the finalizing button is not visible.  A scroll-bar is added.

Verification steps:

1. Navigate to Fusion Server admin page (127.0.0.1/admin)
2. Click on tool bar button ("gear" icon on right-hand side of page)
3. Select "Glc Assembly" - a pop-up window appears titled "GLC assembly tool"
4. Fill in 'name', 'title', and dummy text into 'polygon' fields.
5. Click on 'Layers' "3d" radio button
6. Scroll to the bottom of the page.
7. See "Assemble" button (greyed-out) at bottom of window. 
8. That one can see the button is success.


docs/geedocs/5.2.4/answer/3481558.html
```diff
@@ -69,26 +69,54 @@ <h2>Command reference</h2>
       </tr>
```
tst-nfarah: This document changes are for this PR ? They seems to be for different issue

Issue #1201, opened 2019-02-19
Fusion uses incorrect or no date if using .kip files as source


tst-ccamp:
**Reproduction steps:**
1. If necessary get .kip imagery source by copying the raster.kip directory into your `/gevol/src` directory (see example command below)
1. Create an imagery resource from a valid .kip directory, e.g.`genewmageryresource -o Resources/Imagery/BlueMarble_kip /opt/gevol/src/Imagery/BlueMarble.kip --sourcedate "2001-01-02 00:00:00" --provider NASA`
1. Create an imagery project and 3D database containing the above resource
1. Push/publish database
1. Open globe in Google Earth EC

Expected: The correct date, 2001-01-02, for the image under the cursor is displayed at the bottom of the Earth client
Actual: No date or a different date is displayed

It seems that if only the `raster.kip` directory is copied into the source directory, then no date is displayed. If the entire `ImageResource.kiasset` directory is copied it will use the previously set directory from the source data. Instead, it should be using the date that is specified either on the command line or in the Fusion UI, resource editor.

tst-lsavoie:
Closed via PR #1234

PR #1234, opened 2019-03-21
#1201 - Fix issue with setting source date when using .kip directories

tst-ccamp:
Verification steps:
1. Build 2 imagery resources with date set (suggestion: Blue Marble and usgsLanSat from tutorial)
1. Copy the 2 prebuilt resources into source directory, for Blue Marble just get the product raster.kip directory (you can rename it), for the USGS image get the whole asset (name.kiasset directory)
1. Create imagery resources using command line utility e.g. `genewimageryresource -o Resources/Imagery/usgsLanSat_kip1 /gevol/src/Imagery/usgsLanSat.kiasset/product.kia/ver001/raster.kip`
1. Before building the images, set a unique date for the new Blue Marble resource
1. Create a new Imagery project and database for the 2 new resources
1. Build and publish database
1. Load published database in Google Earth EC
1. The date for Blue Marble should be the new date set for the resource while the USGS image should be the original date that was set

Notes:
* If you end up changing a date, you will need to clean the resource version and rebuild the project and database before the new date will be used

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
@@ -190,9 +190,12 @@ void PacketGenAssetVersionImplD::UpdateChildren(
                                     inset.alphaRPFile,
                                     std::string(), /* cached blend */
                                     std::string()  /* cached blend alpha */));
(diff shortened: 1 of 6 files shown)


tst-dsugar:
This looks like it was really simple to solve, hopping the acquisition date across like that, but completely non obvious ...

tst-dsugar:
I was able to try it.

Issue #1236, opened 2019-03-22
Polygons and Polygon constructors should handle a KML string


tst-blaird:
In polygontoqtnodes.h, the Polygons and Polygon constructors expect a KML filename from which it creates an ifstream object to read the KML file.  A very small change would also allow it to read KML that's in an istringstream object.  This should be added for situations where the KML string has been read from a portable file.

tst-blaird:
Resolved by PR #1239.

PR #1239, opened 2019-03-26
#1236 Polygons constructor can accept KML as a string

tst-blaird:
Added the ability for Polygons constructor to accept a KML string.  The existing Polygons constructor would only accept a string that was a KML file.  I also added unit tests for the new constructor.

Testing should include:
1. The polygontoqtnodes_unittest should pass.
2. Using one or more KML files, run the gepolygontoqtnodes utility and ensure the result is the same as before these changes.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/polygontoqtnodes.h
+++ b/earth_enterprise/src/fusion/portableglobe/polygontoqtnodes.h
@@ -90,7 +90,7 @@ class Polygons {
 
   explicit Polygons(const std::string& kml_file);
 
-  explicit Polygons(const kml_as_string&, const std::string& kml_string);
+  Polygons(const kml_as_string&, const std::string& kml_string);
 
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/fusion/portableglobe/polygontoqtnodes.h
```diff
@@ -80,6 +90,8 @@ class Polygons {
 
   explicit Polygons(const std::string& kml_file);
 
+  explicit Polygons(const kml_as_string&, const std::string& kml_string);
```
tst-rwildes: explicit is only needed on single parameter constructors.

tst-blaird: Ok.  The PR is updated with this change.

tst-blaird:
PR approved by two reviewers and passed Travis with the unit tests.  Merging.

PR #1243, opened 2019-03-26
Fixes #279 - Admin Console password will be preserved on GEE server upgrade

### LINKED ISSUE #279: Server upgrade resets password
Upon upgrading GEE Server, the password for the Admin Console (username: geapacheuser) is reset to the default.  This might be a security issue for users who are not expecting it to be reset.

May be good idea to preserve the existing password by not overwriting /opt/google/gehttpd/conf.d/.htpasswd

Note: this affects both the server install script (install_server.sh) and the rpm upgrades

tst-jzmolek:
- The install_server.sh script has been updated to keep the password file from being overwritten if it already exists
- The server RPM pre and post install scripts will preserve and restore the password file as well
- Release notes and applicable copyright info also updated

Test reproduction steps for install_server.sh (tested on Ubuntu 16.04):
1. Start with a previous version of GEE server installed, but with the service stopped
2. Change the Admin Console password to a non-default one (using htpasswd)
3. Run the install_server.sh script, and start the server
4. Verify through the web interface that the Admin Console password did not revert to the default

Test reproduction steps for RPM install (tested on CentOS 7):
1. Start with a previous version of GEE server installed
2. Change the Admin Console password to a non-default one (using htpasswd)
3. Build the RPMs, then install them
4. Verify through the web interface that the Admin Console password did not revert to the default

### DIFF
--- a/earth_enterprise/src/installer/install_server.sh
+++ b/earth_enterprise/src/installer/install_server.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2017 Google Inc.
+# Copyright 2017 Google Inc., 2019 Open GEE Contributors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


earth_enterprise/src/installer/install_server.sh
```diff
@@ -459,7 +459,15 @@ copy_files_to_target()
...
   cp -rf "$TMPINSTALLDIR/server/opt/google/lib" "$BASEINSTALLDIR_OPT"
   if [ $? -ne 0 ]; then error_on_copy=1; fi
-  cp -rf "$TMPINSTALLDIR/server/opt/google/gehttpd" "$BASEINSTALLDIR_OPT"
+  
+  if [ -f "$BASEINSTALLDIR_OPT/gehttpd/conf.d/.htpasswd" ]
+  then
+    # Preserve Admin Console password
+    rsync -rl "$TMPINSTALLDIR/server/opt/google/gehttpd" "$BASEINSTALLDIR_OPT" --exclude conf.d/.htpasswd
+  else
+    cp -rf "$TMPINSTALLDIR/server/opt/google/gehttpd" "$BASEINSTALLDIR_OPT"
```
tst-dsugar: I could suggest rsync for both for consistency

earth_enterprise/src/installer/install_server.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2017 Google Inc.
+# Copyright 2017 Google Inc., 2019 Open GEE Contributors
```
tst-dsugar: that should be 2018-2019

tst-cjeffries (approved):
It all looks good to me.

earth_enterprise/src/installer/install_server.sh
tst-jzmolek: done

earth_enterprise/src/installer/install_server.sh
tst-jzmolek: done

tst-dsugar (approved):

Issue #1246, opened 2019-03-28
Update master to use version 5.3.1


tst-rwildes:
Files defined in release processing wiki need to have version numbers changed from `5.3.0` to `5.3.1`.  Also 5.3.1 documentation folder needs to be created

tst-rwildes:
All changes have been completed to make master version `5.3.1` now.

PR #1249, opened 2019-03-28
link up 5.3.1 doc folder

tst-rwildes:
Addresses issue: #1246 This PR hooks up the doc folder for 5.3.1


docs/_data/versions.yml
```diff
@@ -12,3 +12,5 @@
   stability: latest stable
 - version: 5.3.0
   stability: unstable
```
tst-jlarocco: Should this be changed to "stable"?

tst-rwildes: > Should this be changed to "stable"?

Not yet.  Once we have a final build that passes all testing then we can declare it stable and at that time it will be changed.

docs/geedocs/5.3.1/answer/7160007.html
```diff
@@ -18,6 +18,7 @@
 <!-- 7160002.html -->
 <div class="container">
 <div class="sidebar1">
+  <p class="sidebar-toc"><a href="../answer/7160008.html">Open GEE 5.3.1</a></p>
   <p class="sidebar-toc"><a href="../answer/7160007.html" class="current-file">Open GEE 5.3.0</a></p>
```
tst-blaird: Does the `class="current-file"` attribute need to move up to the 5.3.1 line?

tst-blaird: Never mind, answered my own question with `no`.  😃

PR #1250, opened 2019-03-29
WIP - Parallel build index

tst-dsheremata:
Converted this WIP branch into a WIP PR, to make it easier to collaborate on in the interim.

### DIFF
--- a/earth_enterprise/src/common/khCache.h
+++ b/earth_enterprise/src/common/khCache.h
@@ -133,22 +133,19 @@ class khCache {
     item->next = head;
     item->prev = 0;
     head = item;
-    if (item->next) {
+    if (item->next)
       item->next->prev = item;
-    }
     else
       tail = item;
     ++numItems;
   }
   void Unlink(Item *item) {
-    if (item->prev) {
+    if (item->prev)
       item->prev->next = item->next;
-    }
     else
       head = item->next;
-    if (item->next) {
+    if (item->next)
       item->next->prev = item->prev;
-    }
     else
       tail = item->prev;
     item->next = 0;
--- a/earth_enterprise/src/common/khCache.h
+++ b/earth_enterprise/src/common/khCache.h
@@ -20,6 +20,7 @@
 
 #include <map>
 #include <vector>
+#include "khThread.h"
 
 // #define SUPPORT_VERBOSE
 #ifdef SUPPORT_VERBOSE
@@ -55,6 +56,7 @@ class khCacheItem {
   // private and unimplemented - no copying allowed, would mess up links
   khCacheItem(const khCacheItem&);
   khCacheItem& operator=(const khCacheItem&);
+  khMutex mutex;
  public:
   khCacheItem *next;
   khCacheItem *prev;
@@ -62,6 +64,8 @@ class khCacheItem {
   Value val;
   khCacheItem(const Key &key_, const Value &val_)
       : next(0), prev(0), key(key_), val(val_) { }
+  void Lock(void) { mutex.Lock(); }
+  void Unlock(void) { mutex.Unlock(); }
 };
 
 // #define CHECK_INVARIANTS
@@ -131,19 +135,28 @@ class khCache {
     item->next = head;
     item->prev = 0;
     head = item;
-    if (item->next)
+    if (item->next) {
+      item->next->Lock();
       item->next->prev = item;
+      item->next->Unlock();
+    }
     else
       tail = item;
     ++numItems;
   }
   void Unlink(Item *item) {
-    if (item->prev)
+    if (item->prev) {
+      item->prev->Lock();
       item->prev->next = item->next;
+      item->prev->Unlock();
+    }
     else
       head = item->next;
-    if (item->next)
+    if (item->next) {
+      item->next->Lock();
       item->next->prev = item->prev;
+      item->next->Unlock();
+    }
     else
       tail = item->prev;
     item->next = 0;
@@ -197,6 +210,7 @@ class khCache {
     // delete any previous item
     Item *item = FindItem(key);
     if (item) {
+      item->Lock();
       Unlink(item);
 #ifdef SUPPORT_VERBOSE
       if (verbose) notify(NFY_ALWAYS, "Deleting previous %s from cache", key.c_str());
@@ -206,7 +220,9 @@ class khCache {
 
     // make a new item, link it in and add to map
     item = new Item(key, val);
+    item->Lock();
     Link(item);
+    item->Unlock();
     map[key] = item;
 #ifdef SUPPORT_VERBOSE
     if (verbose) notify(NFY_ALWAYS, "Adding %s to cache", key.c_str());
@@ -225,6 +241,7 @@ class khCache {
 
     Item *item = FindItem(key);
     if (item) {
+      item->Lock();
       Unlink(item);
       map.erase(key);
 #ifdef SUPPORT_VERBOSE
@@ -245,8 +262,10 @@ class khCache {
     Item *item = FindItem(key);
     if (item) {
       // move it to the head of the list
+      item->Lock();
       Unlink(item);
       Link(item);
+      item->Unlock();
 
       CheckListInvariant();
 
@@ -270,6 +289,7 @@ class khCache {
       if (item) {
         Item *tokill = item;
         item = item->prev;
+        tokill->Lock();
         Unlink(tokill);
         map.erase(tokill->key);
 #ifdef SUPPORT_VERBOSE
--- a/earth_enterprise/src/common/khCache.h
+++ b/earth_enterprise/src/common/khCache.h
@@ -20,7 +20,7 @@
 
 #include <map>
 #include <vector>
-#include "khThread.h"
+#include <mutex>
 
 // #define SUPPORT_VERBOSE
 #ifdef SUPPORT_VERBOSE
@@ -56,7 +56,6 @@ class khCacheItem {
   // private and unimplemented - no copying allowed, would mess up links
   khCacheItem(const khCacheItem&);
   khCacheItem& operator=(const khCacheItem&);
-  khMutex mutex;
  public:
   khCacheItem *next;
   khCacheItem *prev;
@@ -64,8 +63,6 @@ class khCacheItem {
   Value val;
   khCacheItem(const Key &key_, const Value &val_)
       : next(0), prev(0), key(key_), val(val_) { }
-  void Lock(void) { mutex.Lock(); }
-  void Unlock(void) { mutex.Unlock(); }
 };
 
 // #define CHECK_INVARIANTS
@@ -84,6 +81,7 @@ class khCache {
   bool verbose;
 #endif
   uint numItems;
+  std::recursive_mutex mtx;
 
   bool InList(Item *item) {
     Item *tmp = head;
@@ -136,9 +134,7 @@ class khCache {
     item->prev = 0;
     head = item;
     if (item->next) {
-      item->next->Lock();
       item->next->prev = item;
-      item->next->Unlock();
     }
     else
       tail = item;
@@ -146,16 +142,12 @@ class khCache {
   }
   void Unlink(Item *item) {
     if (item->prev) {
-      item->prev->Lock();
       item->prev->next = item->next;
-      item->prev->Unlock();
     }
     else
       head = item->next;
     if (item->next) {
-      item->next->Lock();
       item->next->prev = item->prev;
-      item->next->Unlock();
     }
     else
       tail = item->prev;
@@ -194,6 +186,7 @@ class khCache {
     clear();
   }
   void clear(void) {
+    std::lock_guard<std::recursive_mutex> lock(mtx);
     CheckListInvariant();
     // Delete all the items.
     for (typename MapType::iterator i = map.begin(); i != map.end(); ++i) {
@@ -205,12 +198,12 @@ class khCache {
     head = tail = 0;
   }
   void Add(const Key &key, const Value &val, bool prune = true) {
+    std::lock_guard<std::recursive_mutex> lock(mtx);
     CheckListInvariant();
 
     // delete any previous item
     Item *item = FindItem(key);
     if (item) {
-      item->Lock();
       Unlink(item);
 #ifdef SUPPORT_VERBOSE
       if (verbose) notify(NFY_ALWAYS, "Deleting previous %s from cache", key.c_str());
@@ -220,9 +213,7 @@ class khCache {
 
     // make a new item, link it in and add to map
     item = new Item(key, val);
-    item->Lock();
     Link(item);
-    item->Unlock();
     map[key] = item;
 #ifdef SUPPORT_VERBOSE
     if (verbose) notify(NFY_ALWAYS, "Adding %s to cache", key.c_str());
@@ -237,11 +228,11 @@ class khCache {
   }
 
   void Remove(const Key &key, bool prune = true) {
+    std::lock_guard<std::recursive_mutex> lock(mtx);
     CheckListInvariant();
 
     Item *item = FindItem(key);
     if (item) {
-      item->Lock();
       Unlink(item);
       map.erase(key);
 #ifdef SUPPORT_VERBOSE
@@ -259,13 +250,12 @@ class khCache {
   }
 
   bool Find(const Key &key, Value &val) {
+    std::lock_guard<std::recursive_mutex> lock(mtx);
     Item *item = FindItem(key);
     if (item) {
       // move it to the head of the list
-      item->Lock();
       Unlink(item);
       Link(item);
-      item->Unlock();
 
       CheckListInvariant();
 
@@ -277,6 +267,7 @@ class khCache {
   }
 
   void Prune(void) {
+    std::lock_guard<std::recursive_mutex> lock(mtx);
     CheckListInvariant();
     Item *item = tail;
     while (item && (map.size() > targetMax)) {
@@ -289,7 +280,6 @@ class khCache {
       if (item) {
         Item *tokill = item;
         item = item->prev;
-        tokill->Lock();
         Unlink(tokill);
         map.erase(tokill->key);
 #ifdef SUPPORT_VERBOSE
@@ -303,6 +293,7 @@ class khCache {
 
   // Get old items from the LRU cache, excluding recent "toKeep" items.
   void GetOldKeys(size_t toKeep, std::vector<Key> *oldKeys) {
+    std::lock_guard<std::recursive_mutex> lock(mtx);
     if (map.size() <= toKeep) {
       return;
     }
--- a/earth_enterprise/src/common/khMTTypes.h
+++ b/earth_enterprise/src/common/khMTTypes.h
@@ -21,8 +21,10 @@
 #include "khThread.h"
 #include "khGuard.h"
 #include <deque>
+#include <memory>
 #include <set>
 #include <unordered_map>
+#include <mutex>
 
 template <class T>
 class khAtomic {
@@ -148,6 +150,14 @@ class khMTSet
     khLockGuard guard(mutex);
     return std::vector<T>(set.begin(), set.end());
   }
+  void clear(void) {
+    khLockGuard guard(mutex);
+    set.clear();
+  }
+  size_t size(void) {
+    khLockGuard guard(mutex);
+    return set.size();
+  }
 };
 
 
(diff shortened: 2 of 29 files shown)


earth_enterprise/src/fusion/autoingest/Asset.h
```diff
@@ -36,7 +37,7 @@
  ***  ... = asset->config.layers.size();
  ***
  ******************************************************************************/
-class AssetImpl : public khRefCounter, public AssetStorage {
+class AssetImpl : public khMTRefCounter, public AssetStorage {
```
tst-jhaight: I'm not positive if this change was necessary or not but I switched to making all AssetImpl objects use the multi-thread ref counter to be safe.

earth_enterprise/src/common/khTypes.h
```diff
@@ -19,6 +19,26 @@
 #define GEO_EARTH_ENTERPRISE_SRC_COMMON_KHTYPES_H_
 
 #include <assert.h>
+#include <mutex>
+#include <vector>
+
+template <typename T>
```
tst-jhaight: The class below is really only protecting a pretty small section of code because vector's push_back isn't virtual and a lot of vector references are passed around.  If you change MTVector to private inheritance you'll start seeing all of these because nothing will compile.

earth_enterprise/src/common/khCache.h
```diff
@@ -80,6 +81,7 @@ class khCache {
   bool verbose;
 #endif
   uint numItems;
+  std::recursive_mutex mtx;
```
tst-jhaight: I made this recursive instead of the regular mutex because both the public `Add` and `Remove` methods call `Prune` which is also public.

The cache definitely needs some form of protection on public methods when we move to use anything in parallel.

earth_enterprise/src/common/khTypes.h
tst-jhaight: Another pretty big shortcoming here is for loops over the container items can't really be protected without modification of the calling logic.

earth_enterprise/src/fusion/autoingest/AssetStorage.idl
```diff
@@ -88,11 +88,11 @@ class AssetVersionStorage {
   AssetDefs::State state;
   double       progress;
   bool         locked;
-  std::vector<std::string> inputs;
-  std::vector<std::string> children;
-  std::vector<std::string> parents;
-  std::vector<std::string> listeners;
-  std::vector<std::string> outfiles;
+  MTVector<std::string> inputs;
```
tst-jhaight: IIRC `inputs` was the attribute I first hit problems with during parallel builds.  I made all 5 use the new type at the same time just to be safe.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1153,16 +1155,20 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
...
+      packgenassets.push_back(packgen->GetAssetRef());
+      packgenvers.push_back(packgen);
+    }
   }
 
   curTime = std::chrono::high_resolution_clock::now();
   timeDiff = std::chrono::duration<double>(curTime - prevTime);
   notify(NFY_INFO2, "BuildIndex finished looping through insets - %f seconds elapsed",
          timeDiff.count());
+  assert(MiscConfig::Instance().ContLoadAndThink);
```
tst-jhaight: I added this to cut off large builds at the same point.  When fusion is killed here, nothing is written out to file yet so there is no need to clean up or redo setup for the next build.  I also use this to help get metrics on how long a specific build took.  Definitely don't plan on keeping this.

earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp
```diff
@@ -606,7 +606,8 @@ khAssetManager::SubmitTask(const std::string &verref, const TaskDef &taskdef,
 {
   // assert that we're already locked
   assert(!mutex.TryLock());
-
+  static std::mutex theMutex;
```
tst-jhaight: Attempting to run BuildIndex in parallel I was able to catch segfaults stemming from here.  The `OnStateChange` calls wound up eventually getting in here from different threads at the same time during load & think.  The simple check on line 608 passes because one of the threads _is_ holding the mutex.  I got a little heavy handed here.  This could be more granular but not sure if it is worth it.

tst-lsavoie:
From code inspection, here are the places I can think of that might cause a problem when called from multiple threads at the same time:
- khCache (you've already handled this)
- `parentsToNotify` and `listenersToNotify` in AssetVersionD.h. If I understand how the code works correctly, the code that is being parallelized never adds items to either of these, so we should be OK, but technically they are being used from multiple threads at the same time
- The `dirtyMap`s in MutableAssetHandleD_. There could be trouble in MutalbeAssetHandleD_::OnBind or khAssetManager::ApplyPending was called from different threads at the same time.
- `Asset::cache` and `AssetVersion::cache` - you could end up with two different caches if these were called concurrently. With the code we're parallelizing the caches will certainly have been created before multiple threads start running, but it might be wise to be safe.
- Any of the `Asset::DoBind` or `AssetVersion::DoBind` functions. Those are thread safe already because the cache is thread safe, but if it was called twice at the same time you could end up loading the asset from disk multiple times and have duplicate assets in memory. I think everything would work correctly even if this happened, but it's probably best to protect against it.

earth_enterprise/src/fusion/autoingest/Asset.h
```diff
@@ -93,6 +94,7 @@ template <>
 inline void
 Asset::DoBind(const std::string &ref, bool checkFileExistenceFirst) const
 {
+  khTargetedLock lock(ref);
```
tst-jhaight: This was added in order to prevent two threads from hitting this section simultaneously for the same file and both independently loading the file into memory.  Is this overkill to prevent something that *shouldn't* cause any problems?

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,173 @@ class khMTSet
...
+ public:
+  khMTMap(void) { }
+
+  void insert(const std::pair<K,V> &val) {
+    std::lock_guard<std::mutex> guard(mutex);
+    (void)map.insert(val);
+  }
+  void erase(const K &val) {
+    std::lock_guard<std::mutex> guard(mutex);
+    if (map.find(val) != map.end()) {
```
tst-lsavoie: Is this check necessary? I assume just calling `map.erase` will include this check?

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,173 @@ class khMTSet
...
+    // forward to private copy constructor to protect vector from modification
+    MTVector(const MTVector& a) : MTVector(a, std::lock_guard<std::mutex>(a.mtx)) { }
+    const T& operator[] (const size_t idx) const {
+      std::lock_guard<std::mutex> lock(mtx);
+      return Base::operator[](idx); 
+    }
+
+    bool operator==(const MTVector& x) const {
+      std::lock_guard<std::mutex> lock(mtx);
+      return true;
```
tst-lsavoie: Is this class intended to be permanent? There's several unimplemented functions, like this one that says all MTVectors are the same.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,173 @@ class khMTSet
...
+      static FileLocks file_locks;
+      return file_locks;
+    } 
+  public:
+    khTargetedLock(const std::string& filename) : s(filename) {
+      FileLocks& file_mutexes = getFileLocks();
+      {
+        std::lock_guard<std::mutex> global_lock(globalMutex());
+        if (file_mutexes.find(s) == file_mutexes.end()) {
+          file_mutexes[s] = std::make_shared<std::mutex>();
```
tst-lsavoie: I don't think there's a need for a shared pointer here. `file_mutexes` will always have a pointer, so the object will never be deleted.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1153,10 +1155,13 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
       genInfo[i].PopulatePacketLevels(packgen);
     }
 
-    // remember this new PacketGen version
-    packgenkids.push_back(packgen->GetRef());
-    packgenassets.push_back(packgen->GetAssetRef());
-    packgenvers.push_back(packgen);
+#pragma omp ordered
```
tst-lsavoie: I suspect using `ordered` here is overkill but I think it's fine.

earth_enterprise/src/common/khMTTypes.h
tst-jhaight: oops.  No this shouldn't be here but I forgot about it while fixing other problems.  I will fix this to do a proper comparison.

earth_enterprise/src/common/khMTTypes.h
tst-jhaight: I went with the `shared_ptr` so the destructor could use the reference count to know when it is safe to remove the mutex from `file_mutexes`.  I wanted to try to keep `file_mutexes` from growing too large.  If thread 1 and 2 are both about to load file X into memory, thread 1 creates the mutex and then 2 would destroy it (assuming no other thread comes in before 2 finishes its work).  Is this making things too complicated?

tst-lsavoie: Ah, OK. I didn't look closely enough at the destructor.

I am a bit worried about this, though. It does seem pretty complicated, though it may be necessary complication.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,173 @@ class khMTSet
...
+        }
+        theFileMutex = file_mutexes[s];
+      }
+      theFileMutex->lock();
+    }
+    ~khTargetedLock(void) {
+      theFileMutex->unlock();
+      std::lock_guard<std::mutex> global_lock(globalMutex());
+      if (theFileMutex.use_count() == 2) {
```
tst-lsavoie: I've heard use_count is just a rough count - it's not guaranteed to be exactly correct.

earth_enterprise/src/common/khMTTypes.h
tst-jhaight: Yeah I'm not in love with this idea and I'm open to suggestions.  This seemed much less invasive than other ideas I could come up with.  One other idea was altering the auto-generated code for load to lock and then re-check a cache (if provided as an argument) and immediately add the loaded file to the cache before releasing the lock.

tst-lsavoie: Some (possibly bad) ideas I came up with:
- Just have a single mutex for all files. We'd only be able to read one file at a time, but that's probably what we do anyway, and reading two files from the same disk at the same time can perform badly.
- Can we do something more optimistic? Maybe we read the file unconditionally but we don't put it in the cache if it's already there.

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -91,11 +92,12 @@ AssetVersionImplD::StateChangeNotifier::NotifyParents(
     std::shared_ptr<StateChangeNotifier> notifier) {
   notify(NFY_VERBOSE, "Iterate through parents");
   int i = 1;
-  for (const std::string & ref : parentsToNotify) {
+  std::vector<std::string> theRefs = parentsToNotify.keys();
```
tst-lsavoie: There's a potential issue here where we call `parentsToNotify.clear()` at the end of this function - that could clear out things that were added after you called `keys()`. On the other hand, this function is only called from the destructor, so it's guaranteed to run in a single thread anyway. (On this branch it may be called in 2 places, but the one that's not in the destructor has been removed on @tst-sbayard's version of this branch).

earth_enterprise/src/common/khMTTypes.h
tst-jhaight: I like either of those better than my current implementation since it makes everything much less complicated.  I'm leaning toward using your first idea of a single load lock.

tst-lsavoie: My only concern with the single lock is that reading only a single file at once might be a performance issue, especially on high performance storage systems. However, it seems like the disk IO should be a bigger bottleneck than the serialization.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,173 @@ class khMTSet
     khLockGuard guard(mutex);
     return std::vector<T>(set.begin(), set.end());
   }
+  void clear(void) {
+    khLockGuard guard(mutex);
+    set.clear();
+  }
+  size_t size(void) {
+    khLockGuard guard(mutex);
```
tst-lsavoie: Minor: you might be able to get away with not using the mutex for the size function. If you call size while an insertion is happening I don't know if we care whether it gives the size before or after the insertion.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,173 @@ class khMTSet
...
+    const_iterator begin(void) const {
+      return Base::begin();
+    }
+
+    const_iterator end(void) const {
+      return Base::end();
+    }
+
+    size_t size(void) const {
+      std::lock_guard<std::mutex> lock(mtx);
```
tst-lsavoie: Minor - probably don't need a mutex here.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -189,7 +189,7 @@ class MutableAssetHandleD_ : public virtual Base_ {
       uint32 numDotNew=0;
       for (std::vector<std::string>::iterator it = toDelete.begin();
            it != toDelete.end(); ++it) {
-        if (dirtyMap.find(*it) != dirtyMap.end()) {
+        if (dirtyMap.contains(*it)) {
```
tst-lsavoie: This function was removed in the master branch. You could remove it here as well to make your work easier.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -265,26 +263,26 @@ class MutableAssetHandleD_ : public virtual Base_ {
...
       }
     }
   }
 
 
   static bool SaveDirtyToDotNew(khFilesTransaction &savetrans,
                                 std::vector<std::string> *saveDirty) {
-    for (typename DirtyMap::const_iterator d = dirtyMap.begin();
-         d != dirtyMap.end(); ++d) {
+    std::vector<std::string> dirtyKeys = dirtyMap.keys();
```
tst-lsavoie: The function that calls this function clears the dirtyMap after this runs, which could be a problem. We only want to clear the keys that we saved, not any that were added since the function started.

tst-lsavoie:
I'm definitely getting some missing inputs in the packgen khassetver.xml files.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,137 @@ class khMTSet
...
+    // forward to private copy constructor to protect vector from modification
+    MTVector(const MTVector& a) : MTVector(a, std::lock_guard<std::mutex>(a.mtx)) { }
+
+    const T& operator[] (const size_t idx) const {
+      std::lock_guard<std::mutex> lock(mtx);
+      return Base::operator[](idx); 
+    }
+
+    bool operator==(const MTVector& x) const {
+      std::mutex* firstMtx = &x.mtx;
```
tst-lsavoie: Good work here - I'm not sure I would have thought of locking both mutexes and acquiring them in a fixed order to avoid deadlock. Might be good to include a comment that that's what you're doing.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -265,26 +174,27 @@ class MutableAssetHandleD_ : public virtual Base_ {
...
     }
   }
 
 
   static bool SaveDirtyToDotNew(khFilesTransaction &savetrans,
                                 std::vector<std::string> *saveDirty) {
-    for (typename DirtyMap::const_iterator d = dirtyMap.begin();
-         d != dirtyMap.end(); ++d) {
+    DirtyMap tempDirty = std::move(dirtyMap);
+    std::vector<std::string> dirtyKeys = tempDirty.keys();
```
tst-lsavoie: I think there's still an issue here. Once you call std::move on dirtyMap it's no longer valid, so it needs to be reinitialized. I think that simply calling dirtyMap.keys() and then dirtyMap.clear() while holding a lock would do the trick. Similarly, if you just call dirtyMap.keys() and then remove the keys one by one so that you only remove the keys that you've handled would also work.

earth_enterprise/src/fusion/autoingest/AssetVersion.h
```diff
@@ -114,10 +114,9 @@ class AssetVersionImpl : public khMTRefCounter, public AssetVersionStorage {
 
   template <class outIter>
   outIter GetInputs(outIter oi) const {
-    for (std::vector<std::string>::const_iterator i = inputs.begin();
-         i != inputs.end(); ++i) {
-      oi++ = *i;
-    }
+    auto theClosure = [&oi](const std::string& v) { oi++ = v; };
+    outfiles.doForEach(theClosure);
+    //inputs.copyTo(oi);
```
tst-jkent: Minor: Should this comment be removed?

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,166 @@ class khMTSet
...
+// WARNING: Use this with extreme caution
+// It protects some obvious problems but creates some less obvious ones
+template <typename T>
+class MTVector {
+  public:
+    typedef std::vector<T> Base;
+    typedef typename Base::const_iterator const_iterator;
 
+  private:
+    mutable std::mutex mtx;
```
tst-jhaight: Note for reviewers: we could change this to be a pthread read/write lock.  From my testing on an 8 core machine there was minimal benefit to doing this, so I left it out.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +151,166 @@ class khMTSet
...
+      thekeys.push_back(p.first);
+    }
+    return thekeys;
+  }
+  bool contains(const K& key) {
+    std::lock_guard<std::mutex> guard(mutex);
+    return map.find(key) != map.end();
+  }
+  bool empty(void) const {
+    std::lock_guard<std::mutex> guard(mutex);
```
tst-lsavoie: Minor: probably no lock required here

earth_enterprise/src/common/khMTTypes.h
tst-lsavoie: Is there any significant downside? It might be nice to have the code just in case.

earth_enterprise/src/common/khxml/khdom.h
```diff
@@ -282,6 +289,18 @@ ToElementWithChildName(khxml::DOMElement *elem,
...
 }
 
+
+template <class T>
+void
+ToElementWithChildName(khxml::DOMElement *elem,
+                       const std::string &childTagName,
+                       const MTVector<T> &val)
+{
+  auto theClosure = [&](const std::string& v) { AddElement(elem, childTagName, v); };
```
tst-lsavoie: Minor: I prefer writing lambdas inline in cases like this. I think it's easier to read.
```
  val.doForEach([&](const std::string& v) {
    AddElement(elem, childTagName, v);
  });
```

tst-lsavoie: The same comment applies pretty much everywhere you've used a function like this, though it's a stylistic issue so it's not a huge deal.

earth_enterprise/src/common/khxml/khdom.h
```diff
@@ -864,6 +897,24 @@ FromElementWithChildName(khxml::DOMElement *elem,
   }
 }
 
+template <class T>
+void
+FromElementWithChildName(khxml::DOMElement *elem,
+                         const std::string &childTagName,
+                         MTVector<T> &vec)
+{
+  vec.clear();
```
tst-lsavoie: Technically there are race conditions here because another thread could be adding stuff to the vector after you clear it. You'll always end up with a valid vector but it may not be what you expect after calling this function. Given what our code does I don't think that will be a problem, but I thought I'd bring it up.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketLevel.src
```diff
@@ -128,14 +109,19 @@ void PacketLevelAssetVersionImplD::DoSubmitTask(void) {
...
+  auto theClosure = [&](const std::string& v)
+  {
+    std::shared_ptr<AssetVersion> p = std::make_shared<AssetVersion>(v);
+    if (*p)
+      goodVersions.push(p);
+  };
+  ver->inputs.doForEach(theClosure);
+  while (!goodVersions.empty()) {
+    AssetVersion& input = *(goodVersions.front().get());
+    auto anotherClosure = [&inputkids](const std::string& v) { inputkids.push_back(v); };
```
tst-lsavoie: Minor: could improve naming here (or just make this inline).

tst-lsavoie:
I can't say that there are obviously no problems with this PR, but there are no obvious problems.

tst-lsavoie:
I've tested the latest updates and the xml files come out the same in parallel and in serial (not counting things like times and unique IDs). The binary files, however, are pretty different, and I noticed in one of the log files that at least one of the output files is a slightly different size. I don't know if that means anything, but it does worry me a bit.

PR #1252, opened 2019-03-30
Move build number to release rpm

### LINKED ISSUE #1251: Build number should not be part of the RPM version string
The build number is currently appended to the RPM version string after the release version and with no way to distinguish where the version number ends and the build number begins.  This causes issues with point releases where the build number becomes part of the comparison with the version number and causes invalid compare results.  Instead the build number should be a moved to the rpm release field (currently we are just hard coding this to `1`) where it will properly serve as a secondary version (as it should be used).

tst-rwildes:
This fixes #1251 

This will not append the build number to the rpm version.  Instead build number will be put into the rpm release field and be used as a secondary version as it should be.   It is possible (but rarely done) to do multiple releases within a given version (mostly for critical bug fixes).

### DIFF
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -85,13 +85,16 @@ def rpmPlatformString = org.opengee.os.Platform.rpmPlatformString
 def gee_long_version_txt = "${stagedInstall_path}/gee_long_version.txt"
 def openGeeVersionFile = new File( gee_long_version_txt )
 
-def getOsPackageVersionFromOpenGeeVersion(value) {
-    return value.replaceAll(/-.*release/, '').
-        replaceAll('-', '.')
+ext.getOsPackageVersionFromOpenGeeVersion = { value -> 
+    return value.split('-')[0]
 }
 
-def getOsPackageVersionFromOpenGeeVersionFile(file) {
-    return getOsPackageVersionFromOpenGeeVersion(
+ext.getOsPackageBuildFromOpenGeeVersion = { value ->
+    return value.split('-')[1]
+}
+
+def getOsPackageVersionOrBuildFromOpenGeeVersionFile(file, getString) {
+    return getString(
         file.readLines().
         collect {
             // Strip comments:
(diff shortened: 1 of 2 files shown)


tst-rwildes:
What RPM info not looks like with these changes:
`
$ rpm -qip earth_enterprise/rpms/build/distributions/opengee-fusion-5.3.0.1-891.4.1.el7.x86_64.rpm
Name        : opengee-fusion
Epoch       : 0
Version     : 5.3.0.1
Release     : 891.4.1.el7
Architecture: x86_64
Install Date: (not installed)
Group       : Application/Productivity
Size        : 117287718
License     : ASL 2.0
Signature   : (none)
Source RPM  : opengee-fusion-5.3.0.1-891.4.1.el7-src.rpm
Build Date  : Sat 30 Mar 2019 04:18:29 AM UTC
Build Host  : ip-10-0-2-165.ec2.internal
Relocations : (not relocatable)
Packager    : gee-oss@googlegroups.com
Vendor      : The OpenGEE Contributors
URL         : https://github.com/google/earthenterprise
Summary     : Open GEE Fusion UI and Command-line Utilities
Description :
Command line utitiles and Qt UI for building Open GEE globes.

The utilities in this package allow Open GEE globe and map databases to be
built from raster, vector, and location properties data.
`

earth_enterprise/rpms/build.gradle
```diff
@@ -176,20 +193,34 @@ def rpmCommandFilter = {
     return rpmCapabilityMap.containsKey(it) ? rpmCapabilityMap[it] : it
 }
 
+// A lamda to figure out the best way to get the build number
```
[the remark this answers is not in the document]
tst-rwildes: Fixed!  Thanks!

earth_enterprise/rpms/build.gradle
```diff
@@ -685,7 +716,7 @@ task openGeeFusionRpm (type: GeeRpm, dependsOn: openGeeFusionInitScripts) {
 
     packageName = 'opengee-fusion'
     version = ospackage.version
-    release = "1.${rpmPlatformString}"
+    release = ospackage.release
```
tst-dsugar: where is this coming from now, as release is more specifically about rpm build/package revision...

tst-rwildes: It is the build number (i.e. what is returned as part of `earth_enterprise/src/scons/getversion.py -l`).  They do say that the main intent is for packaging but then also mention it can be used for hot fixing which in a sense is what we are doing by putting the build number in there  Also, we don't have separate repositories/build process for packaging the binaries vs building the Open GEE binaries and so where not using this at all (just hard coding it to `1`).  Because of all this it seems okay to hijack it for the purposes of distinguishing what build within a release cycle this package was created from.  It does bend the rules a bit but gives us the behavior we need/desire.

PR #1265, opened 2019-04-08
[WIP] Replace string references to assets with SharedString objects.

tst-sbayard:
Replacing string references to assets with SharedString objects in order to avoid storing duplicate copies of 
potentially tens of thousands of strings and reduce time spent on string comparisons.


tst-lsavoie:
Superseded by PR #1266

PR #1266, opened 2019-04-09
Simple shared strings

### LINKED ISSUE #1277: Change asset and asset version keys to ints
Fusion uses strings to refer to assets and asset versions (the string corresponds to the asset's location under the asset root). This results in lots of duplicate, possibly long strings since every asset needs to hold the string referring to every other asset it has a relationship with (children, parents, inputs, and listeners). For large builds this can take lots of memory and use extra processing time since string comparisons are inefficient. We should update Fusion to use ints instead of strings when referring to assets and just store a single table relating strings to assets. This should reduce memory consumption and improve performance.

tst-sbayard:
Replacing duplicate asset and version references (strings) with objects that will use far less space.

Fixes #1277


earth_enterprise/src/common/SharedString.h
```diff
@@ -0,0 +1,123 @@
...
+// thousands of duplicates of each string in very large projects.
+//
+// With this class, only an integer identifier is stored in each instance (thus
+// drastically reducing the amount of memory needed in large projects), and
+// that identifier is used to retrieve the string value when it is requested.
+// Since the string-to-identifier ratio is 1:1, equality comparisons are also 
+// much faster as we only need to compare the identifiers.
+class SharedString {
+protected:
+   class RefStorage {
```
tst-lsavoie: This is nit-picky, but SharedString is generic enough that it may not always hold asset refs. It might be a good idea to rename this (and other parts of the code that talk about refs) to StringStorage or something like that.

earth_enterprise/src/common/SharedString.h
```diff
@@ -0,0 +1,129 @@
...
+
+    operator std::string() const {
+        return refStore.RefFromKey(key);
+    }
+
+    std::string toString() const {
+        return refStore.RefFromKey(key);
+    }
+
+    bool operator<(const SharedString &other) const {
```
tst-basilhuffman: Maybe add operator> for the sake of consistency

earth_enterprise/src/common/SharedString.h
```diff
@@ -0,0 +1,129 @@
...
+            keyFromRefTable.insert(std::pair<std::string, uint32_t>("", 0));
+				}
+    };
+
+    static RefStorage refStore;
+    uint32_t key;
+
+    friend std::ostream & operator<<(std::ostream &out, const SharedString & ref);
+  public:
+    SharedString(): key(0) {}
```
tst-lsavoie: I think there may be some issues with empty shared strings. Consider the following cases:
```
SharedString s();
s.toString(); // What does this return?
```

```
SharedString s("Not empty");
s = std::string();
assert(s.empty());  // Does this succeed?
```

The simplest fix might be to add an item to the refStore with key == 0 and ref == "".

earth_enterprise/src/common/SharedString.h
tst-sbayard: Renamed as suggested.

earth_enterprise/src/common/SharedString.h
tst-sbayard: It would be more consistent, per se. I still have a preference to only add what is necessary at this point, so I'll probably leave it as is unless it becomes a necessity.

earth_enterprise/src/common/SharedString.h
```diff
@@ -0,0 +1,125 @@
...
+// With this class, only an integer identifier is stored in each instance (thus
+// drastically reducing the amount of memory needed in large projects), and
+// that identifier is used to retrieve the string value when it is requested.
+// Since the string-to-identifier ratio is 1:1, equality comparisons are also 
+// much faster as we only need to compare the identifiers.
+class SharedString {
+protected:
+   class StringStorage {
+      private:
+        std::unordered_map<uint32_t, std::string> refFromKeyTable;
```
tst-lsavoie: For consistency we probably ought to change the names here and the names of the functions - something like `strFromKeyTable`. Basically, everything named ref should have the name changed to string or something like that.

tst-ccamp: Wouldn't you typically name these "KeyToValue" rather than "ValueFromKey"? It seems a bit awkward.

earth_enterprise/src/common/SharedString.h
```diff
@@ -0,0 +1,129 @@
...
+        std::unordered_map<std::string, uint32_t> keyFromRefTable;
+        static uint32_t nextID;
```
tst-ccamp: Just thinking, would it be better to use a hash instead of an ID? That way, you could actually check to see if the string is already in the lookup table without doing a search. I guess it depends on the performance of the hash algorithm as to which would work better.

earth_enterprise/src/common/SharedString.h
```diff
@@ -0,0 +1,129 @@
...
+      public:
+        const std::string & RefFromKey(const uint32_t &key) {
+					notify(NFY_INFO2, "SharedString::StringStorage::RefFromKey called with %u\n", key);
```
tst-ccamp: Seems like this would be better as a debug rather than an info

tst-ccamp: Also, spacing is really off on the notify statements

tst-sbayard: I think the spacing was a problem with how my tabs were set up in `vim`. But I removed these `notify` calls since they ended up being really excessive. Like, in testing a project with about 1500 imagery assets, the log file was growing at about 1GB per minute when `gesystemmanager` was doing planning.

earth_enterprise/src/common/SharedString.h
tst-lsavoie: The reasoning behind this is that the key goes next to the word key and the value goes next to the word value:
```
value = valueFromKey(key);
// As opposed to
value = keyToValue(key);
```
It reduces the mental friction of figuring out which function to use. It seems like a small thing, but in a class like this where you convert in both directions, it helps.

earth_enterprise/src/common/SharedString.h
tst-sbayard: Since `unordered_map` uses hashing internally, I'm not sure how much improvement that would buy when looking things up.

tst-lsavoie:
@tst-sbayard it might be a good idea to update getop to report the size of the stringStore alongside the size of the caches.

earth_enterprise/src/common/SharedString.h
```diff
@@ -0,0 +1,124 @@
...
+// a cached asset or asset version. Previously, each asset stored the full names
+// of all of its children, parents, listeners, and inputs, resulting in tens of
+// thousands of duplicates of each string in very large projects.
+//
+// With this class, only an integer identifier is stored in each instance (thus
+// drastically reducing the amount of memory needed in large projects), and
+// that identifier is used to retrieve the string value when it is requested.
+// Since the string-to-identifier ratio is 1:1, equality comparisons are also 
+// much faster as we only need to compare the identifiers.
+class SharedString {
```
tst-jhaight: If it is easy, it may be worth using SharedString in place of std::string for the DirtyMap typedef in `fusion/autoingest/sysman/AssetHandleD.h`.  Maybe also change to unordered_map while you're at it?

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -415,7 +327,7 @@ class MutableDerivedAssetHandleD_ : public DerivedBase_, public MutableBase_
  public:
```
tst-jhaight: For some reason it won't let me comment on line 411/323.  I _think_ that line is doing a little extra work since this->ref is a std::string and dirtyMap is SharedString now.   We can probably also replace `insert(std::make_pair` with `dirtyMap.emplace` since we have c++11 and save a little more

tst-sbayard: I haven't followed the inheritance chain back, but I assume that this inherits from AssetHandle which is where `ref` is defined. But that's been changed to a `SharedString` (unless I'm missing something), so I think there's no extra conversion there. But good point about `emplace`. I can switch he insert calls to that.

tst-jhaight: I think you're right.  I followed back to a different ref the first time which was a std::string but I think the one in AssetHandle is the correct one.  I'll leave the emplace up to you.  Probably many places could benefit from that now that it is available.  I won't be too offended if you decide to draw the line where things stand

PR #1269, opened 2019-04-11
1192 update openjpeg to 2.3.1

tst-jlarocco:


tst-jlarocco:
> Did we verify that OpenJpeg patch (src/third_party/gdal/gdal-jp2-V2.3.0.patch) we apply in build to GDAL was still applied correctly after this library upgrade ?

OpenGEE built okay and I was able to load and build JP2 files, but I didn't specifically check if the patch was applied.  Is there a way to check if the patch was applied?

tst-jlarocco:
> > Did we verify that OpenJpeg patch (src/third_party/gdal/gdal-jp2-V2.3.0.patch) we apply in build to GDAL was still applied correctly after this library upgrade ?
> 
> OpenGEE built okay and I was able to load and build JP2 files, but I didn't specifically check if the patch was applied. Is there a way to check if the patch was applied?
> 
> EDIT: I'm rebuilding now (I had removed the build directory), I'll check the configure script that the patch runs against.

Just checked GDAL's configure script and frmts/openjpeg/openjpegdataset.cpp and they both had te patch applied correctly.

tst-nfarah:
> > > Did we verify that OpenJpeg patch (src/third_party/gdal/gdal-jp2-V2.3.0.patch) we apply in build to GDAL was still applied correctly after this library upgrade ?
> > 
> > 
> > OpenGEE built okay and I was able to load and build JP2 files, but I didn't specifically check if the patch was applied. Is there a way to check if the patch was applied?
> > EDIT: I'm rebuilding now (I had removed the build directory), I'll check the configure script that the patch runs against.
> 
> Just checked GDAL's configure script and frmts/openjpeg/openjpegdataset.cpp and they both had te patch applied correctly.

yeah verify `configure` and `frmts/openjpeg/openjpegdataset.cpp` would be one way to do it. Since we are modifying this let's rename that patch to be `gdal-jp2-V2.3.0.patch`

Issue #1279, opened 2019-04-18
getversion.py returns incorrect version on Windows


tst-blaird:
I did a checkout of the release_5.3.0 branch and ran getversion.py.  It returned 5.2.5 when it should have returned 5.3.0.

tst-blaird:
Resolved by [PR 1280](https://github.com/google/earthenterprise/pull/1280).

PR #1280, opened 2019-04-18
1279 Fix getversion.py for Windows and update OpenGEE copyright year.

tst-blaird:
Address #1279

Need a different parameters for _GitTagRealCommitId() depending on which platform getversion.py is running on.  Also during Portable Server testing on Windows I found that the copyright year needed to be updated.

To verify the change, checkout this PR and run it's getversion.py on both Linux and Windows.  For this branch it should return `5.3.0`.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/local/preview/developers/terms.html
+++ b/earth_enterprise/src/fusion/portableglobe/servers/local/preview/developers/terms.html
@@ -1,6 +1,6 @@
 <!--
   Copyright 2017 Google Inc.
-  Copyright 2018 the OpenGEE Contributors.
+  Copyright 2018-2019 the OpenGEE Contributors.
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -160,7 +160,7 @@
       <p>
 	Copyright 2017 Google Inc.
     <br>
-    Copyright 2018 the OpenGEE Contributors.
+    Copyright 2018-2019 the OpenGEE Contributors.
       <p>
 	Licensed under the Apache License, Version 2.0 (the "License");
 	you may not use this file except in compliance with the License.
--- a/earth_enterprise/src/scons/getversion.py
+++ b/earth_enterprise/src/scons/getversion.py
@@ -125,10 +125,24 @@ def _GitPreviousReleaseTag(tailTagName):
     return ''
 
 
+def _GitTagRealCommitIdWindows(tagName):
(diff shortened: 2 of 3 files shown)


earth_enterprise/src/scons/getversion.py
```diff
@@ -125,10 +125,24 @@ def _GitPreviousReleaseTag(tailTagName):
...
+    return commitId
+
 def _GitTagRealCommitId(tagName):
     """use shell command to retrieve commit id of where the tag points to"""
-    # for some reason .hexsha was not returning the same id....
-    return os.popen("git rev-list -n 1 '{0}'".format(tagName.replace("'", "'\"'\"'"))).read().strip()
+    if os.name is 'nt':
+        return _GitTagRealCommitIdWindows(tagName)
+    else:
+        return _GitTagRealCommitIdLinux(tagName)
```
tst-nfarah: Would this work on MacOS ? May be it's not an issue since there seems to be some [build issues with Portable](https://github.com/google/earthenterprise/wiki/Portable-Server#portable-server-on-mac-os) in Mac

earth_enterprise/src/fusion/portableglobe/servers/local/preview/developers/terms.html
```diff
@@ -1,6 +1,6 @@
 <!--
   Copyright 2017 Google Inc.
-  Copyright 2018 the OpenGEE Contributors.
+  Copyright 2018-2019 the OpenGEE Contributors.
```
tst-nfarah: Does the addressed bug existed in previous version ? (5.2.5). If yes, let's document this fix in release notes.

tst-blaird: I did a checkout of OpenGEE branch release_5.2.5 and ran `getversion.py` on both Ubuntu and Windows.  Both reported 5.2.5.  So it appears the issue is either new to or first appeared with 5.3.0.  So no need for a release note, correct?

earth_enterprise/src/scons/getversion.py
tst-blaird: I don't have a Mac setup to be able to test this.  We have a disclaimer about MacOS running Portable Server on Mac OS not being tested so I think we're ok.

earth_enterprise/src/fusion/portableglobe/servers/local/preview/developers/terms.html
tst-nfarah: Ok, then it's introduced in 5.3.0 and fixed in 5.3.0 and we don't need to document it.

PR #1286, opened 2019-04-29
Limit cache sizes based on memory usage

tst-jkent:
Added option for limiting cache sizes based on total memory used by the objects in each cache.
Prune function is modified to accommodate for this new option.


earth_enterprise/src/common/MemoryMonitor.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "MemoryMonitor.h"
+
+MemoryMonitor* MemoryMonitor::memoryMonitor = 0;
+
+MemoryMonitor* MemoryMonitor::Instance() {
```
tst-basilhuffman: Probably better to use a Meyer's singleton:

`static MemoryMonitor& MemoryMonitor::Instance()
{
    static MemoryMonitor _instance;
    return _instance;
}`

earth_enterprise/src/common/MemoryMonitor.cpp
```diff
@@ -0,0 +1,82 @@
...
+            break;
+        }
+        tokens.clear();
+    }
+    memFile.close();
+    if (memoryVars[0] == 0) {
+        memoryUsed = 0;
+    }
+    else {
+        memoryUsed = (((float) (memoryVars[0] - memoryVars[1] - memoryVars[2] - memoryVars[3]))/memoryVars[0])*100;
```
tst-basilhuffman: might be more readable to use `std::accumulate` and safer to use `static_cast`

earth_enterprise/src/common/khCache.h
```diff
@@ -28,6 +28,10 @@
 #include <notify.h>
 #endif
 
+#include "fusion/autoingest/MiscConfig.h"
+//#include <typeinfo>
```
tst-lsavoie: Minor: remove commented out code.

earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp
```diff
@@ -109,6 +109,8 @@ khAssetManager::ApplyPending(void)
 
   notify(NFY_INFO, "Asset cache size = %d", Asset::CacheSize());
   notify(NFY_INFO, "Version cache size = %d", AssetVersion::CacheSize());
+  notify(NFY_INFO, "Asset cache file size = %lu", Asset::CacheObjectSize());
+  notify(NFY_INFO, "Version cache file size = %lu", AssetVersion::CacheObjectSize());
```
tst-sbayard: I feel like the wording here might give the wrong impression. It kind of sounds to me like there's a file on disk that represents each cache, and I believe the cache is only in-memory. I think something like `Asset cache memory footprint` or `Asset cache size in memory` along with units (bytes, presumably?) might be clearer.

earth_enterprise/src/common/khCache.h
```diff
@@ -260,7 +274,7 @@ class khCache {
   void Prune(void) {
     CheckListInvariant();
     Item *item = tail;
-    while (item && (map.size() > targetMax)) {
+    while (item && ( (cacheObjectSizes > targetMax) && limitCache )) {
```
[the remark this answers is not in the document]
tst-jkent: Yes to both. In the the constructor calls for khCache, instead of passing the value from the old options, I just use the new options.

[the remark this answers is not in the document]
tst-sbayard: Sounds good. In that case, I think `targetMax` and the return value of `size_type capacity(void) const { return targetMax; }` should be changed to be `uint64`.

tst-lsavoie:
@tst-jkent some thoughts about this - it would be great to have the option to limit the cache by number of objects OR object size. Also, khCache is used in multiple places - I would think we'd only want to use the new code for the Asset and AssetVersion classes and leave the other khCache uses the same as they are now. I'm not sure how important that is.

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -221,10 +221,14 @@ main(int argc, char *argv[])
             break;
           }
         }
+
         outline("");
         outline("Number of cached assets: %u", taskLists.num_assets_cached);
         outline("Number of cached asset versions: %u",
                 taskLists.num_assetversions_cached);
+        outline("Total size of cached assets: %lu", taskLists.asset_cache_size);
```
tst-sbayard: After running it, I found I had to carefully count the number of digits to be sure of the size (`22631488` in this case).
It might be a nice quality of life change to make this human readable by adding a function like this to convert the number of bytes to a human readable form for output (found on Google, modified slightly but not tested):

`std::string GetReadableSize(double size/*in bytes*/) {
    int i = 0;
    const char* units[] = {"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"};
    while (size > 1024) {
        size /= 1024;
        i++;
    }
    char buf[100];
    sprintf(buf, "%.*f %s", i, size, units[i]);
    return std::string(buf);
}`

earth_enterprise/src/common/khCache.h
```diff
@@ -260,7 +305,7 @@ class khCache {
   void Prune(void) {
     CheckListInvariant();
     Item *item = tail;
-    while (item && (map.size() > targetMax)) {
+    while (item && ( ((cacheObjectSizes > maxCacheMemory) && limitCacheMemory) || ((map.size() > targetMax) && !limitCacheMemory) )) {
```
tst-basilhuffman: easier to read if these clauses were separated across a couple lines

earth_enterprise/src/fusion/autoingest/Asset.h
```diff
@@ -68,6 +68,19 @@ class AssetImpl : public khRefCounter, public AssetStorage, public StorageManage
   virtual ~AssetImpl(void) { }
   const SharedString & GetRef(void) const { return name; }
 
+  const uint64 GetSize() {
```
tst-basilhuffman: same: split across multiple lines, perhaps provide comments as to what each is

earth_enterprise/src/fusion/autoingest/AssetVersion.h
```diff
@@ -114,6 +114,27 @@ class AssetVersionImpl : public khRefCounter, public AssetVersionStorage, public
     return verref.AssetRef();
   }
 
+  const uint64 GetSize() {
```
tst-basilhuffman: same as above

earth_enterprise/src/fusion/autoingest/AssetVersion.h
```diff
@@ -114,6 +114,27 @@ class AssetVersionImpl : public khRefCounter, public AssetVersionStorage, public
...
+    + sizeof(beginTime) + sizeof(progressTime) + sizeof(endTime) + sizeof(taskid));
+  }
+
+  template<class T>
+  const uint64 GetVectorSize(std::vector<T> vec) {
+    return (vec.capacity() * sizeof(T));
+  }
+  const uint64 GetVectorSize(std::vector<std::string> vec) {
+    uint64 total = 0;
+    for(uint i = 0; i < vec.size(); i++) {
```
tst-basilhuffman: use range loop

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -73,6 +73,22 @@ usage(const std::string &progn, const char *msg = 0, ...)
   exit(1);
 }
 
+std::string
+readableMemorySize(uint64 size) {
+  double readable = static_cast<double>(size);
+  const char* units[] = {"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"};
```
tst-basilhuffman: use std::array

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -73,6 +73,22 @@ usage(const std::string &progn, const char *msg = 0, ...)
...
+  int i = 0;
+  char buf[100];
+
+  while (readable > 1024) {
+    readable /= 1024;
+    i++;
+  }
+
+  sprintf(buf, "%.2f %s", readable, units[i]);
+  return std::string(buf);
```
tst-basilhuffman: try `return std::string(to_string(readable) + ' ' + units[i]);`

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -73,6 +73,22 @@ usage(const std::string &progn, const char *msg = 0, ...)
   exit(1);
 }
 
+std::string
+readableMemorySize(uint64 size) {
+  double readable = static_cast<double>(size);
+  const char* units[] = {"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"};
+  int i = 0;
```
tst-basilhuffman: `uint8`

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
tst-basilhuffman: use stringstream i.e.
`std::stringstream ss;`
`ss << readable << ' ' << units[i];`
`return ss.str();`

tst-basilhuffman:
small changes, mostly cosmetic to use more c++11 constructs, won't effect logic

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -73,6 +73,23 @@ usage(const std::string &progn, const char *msg = 0, ...)
...
+  const std::array<std::string, 9> units = {"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"};
+  uint8 i = 0;
+  std::stringstream memoryUsed;
+
+  while (readable > 1024) {
+    readable /= 1024;
+    i++;
+  }
+
+  memoryUsed << round(readable) << ' ' << units[i];
```
tst-basilhuffman: `memoryUsed << std::fixed << memoryUsed.setprecision(2) << ' ' << units[i];`

earth_enterprise/src/common/SharedString.h
```diff
@@ -79,6 +79,10 @@ class SharedString {
       return strStore.size();
     }
 
+    static uint64 GetSharedStringSize() {
+      return sizeof(key) + sizeof(strStore);
```
tst-lsavoie: I would take out `sizeof(strStore)` - since it's static there's only one of them, so it doesn't really affect the size of individual strings.

earth_enterprise/src/common/khCache.h
```diff
@@ -162,24 +171,59 @@ class khCache {
  public:
   typedef typename MapType::size_type size_type;
   size_type size(void) const { return map.size(); }
-  size_type capacity(void) const { return targetMax; }
+  uint64 capacity(void) const { return targetMax; }
+  uint64 getMemoryUse(void) const { return cacheObjectSizes; }
+  uint64 itemSizeMinusKeyAndValue(void) const { return sizeof(khCacheItem<Key, Value>) - sizeof(Key) - sizeof(Value); }
```
tst-lsavoie: Minor: this could be a constant instead of a function, which might be a bit faster.

earth_enterprise/src/common/khCache.h
```diff
@@ -162,24 +171,59 @@ class khCache {
...
+  }
+  uint64 getKeyOrValueSize(const SharedString shstr) {
+    return shstr.GetSharedStringSize();
+  }
+  template<class T>
+  uint64 getKeyOrValueSize(const khRefGuard<T> obj) {
+    return (sizeof(obj) + obj.getRefGuardSize());
+  }
+  void updateObjectSize(const Key &key) {
+    if ( FindItem(key) != 0 ) {
```
tst-lsavoie: Minor: 2 FindItem calls is redundant - you can move the call outside the if statement.

earth_enterprise/src/common/khCache.h
```diff
@@ -162,24 +171,59 @@ class khCache {
...
+      Item *item = FindItem(key);
+      uint64 size = khCacheItemSize + getKeyOrValueSize(item->key) + getKeyOrValueSize(item->val);
+      if (size > item->size) {
+        cacheObjectSizes += (size - item->size);
```
tst-lsavoie: I'm pretty sure that both of these cases simplify to `cacheObjectSizes = cacheObjectSizes - item->size + size`.

earth_enterprise/src/common/khCache.h
```diff
@@ -162,24 +171,59 @@ class khCache {
...
 #ifdef SUPPORT_VERBOSE
           , bool verbose_ = false
 #endif
           ) : head(0), tail(0), targetMax(targetMax_),
 #ifdef SUPPORT_VERBOSE
               verbose(verbose_),
 #endif
-              numItems(0)
+              numItems(0), cacheObjectSizes(0),
+              limitCacheMemory(0),
```
tst-lsavoie: Minor: should be `false` instead of `0`

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -45,22 +45,25 @@ class StorageManager
...
-        assetType(type) {}
+        assetType(type) { SetCacheMemoryLimit(enabled, maxMemory); }
     ~StorageManager() = default;
 
     inline uint32 CacheSize() const { return cache.size(); }
-    inline uint32 CacheCapacity() const { return cache.capacity(); }
+    inline uint64 CacheCapacity() const { return cache.capacity(); }
     inline uint32 DirtySize() const { return dirtyMap.size(); }
+    inline uint64 CacheMemoryUse() const { return cache.getMemoryUse(); }
+    inline void SetCacheMemoryLimit(bool enabled, uint64 maxMemory) { cache.setCacheMemoryLimit(enabled, maxMemory); }
```
tst-lsavoie: Minor: I don't know that it makes sense for this function to be public, but perhaps it's necessary for testing.

earth_enterprise/src/fusion/autoingest/StorageManager_unittest.cpp
```diff
@@ -77,7 +78,7 @@ class StorageManagerTest : public testing::Test {
  protected:
   StorageManager<TestItem> storageManager;
  public:
-  StorageManagerTest() : storageManager(CACHE_SIZE, "test") {}
+  StorageManagerTest() : storageManager(CACHE_SIZE, 0, MEMORY_LIMIT, "test") {}
```
tst-lsavoie: Minor: false instead of 0

earth_enterprise/src/common/khCache.h
```diff
@@ -260,7 +305,9 @@ class khCache {
   void Prune(void) {
     CheckListInvariant();
     Item *item = tail;
-    while (item && (map.size() > targetMax)) {
+    while (item && (
+    ( (cacheObjectSizes > maxCacheMemory) && limitCacheMemory ) ||
+    ( (map.size() > targetMax) && !limitCacheMemory ) )) {
```
tst-lsavoie: I'm confused here - you're removing objects from the cache but I don't see you updating the cache size. I'm afraid this might be an infinite loop. It seems you would need to update the cache size every time you add or remove an object.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-jkent: SetCacheMemoryLimit is used by the PurgeCacheBasedOnMemoryUtilization unit test to verify that objects are pruned from the cache if the total memory used by the cache exceeds the set limit.

earth_enterprise/src/common/khCache.h
tst-jkent: cacheObjectSizes is updated whenever the deconstructor for MutableAssetHandleD or MutableDerivedAssetHandleD is called. I've tested to verify that when objects are added to the cache, the initial size of the object is calculated and cacheObjectSizes is then updated. Also, when objects are pruned, cacheObjectSizes is updated.

tst-jkent: Okay. Just ran some more specific tests and I see what you mean. Working to fix this now.

tst-lsavoie:
Our conversation reminded me - we need to update the docs and the release notes.

earth_enterprise/src/fusion/autoingest/StorageManager_unittest.cpp
```diff
@@ -170,6 +179,18 @@ TEST_F(StorageManagerTest, PurgeCache) {
   }
 }
 
+TEST_F(StorageManagerTest, PurgeCacheBasedOnMemoryUtilization) {
+  storageManager.SetCacheMemoryLimit(1, MEMORY_LIMIT);
```
tst-lsavoie: `true` instead of `1`

earth_enterprise/src/fusion/autoingest/StorageManager_unittest.cpp
```diff
@@ -170,6 +179,18 @@ TEST_F(StorageManagerTest, PurgeCache) {
...
+  size_t i;
+  for(i = 0; i < CACHE_SIZE + 2; ++i) {
+    stringstream s;
+    s << "asset" << i;
+    Get<TestHandle>(storageManager, s.str(), false, true, false);
+    ASSERT_EQ(storageManager.DirtySize(), 0) << "Storage manager has unexpected item in dirty map";
+  }
+  ASSERT_LE(storageManager.CacheSize(), i+1) << "Unexpected number of items in cache";
```
tst-lsavoie: If the memory limit is 0 shouldn't the cache size also be 0?

earth_enterprise/src/fusion/autoingest/StorageManager_unittest.cpp
```diff
@@ -170,6 +179,18 @@ TEST_F(StorageManagerTest, PurgeCache) {
...
+}
+
```
tst-lsavoie: Thinking about some other tests we could add:
- Memory limit that is greater than 0
- Check that the value returned from StorageManager::CacheMemoryUse makes sense

docs/geedocs/5.2.5/answer/176738.html
```diff
@@ -256,7 +256,8 @@ <h2 id="cache_configuration">Cache configuration</h2>
...
 
-Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
+Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>
+<p>Also included within <code>misc.xml</code> is an experimental option, which allows the user to limit the caches by memory usage instead of number of objects. This option is controlled by <code>LimitMemoryUtilization</code> and its default value is false. The maximum memory usage for each cache can be set with <code>MaxAssetCacheMemorySize</code> and <code>MaxVersionCacheMemorySize</code>. This size is measured in bytes and the default setting is 1GB. While this option is enabled, the caches will not be limited by the number of objects in the cache.</p>
```
tst-lsavoie: This documentation really needs to go into the 5.3.1 version. I haven't figured out how to handle that yet.

earth_enterprise/src/common/khCache.h
```diff
@@ -260,7 +304,9 @@ class khCache {
   void Prune(void) {
     CheckListInvariant();
     Item *item = tail;
-    while (item && (map.size() > targetMax)) {
+    while (item && (
```
tst-lsavoie: Minor: you could make this a bit clearer by moving the boolean logic here into a function, something like
```
while (item && cacheIsTooBig())
```
Then you can make the function as expressive as you like:
```
bool cacheIsTooBig() {
  if (limitCacheMemory) return cacheMemoryUse > maxCacheMemory;
  else return map.size() > targetMax;
}
```
or something like that.

docs/geedocs/5.2.5/answer/176738.html
```diff
@@ -256,7 +256,7 @@ <h2 id="cache_configuration">Cache configuration</h2>
 The asset cache size is controlled by <code>AssetCacheSize</code> and the asset version cache size is controlled by <code>VersionCacheSize</code>.
 Note that the asset version cache is generally larger than the asset cache because it may contain multiple versions of the same asset.</p>
 
-Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
+Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>
```
tst-lsavoie: Need to remove this here since there's a separate PR.

tst-jkent: This is what was originally in the 5.2.5 doc file, this is not the new addition. It appears I removed a period at the end of the original by mistake.

earth_enterprise/src/common/khRefCounter.h
```diff
@@ -84,7 +84,7 @@ class khRefGuard {
 
   // expose refcount function from my shared object
   inline uint32 refcount(void) const  { return ptr ? ptr->refcount() : 0; }
-
+  inline uint64 getRefGuardSize(void) const { return ptr->GetSize(); }
```
tst-lsavoie: Minor: could include the size of the pointer itself: `return sizeof(ptr) + ptr->GetSize()`. Won't make a huge difference, though.

earth_enterprise/src/common/khCache.h
```diff
@@ -172,14 +186,51 @@ class khCache {
...
   }
+  template<class T>
+  uint64 getKeyOrValueSize(const T obj) {
+    return sizeof(obj);
+  }
+  uint64 getKeyOrValueSize(const std::string str) {
+    return (sizeof(str) + (str.capacity() * sizeof(char)));
+  }
+  uint64 getKeyOrValueSize(const SharedString shstr) {
+    return shstr.GetSharedStringSize();
```
tst-lsavoie: Minor: shared strings don't include a pointer, so for this you can just `return sizeof(shstr)`. It won't make a huge difference (maybe no difference at all), but it might be a little more accurate.

earth_enterprise/src/fusion/autoingest/AssetVersion.h
```diff
@@ -113,6 +113,47 @@ class AssetVersionImpl : public khRefCounter, public AssetVersionStorage, public
...
+    + sizeof(beginTime)
+    + sizeof(progressTime)
+    + sizeof(endTime)
+    + sizeof(taskid)
+    + sizeof(timestamp)
+    + sizeof(filesize));
+  }
+
+// Get the total memory used by a template vector
+  template<class T>
```
tst-lsavoie: It would be nice to put these in a common location so we don't have to duplicate the code.

tst-lsavoie:
The only big thing I can think of that might be missing here is code to calculate the sizes of classes that are derived from AssetImpl or AssetVersionImpl. For example, one inheritance chain goes AssetVersionStorage -> AssetVersionImpl -> AssetVersionImplD -> CompositeAssetVersionImplD -> RasterProjectAssetVersionImplD. We're only calculating size at the AssetVersionImpl level and below, but the other classes can add fields and we aren't accounting for those right now.

earth_enterprise/src/common/khRefCounter.h
tst-jkent: Was calculating this but moved the sizeof(ptr) call into this function for more clarity.

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "khTypes.h"
+#include "khRefCounter.h"
+#include "khMetaData.h"
+#include <notify.h>
+
+#ifndef _INCL_GUARD
```
tst-lsavoie: We should name our include guards something specific to the file, like CACHE_SIZE_CALCULATIONS_H.

Also, I prefer to have everything (including the other include statements) inside the include guard. It won't change anything except possibly make compilation a bit faster.

earth_enterprise/src/common/khCache.h
```diff
@@ -22,6 +22,10 @@
 #include <vector>
 
 #include "khTypes.h"
+#include "fusion/autoingest/MiscConfig.h"
```
tst-lsavoie: I don't think you need the MiscConfig, khRefCounter, or SharedString includes any more.

earth_enterprise/src/common/khRefCounter.h
```diff
@@ -84,7 +84,8 @@ class khRefGuard {
 
   // expose refcount function from my shared object
   inline uint32 refcount(void) const  { return ptr ? ptr->refcount() : 0; }
-
+  inline uint64 getPtrSize(void) const { return ptr->GetSize(); }
+  //inline T* getPtr(void) const { return ptr; }
```
tst-lsavoie: Remove commented out code.

earth_enterprise/src/common/khRefCounter.h
```diff
@@ -84,7 +84,8 @@ class khRefGuard {
 
   // expose refcount function from my shared object
   inline uint32 refcount(void) const  { return ptr ? ptr->refcount() : 0; }
-
+  inline uint64 getPtrSize(void) const { return ptr->GetSize(); }
```
tst-lsavoie: You need to check that `ptr` is not null before dereferencing it here.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -45,22 +45,26 @@ class StorageManager
     using HandleType = khRefGuard<AssetType>;
     using AssetKey = SharedString;
 
-    StorageManager(uint cacheSize, const std::string & type) :
+    StorageManager(uint cacheSize, bool enabled, uint64 maxMemory, const std::string & type) :
```
tst-lsavoie: Minor: I'd prefer a more specific name than `enabled` here - something like `limitCacheByMemory`.

earth_enterprise/src/fusion/autoingest/tools/SConscript
```diff
@@ -22,7 +22,7 @@ Import('env')
 # ***  apps that are NOT installed
 # ****************************************************************************
 env.executable('gefindsrconvol', ['gefindsrconvol.cpp'],
-               LIBS=['geautoingest', 'gecommon', 'gemiscconfig', 'geutil'])
+               LIBS=['geautoingest', 'gecommon', 'gemiscconfig', 'geutil', 'qt-mt'])
```
tst-lsavoie: Any reason for this change? This doesn't seem related to the cache stuff.

earth_enterprise/src/fusion/rasterfuse/SConscript
```diff
@@ -31,7 +31,7 @@ gepackgen = env.executable('gepackgen',
                            LIBS=['geautoingest', 'khrasterffio', 'gegdal',
                                  'geraster', 'khffio', 'gefconfigutil',
                                  'geutil', 'gecommon', 'geprotobuf',
-                                 'qt-mt', 'geos'])
+                                 'qt-mt', 'geos', 'gemiscconfig'])
```
tst-lsavoie: I don't think this will be necessary if you remove MiscConfig.h from khCache, but I could be wrong.

earth_enterprise/src/fusion/rasterfuse/SConscript
```diff
@@ -41,7 +41,7 @@ pyr_info = env.executable('pyr_info', ['pyrinfo.cpp', '../tools/geImageWriter.o'
 # ***** They share a .cpp file and SCons whines if they override the LIBS
 # ***** themselves.
 env = env.Clone();
-env.Append(LIBS=['geraster', 'gegdal', 'gegdalutil', 'gdal', 'geutil', 'gecommon', 'qt-mt', 'geos'])
+env.Append(LIBS=['geraster', 'gegdal', 'gegdalutil', 'gdal', 'geutil', 'gecommon', 'qt-mt', 'geos', 'gemiscconfig'])
```
tst-lsavoie: Same: this may no longer be necessary.

earth_enterprise/src/fusion/autoingest/tools/SConscript
tst-jkent: This was added due to some compile errors I encountered when calculating the size of QStrings in CacheSizeCalculations.

tst-basilhuffman:
minor comments, mostly performance related

PR #1289, opened 2019-05-02
WIP Experimental branch merging the SharedString optimizations into the parallelized inset traversals

husf-dsheremata:


earth_enterprise/integration_test/specs/state_propagation.spec
```diff
@@ -379,7 +379,7 @@ Clean resource after build
 * Build imagery resource "USGSLanSat_MultiProject"
 * Build imagery project "StatePropagationTest_MultiProject1"
 * Verify that the state of imagery project "StatePropagationTest_MultiProject1" is "Waiting"
-* Verify that the state of imagery project "StatePropagationTest_MultiProject2" is "Succeeded"
+* Verify that the state of imagery project "StatePropagationTest_Database" is "Succeeded"
```
tst-lsavoie: Looks like this change was not intended to be permanent.

earth_enterprise/src/common/khCache.h
```diff
@@ -280,6 +287,22 @@ class khCache {
     }
     CheckListInvariant();
   }
+
```
tst-lsavoie: This function can be deleted. It was apparently deleted on one branch but not the other.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -119,6 +123,44 @@ class khMTQueue
 };
 
 
+class khRWLock
+{
+  private:
+    pthread_rwlock_t rwlock;
+  public:
+    khRWLock() : rwlock(PTHREAD_RWLOCK_INITIALIZER) { }
```
tst-lsavoie: I heard that `PTHREAD_RWLOCK_INITIALIZER` is only supposed to be used for statically allocated objects. It might be safer to call `pthread_rwlock_init` (or whatever the function name is) inside the constructor.

earth_enterprise/src/fusion/autoingest/AssetVersionRef.h
```diff
@@ -33,6 +33,8 @@ class AssetVersionRef {
  public:
   AssetVersionRef(const std::string &ref_);
   AssetVersionRef(const SharedString &ref_);
+  AssetVersionRef(const SharedString &ref_, uint vernum);
+  AssetVersionRef(const SharedString &ref,  const std::string &version_);
```
tst-lsavoie: Probably ought to delete this line since the implementation is commented out.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -142,28 +143,117 @@ class MutableAssetHandleD_ : public virtual Base_ {
...
-  typedef std::map<SharedString, Base> DirtyMap;
+  typedef khMTMap<SharedString, Base> DirtyMap;
   static DirtyMap dirtyMap;
 
   // Test whether an asset is a project asset version.
   static inline bool IsProjectAssetVersion(const std::string& assetName) {
     return assetName.find(kProjectAssetVersionNumPrefix) != std::string::npos;
   }
 
+  // Purge the cache if needed and keep recent "toKeep" items.
```
tst-lsavoie: This function should also be deleted.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +190,168 @@ class khMTSet
...
+    const T& operator[] (const size_t idx) const {
+      khReadGuard lock(mtx);
+      return Base::operator[] (idx); 
+    }
+
+    bool operator==(const MTVector& x) const {
+      // Since using a read lock, no need to worry about lock ordering or things like a==a
+      khReadGuard lock1(mtx);
+      khReadGuard lock2(x.mtx);
+      return std::equal(Base::begin(), Base::end(), x.begin());
```
tst-sbayard: Minor: I would take out `Base::` for the `begin()` and `end()` calls here. They're not necessary, and in the (albeit unlikely) event that `begin` and `end` were overridden in MTVector, this would have unexpected results.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +190,168 @@ class khMTSet
...
+    }
+
```
tst-sbayard: Seems like the other comparison operators would need to be overridden if we're applying read locks during comparisons. Also, wouldn't the assignment operator need to be overridden and use a write lock?

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +190,168 @@ class khMTSet
...
+      return Base::operator==(x);
+    }
+
+    void clear(void) {
+      khWriteGuard lock(mtx);
+      Base::clear();
+    }
+
+    size_t size(void) const {
+      khWriteGuard lock(mtx);
```
tst-sbayard: I think this should be `khReadGuard`.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +190,168 @@ class khMTSet
...
+      return Base::size();
+    }
+
+    bool empty(void) const {
+      khWriteGuard lock(mtx);
```
tst-sbayard: Same here, should be `khReadGuard`.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +190,168 @@ class khMTSet
...
+    void shrink_to_fit(void) {
+      khWriteGuard lock(mtx);
+      Base::shrink_to_fit();
+    }
+    
+    
+    void doForEach(std::function<void (const T&)> func) const {
+      if (!func) return; // can't do anyting with an invalid function
+      khReadGuard lock(mtx);
+      for_each(Base::begin(), Base::end(), func);
```
tst-sbayard: Minor: don't need to specify `Base::` for `begin()` and `end()` here.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +190,168 @@ class khMTSet
...
+    }
+
+    bool doForEachUntil(std::function<bool (const T&)> func) const {
+      if (!func) return false; // can't do anyting with an invalid function
+      khReadGuard lock(mtx);
+      bool loopedThroughAll = true;
+      for (uint i = 0; i < Base::size(); ++i) {
+        if (!func(operator [] (i))) {
```
tst-sbayard: I'm not clear on the need to iterate via indices instead of an iterator. Also, calling `operator[]` here is going to attempt to apply another read lock. It should work, but that's a little bit of additional overhead for each access.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -148,9 +190,168 @@ class khMTSet
...
+          loopedThroughAll = false;
+          break;
+        }
+      }
+      return loopedThroughAll;
+    }
 
+// The remaining methods are bad news but are currently needed to compile
```
tst-sbayard: It seems like the 4 operators below are using the dereferencing operator backward since `this` is a pointer.

tst-sbayard (changes requested):
The comments included in this review only cover khCache.h. Trying to give intermediate feedback since so many files are touched.

earth_enterprise/src/common/khCache.h
tst-dsheremata: Which was that other branch?  I'm still seeing it on  release_5.2.5.1_experimental

earth_enterprise/src/common/khMTTypes.h
tst-dsheremata: It turns out that this was required for the template compilation to work with the new subclass.  I've been unable to find any workaround.

earth_enterprise/src/common/khCache.h
```diff
@@ -92,41 +91,7 @@ class khCache {
     }
     return false;
   }
-  void CheckListInvariant(void) {
```
tst-lsavoie: We should arguably leave this function in since we might want to enable it for testing at some point.

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -24,6 +24,7 @@
 #include <memory>
 #include <set>
 #include <unordered_map>
+//#include <map>
```
tst-lsavoie: Remove commented out code

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -203,6 +204,7 @@ template <class K, class V>
 class khMTMap
 {
  protected:
+  //typedef std::unordered_map<K,V> Map;
```
tst-lsavoie: Remove commented out code

earth_enterprise/src/common/khMTTypes.h
```diff
@@ -258,88 +260,85 @@ class khMTMap
 // WARNING: Use this with extreme caution
 // It protects some obvious problems but creates some less obvious ones
 template <typename T>
-class MTVector {
+class MTVector : public std::vector<T>{
```
tst-lsavoie: Inheriting from std::vector makes me crazy nervous. We've already discussed that, though.

earth_enterprise/src/fusion/autoingest/AssetVersion.h
```diff
@@ -106,13 +106,13 @@ class AssetVersionImpl : public khMTRefCounter, public AssetVersionStorage {
...
 
-  std::string GetRef(void) const { return name; }
+  const SharedString & GetRef(void) const { return name; }
   std::string GetAssetRef(void) const {
     AssetVersionRef verref(name);
     return verref.AssetRef();
   }
 
-  template <class outIter>
+  template <class outIter>
```
tst-lsavoie: Minor: might as well get rid of that extra space at the end of the line.

earth_enterprise/src/fusion/autoingest/plugins.src/MapLayer.src
```diff
@@ -33,7 +33,7 @@ std::vector<std::string> MapLayerAssetVersionImpl::GetPOIFiles(void) const {
   std::vector<std::string> results;
   if (children.size() > 2) {
     int closureCount = 0;
-    auto theClosure = [&results, &closureCount](const std::string& v)
+    auto collectPOIOutputFiles = [&results, &closureCount](const std::string& v)
```
tst-lsavoie: Good call.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -20,9 +20,11 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <utility>
```
tst-lsavoie: Is this needed? We should remove it if it's not used.

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -44,19 +44,16 @@ AssetVersionImplD::StateChangeNotifier::GetNotifier(
...
 void
-AssetVersionImplD::StateChangeNotifier::AddParentsToNotify(const std::vector<std::string> & parents) {
+AssetVersionImplD::StateChangeNotifier::AddParentsToNotify(const std::vector<SharedString> & parents) {
+   //std::copy(parents.begin(), parents.end(), std::inserter(parentsToNotify, parentsToNotify.end()));
   for (const auto& p : parents)
     parentsToNotify.insert(p);
-  if (!MiscConfig::Instance().ConsolidateListenerNotifications) {
-    // If the user has opted not to consolidate notifications, send
-    // notifications immediately
     NotifyParents(nullptr);
```
tst-lsavoie: This line should be removed. It was only called if someone disabled the consolidate listener notifications optimization, but that is the default behavior now. With the current code we don't want to notify parents right away - we want to wait until we know all the parents to notify.

earth_enterprise/src/fusion/autoingest/sysman/AssetVersionD.cpp
```diff
@@ -116,22 +107,23 @@ AssetVersionImplD::StateChangeNotifier::NotifyListeners(
     std::shared_ptr<StateChangeNotifier> notifier) {
   notify(NFY_VERBOSE, "Iterate through listeners");
   int i = 1;
-  std::vector<std::string> theRefs = listenersToNotify.keys();
-  for (const std::string& ref : theRefs) {
+  //for (const std::pair<SharedString, InputStates> & elem : listenersToNotify) {
+    for (const SharedString & ref : listenersToNotify.keys()) {
```
tst-lsavoie: Minor: indentation and remove commented out code.

tst-lsavoie (changes requested):
I just did a quick visual review and this looks reasonable in general. There is one thing that needs to be fixed (the call to NotifyParents that needs to be removed), but other than that, I think this is fine.

earth_enterprise/src/common/khMTTypes.h
tst-sbayard: Just for posterity, I'll register my agreement with @tst-lsavoie here and reiterate that the risks involved with this were a topic of a fair bit of discussion. However, since this code is considered experimental, with the understanding that anything that would be merged into master would need a more robust implementation, I'm okay with having it implemented this way in this branch.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -290,12 +290,20 @@ ()
...
+    MTVector<SharedString> inputarg = $actualinputarg;
     // bind my input versions refs
-    std::vector<std::string> boundInputs;
-    boundInputs.reserve(inputarg.size());
-    std::transform(inputarg.begin(), inputarg.end(), back_inserter(boundInputs),
-                   ptr_fun(&AssetVersionRef::Bind));
+    // Make a copy using the copy constructor
+    MTVector<SharedString> boundInputs(inputarg);
+
+//TODO - double check that this works
```
tst-sbayard: Minor: could remove these comments, and the ones a few lines below, if they're not needed.

tst-sbayard (approved):
As an experimental branch, I think this is good.

earth_enterprise/src/common/khCache.h
tst-dsheremata: Ah - got it, just remove the invocations, not the implementation.

earth_enterprise/src/common/khMTTypes.h
tst-dsheremata: Also for posterity, here are some of the rationales that I considered before using this approach: 
1) Stroupstrup himself actually  subclassess std::vectors in    [https://isocpp.org/files/papers/4-Tour-Algo-draft.pdf](A Tour of C++: Containers and Algorithms) 
2) A prominent (but not prevalent) argument that it's OK so long as no state is maintained in the subclass.  This is the only way to address to the lack no virtual destructors in the std container classes - so we're good here.

tst-dsheremata: (also, makes me crazy nervous too!)

earth_enterprise/src/common/khMTTypes.h
tst-dsheremata: Good call - done.

tst-lsavoie (approved):
My comments have been addressed. I'm happy with the state of the code at this point.

PR #1295, opened 2019-05-13
Storage Manager

tst-lsavoie:
This PR starts on refactoring the code such that all the code related to storing and retrieving items from disk is in the same location. It creates a StorageManager that is responsible for reading and writing XML, caching, and keeping track of modified assets. In the future this could be used to more easily change the storage backend to something other than XML files or to implement per-asset locking.


tst-lsavoie:
Will create an updated PR shortly.

PR #1309, opened 2019-05-17
Issue 1308: Purge XML by default

### LINKED ISSUE #1308: Memory leak causes gesystemmanager to be killed by kernel
This is for the release_5.3.0 branch.
To reproduce:
1. Add imagery one at a time to the same database (build a resource, add to a single project, and rebuild the database each time).
1. As assets are building, the memory usage of `gesystemmanager` continuously increases and does not decrease.

tst-lsavoie:
This makes XML purging the default behavior. The end result is that 5.3.0 starts with higher memory utilization than 5.2.5, but it's memory usage remains steady, whereas in 5.2.5 memory usage climbs continually.

Fixes #1308.

### DIFF
--- a/docs/geedocs/5.3.0/answer/176738.html
+++ b/docs/geedocs/5.3.0/answer/176738.html
@@ -280,16 +280,12 @@ MAX_HEAP_SIZE=&lt;value&gt;
 BLOCK_SIZE=&lt;value&gt;
 <br><br>
 # whether or not to purge the XML cache; valid values are 0 or 1<br>
-# default is 0<br>
-PURGE=0
+# default is 1<br>
+PURGE=1
 <br><br>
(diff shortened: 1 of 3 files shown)


tst-lsavoie:
@tst-nfarah since this bug was introduced in 5.3.0 I didn't think it was needed in the release notes.

tst-nfarah:
@tst-lsavoie good point. You're right we don't need to document it.

> @tst-nfarah since this bug was introduced in 5.3.0 I didn't think it was needed in the release notes.

tst-rwildes:
> We need this added to list of fixed issues documented in release notes.

I think this issue was introduced in 5.3.0 so no need to document in release notes.  The feature is in the release notes and @tst-lsavoie  updated the documentation regarding the feature to reflect the changes.  I think that is good enough.

PR #1321, opened 2019-05-23
Speed up parallel buildindex

tst-sbayard:
- Minor tweaks to speed up processing by avoiding the need for a critical section in the main loop in `BuildIndex`.
- One debug message was using 1-based index and all others used 0-based. Changed all to 1-based.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
@@ -1170,7 +1170,7 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
   }
 
   // Populate packgenkids, packgenassets, and packgenvers from prelim_packgenvers
-  for(auto & packgen : prelim_packgenvers) {
+  for(const auto & packgen : prelim_packgenvers) {
     if( !packgen->GetRef().empty() ) {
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1157,9 +1165,13 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
...
 
-#pragma omp critical 
-    {
-      // remember this new PacketGen version
+    // remember this new PacketGen version
+    prelim_packgenvers[i] = packgen;
+  }
+
+  // Populate packgenkids, packgenassets, and packgenvers from prelim_packgenvers
+  for(auto & packgen : prelim_packgenvers) {
```
tst-lsavoie: I'm just thinking out loud; there's no need to take any action, but there's got to be some algorithm for combining vectors in parallel.

tst-basilhuffman: extremely minor: `const auto & packgen`

tst-sbayard: Good point and easy to fix. Should be updated now.

PR #1256, opened 2019-04-01
#1255 - Limits internal version string to 3 parts to fix patch releases

tst-ccamp:
Fixes #1255 

Internally, Fusion will only track 3 parts of the version string in order to maintain compatibility with the asset root and with client/server messages.

Verification steps
1. Install 5.2.5
1. Do build
1. Perform an upgrade to 5.2.5.2

Issue #1255, opened 2019-04-01
Version errors with GEE patch releases


tst-ccamp:
New asset roots are created with the 3-digit version
Causes error:
```
[2019-03-28 21:57:14] Fatal: This machine is running fusion software version 5.2.5.2.
/gevol/assets is configured to use version 5.2.5.
Unable to proceed.
[centos@geep-montenegro log]$ sudo /opt/google/bin/geupgradeassetroot 
No upgrade needed.
```

If you fix version on asset root by updating it to 5.2.5.2, then the gefusion startup script doesn't start:
```
Mar 28 22:05:08 geep-montenegro.partners.t-sciences.com gefusion[9210]: This machine is running fusion software version 5.2.5.
Mar 28 22:05:08 geep-montenegro.partners.t-sciences.com gefusion[9210]: /gevol/assets is configured to use version 5.2.5.2
Mar 28 22:05:08 geep-montenegro.partners.t-sciences.com gefusion[9210]: .
Mar 28 22:05:08 geep-montenegro.partners.t-sciences.com gefusion[9210]: In order to use /gevol/assets from this machine, you must upgrade
Mar 28 22:05:08 geep-montenegro.partners.t-sciences.com gefusion[9210]: the fusion software on this machine to version 5.2.5.2
```

Disabling the check_configuration call in /etc/init.d/gefusion allows the service to start, but now there's a version mismatch in the client tools:
```
$ getop
Fusion Fatal:    Invalid protocol version (client/server mismatch)
```
Notes:

* I suspect this has to do with parts of the code, like gesystemmanager using a 4-digit version in its check instead of the proper 3-digit version

PR #1324, opened 2019-05-26
Issue #1255 - Limits internal version string to 3 parts to fix variou…

shu-hung:
…s errors with patch releases (#1256) thanks


tst-ccamp:
We don't want to put this into 5.3.0 because it wasn't included with testing that version. If there is going to be a release_5.3.0.1 branch, that's where it would go. It's been merged to master.

PR #1330, opened 2019-06-03
refactor overlap calculation code

tst-basilhuffman:
Initial commit, will build and cancel sample imagery but integration tests fail


docs/geedocs/5.2.5/answer/176738.html
```diff
@@ -257,6 +257,10 @@ <h2 id="cache_configuration">Cache configuration</h2>
 Note that the asset version cache is generally larger than the asset cache because it may contain multiple versions of the same asset.</p>
 
 Also included within <code>misc.xml</code> is an experimental option, which allows for the user to disable cache purging while projects build. This is controlled by <code>DisablePacketLevelVersionCachePurge</code> and its default value is <code>false</code>.
+
+<h2 id="overlap_calculation">Overlap Calculation</h2>
```
tst-lsavoie: Thanks for updating the docs. These changes are going to end up in 5.3.1, so that's where the documentation should go, not 5.2.5. The only way to do that right now is to have a separate PR for the docs.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -996,28 +989,30 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
...
+      extraArg.genInfo.push_back(&genInfo[idx]);
+    }
 
+      // reuse or make a new PacketGenAssetVersion
+      MutableAssetVersionD packgen =
+        PacketGenFactory::ReuseOrMakeAndUpdate(genInfo[i].assetname,
+                                               type,
+                                               khMetaData(),
+                                               packetGenConfig,
+                                               extraArg,
```
tst-lsavoie: Not a huge deal, but I think the idea behind `extraArg` is to include any extra arguments that are specific to this type of asset. I think it would be cleaner to move these two new arguments into extraArg. That way you wouldn't have to change the function signature.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1038,6 +1035,25 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
   packgenkids.reserve(numInsets);
   packgenassets.reserve(numInsets);
   packgenvers.reserve(numInsets);
+  std::vector<uint> totalNeededIndexes;
+
+// calculate total overlap before entering here
```
tst-lsavoie: Looks like this is debug code that was not intended to be permanent.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.cpp
```diff
@@ -207,6 +207,12 @@ void CalcPacketGenInfo(const khTilespace            &tilespace,
 
   // populate list - inital coverage estimate from inset degExtents
   genInfos.reserve(insetInfos.size());
+  /*
```
tst-lsavoie: Looks like this was not intended to be permanent.

tst-lsavoie:
I put a little thought into this and I think this is pretty close to what we need. The overall idea is that we want to make it easy to swap out overlap algorithms. Separating the overlap calls into a function is a step towards accomplishing that. My original thought was that calculating the overlap all at once would be the most generic method since we can simply replace that all-at-once calculation with a different all at once calculation. After looking at this I think there's a simpler way.

In general, this is what needs to happen:

1. Perform some general calculations to get partial overlap data (this happens in BuildIndex).
1. Pass this partial data down to UpdateChildren.
1. Calculate the exact overlap for each level in UpdateChildren.

I think you can accomplish this with just a couple of function name changes. We could call the function called from BuildIndex something like `PreprocessOverlapForInset` and the function called from UpdateChildren something like `GetOverlapForLevel`. For this version of the code they would just call the same CalculateOverlap function that you've already written, but when we start working on overlap calculations they could do different things. That also eliminates the need for a flag since the calculations will be the same for this version, they'll just be organized differently.

To take it a step farther you could create an overlap class and put the two functions into it, then just pass the whole object from BuildIndex to UpdateChildren.

In summary, here's what I think needs to be done to finish up this ticket:

1. Put the two overlap calculation calls inside wrapper functions that describe what they're doing (as I described above). As I mentioned, you could create a class for this.
1. Remove the changes to the function signature for `ReuseOrMakeAndUpdate`.
1. Remove the new flag in MiscConfig.
1. Remove debugging code as noted in comments.
1. Move the doc changes to a separate PR so they can be in the 5.3.1 docs.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -85,9 +88,11 @@ PacketGenFactory::ReuseOrMakeAndUpdateSubAsset(
 
 template <typename ProductAssetVersion>
 void PacketGenAssetVersionImplD::UpdateChildren(
-    const PacketGenExtraUpdateArg<ProductAssetVersion> &extra) {
+    const PacketGenExtraUpdateArg<ProductAssetVersion> &extra,
+    bool skip, const std::vector<uint>& _neededIndexes) {
```
tst-sbayard: I feel like `skip` may be ambiguous for anyone looking at the function signature. Suggest renaming here and other uses to something that makes it clear what is being skipped.

tst-basilhuffman: this is getting removed

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -122,21 +124,23 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
+
+    // there are situations where this can happen and must be accounted for
+    if (calc == nullptr)
+    {
+        calc.reset(new OverlapCalculator<ProductAssetVersion>(env));
+        calc->PreprocessForInset();
     }
 
+    neededIndexes = std::move(calc->GetOverlapForLevel());
+    //neededIndexes = std::move(CalculateOverlap(env));
```
tst-sbayard: Should remove this commented-out code.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1057,19 +1054,15 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
...
-    }
+    overlapEnvelope<RasterProductAssetVersion> env (
+        type, genInfo[i].coverage, insetInfoPtrs, i,
+        beginMinifyLevel, endMinifyLevel);
+
+    std::shared_ptr<OverlapCalculator<RasterProductAssetVersion>> calc =
+        std::make_shared<OverlapCalculator<RasterProductAssetVersion>>(env);
+    calc->PreprocessForInset();
+    neededIndexes = std::move(calc->GetOverlapForLevel());
+    //neededIndexes = std::move(CalculateOverlap(env));
```
tst-sbayard: Remove commented-out code.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.h
```diff
@@ -168,6 +168,55 @@ void CalcPacketGenInfo(const khTilespace            &tilespace,
...
+    void PreprocessForInset()
+    {
+        neededIndexes.clear();
+        CalculateOverlap();
+    }
+
+    std::vector<uint> GetOverlapForLevel() { return neededIndexes; }
+
+};
+
```
tst-sbayard: Should remove the commented-out code below if it's not needed.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -33,6 +33,7 @@
 #include "autoingest/plugins/RasterDBRootAsset.h"
 #include "autoingest/plugins/MercatorRasterProductAsset.h"
 #include "autoingest/plugins/MapLayerJSAsset.h"
+#include "autoingest/MiscConfig.h"
```
tst-lsavoie: Minor: I think this can be removed.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -122,21 +124,23 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
+                         config.beginMinifyLevel, config.endMinifyLevel);
+
+    // there are situations where this can happen and must be accounted for
+    if (calc == nullptr)
+    {
+        calc.reset(new OverlapCalculator<ProductAssetVersion>(env));
+        calc->PreprocessForInset();
     }
 
+    neededIndexes = std::move(calc->GetOverlapForLevel());
```
tst-lsavoie: I believe the level needs to be passed as a parameter here. If I'm not mistaken that's one of the important pieces of information for figuring out the overlap for each level.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -942,19 +943,15 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
...
-                              neededIndexes,
-                              beginMinifyLevel,
-                              endMinifyLevel);
-    }
+
+    overlapEnvelope<MercatorRasterProductAssetVersion> env (
+        type, genInfo[i].coverage, insetInfoPtrs, i,
+        beginMinifyLevel, endMinifyLevel);
+
+    std::shared_ptr<OverlapCalculator<MercatorRasterProductAssetVersion>> calc =
```
tst-lsavoie: Very minor: I suspect this could be a unique_ptr instead of a shared_ptr, which has slightly better performance.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -996,28 +992,29 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
...
+    }
 
+      // reuse or make a new PacketGenAssetVersion
+      MutableAssetVersionD packgen =
+        PacketGenFactory::ReuseOrMakeAndUpdate(genInfo[i].assetname,
+                                               type,
+                                               khMetaData(),
+                                               packetGenConfig,
+                                               extraArg,
+                                               calc);
```
tst-lsavoie: I'd really like to see calc become a member of extraArg. That way you don't need an extra parameter.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
tst-basilhuffman: unless i misunderstand it, is already being covered within khinsetcoverage, but i am passing it down either way

earth_enterprise/src/fusion/autoingest/AssetGenDaemonH.pl
```diff
@@ -98,7 +103,7 @@ ()
 
 #include <$header>
 #include <sysman/AssetD.h>
-
+#include <memory>
```
tst-lsavoie: Minor: I think you can remove this include now.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -868,9 +865,10 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
   if (type == AssetDefs::Imagery) {
     // first we know that we don't want skip tiles during minification
     // so set each inset to start after that
-    for (uint i = 0; i < genInfo.size(); ++i) {
-      genInfo[i].beginSkipTransparentLevel =
-        std::max(endMinifyLevel, genInfo[i].coverage.beginLevel());
+
+    for (auto& i : genInfo) {
+        i.beginSkipTransparentLevel =
+        std::max(endMinifyLevel, i.coverage.beginLevel());
```
tst-lsavoie: Minor: would be good to have indentation for the second line.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.cpp
```diff
@@ -330,6 +331,31 @@ void CalcPacketGenInfo<InsetInfo<MercatorRasterProductAssetVersion> > (
...
+        FindNeededImageryInsets(env.gencov,
+                                env.insets,
+                                env.numInsets,
+                                neededIndexes,
+                                env.beginMinifyLevel,
+                                env.endMinifyLevel);
+    }
+    else
+    {
+        FindNeededImageryInsets(env.gencov,
```
tst-lsavoie: Shouldn't this second call be `FindNeededTerrainInsets`?

tst-lsavoie:
This is performance critical code, so I'm being extra picky.

Looking at the old code, this is roughly what happens:
```
Inside BuildIndex
For each genInfo:
  Call FindNeededImageryInsets // Returns all insets that overlap with the current inset at any level
  Call PacketGenFactory::ReuseOrMakeAndUpdate which eventually calls UpdateChildren
  
Inside UpdateChildren
For each level:
  Call FindNeededImageryInsets // Returns all insets that overlap with the current inset at a specific level
```

In the new code, this is what happens:
```
Inside BuildIndex
For each genInfo:
  Call calc->PreprocessForInset
    Which calls FindNeededImageryInsets // Returns all insets that overlap with the current inset at any level
  Call PacketGenFactory::ReuseOrMakeAndUpdate which eventually calls UpdateChildren

Inside UpdateChildren
For each level:
  Call calc->GetOverlapForLevel (does not call FindNeededImageryInsets)
```

The code in BuildIndex does the same thing in both versions, but the second call to FindNeededImageryInsets in UpdateChildren is missing in the new version. If I understand correctly, that means we'll have extra insets and do unnecessary processing. The ultimate goal here is to refactor the code such that it works exactly the same but it's easy to swap out different methods of calculating overlap.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1057,19 +1054,14 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
...
-    }
+    overlapEnvelope<RasterProductAssetVersion> env (
+        type, genInfo[i].coverage, insetInfoPtrs, i,
+        beginMinifyLevel, endMinifyLevel, i);
+
+    std::shared_ptr<OverlapCalculator<RasterProductAssetVersion>> calc =
+        std::make_shared<OverlapCalculator<RasterProductAssetVersion>>(env);
+    calc->PreprocessForInset();
+    neededIndexes = std::move(calc->GetOverlapForLevel());
     neededIndexes.push_back(i); // each inset always intersects itself
```
tst-lsavoie: I think there may be a problem here. Since we're adding the inset outside of the calc object the version of neededIndexes that's inside the object will not have the extra element.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
tst-basilhuffman: done

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.cpp
tst-basilhuffman: fat fingered. fixed

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
tst-basilhuffman: fixed

tst-basilhuffman:
> This is performance critical code, so I'm being extra picky.
> 
> Looking at the old code, this is roughly what happens:
> 
> ```
> Inside BuildIndex
> For each genInfo:
>   Call FindNeededImageryInsets // Returns all insets that overlap with the current inset at any level
>   Call PacketGenFactory::ReuseOrMakeAndUpdate which eventually calls UpdateChildren
>   
> Inside UpdateChildren
> For each level:
>   Call FindNeededImageryInsets // Returns all insets that overlap with the current inset at a specific level
> ```
> 
> In the new code, this is what happens:
> 
> ```
> Inside BuildIndex
> For each genInfo:
>   Call calc->PreprocessForInset
>     Which calls FindNeededImageryInsets // Returns all insets that overlap with the current inset at any level
>   Call PacketGenFactory::ReuseOrMakeAndUpdate which eventually calls UpdateChildren
> 
> Inside UpdateChildren
> For each level:
>   Call calc->GetOverlapForLevel (does not call FindNeededImageryInsets)
> ```
> 
> The code in BuildIndex does the same thing in both versions, but the second call to FindNeededImageryInsets in UpdateChildren is missing in the new version. If I understand correctly, that means we'll have extra insets and do unnecessary processing. The ultimate goal here is to refactor the code such that it works exactly the same but it's easy to swap out different methods of calculating overlap.

So, for now, UpdateChildren **should** be making that call to `FindNeededImageryInsets`?

tst-lsavoie:
Yes.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -120,22 +120,19 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
-                              neededIndexes,
-                              config.beginMinifyLevel,
-                              config.endMinifyLevel);
-    }
+    auto gencov = khInsetCoverage(levCov);
+    //overlapEnvelope<ProductAssetVersion> env(type, gencov,
+    //                     extra.insetInfo, extra.insetInfo.size(),
+    //                     config.beginMinifyLevel, config.endMinifyLevel, level);
+
+    extra.calc->env.gencov = gencov;
```
tst-sbayard: Since `extra.calc->env.gencov` is a `khInsetCoverage&`, it seems like `extra.calc->env.gencov = gencov;` is going to end up modifying the object that `extra.calc->env.gencov` references ( `genInfo[i].coverage` from *RasterProject.src) instead of modifying the reference itself.
Not sure if that's what we want to do.

tst-sbayard: Resolved by latest commit.

earth_enterprise/src/common/khInsetCoverage.h
```diff
@@ -39,6 +39,7 @@ class khInsetCoverage {
   inline uint beginLevel(void) const { return beginLevel_; }
   inline uint endLevel(void)   const { return endLevel_; }
   inline uint numLevels(void)  const { return endLevel() - beginLevel(); }
+  inline uint numVLvls(void)  const { return vec0Level; }
```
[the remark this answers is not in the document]
tst-sbayard: Leaving in for now as it may be useful for upcoming chanes.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -868,9 +865,10 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
   if (type == AssetDefs::Imagery) {
     // first we know that we don't want skip tiles during minification
     // so set each inset to start after that
-    for (uint i = 0; i < genInfo.size(); ++i) {
-      genInfo[i].beginSkipTransparentLevel =
-        std::max(endMinifyLevel, genInfo[i].coverage.beginLevel());
+
+    for (auto& i : genInfo) {
+        i.beginSkipTransparentLevel =
```
tst-lsavoie: Minor: would be good to reduce this indentation by 2 spaces so that the next line is indented more than this one.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -942,19 +940,15 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
...
-                              neededIndexes,
-                              beginMinifyLevel,
-                              endMinifyLevel);
-    }
+
+    overlapEnvelope<MercatorRasterProductAssetVersion> env (
+        type, genInfo[i].coverage, insetInfoPtrs, i,
+        beginMinifyLevel, endMinifyLevel, i);
+
+    std::shared_ptr<OverlapCalculator<MercatorRasterProductAssetVersion>> calc =
```
tst-lsavoie: Minor: thinking about this a bit more: do you need a shared_ptr at all here? I'm not sure what the shared_ptr is getting us since we don't have lots of parts of the code using this over and over. I suspect just passing around the object (rather than a shared pointer to the object) would work fine.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -120,22 +120,19 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
-                              config.endMinifyLevel);
-    } else {
-      FindNeededTerrainInsets(khInsetCoverage(levCov),
-                              extra.insetInfo, numInsets,
-                              neededIndexes,
-                              config.beginMinifyLevel,
-                              config.endMinifyLevel);
-    }
+    auto gencov = khInsetCoverage(levCov);
+    //overlapEnvelope<ProductAssetVersion> env(type, gencov,
```
tst-lsavoie: Remove commented out code.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -996,27 +989,27 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
...
     }
 
-    // reuse or make a new PacketGenAssetVersion
-    MutableAssetVersionD packgen =
-      PacketGenFactory::ReuseOrMakeAndUpdate(genInfo[i].assetname,
-                                             type,
-                                             khMetaData(),
-                                             packetGenConfig,
-                                             extraArg);
+      // reuse or make a new PacketGenAssetVersion
```
tst-lsavoie: Minor: I think the extra indentation was added here by mistake.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -120,22 +120,19 @@ void PacketGenAssetVersionImplD::UpdateChildren(
     prevTime = curTime;
 
     // Determine which insets intersect us at this level
-    const khLevelCoverage levCov(config.coverage.levelCoverage(level));
+    khLevelCoverage levCov(config.coverage.levelCoverage(level));
```
tst-lsavoie: Any reason this is not const anymore?

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1132,20 +1123,23 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
...
 
-    // reuse or make a new PacketGenAssetVersion
-    MutableAssetVersionD packgen =
-      PacketGenFactory::ReuseOrMakeAndUpdate(genInfo[i].assetname,
-                                             type,
-                                             khMetaData(),
-                                             packetGenConfig,
-                                             extraArg);
+    // reuse or make a new PacketGenAssetVersion --> update children
+    // refactor this method call if wanting to do all calculations at once
```
tst-lsavoie: Minor: I think we can remove this comment.

Also minor: I think the extra indentation was added below by accident.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.cpp
tst-lsavoie: I'm not sure what happened but this still calls the imagery version, not the terrain version.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.h
```diff
@@ -168,6 +168,60 @@ void CalcPacketGenInfo(const khTilespace            &tilespace,
...
+    overlapEnvelope<ProductAssetVersion>& env;
+
+    void PreprocessForInset()
+    {
+        neededIndexes.clear();
+        CalculateOverlap();
+    }
+
+    std::vector<uint> GetOverlapForLevel(bool recalculate = false)
```
tst-lsavoie: Minor: you could get rid of the boolean by just returning neededIndexes from PreprocessForInset and not calling GetOverlapForLevel at all from BuildIndex.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1057,19 +1054,14 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
...
-                              endMinifyLevel);
-    } else {
-      FindNeededTerrainInsets(genInfo[i].coverage,
-                              insetInfoPtrs, i,
-                              neededIndexes,
-                              beginMinifyLevel,
-                              endMinifyLevel);
-    }
+    overlapEnvelope<RasterProductAssetVersion> env (
+        type, genInfo[i].coverage, insetInfoPtrs, i,
```
tst-lsavoie: Passing `i` twice here seems unnecessary. It would be nice to only do it once.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.h
```diff
@@ -168,6 +168,60 @@ void CalcPacketGenInfo(const khTilespace            &tilespace,
...
+                    numInsets(_numInsets),
+                    beginMinifyLevel(_beginMinifyLevel), endMinifyLevel(_endMinifyLevel),
+                    level(_level) {}
+};
+
+template <typename ProductAssetVersion>
+class OverlapCalculator
+{
+private:
+    std::vector<uint> neededIndexes;
```
tst-lsavoie: I think we can get away without storing neededIndexes permanently in this object for now since we're clearing it every time we use it. That will save us some memory, which is an important concern right now. Here's my recommendation.

In the `OverlapCalculator` class, remove the `neededIndexes` member variable.

Change `PreprocessForInset` from
```
    void PreprocessForInset()
    {
        neededIndexes.clear();
        CalculateOverlap();
    }
```
to
```
    std::vector<uint> PreprocessForInset()
    {
        std::vector<uint> neededIndexes;
        CalculateOverlap(neededIndexes);
        return neededIndexes;
    }
```

In BuildIndex, change:
```
    std::shared_ptr<OverlapCalculator<RasterProductAssetVersion>> calc =
        std::make_shared<OverlapCalculator<RasterProductAssetVersion>>(env);
    calc->PreprocessForInset();
    neededIndexes = std::move(calc->GetOverlapForLevel());
```
to
```
    std::shared_ptr<OverlapCalculator<RasterProductAssetVersion>> calc =
        std::make_shared<OverlapCalculator<RasterProductAssetVersion>>(env);
    neededIndexes = std::move(calc->PreprocessForInset());
```

There are a few other changes that will be necessary to make this work but I think those are the main ones. This will also allow you to get rid of the boolean parameter to GetOverlapForLevel and just always recalculate when that is called.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
tst-lsavoie: Making a member variable public and modifying it external to the class is not a great pattern. It would be best to handle this another way. Maybe gencov shouldn't be part of the envelope but should be passed to the calls that need it.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonH.pl
tst-lsavoie: I don't think this include is needed.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -120,22 +120,15 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
-                              extra.insetInfo, numInsets,
-                              neededIndexes,
-                              config.beginMinifyLevel,
-                              config.endMinifyLevel);
-    }
+    auto gencov = khInsetCoverage(levCov);
+
+    extra.calc->setBeginMinifyLevel(config.beginMinifyLevel);
+    extra.calc->setEndMinifyLevel(config.endMinifyLevel);
+    neededIndexes = std::move(extra.calc->GetOverlapForLevel(gencov));
```
tst-lsavoie: I just realized that the `extra.insetInfo` variable is different here than it is in BuildIndex. I think this will work but it may take longer. I think the most efficient way to do this is to pass in the entire envelope each time to make sure that we get the right inputs.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.h
```diff
@@ -168,6 +168,60 @@ void CalcPacketGenInfo(const khTilespace            &tilespace,
...
+{
+private:
+    void CalculateOverlap(std::vector<uint>&, const khInsetCoverage&);
+    overlapEnvelope<ProductAssetVersion> env;
+
+public:
+    OverlapCalculator(const overlapEnvelope<ProductAssetVersion>& _env) :
+        env(_env) {}
+
+    void setBeginMinifyLevel(uint _begin) { env.beginMinifyLevel = _begin; }
```
tst-lsavoie: Minor: I don't think you need these two functions any more.

tst-lsavoie:
I just realized this was targeted to 5.2.5.1. I've retargeted it to master (where we want to merge it), but that has created some merge conflicts.

tst-lsavoie:
I'm fine with including my commits on this PR.

PR #1334, opened 2019-06-05
Resume using graph operations

### LINKED ISSUE #1329: Slow resume for large builds
**Describe the bug**

Resume operations for large builds are very slow, especially after packgen tasks have been created.

**To Reproduce**

1. Create a large imagery project with at least several thousand resources.
1. Build the project until it starts executing packgen tasks.
1. Cancel the project
1. Resume the project

The resume will be much slower than the cancel even though they are essentially the same operation.

**Expected behavior**

The resume operation completes in a reasonable amount of time.

**Desktop**

- Server OS: CentOS 6
- Version: 5.2.5.1

**Additional context**

The issue seems to be that setting a "bad" state is easier than setting a "good" state. If any input to a resource is bad, then the resource is bad. However, all inputs must be good for a resource to be good. Thus, cancel will always be faster than resume.

However, from a larger perspective, both operations are mostly comparing and setting a bunch of enums, so they should both be fairly fast. The code can be improved significantly in this regard.

tst-lsavoie:
Rewrites the resume operation to use graph operations for increased speed.

To test:

1. Set `<GraphOperations>1</GraphOperations>` in `/gevol/assets/.config/misc.xml`
1. Restart Fusion.
1. Start building an imagery project.
1. Cancel the imagery project.
1. Resume the imagery project using the geresume command.

Alternatively, run the integration tests: https://github.com/google/earthenterprise/tree/master/earth_enterprise/integration_test

Fixes #1329.


earth_enterprise/BUILD_Ubuntu.md
```diff
@@ -22,7 +22,7 @@ Execute the following:
 
 ```bash
 curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
-sudo apt install git-lfs
+sudo apt-get install git-lfs
```
tst-basilhuffman: minor: is there a reason for changing?

tst-basilhuffman:
lgtm. just a minor comment

earth_enterprise/BUILD_Ubuntu.md
tst-lsavoie: I'm just making it consistent with the other lines. It just seemed odd to use apt here and apt-get everywhere else. Doesn't really make a difference.

PR #1337, opened 2019-06-07
Parallelize Section of Packgen Code

tst-jkent:
Parallelize loop in packgen code.
Use thread safe RefGuards.


earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -285,90 +284,80 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
-           idx != mergeIndexes.end(); ++idx) {
-        const InsetInfo<ProductAssetVersion> &inset(*extra.insetInfo[*idx]);
-        const PacketGenInfo &genInfo(*extra.genInfo[*idx]);
-        AssetVersion blendVer;
+      levInputs.resize(mergeIndexes.size() * 2);
+      levInputVers.resize(mergeIndexes.size() * 2);
+      int i = 0;
+      std::vector<int> emptyElements;
+      emptyElements.reserve(mergeIndexes.size());
+      #pragma omp parallel for schedule(static) num_threads(threadsToUse)
```
tst-ccamp: Can you add a comment explaining this pragma

tst-jkent: Will do.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -285,90 +284,79 @@ void PacketGenAssetVersionImplD::UpdateChildren(
       prevTime = curTime;
 
       // populate inputs and inputVers & config.insets
```
tst-sbayard: For `levInputs`, `levInputVers`, and 'packetLevelConfig.insets`:
- the calls to `reserve` are no longer necessary since `resize` is being called for each
- I _think_ that the `resize` size for `levInputs` and `levInputVers` should be `mergeIndexes.size() * 2 + 1`. The final `push_back` call after the merge index loop would cause a reallocation and copying of all of the elements otherwise.
- similarly for `packetLevelConfig.insets.resize()`, I think the value should be `mergeIndexes.size() + 1`.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -285,90 +284,79 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
 
-          packetLevelConfig.insets.push_back
-            (PacketLevelConfig::Inset(
-                inset.dataRPFile,
-                inset.alphaRPFile,
-                blendVer.Ref(), /* cached blend */
-                cache_alpha_exist ?
-                    blendVer.Ref() : SharedString()  /* cached blend alpha */));
-        }
+      for (size_t idx = 0; idx < emptyElements.size(); ++idx) {
```
tst-sbayard: I think this loop needs to be iterating emptyElements in reverse. Otherwise the wrong elements will be getting erased from `levInputs` and `levInputVers` as elements get shifted after each `erase` call.

tst-lsavoie:
I'm OK with including my commits in this PR.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -412,6 +388,8 @@ void PacketGenAssetVersionImplD::UpdateChildren(
         prevTime = curTime;
       }
 
+      std::remove(levInputs.begin(), levInputs.end(), "");
```
tst-lsavoie: Always good to use standard library functions.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -45,6 +45,7 @@ class PacketGenExtraUpdateArg {
 #include <autoingest/plugins/RasterProductAsset.h>
 #include <autoingest/plugins/MercatorRasterProductAsset.h>
 #include <chrono>
+#include <typeinfo>
```
tst-lsavoie: Is this still needed? I don't see anywhere it's used, but I could have missed something.

tst-lsavoie:
We'll need documentation updates on this. We'll have to do it like the other PRs where the doc updates are in a separate PR.

tst-lsavoie:
Oh, and there's some merge conflicts that need to be fixed.

tst-lsavoie:
One other thing that we missed - we need to add a lock when accessing the internal storage for shared strings, since that may be updated by multiple threads at once. I don't think the SharedStrings were finished when I started on this.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -247,90 +245,65 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
+        const InsetInfo<ProductAssetVersion> &inset(*extra.insetInfo[mergeIndexes[idx]]);
+        const PacketGenInfo &genInfo(*extra.genInfo[mergeIndexes[idx]]);
+        int i = idx*2;
+        levInputs[i] = (inset.combinedrp_).Ref();
+        levInputVers[i] = inset.combinedrp_;
+
         if (!inset.alphaRPFile.empty() &&
             (genInfo.packetLevels.size() > minifySrcLevel)) {
-          blendVer = genInfo.packetLevels[minifySrcLevel];
+          levInputs[i+1] = (genInfo.packetLevels[minifySrcLevel]).Ref();
```
tst-lsavoie: Technically there's a difference between this and the old version. The old version worked like this:
```
        if (!inset.alphaRPFile.empty() &&
            (genInfo.packetLevels.size() > minifySrcLevel)) {
          blendVer = genInfo.packetLevels[minifySrcLevel];
        }

        if (!blendVer) {
          levInputs.push_back(inset.combinedrp_->GetRef());
          levInputVers.push_back(inset.combinedrp_);
        }
```
This change shortens it to
```
        if (!inset.alphaRPFile.empty() &&
            (genInfo.packetLevels.size() > minifySrcLevel)) {
          levInputs[i+1] = (genInfo.packetLevels[minifySrcLevel]).Ref();
          levInputVers[i+1] = genInfo.packetLevels[minifySrcLevel];
        }
```
There's a chance that `blendVer` might still evaluate to false even after `blendVer = genInfo.packetLevels[minifySrcLevel];`. In that case `levInputs` and `levInputVers` would never be set, so the outcome of the two versions of the code is different. I'm not sure if it matters in this case, but it's safest to keep the behavior the same.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -247,90 +245,65 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
+      int threadsToUse = MiscConfig::Instance().PackGenThreads;
+      levInputs.resize(mergeIndexes.size() * 2 + 1);
+      levInputVers.resize(mergeIndexes.size() * 2 + 1);
+
+      // Run the following loop in parallel using the amount of threads set in misc.xml.
+      #pragma omp parallel for schedule(static) num_threads(threadsToUse)
+      for (size_t idx = 0; idx < mergeIndexes.size(); ++idx) {
+        const InsetInfo<ProductAssetVersion> &inset(*extra.insetInfo[mergeIndexes[idx]]);
+        const PacketGenInfo &genInfo(*extra.genInfo[mergeIndexes[idx]]);
+        int i = idx*2;
```
tst-lsavoie: Might be a good idea to use something other than `i` here. If we have `i` and `idx` in the same loop, people are going to get confused.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -247,90 +245,70 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
+        levInputVers[inputs_idx] = inset.combinedrp_;
+
         if (!inset.alphaRPFile.empty() &&
             (genInfo.packetLevels.size() > minifySrcLevel)) {
           blendVer = genInfo.packetLevels[minifySrcLevel];
         }
 
-        if (!blendVer) {
-          // This inset doesn't have a cached blend file. This
+        if (blendVer) {
```
tst-sbayard: I would suggest changing the general flow of logic back the way it was unless there's a compelling reason to keep it the way it is now. The if/else conditions are kind of backward to what they were before, and it makes it fairly difficult to compare the old and new versions and feel confident that they're doing the same thing. 
I think it would be best to keep it as
```
if (!blendVer) {
    // Do same stuff as before but with the thread-safe constructs.
}
else {
    // Do same stuff as before but with the thread-safe constructs.
    // Of course, collapse the two identical cases into one as you've already done.
}
```

earth_enterprise/src/common/SharedString.h
```diff
@@ -42,6 +43,7 @@ class SharedString {
         std::unordered_map<uint32_t, std::string> refFromKeyTable;
         std::unordered_map<std::string, uint32_t> keyFromRefTable;
         static uint32_t nextID;
+        mutable std::mutex m;
```
tst-sbayard: We may want to consider using shared_mutex and shared_lock/unique_lock where appropriate. That would allow different threads to call `size` and `RefFromKey` simultaneously. With the sheer number of calls that will be made to `RefFromKey`, I think this might be worthwhile.
Caveat: I actually haven't used it before, so I'm speaking hypothetically here, but it might be good.

tst-sbayard: @tst-jkent pointed out that shared_mutex is only available in C++17.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -247,90 +245,67 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
-      for (std::vector<uint>::const_iterator idx = mergeIndexes.begin();
-           idx != mergeIndexes.end(); ++idx) {
-        const InsetInfo<ProductAssetVersion> &inset(*extra.insetInfo[*idx]);
-        const PacketGenInfo &genInfo(*extra.genInfo[*idx]);
-        AssetVersion blendVer;
+      packetLevelConfig.insets.resize(mergeIndexes.size());
+      int threadsToUse = MiscConfig::Instance().PackGenThreads;
+      levInputs.resize(mergeIndexes.size() * 2 + 1);
+      levInputVers.resize(mergeIndexes.size() * 2 + 1);
+      AssetVersion blendVer;
```
tst-sbayard: As discussed, it would be good to move the `blendVer` declaration inside the for loop.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -362,6 +338,8 @@ void PacketGenAssetVersionImplD::UpdateChildren(
                   prevLevelVer.Ref() : SharedString() /* cached blend alpha */));
       }
 
+      std::remove(levInputs.begin(), levInputs.end(), "");
+      std::remove(levInputVers.begin(), levInputVers.end(), nullptr);
```
tst-sbayard: As discussed, the `std::remove` calls here are leaving the vectors with duplicate items at the end. Add `erase` calls to remove the extraneous items.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -362,6 +338,8 @@ void PacketGenAssetVersionImplD::UpdateChildren(
                   prevLevelVer.Ref() : SharedString() /* cached blend alpha */));
       }
 
+      levInputs.erase(std::remove(levInputs.begin(), levInputs.end(), ""));
+      levInputVers.erase(std::remove(levInputVers.begin(), levInputVers.end(), nullptr));
```
tst-sbayard: I think you need to include `levInputs.end()` and `levInputVers.end()` in the calls to `erase()`. Otherwise it will only erase once item. That will only be correct in cases where there is only one blank item in the vector.

earth_enterprise/src/common/SharedString.h
```diff
@@ -42,6 +43,7 @@ class SharedString {
         std::unordered_map<uint32_t, std::string> refFromKeyTable;
         std::unordered_map<std::string, uint32_t> keyFromRefTable;
         static uint32_t nextID;
+        mutable std::mutex string_storage_mutex_;
       public:
         const std::string & RefFromKey(const uint32_t &key) {
           auto refIter = refFromKeyTable.find(key);
```
tst-lsavoie: We need a lock everywhere that refFromKeyTable or keyFromRefTable is accessed in any way.. That includes this function and parts of the KeyFromRef function that are not currently locked.

Also, I agree with Stepehen that a read/write lock would be best here.

earth_enterprise/src/common/SharedString.h
```diff
@@ -42,30 +43,42 @@ class SharedString {
...
         }
         StringStorage(){
+          rc = pthread_rwlock_wrlock(&rwlock);
           refFromKeyTable.insert(std::pair<uint32_t, std::string>(0, ""));
           keyFromRefTable.insert(std::pair<std::string, uint32_t>("", 0));
+          rc = pthread_rwlock_unlock(&rwlock);
         }
         size_t size() const {
+          rc = pthread_rwlock_rdlock(&rwlock);
           return refFromKeyTable.size();
```
tst-lsavoie: I think there's a bug here since you never release the read lock.

earth_enterprise/src/common/SharedString.h
```diff
@@ -42,30 +43,42 @@ class SharedString {
         std::unordered_map<uint32_t, std::string> refFromKeyTable;
         std::unordered_map<std::string, uint32_t> keyFromRefTable;
         static uint32_t nextID;
+        mutable pthread_rwlock_t rwlock;
+        mutable int rc = pthread_rwlock_init(&rwlock, NULL);
```
tst-lsavoie: Minor: might be a good idea to move this to the constructor.

earth_enterprise/src/common/SharedString.h
```diff
@@ -42,30 +43,42 @@ class SharedString {
...
       public:
         const std::string & RefFromKey(const uint32_t &key) {
+          rc = pthread_rwlock_rdlock(&rwlock);
```
tst-lsavoie: Locking and unlocking is typically done in some kind of wrapper class. The reasoning is that if something bad happens (an exception is thrown in the middle of the function) the lock will still be released. There's an example of how to do this in the parallel-system-manager branch: https://github.com/google/earthenterprise/blob/parallel-system-manager/earth_enterprise/src/common/khMTTypes.h (lines 125-160). To acquire a read lock you just create an instance of the khReadGuard, passing a khRWLock as a parameter. The lock is released when the read guard goes out of scope. I'll leave it up to you as to whether you want to make that change here.

earth_enterprise/src/common/SharedString.h
```diff
@@ -37,35 +38,77 @@
 // much faster as we only need to compare the identifiers.
 class SharedString {
 protected:
-   class StringStorage {
+    class stringStorageRWLock
+    {
+      private:
+        pthread_rwlock_t rwlock;
```
tst-basilhuffman: I thought Justin already wrote this but I am not seeing it. Might be easier to do this with boost. Wouldn't be a bad idea to add it to khThread also

tst-jkent: I had previously tried to implement boost's read/write locks, but quite a few linker issues regarding the unit tests occurred when adding that dependency to SharedString. Perhaps moving this to khThread may avoid that.

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -39,9 +38,9 @@ inline uint64 GetObjectSize(const std::vector<T> &vec) {
 inline uint64 GetObjectSize(const std::string &str) {
     return sizeof(str) + (str.capacity() * sizeof(char));
 }
-// determine amount of memory used by a shared_ptr and the object it points to
+// determine amount of memory used by a khRefGuard and the object it points to
 template<class T>
-inline uint64 GetObjectSize(const std::shared_ptr<T> &guard) {
-    return sizeof(guard) + (guard ? guard->GetSize() : 0);
+inline uint64 GetObjectSize(const khRefGuard<T> &guard) {
```
tst-lsavoie: Something has gone wrong with your merge. We should be using the `shared_ptr` version of this function, not the `khRefGuard` version.

tst-lsavoie:
Replaced by PR #1397.

Issue #1339, opened 2019-06-17
run_historical script results in Blocked Databases


tst-cjeffries:
The `run_historical.sh` tutorial script results in the first database succeeding, with the following databases becoming blocked. The blocked databases, have failed resources due to a `DelayedBuildChildren` error:

`[2019-05-20 15:19:09] Warning: Exception during OnStateChange: No such file: /gevol/assets/Tutorial/Resources/Imagery/i3_15Meter.kiasset/CombinedRP.kia/packgen.kia/ver002/khassetver.xml`
`[2019-05-20 15:19:09] Warning: FATAL ERROR: DelayedBuildChildren, logging to file: /gevol/assets/Tutorial/Projects/Imagery/SFinset_2.kiproject/ver001/logfile`

**To Reproduce**

Run the run_historical.sh script after creating the needed prerequisite resources.

**Expected behavior**

All databases should finish building successfully.

**Desktop**

Ubuntu 16 (and likely others)

PR #1340, opened 2019-06-17
Issue #1339: Handle DelayedBuildChildren error variant

tst-cjeffries:
Add error handling to alleviate a version of the DelayedBuildChildren error.

To test:

1. Build and install OpenGEE (can reliably reproduced on Ubuntu16).
2. Run the historical and tutorial scripts.
3. Verify that the databases are correct, and that `/opt/google/log/gesystemmanager.log` shows that the error was caught for the SFinset_2-5 databases (it should be caught many times).
4. Run the gauge tests (it may be wise to reinstall and clear out the tutorial databases before this step).

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
+++ b/earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
@@ -323,8 +323,8 @@ ${name}Factory::ReuseOrMakeAndUpdate(
                 for (std::vector<std::string>::const_iterator iv =
                      version->inputs.begin();
                      iv != version->inputs.end(); ++iv) {
-                     notify(NFY_NOTICE, "             %s", iv->c_str());
-                  }
+                    notify(NFY_NOTICE, "             %s", iv->c_str());
+                }
 #endif
                 // Tell the storage manager that we're going to use this
                 // version again.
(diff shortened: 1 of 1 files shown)


tst-lsavoie:
It looks like this code will work. The only potential problem I see is that you could create extra versions of an asset (you create a version when the build starts, then when DelayedBuildChildren is called you can't load that version so you create a new version). I think we can live with that, though, especially since this is relatively rare.

I didn't do any testing.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -302,38 +302,44 @@ ()
...
-                for (const auto & bi : boundInputs) {
-                    notify(NFY_NOTICE, "             %s", bi.c_str());
-                }
-                notify(NFY_NOTICE, "         version inputs:");
-                for (std::vector<std::string>::const_iterator iv =
-                     version->inputs.begin();
-                     iv != version->inputs.end(); ++iv) {
-                    notify(NFY_NOTICE, "             %s", iv->c_str());
-                }
+                  notify(NFY_NOTICE,
```
tst-dsugar: besides being a spacing/tab error, do we really want to carry this code block?

tst-cjeffries: I just further indented the existing formatting. How would you prefer it? I try not to change the existing code much, in hopes of keeping it consistent.

tst-cjeffries: Cleaned up some of the spacing to reduce the noise.

PR #1343, opened 2019-06-17
Issue #1342: missed shared string

### LINKED ISSUE #1342: Use shared strings as cache keys
We previously updated Fusion to use one copy of each asset and asset version name. However, we missed one update. This bug is to fix the missed update.

In storage manager, The declaration:
```
using CacheType = khCache<std::string, HandleType>;
```
should be
```
using CacheType = khCache<AssetKey, HandleType>;
```

tst-lsavoie:
Fixes #1342 by using shared strings as keys in the asset and asset version caches.


tst-jkent:
I've reviewed and it all looks good to me.
Only thing I'm thinking is that it might be a good idea to run some tests to verify that the changes to SaveToDotNew work as expected.

PR #1354, opened 2019-06-20
Issue 1342 missed shared string

tst-lsavoie:


tst-lsavoie:
This was previously reviewed, tested, and merged to master. No need for further reviews.

PR #1355, opened 2019-06-22
Change the request to quit the portable server into a POST

rafd123:
Fixes google#1281

IE appears to perform aggressive caching, which was preventing the `GET` request to `cmd=quit` from making it to the server.

Changing the command to be a `POST` request prevents this; given that `cmd=quit` has side effects, it seems like it should be a `POST` anyhow.


rafd123:
CLA: I signed it.

rafd123:
🎩

PR #1357, opened 2019-06-25
Using and documenting correct FQDN config for issue 1132

### LINKED ISSUE #1132: GEE Server - Determine hostname more reliably than /etc/hosts file
When geserver service starts, it gets the hostname to use for the default virtual host entries by reading the /etc/hosts file rather than using the `hostname` command, like Fusion and the installer scripts. This usually has the effect of causing the default virtual hosts to point to `localhost` rather than the hostname. 

**AC**
1. Update GE Server startup service to determine the hostname using the `hostname` utility rather than through /etc/hosts
1. Document the change in behavior so users know to make sure the server name is still correct after an upgrade

tst-jzmolek:
Due to occasional errors determining the FQDN on server startup believed to be caused by possible OS misconfiguration, documentation has been updated for how network settings should be configured as well as code updates to use the correct hostname for virtual hosts if the local host happens to get eslected.

Reproduction steps:  Verify that by following the instructions the Apache server starts up without errors determining the hostname/FQDN, and that the virtual host URLs use the correct FQDN by executing `/opt/google/bin/geserveradmin --listvhs`

Tested on Ubuntu 14, Ubuntu 16, and CentOS 7.
Fixes #1132


earth_enterprise/src/server/wsgi/common/geAbstractionFetcher.py
```diff
@@ -0,0 +1,69 @@
...
+  return thisHost
+
+
+def Usage(cmd):
+  """Show proper usage for this command."""
+  print "usage: %s [hostname [-f]]" % cmd
+  exit(1)
+
+
+def main(args):
```
tst-cjeffries: This is like a full on mini-module. Nice job!

tst-cjeffries:
Everything looks good! Approved.

docs/geedocs/5.3.2/answer/HostnameFqdnConfiguration.html
```diff
@@ -0,0 +1,53 @@
...
+<p>Ideally the network where GEE is installed has a DNS set up to resolve hostnames so that IP traffic can be
+routed appropriately. However in cases where there is no DNS, static hostnames can be used as well. With static
+hostnames the <code>hostname</code> or <code>hostnamectl</code> command should be used to set the hostname to its
+FQDN. The <code>/etc/hostname</code> file should be updated to match as well so the name stays after reboots. If
+setting the hostname to the FQDN is not possible then setting the FQDN in the <code>/etc/hosts</code> file is
+necessary. When using a static hostname some operating systems may still give a warning on Apache startup even if
+it correctly determined the FQDN. Adding the FQDN to <code>/etc/hosts</code> would fix this as well. If adding the
+FQDN to <code>/etc/hosts</code> is needed, then the FQDN followed by just the host portion should be added to any
+line that contains the IP address the server will be listening on, like this: </p>
+<p><code>127.0.0.1	myserver.mydomainname.com myserver localhost</code></p>
```
tst-ccamp: You should add a sentence specifying that this line with the hostname should be the first line of the file

PR #1364, opened 2019-06-28
Clean and Validate KML Pasted in Cutter

restouffer:
KML pasted into Cutter to define a region is checked in Javascript. That XML is then implicitly trusted when the process to build a globe is started. This PR adds a server-side validity check and saves the XML in a standardized form.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
@@ -31,6 +31,7 @@ import shutil
 import sys
 import time
 import urllib
+from lxml import etree
 
 from common import form_wrap
 from common import postgres_manager_wrap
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
```diff
@@ -31,6 +31,7 @@
 import sys
 import time
 import urllib
+from lxml import etree
```
tst-dsugar: This appears to be a new dependency.

tst-dsugar: I agree with @tst-blaird  that a very minor change maybe can still be done on the front end, too, though this may already be caught as part of this backend check using lxml.  It does look like python lxml at least is already packaged in debian, ubuntu, and centos.  This should be confirmed and tested on centos6, though.

restouffer: lxml is used extensively in the repository, so it might be worth making that dependency explicit even if this change is not merged.

tst-dsugar:
LGTM too

PR #1368, opened 2019-07-09
Fix 1211: Copy download_tutorial.sh instead of link

tst-cjeffries:
Adjusts scons build scripts to copy the `download_tutorial.sh` script instead of linking it. Thus when it is installed it is an actual file.

To test:

1. Build OpenGEE, including the RPMs.
2. Install using the installation scripts.
3. Verify that `download_tutorial.sh` is a file, not a link.
4. Uninstall OpenGEE.
5. Install using the RPMs.
6. Verify that `download_tutorial.sh` is a file, not a link.

### DIFF
--- a/earth_enterprise/tutorial/SConstruct
+++ b/earth_enterprise/tutorial/SConstruct
@@ -92,6 +92,5 @@ if env.GetOption('clean'):
     # clean the search and getomcat install folders.
     commands.getstatusoutput("rm -rf " + tutorials_dir)
 else:
-    # makeSoftLinks(root_dir, tutorial_src_dir, ['.svn'])
     copyFiles(root_dir, tutorial_src_dir, ['.svn'])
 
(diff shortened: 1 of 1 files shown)


earth_enterprise/tutorial/SConstruct
```diff
@@ -62,10 +62,35 @@ def makeSoftLinks(dest_dir, src_dir, excluded_list):
...
+                                    dest_filepath + '"')
+
 # Clean install dir (note: the current sconscripts leave stuff behind, like
 # directories).
 if env.GetOption('clean'):
     # clean the search and getomcat install folders.
     commands.getstatusoutput("rm -rf " + tutorials_dir)
 else:
-    makeSoftLinks(root_dir, tutorial_src_dir, ['.svn'])
+    copyFiles(root_dir, tutorial_src_dir, ['.svn'])
```
tst-dsugar: I am guessing historically geep was originally hosted in svn rather than git.  But do we have .svn directories today??  If not, and we don't need the exclude list anymore, could we simplify this to using a cp -R or maybe an existing python function that does recursive copying?

tst-cjeffries: Sure. We can certainly refactor it. I wrote it this way to be in line with other scons stuff.

PR #1374, opened 2019-07-17
Update docs resume

tst-basilhuffman:
updated docs to reflect new resume operations


tst-lsavoie:
Can we try and make the history cleaner on this PR? The PR includes 10 merge commits and only one with actual data.

docs/geedocs/5.3.2/answer/176738.html
```diff
@@ -309,6 +309,9 @@ <h2 id="status_request_timeout">Status request timeout</h2>
    spend waiting to fulfill a status request before it times out. This timeout can be configured manually by setting
    <code>MutexTimedWaitSec</code> in <code>/gevol/assets/.config/misc.xml</code></p>
 <code>&lt;MutexTimedWaitSec&gt;60&lt;/MutexTimedWaitSec&gt;</code>
+<br><br>
+<h2 id="utilizing_graph_operations_in_resume_operations">Utilizing graph operations in resume operations</h2>
+<p>Included in <code>/gevol/assets/.config/misc.xml</code> is a <code>GraphOperations</code> option, and it may be set to <code>true</code> or <code>false</code>. It is an experimental operation which allows for resume operations to be optimized via graph operations.
```
tst-lsavoie: My thoughts:
- I think it needs to be set to `1` or `0`, not `true` or `false`.
- I'm not sure that we need to mention that it uses graph operations, as the average user won't care about that. We can just say it's an experimental version that improves performance.
- I'd like to make the experimental part clearer by putting it into its own sentence.
- The GraphOperations option will likely be used for more than just resume as we update more operations.

So the second sentence would get split into two sentences and look something like this: "It enables optimized versions of some operations. These optimizations are experimental and may contain bugs."

docs/geedocs/5.3.2/answer/176738.html
```diff
@@ -309,6 +309,10 @@ <h2 id="status_request_timeout">Status request timeout</h2>
    spend waiting to fulfill a status request before it times out. This timeout can be configured manually by setting
    <code>MutexTimedWaitSec</code> in <code>/gevol/assets/.config/misc.xml</code></p>
 <code>&lt;MutexTimedWaitSec&gt;60&lt;/MutexTimedWaitSec&gt;</code>
+<br><br>
+
+<h2 id="improving_performance_in_resume_operations">Improving performance in resume operations</h2>
```
tst-lsavoie: I just realized the title specifically mentions resume. That should be changed to be more generic like the body text. Something like "Improving Operation Performance"

docs/geedocs/5.3.2/answer/176738.html
```diff
@@ -309,6 +309,10 @@ <h2 id="status_request_timeout">Status request timeout</h2>
    spend waiting to fulfill a status request before it times out. This timeout can be configured manually by setting
    <code>MutexTimedWaitSec</code> in <code>/gevol/assets/.config/misc.xml</code></p>
 <code>&lt;MutexTimedWaitSec&gt;60&lt;/MutexTimedWaitSec&gt;</code>
+<br><br>
+
+<h2 id="improving_optimization_performance">Improving Optimization Performance</h2>
```
tst-lsavoie: Very minor: "Improving Optimization Performance" doesn't make sense to me. Either "Improving Performance" or "Optimizing Performance" makes more sense.

tst-lsavoie:
All commits are either from @tst-basilhuffman or me. I'm fine with my commits being included.

PR #1377, opened 2019-07-24
Sharedstring to flyweight

### LINKED ISSUE #1378: Use Boost flyweight instead of custom SharedStrings
We've previously created a SharedString class so that we don't hold multiple copies of asset and asset version refs in memory. Boost provides a similar capability through the flyweight library. We should switch to Boost flyweight instead of using custom code.

tst-basilhuffman:
initial commit. notes:

- decided to go with a bridge instead of a typedef since there were some things that needed smoothing
- may need to discuss the meaning of "empty"
- looks overwhelming but the bulk is changing .toString() to std::string(...), as it seemed more logical to me that way

Fixes #1378


earth_enterprise/src/common/khTypes.h
```diff
@@ -62,6 +62,61 @@ typedef double                          float64;
 
 #ifdef __cplusplus
 
+// used to limit the number of strings in memory
+// thread-safe and removes when no longer needed
+#include <boost/flyweight.hpp>
+
+class AssetHandle
```
tst-lsavoie: I think the term AssetHandle is getting too overloaded. Maybe call this an AssetKey?

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -509,7 +509,7 @@ ()
     typedef ${name}AssetVersionImplD VerImpl;
     Mutable${name}AssetVersionD newver(std::make_shared<VerImpl>(this));
 
-    AddVersionRef(newver->GetRef().toString());
+    AddVersionRef(std::string(newver->GetRef()));
```
tst-lsavoie: Minor: do we need the explicit std::string conversion here (and everywhere else it's used?)

tst-basilhuffman: To me, it seemed like a better way to do things than `.toString()` i.e. it looks more like c++ than java

tst-lsavoie: I think it's fine, I'm just not sure it's necessary. It seems like `AddVersionRef(newver->GetRef());` should work with no explicit conversion.

earth_enterprise/src/common/khTypes.h
tst-basilhuffman: looks like AssetKey is present as well, maybe something like `AssetRefKey` ?

tst-lsavoie: Maybe just `AssetRef`?

The AssetKey definition in StorageManager can be removed and replaced with this class, so AssetKey is no longer needed there, but it looks like AssetKey is also used other places.

earth_enterprise/src/common/khTypes.h
```diff
@@ -62,6 +62,61 @@ typedef double                          float64;
...
+{
+    boost::flyweights::flyweight<std::string> fw;
+public:
+    AssetHandle() = default;
+    AssetHandle(const AssetHandle& other) { fw = other.fw; }
```
tst-lsavoie: It looks like a lot of these functions (copy constructor, copy assignment operator, etc) are just the default versions and can be removed.

earth_enterprise/src/common/khTypes.h
```diff
@@ -62,6 +62,61 @@ typedef double                          float64;
...
+    bool operator==(const AssetHandle& other) const { return fw == other.fw; }
+    bool operator!=(const AssetHandle& other) const { return fw != other.fw; }
+    bool operator<(const AssetHandle& other) const { return fw < other.fw; }
+    bool operator>(const AssetHandle& other) const { return fw > other.fw; }
+    bool operator==(const std::string& ref) const { return fw == ref; }
+    bool operator!=(const std::string& ref) const { return fw != ref; }
+    bool operator<(const std::string& ref) const { return fw < ref; }
+    bool operator>(const std::string& ref) const { return fw > ref; }
+
+    bool empty() const { return std::string(fw).size() == 0 ? true : false; }
```
tst-lsavoie: Minor: you don't need the `:?` operator here, you can just `return std::string(fw).size == 0;` or even `return std::string(fw).empty();`.

Thinking out loud here - for efficiency sake it might be faster to check `fw == boost::flyweights::flyweight<std::string>("");` since comparisons between flyweights are likely more efficient than creating a new string.

earth_enterprise/src/common/khTypes.h
```diff
@@ -62,14 +62,83 @@ typedef double                          float64;
...
+
 // Class template to extract properties from a generic type.
 // It is used to overload a function template based on a type condition.
 template <int v>
 struct Int2Type {
     enum {value = v};
 };
 
-// defined ptr_uint_t to the appropriate type for this platform
+// defined ptr_uint_t SharedStringto the appropriate type for this platform
```
tst-lsavoie: Looks like this comment change was a mistake and should be backed out.

earth_enterprise/src/common/khsimple_strconv.h
```diff
@@ -82,7 +81,7 @@ FromString(const std::string &str, T &val)
 
 inline
 void
-FromString(const std::string &str, SharedString &val)
+FromString(const std::string &str, AssetRefKey& val)//SharedString &val)
```
tst-lsavoie: Remove comment.

earth_enterprise/src/common/sharedstring_unittest.cpp
```diff
@@ -1,118 +0,0 @@
-// Copyright 2019 The Open GEE Contributors
```
tst-lsavoie: It wouldn't hurt to have some very basic unit tests for AssetRefKey since it does involve some code.

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -249,7 +249,7 @@ main(int argc, char *argv[])
                 readableMemorySize(taskLists.asset_cache_memory).c_str());
         outline("Total memory used by cached asset versions: %s",
                 readableMemorySize(taskLists.version_cache_memory).c_str());
-        outline("Number of strings cached: %u", taskLists.str_store_size);
+        outline("");
```
tst-lsavoie: We should just remove this rather than making it an empty line.

earth_enterprise/src/common/khTypes.h
```diff
@@ -62,14 +62,83 @@ typedef double                          float64;
 
 #ifdef __cplusplus
 
+
+#include <boost/flyweight.hpp>
+// wrapper class for boost flyweight
+// relies on a lot of underlying boot functionalty
```
tst-lsavoie: `boost`, not `boot`

tst-lsavoie:
One other comment - there's been some merges to master, so conflicts need to be fixed.

PR #1384, opened 2019-07-31
Consolidate Find, FindVersion and ValidateRefForInput logic

tst-jkent:
Take the auto gen logic for Find, FindVersion and ValidateRefForInput and migrate them to Storage Manager.


earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -440,7 +440,7 @@ void RasterProjectAssetImplD::HandleEditRequest(
 
   // Custom FindMake - modified to set/preserve indexVersion & fuid_channel_.
   MutableRasterProjectAssetD project =
-      RasterProjectFactory::Find(projname, req.type);
+      Find(projname, req.type, kProjectSubtype);
```
tst-lsavoie: One way to do this is to make Find a template function, like this:
```
template <class Type>
Type Find(const std::string &ref_ )
{
    try {
        Asset asset(ref_);
        if (asset &&
            (asset->type == AssetDefs::Map) &&
            (asset->subtype == Type::Name())) {
            return Type(ref_);
        }
    } catch (...) {
        // do nothing - don't even generate any warnings
    }
    return Type();
}
```
Then you'd have to add a new autogenerated function `Name` that returns the type name. You would call it like this:
```
MutableRasterProjectAssetD project =
      Find<MutableRasterProjectAssetD>(projname, req.type);
```
There's still the question of whether this is the right way to approach enhancing the storage manager, but I'm curious to see if this would work.

tst-lsavoie: It does make the code simpler and smaller and eliminate some autogeneration, though.

earth_enterprise/src/fusion/autoingest/sysman/AssetD.h
```diff
@@ -58,5 +58,55 @@ class AssetImplD : public virtual AssetImpl
 typedef DerivedAssetHandle_<Asset, AssetImplD> AssetD;
 typedef MutableAssetHandleD_<AssetD> MutableAssetD;
 
+template<class T>
```
tst-lsavoie: Minor: I prefer something more descriptive than `T` for the template parameter. Maybe `AssetType`?

Same for other functions.

earth_enterprise/src/fusion/autoingest/sysman/AssetD.h
```diff
@@ -58,5 +58,55 @@ class AssetImplD : public virtual AssetImpl
...
+        return T(SharedString(ref));
+    }
+  } catch (...) {
+      // do nothing - don't even generate any warnings
+  }
+  return T();
+}
+
+template<class T>
+T FindVersion(const std::string & ref, const AssetDefs::Type & type, const std::string & subtype)
```
tst-lsavoie: AssetVersionD.h might be a better place for FindVersion.

tst-lsavoie:
We need to remove these 3 functions so that they are no longer auto-generated.

earth_enterprise/src/fusion/autoingest/plugins.src/Database.src
```diff
@@ -228,17 +228,15 @@ void DatabaseAssetImplD::ConstructInputsFromRequest(
                             std::vector<SharedString>* inputs) {
   // make sure the project asset_ref_string's are valid
   if (!req.config.imageryProject.empty()) {
-    RasterProjectFactory::ValidateRefForInput(req.config.imageryProject,
-                                              AssetDefs::Imagery);
+    ValidateRefForInput<MutableDatabaseAssetD>(req.config.imageryProject, AssetDefs::Imagery, kDatabaseSubtype);
```
tst-lsavoie: Based on the contents of RasterProjectFactory, the template parameter should be RasterProjectAssetVersionD.

earth_enterprise/src/fusion/autoingest/plugins.src/Database.src
```diff
@@ -228,17 +228,15 @@ void DatabaseAssetImplD::ConstructInputsFromRequest(
...
   }
   if (!req.config.vectorProject.empty()) {
-    VectorProjectFactory::ValidateRefForInput(req.config.vectorProject);
+    ValidateRefForInput<MutableDatabaseAssetD>(req.config.vectorProject, AssetDefs::Vector, kDatabaseSubtype);
     inputs->push_back(req.config.vectorProject);
   }
   if (!req.config.terrainProject.empty()) {
-    RasterProjectFactory::ValidateRefForInput(req.config.terrainProject,
-                                              AssetDefs::Terrain);
+    ValidateRefForInput<MutableDatabaseAssetD>(req.config.terrainProject, AssetDefs::Terrain, kDatabaseSubtype);
```
tst-lsavoie: Template parameter should be RasterProjectAssetVersionD.

earth_enterprise/src/fusion/autoingest/plugins.src/Database.src
tst-lsavoie: There are likely other places like this where some of the parameters aren't correct. Would you take a pass through and make sure the parameters you're passing are the same as the ones used in the Factory that used to be called?

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -190,7 +139,8 @@ ()
 {
     // keep hold of it as a mutable so we can change/create it and
     // have the changes automatically saved
-    Mutable${name}AssetD asset = Find(ref_ $forwardtypearg);
+    using ::Find;
+    Mutable${name}AssetD asset = Find<Mutable${name}AssetD>(ref_, $typeref, ToString("$subtype"));
```
tst-lsavoie: This is very minor, but you can get the same effect by leaving out the `using` clause by putting the scope resolution operator on the beginning of the Find call:
```
Mutable${name}AssetD asset = ::Find<Mutable${name}AssetD>(ref_, $typeref, ToString("$subtype"));
``

File: earth_enterprise/src/fusion/autoingest/StorageManager.h [hunk 7]
```diff
@@ -87,6 +87,9 @@ class StorageManager : public StorageManagerInterface<AssetType> {
     
     // Pass a handle to a non-const so callers can modify it.
     AssetHandle<AssetType> GetMutable(const AssetKey &);
+
```
tst-lsavoie: Remove commented out code.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -286,5 +289,61 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
   return true;
 }
 
+template<class Asset>
+Asset Find(const std::string & ref, const AssetDefs::Type & type, const std::string & subtype)
+{
+  notify(NFY_WARN, "Find: %s\t%s\t%s", ref.c_str(), ToString(type).c_str(), subtype.c_str());
```
tst-lsavoie: I assume these notify statements are for debugging. We should remove them now (all instances in all 3 functions).

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -105,57 +105,6 @@ ()
 // ****************************************************************************
 // ***  ${name}Factory - Auto generated
 // ****************************************************************************
-${name}AssetD
```
tst-lsavoie: We need to remove these from the .h file as well, I think.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -286,5 +289,61 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
...
+            return Version(SharedString(ref));
+        }
+    } catch (...) {
+        // do nothing - don't even generate any warnings
+    }
+    return Version();
+}
+
+template<class Asset, class Version>
+void ValidateRefForInput(const std::string & ref, const AssetDefs::Type & type, const std::string & subtype)
```
tst-lsavoie: It occurs to me that the asset and version types passed in here have to match, so a programmer that's not being careful enough could write code that will never work, and I don't think the compiler will find the error. Thinking out loud, we could get around that by defining the version type inside the asset class:
```
class CombinedRPAssetD {
  public:
    using Version = CombinedRPAssetVersionD;
};
```
Then you can just have one template parameter for the asset type and get the version parameter from that with `Asset::Version`.

I'll leave it up to you to decide whether to implement that here.

earth_enterprise/src/fusion/autoingest/plugins.src/Database.src
```diff
@@ -228,17 +228,15 @@ void DatabaseAssetImplD::ConstructInputsFromRequest(
                             std::vector<SharedString>* inputs) {
   // make sure the project asset_ref_string's are valid
   if (!req.config.imageryProject.empty()) {
-    RasterProjectFactory::ValidateRefForInput(req.config.imageryProject,
-                                              AssetDefs::Imagery);
+    ValidateRefForInput<MutableDatabaseAssetD, MutableDatabaseAssetVersionD>(req.config.imageryProject, AssetDefs::Imagery, kDatabaseSubtype);
```
tst-lsavoie: There's still an issue here. Even though this call is inside `MutableDatabaseAssetD`, it's calling `RasterProjectFactory`. So the call should be something like `ValidateRefForInput<MutableRasterProjectAssetD, MutableRasterProjectAssetVersionD>`. Same thing with the other calls.

earth_enterprise/src/fusion/autoingest/plugins.src/MapLayer.src
```diff
@@ -73,8 +73,9 @@ MapLayerAssetImplD::HandleEditRequest(const MapLayerEditRequest &req)
   std::vector<SharedString> productrefs;
   productrefs.reserve(req.config.subLayers.size());
   for (uint s = 0; s < req.config.subLayers.size(); ++s) {
-    VectorProductFactory::ValidateRefForInput
-      (req.config.subLayers[s].asset_ref);
+    //VectorProductFactory::ValidateRefForInput
```
tst-lsavoie: Remove commented out code.

earth_enterprise/src/fusion/autoingest/plugins.src/MapProject.src
```diff
@@ -67,7 +67,8 @@ MapProjectAssetImplD::HandleEditRequest(const MapProjectEditRequest &req)
          = req.config.layers.begin();
        layer != req.config.layers.end(); ++layer) {
     if (!layer->skipLayer) {
-      MapLayerFactory::ValidateRefForInput(layer->assetRef);
+      //MapLayerFactory::ValidateRefForInput(layer->assetRef);
```
tst-lsavoie: Remove commented out code.

tst-lsavoie:
Replaced by PR #1396.

PR #1385, opened 2019-08-01
Move load save into storage manager

tst-sbayard:
As part of the code clean-up and simplification effort, asset loading and saving code is being moved into the storage manager.


earth_enterprise/src/fusion/autoingest/Asset.h
```diff
@@ -38,7 +38,7 @@
  ***
  ******************************************************************************/
 class AssetImpl : public AssetStorage, public StorageManaged {
-  friend class AssetHandle_<AssetImpl>;
+  friend class AssetHandle_<AssetImpl, AssetStorage>;
```
tst-lsavoie: Minor: you might be able to get away from all of these extra template parameters by adding `using Base = AssetStorage;` in this class and then referring to AssetType::Base in the storage manager.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -420,68 +420,17 @@ ()
...
-
-    return result;
+    return "${name}Asset";
 }
 
-extern void ToElement(DOMElement *elem, const AssetStorage &self);
-
-bool
-${name}AssetImplD::Save(const std::string &filename) const
+void ${name}AssetImplD::SerializeConfig(DOMElement *top) const
```
tst-lsavoie: Minor: just throwing ideas out there. If we had functions that translated between generic XML and something like a key/value store we could update AddConfig so that it gets its data from the key/value store and not directly from the XML objects. That way only the serializer needs to know about XML.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -38,9 +43,55 @@ class StorageManaged {
     StorageManaged() : timestamp(0), filesize(0) {}
 };
 
+template<class AssetType, class AssetStorageType> class AssetHandleInterface;
+
+template<class AssetType, class AssetStorageType>
+class AssetSerializerLocalXML
```
tst-lsavoie: Minor: maybe ought to move this class to a separate file.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -188,9 +241,9 @@ StorageManager<AssetType>::Get(
   return entry;
 }
 
-template<class AssetType>
-typename StorageManager<AssetType>::PointerType
-StorageManager<AssetType>::GetEntryFromCacheOrDisk(const AssetKey & ref) {
+template<class AssetType, class AssetStorageType>
```
tst-lsavoie: Should update this function to use the serializer as well instead of calling Load directly.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -264,15 +317,17 @@ void StorageManager<AssetType>::Abort() {
...
+bool StorageManager<AssetType, AssetStorageType>::SaveDirtyToDotNew(
     khFilesTransaction &savetrans,
     std::vector<AssetKey> *saved) {
   notify(NFY_INFO, "Writing %lu %s records", dirtyMap.size(), assetType.c_str());
   typename std::map<AssetKey, PointerType>::iterator entry = dirtyMap.begin();
   while (entry != dirtyMap.end()) {
     std::string filename = entry->second->XMLFilename() + ".new";
-    if (entry->second->Save(filename)) {
+    AssetSerializerLocalXML<AssetType, AssetStorageType> serializer;
```
tst-lsavoie: Minor: might perform slightly better if we declare the serializer outside the loop. Another option is to make the serializer a member of StorageManager so you don't have to keep declaring it.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
```diff
@@ -46,7 +46,10 @@ class DerivedAssetHandleD_ : public virtual BaseD_, public ROBase_
   typedef typename BBase::HandleType HandleType;
  public:
   virtual HandleType Load(const std::string &boundref) const {
-    return HandleType(Impl::Load(boundref));
```
tst-lsavoie: Is this load function still necessary? It seems that it could be removed.

tst-lsavoie:
CI is failing because the unit tests are failing. Will those be in a separate PR?

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -38,9 +43,55 @@ class StorageManaged {
...
+template<class AssetType, class AssetStorageType> class AssetHandleInterface;
+
+template<class AssetType, class AssetStorageType>
+class AssetSerializerLocalXML
+{
+  public:
+    AssetSerializerLocalXML() {}
+    std::shared_ptr<AssetType> Load(const std::string &boundref)
+    {
+      return AssetType::Load(boundref);
```
tst-lsavoie: Minor: It would be nice if we could get this Load code out of the assets as well. Since AssetType::Load is autogenerated, we could change the autogeneration to make it a part of this class. To go all the way we'd need to move the Load logic inside each asset class implementation into this class as well. I'd be OK with making that a follow up ticket, though.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -38,9 +43,55 @@ class StorageManaged {
...
 };
 
+template<class AssetType, class AssetStorageType> class AssetHandleInterface;
+
+template<class AssetType, class AssetStorageType>
+class AssetSerializerLocalXML
+{
+  public:
+    AssetSerializerLocalXML() {}
+    std::shared_ptr<AssetType> Load(const std::string &boundref)
```
tst-lsavoie: Here and other places you can use AssetPointerType instead of std::shared_ptr. That makes it easier to switch between them.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -38,9 +43,55 @@ class StorageManaged {
...
+      return AssetType::Load(boundref);
+    }
+
+    bool Save(std::shared_ptr<AssetType> asset, std::string filename){
+      extern void ToElement(DOMElement *elem, const AssetStorageType &self);
+
+      std::unique_ptr<GEDocument> doc = CreateEmptyDocument(asset->GetName());
+      if (!doc) {
+          notify(NFY_WARN,
+                "Unable to create empty document: ${name}AssetVersion");
```
tst-lsavoie: This warning should call asset->getName() instead of printing `${name}AssetVersion`.

tst-lsavoie:
I've made a few comments (mostly minor) and I still need to test, but this looks really solid overall.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -21,11 +21,16 @@
...
+#include <memory>
 
 #include "common/khCache.h"
 #include "common/khFileUtils.h"
 #include "common/notify.h"
 #include "common/SharedString.h"
+#include "common/khxml/khxml.h"
+#include "common/khxml/khdom.h"
+
+using namespace khxml;
```
tst-lsavoie: Probably ought to pull up the include below this so that it's above the `using` statement.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -38,9 +43,55 @@ class StorageManaged {
     StorageManaged() : timestamp(0), filesize(0) {}
 };
 
+template<class AssetType, class AssetStorageType> class AssetHandleInterface;
```
tst-lsavoie: Minor: I don't think this line is necessary. It must have been pulled in as part of the merge.

earth_enterprise/src/fusion/autoingest/Asset.h
tst-sbayard: I like that. I wasn't super happy about adding the extra template parameter. It's changed now.

earth_enterprise/src/fusion/autoingest/sysman/AssetHandleD.h
tst-sbayard: Turns out it wasn't necessary. I took that out and the base class version in AssetHandle.h.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
tst-sbayard: I really like that idea, though at this point I'd rather not include it in this PR. I'll put a comment above the SerializeConfig definitions in Asset.h and AssetVersion.h as a note for future development. I think the usefulness of that change would come into play more if/when different methods of serialization are added.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: Moved out to a new file: AssetSerializer.h. I left the name of the new file generic.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: It got moved out to AssetSerializer.h anyway.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: I think those new calls came in after I merged in from master, and I just didn't notice it. Nice catch.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: Agreed 100%, but I'd definitely lean toward doing that as part of follow-up work.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: Changed as suggested.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: Changed as suggested.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: Good catch.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-sbayard: Good catch. Removed it.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
tst-lsavoie: Works for me.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
```diff
@@ -0,0 +1,55 @@
...
+              status = WriteDocument(doc.get(), filename);
+
+              if (!status && khExists(filename)) {
+                  khUnlink(filename);
+              }
+          } else {
+              notify(NFY_WARN, "Unable to create document element %s",
+                    filename.c_str());
+          }
+      } catch (const std::exception &e) {
```
tst-lsavoie: I think it would be a good idea to keep the full error handling here (i.e., catch XMLException and DOMException separately).

tst-lsavoie:
My build is failing on CentOS 6.10 now for some reason, so we'll have to get that fixed before merging.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
```diff
@@ -0,0 +1,55 @@
+#ifndef ASSET_SERIALIZER_H
+#define ASSET_SERIALIZER_H
+
+#include "common/khxml/khxml.h"
+#include "common/khxml/khdom.h"
+
+using namespace khxml;
```
tst-lsavoie: It's generally not a good idea to have a `using` directive in a header file.

tst-lsavoie:
Replaced by PR #1390.

Issue #1376, opened 2019-07-23
Portable Globe Cutter fails the last step when using HTTPS


tst-nfarah:
- Use GEE server on https build a 3D database (ex. using blue marble geotiff tutorial data).
- Using KML polygon file, try to cut a a portable globe.
--> This produce an error on last step: ```
Extract search data ...
Querying search poi ids: target=/globe2
FAILED: Request failed: cannot connect to server
FAILED
```
--> Expected behavior: the globe should have been cut without an error.

When switching over to HTTP the cut worked without an error.

PR #1391, opened 2019-08-07
1376 Use config for Cutter API calls

tst-nrichman:
Fixes issue #1376 .  Uses the advanced configuration to set a URL that will be used when cutter tries to make an openurl call. As these calls happen internally by default, this defaults to localhost.


tst-jlarocco:
Looks good to me!

tst-ccamp:
I see 2 approvals and looks like it's been tested, so merging.

PR #1397, opened 2019-08-13
Parallelize a small section of the packgen processing

tst-lsavoie:
This is an attempt to speed up packgen processing by parallelizing a small section of code.

All commits from @tst-jkent were previously submitted to this repo.


docs/geedocs/5.3.2/answer/176738.html
```diff
@@ -294,6 +294,8 @@ <h2 id="cache_configuration">Cache configuration</h2>
 # default is 3<br>
 PURGE_LEVEL=&lt;value&gt;
 </code>
+<h2 id="packgen_thread_configuration"> PackGen thread configuration</h2>
+Within <code>misc.xml</code>, there is an experimental feature for setting the number of threads to use during the packgen process. The number of threads is controlled by <code>PackGenThreads</code>.
```
tst-dsheremata: The name of this configuration variable is too ambiguous because the actual PackGen task can also be configured to use a variable number of threads.   (See http://www.opengee.org/geedocs/5.2.2/answer/6033487.html and http://www.opengee.org/geedocs/5.2.2/answer/176738.html).   This variable name and the associated docs need to distinguish that this variable controls the number of threads used to analyze the dependencies and define the actual packgen tasks to be executed later in the build.  Perhaps something like "PackGenAnalysisThreads" or "SysmanPackGenThreads"?

tst-lsavoie: Changed to `PackGenAnalysisThreads`. Also rewrote the doc text a bit.

PR #1402, opened 2019-08-15
Log when cache size is exceeded

### LINKED ISSUE #1401: Log when cache size is exceeded
The asset and asset version caches can exceed their specified sizes because they do not remove elements that are currently in use by other parts of the code. This includes any assets that have not yet been written to disk. However, when the number of assets is much higher than the size of the cache the entire cache can fill up with objects that haven't been written to disk. Eventually, every Purge call will have to loop through the entire cache looking for something to remove, which could become a performance issue. This ticket is to add logging statements to the Purge function when the cache size is exceeded. The logging statement should include the configured size of the cache, the actual number of items in the cache, and the amount of time it took to run the Purge function.

tst-lsavoie:
Fixes #1401 by adding a log statement when the cache size is exceeded.

To test:
1. Set the log level to 4 or more.
2. Set the cache sizes very low in misc.xml (I used 10).
3. Run a test (I used the gauge tests).
4. Look for log statements in gesystemmanager.log indicating that there are too many items in the cache(s).
5. Set the logs to be limited by memory utilization and set the memory size very low (I used 5000 bytes).
6. Run a test (I used the gauge tests again).
7. Look for log statements in gesystemmanager.log indicating that the size of the caches have exceeded the specified number of bytes.

### DIFF
--- a/earth_enterprise/src/common/khCache.h
+++ b/earth_enterprise/src/common/khCache.h
@@ -18,6 +18,7 @@
 #ifndef __khCache_h
 #define __khCache_h
 
+#include <cctype>
 #include <chrono>
 #include <map>
 #include <vector>
@@ -80,6 +81,7 @@ class khCache {
   Item *head;
   Item *tail;
   const uint targetMax;
+  const std::string name;
 #ifdef SUPPORT_VERBOSE
   bool verbose;
 #endif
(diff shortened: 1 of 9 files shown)


earth_enterprise/src/common/khCache.h
```diff
@@ -322,6 +326,29 @@ class khCache {
       }
     }
     CheckListInvariant();
+    TimePoint finishTime = std::chrono::high_resolution_clock::now();
+    // Log when the cache size is exceeded
+    if (TooMuchMemory() || TooManyObjects()) {
```
tst-sbayard: Looks like this is being called after the pruning is already done. Seems like `TooMuchMemory()` and/or `TooManyObjects()` would return false after a successful pruning, preventing `LogPruneTime()` from being called.

tst-lsavoie: If `use_count() > 1` is true for lots of items, the Prune function can finish with the cache oversize. That's when we want to print the log message.

tst-sbayard: Okay, I misunderstood the intent. I thought it was to show a log any time the cache size was exceeded and about to be pruned. But the goal is to log when the cache limit is still exceeded after being pruned. Got it.

PR #1406, opened 2019-08-16
Skip Internal Certificate Verification

### LINKED ISSUE #1404: WMS fails using HTTPS
When making a WMS connection to a database served over SSL, the server fails by internally trying to verify its own certificate without being set up properly to do so.

**To Reproduce**
1. GEEP running HTTPS
2. Create map database
3. Publish as WMS with an SSL virtual host
4. Attempt to use WMS service in a WMS client

tst-nrichman:
Fixes #1404. Allows Python to bypass the internal certificate verification by default. Extra security configurations are also added to allow users to redirect the host and verify against specific certificates.

Verification Steps:
1. Install open gee.
2. Obtain an SSL certificate (preferably third party) and set up GEE Server for SSL.
3. Set a virtual host as an SSL server.
4. Publish a database over your SSL virtual host and enable WMS.
5. Ensure you can connect to your database over https with a WMS client. (QGIS is a free, open source tool https://qgis.org/en/site/)

### DIFF
--- a/earth_enterprise/src/server/wsgi/wms/ogc/common/configs.py
+++ b/earth_enterprise/src/server/wsgi/wms/ogc/common/configs.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+#
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Utilities for getting config parameters."""
+
+
+class Configs(object):
+  """Class for loading and provide configuration parameters."""
+
+  def __init__(self, config_file, default_values=None):
+    """Load config key/value pairs from the config file."""
+    if default_values:
+      self._config_values = default_values
+    else:
+      self._config_values = {}
+    try:
+      fp = open(config_file)
+      for line in fp:
+        line = line.strip()
+        if line and line[0] != "#":
+          (key, value) = line.split(" ")
+          self._config_values[key] = value
+      fp.close()
+    except IOError:
+      pass
+    except:
+      pass
+
+  def GetStr(self, key):
+    """Returns the value for the given key from the config file."""
+    try:
+      return self._config_values[key]
+    except KeyError:
+      return ""
+
+  def GetInt(self, key):
+    """Returns the integer value for the given key from the config file."""
+    try:
+      return int(self._config_values[key])
+    except KeyError:
+      return 0
+
+  def GetBool(self, key):
+    """Returns the boolean value for the given key from the config file."""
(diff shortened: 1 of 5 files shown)


earth_enterprise/src/server/wsgi/wms/ogc/common/configs.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+#
+# Copyright 2017 Google Inc.
```
tst-jlarocco: Do these copyright notices need to be updated to 2019 and The OpenGEE Contributors?

earth_enterprise/src/server/wsgi/wms/ogc/wms.cfg
```diff
@@ -0,0 +1,9 @@
+# Toggle to enable and disable certificate validation
+VALIDATE_CERTIFICATE False
```
tst-jlarocco: Should this default to True?

earth_enterprise/src/server/wsgi/wms/ogc/common/configs.py
tst-nrichman: Wasn't sure when I was making the changes, most of the Copyrights have the same year and name. Do these ones need to change because it's a new file?

earth_enterprise/src/server/wsgi/wms/ogc/wms.cfg
tst-nrichman: No. By default it should just work out of the box. Configuring certificate validation is optional.

earth_enterprise/src/server/wsgi/wms/ogc/common/configs.py
tst-ccamp: Since this appears to be a new file, then yes

tst-ccamp: If it was just copied from a different directory, then no

tst-nrichman: This file was copied from fusion/portableglobe/cutter.

earth_enterprise/src/server/wsgi/wms/ogc/common/tiles.py
```diff
@@ -286,6 +286,8 @@ def _SetTransPixelToBgcolor(tile, bgcolor):
 
   return tile
 
+key_file = '/opt/google/gehttpd/conf/privkey.pem'
+cert_file = '/opt/google/gehttpd/conf/fullchain.pem'
```
tst-rwildes: Where do these variables get used?

tst-nrichman: Oops. Leftover from development.

earth_enterprise/src/server/wsgi/wms/ogc/common/wms_connection.py
```diff
@@ -0,0 +1,57 @@
...
+      context.check_hostname = False
+      context.verify_mode = ssl.CERT_NONE
+
+    fp = urllib2.urlopen(url, context=context)
+    response = fp.read()
+  except urllib2.HTTPError, e:
+    logger.warning("Server definitions didn't return any results %s.", e)
+    return {}
+  finally:
+    if fp:
```
tst-rwildes: should set `fp` to `None` at the beginning of the function (before `try`) or you could get an error accessing an uninitialized variable

PR #1413, opened 2019-08-26
Eliminate redundant state propagation

tst-lsavoie:
This change eliminates several cases of redundant state propagation during the new resume operation. The new code for resume implements its own state propagation, but several calls that happen as part of the OnStateChange function invoke the old method of state change propagation. Specifically, I changed the following:

- When there is an error, instead of setting the state of the asset to Failed immediately, I throw an error that is caught by the StateUpdater, which sets the state to Failed.
- For cases in which a state change triggers another state change, I allow OnStateChange to return a new state and continue looping and changing state until no more state changes are needed. This capability is only used in CompositeAssetVersionImplD::OnStateChange when setting the state to Succeeded.
- There is one case that is not easily handled. Some assets cannot create their children until their inputs have completed. In these cases, OnStateChange calls DelayedBuildChildren to create the children of the asset. StateUpdater doesn't currently handle asset trees that change in the middle of an operation (such as by adding children). Thus, in this case, I revert to the legacy state propagation and abandon the operation in StateUpdater.  This is the call to SyncState in CompositeAssetVersionImplD::OnStateChange. When we work on cleaning up the build operation we can handle this in a better way.

Note: in order to run this code you have to set GraphOperations to 1 in misc.xml. That doesn't apply to unit tests, though, as those don't use misc.xml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,11 @@ before_install:
     libboost-all-dev libc6 libc6-dev libcap-dev libfreetype6 libfreetype6-dev
       libgdbm-dev libgeos-dev libgeos++-dev libgif-dev libgtest-dev
       libjpeg-dev libjpeg8-dev libmng-dev libogdi3.2-dev
+      libpcre3 libpcre3-dev
       libperl4-corelibs-perl libpng12-0 libpng12-dev libpq-dev libproj-dev
       libstdc++6 libtool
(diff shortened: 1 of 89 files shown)


tst-basilhuffman:
some integration tests failed the first time around. disappeared upon seccessive tests. that being said, lgtm

PR #1423, opened 2019-08-30
Optimize child state change handling

### LINKED ISSUE #1418: Optimize child state change handling
The function HandleChildStateChange checks all of an assets children and calculates the asset's state based on the states of the children. However, in many cases, the state of the parent asset won't change. For example, if a parent asset has 100 children that are in progress, and one of them finishes, the parent's state won't change because it's still waiting on 99 other children. We should implement optimizations so that we don't unnecessarily scan through all of an asset's children every time one of it's children changes state, similar to the optimizations implemented in HandleInputStateChange.

tst-lsavoie:
Parents are notified whenever their child's state changes. When this happens, they check the states of all of their children to calculate their own state. However, in many cases the parent's state will not change: for example, if a parent is waiting on 100 children and one of them transitions to Succeeded, the parent's state will not change because it is still waiting on 99 other children. Thus, the constant re-checking of child states every time a parent receives a notification is unnecessary.

There is a similar issue when listeners are notified that their input state changed, but their are optimizations in HandleInputStateChange that reduce the extra work. Listeners keep track of the number of inputs they are waiting for and don't recalculate their state until that number goes to 0. This PR implements the same kind of optimization in HandleChildStateChange.

I also removed the PropagateProgress handling that loaded all the parents just so it could call an empty function on them.

This PR also updates the release notes for other work done in 5.3.2 that has not yet been included in release notes.

And, of course, I did some minor clean up and refactoring. The code related to `InNodeVertexIndexMap` is unchanged; it's only moved to a different file to make things cleaner.

Fixes #1418.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/sysman/InNodeVertexIndexMap.h
+++ b/earth_enterprise/src/fusion/autoingest/sysman/InNodeVertexIndexMap.h
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2019 The Open GEE Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef IN_NODE_VERTEX_INDEX_MAP_H
+#define IN_NODE_VERTEX_INDEX_MAP_H
+
+#include "DependentStateTree.h"
+
+// The depth_first_search function needs a way to map vertices to indexes. We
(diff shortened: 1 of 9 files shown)


tst-lsavoie:
My correctness tests have passed. I'm still running performance tests to quantify the improvement, but I'm going ahead and moving this out of WIP so that reviews can start.

earth_enterprise/src/fusion/autoingest/sysman/InNodeVertexIndexMap.h
```diff
@@ -0,0 +1,57 @@
...
+// The depth_first_search function needs a way to map vertices to indexes. We
+// store a unique index inside each vertex; the code below provides a way for
+// boost to access them. These must be defined before including
+// depth_first_search.hpp.
+template <class Graph>
+class InNodeVertexIndexMap {
+  public:
+    typedef boost::readable_property_map_tag category;
+    typedef size_t value_type;
+    typedef value_type reference;
```
tst-sbayard: I'm not seeing 'reference' used anywhere, at least in this file. Is this typedef needed by boost?

tst-lsavoie: Yes, it's used by boost. If I remove any of the typedefs I get a compile error.

PR #1424, opened 2019-08-30
Consolidate make functions

tst-sbayard:
As part of the effort to consolidate common functionality and reduce reliance on auto-generated code, the "Make" methods are being moved to a central location.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/AssetFactory.h
+++ b/earth_enterprise/src/fusion/autoingest/AssetFactory.h
@@ -23,6 +23,56 @@
 // asset creation is handled.
 namespace AssetFactory
 {
+  template<class AssetType>
+  AssetType Find(const std::string & ref, const AssetDefs::Type & type)
+  {
+    assert(type != AssetDefs::Invalid);
+    const std::string subtype = AssetType::Impl::EXPECTED_SUBTYPE;
+    try {
+      AssetType asset(ref);
+      if (asset &&
+          (asset->type == type) &&
+          (asset->subtype == subtype)) {
+          return asset;
+      }
+    } catch (...) {
+        // do nothing - don't even generate any warnings
+    }
+    return AssetType();
+  }
+
+  template<class AssetType>
+  AssetType Find(const std::string & ref) {
+    return Find<AssetType>(ref, AssetType::Impl::EXPECTED_TYPE);
+  }
+
+  template<class VersionType>
+  void ValidateRefForInput(const std::string & ref, const AssetDefs::Type & type)
+  {
+    assert(type != AssetDefs::Invalid);
+    using AssetType = typename VersionType::Impl::AssetType;
+    const std::string subtype = VersionType::Impl::EXPECTED_SUBTYPE;
+    if (AssetVersionRef(ref).Version() == "current") {
+      AssetType asset = Find<AssetType>(ref, type);
+      if (!asset) {
+        throw std::invalid_argument(
+            "No such " + ToString(type) + " " + subtype + " asset: " + ref);
+      }
+    } else {
+      VersionType version = Find<VersionType>(ref, type);
+      if (!version) {
+        throw std::invalid_argument(
+            "No such " + ToString(type) + " " + subtype + " asset version: " +
+            ref);
+      }
+    }
+  }  
+  
+  template<class VersionType>
+  void ValidateRefForInput(const std::string & ref) {
+    return ValidateRefForInput<VersionType>(ref, VersionType::Impl::EXPECTED_TYPE);
+  }
+ 
   template<class MutableDerivedAssetHandleType, class ConfigType>
   MutableDerivedAssetHandleType Make( const std::string &ref_,
                                       AssetDefs::Type type_,
@@ -30,8 +80,9 @@ namespace AssetFactory
                                       const ConfigType &config)
   {
     using Impl = typename MutableDerivedAssetHandleType::Impl;
+    using AssetStorageType = typename Impl::Base;
     return MutableDerivedAssetHandleType(std::make_shared<Impl>
-                                            (AssetStorage::MakeStorage(
+                                            (AssetStorageType::MakeStorage(
                                         ref_, type_, 
                                         Impl::EXPECTED_SUBTYPE,
                                         std::vector<SharedString>(), meta),
@@ -46,8 +97,9 @@ namespace AssetFactory
                                       const ConfigType &config)
   {
     using Impl = typename MutableDerivedAssetHandleType::Impl;
+    using AssetStorageType = typename Impl::Base;
     return MutableDerivedAssetHandleType(std::make_shared<Impl>
-                                            (AssetStorage::MakeStorage(
+                                            (AssetStorageType::MakeStorage(
                                         ref_, type_, 
                                         Impl::EXPECTED_SUBTYPE,
                                         inputs_, meta),
@@ -61,8 +113,9 @@ namespace AssetFactory
                                       const ConfigType &config)
   {
     using Impl = typename MutableDerivedAssetHandleType::Impl;
+    using AssetStorageType = typename Impl::Base;
     return MutableDerivedAssetHandleType(std::make_shared<Impl>
-                                            (AssetStorage::MakeStorage(
+                                            (AssetStorageType::MakeStorage(
                                         ref_, Impl::EXPECTED_TYPE, 
                                         Impl::EXPECTED_SUBTYPE,
                                         inputs_, meta),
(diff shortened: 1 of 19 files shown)


earth_enterprise/src/fusion/autoingest/AssetFactory.h
```diff
@@ -0,0 +1,115 @@
...
+#ifndef ASSET_FACTORY_H
+#define ASSET_FACTORY_H
+
+#include <common/SharedString.h>
+#include <common/khMetaData.h>
+
+// The goal of this class is to have a single place in the code where all
+// asset creation is handled.
+template<class MutableDerivedAssetHandleType>
+class AssetFactory
```
tst-lsavoie: Thinking out loud: since these are static functions, I'm not sure that there's value in wrapping them in a class. It might be fine to make them regular old non-member functions. On the other hand, having the class or a namespace might be a worthwhile organizational tool, especially since the functions have simple names like `Find` and `Make`. I'm undecided.

tst-sbayard: That crossed my mind. I leaned slightly toward consolidating into a class, but maybe a namespace would make more sense.

tst-lsavoie:
@tst-sbayard You've got a commit on here that doesn't have your name on it, so the PR is flagged as not having CLAs signed. I think you can change it using this procedure: https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit. @tst-basilhuffman this means you will have to rebase on the updated branch as well.

earth_enterprise/src/fusion/autoingest/AssetFactory.h
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+  }  
+  
+  template<class VersionType>
+  void ValidateRefForInput(const std::string & ref) {
+    return ValidateRefForInput<VersionType>(ref, VersionType::Impl::EXPECTED_TYPE);
+  }
+ 
+  template<class MutableDerivedAssetHandleType, class ConfigType>
```
tst-lsavoie: There's a potential problem here where a user could pass a `MutableDerivedAssetHandleType` that doesn't match the `ConfigType`. You can get around that by doing something similar to what you did for `AssetStorageType`. If each class has a `using` statement for its config type, then you only have to pass the handle type.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -643,14 +578,14 @@ ()
     if (!IsUpToDate(bound_cofig)) {
         Mutable${name}AssetD self(GetRef());
         Mutable${name}AssetVersionD newver =
-            self->MakeNewVersion(bound_config);
+            MakeNewVersion<Mutable${name}AssetD, ${name}AssetImplD::Config, Mutable${name}AssetVersionD>(self, bound_config); //self->MakeNewVersion(bound_config);
```
tst-lsavoie: Remove commented out code.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonCpp.pl
```diff
@@ -643,14 +578,14 @@ ()
...
 EOF
 }else {
     print $fh <<EOF;
     // now see if I'm up to date
     if (!IsUpToDate()) {
         Mutable${name}AssetD self(GetRef());
-        Mutable${name}AssetVersionD newver = self->MakeNewVersion();
+        Mutable${name}AssetVersionD newver = MakeNewVersion<Mutable${name}AssetD, Mutable${name}AssetVersionD>(self);//self->MakeNewVersion();
```
tst-lsavoie: Remove commented out code.

earth_enterprise/src/fusion/autoingest/AssetFactory.h
tst-lsavoie: Also, there's an opportunity to combine these 3 functions. You can have one function that takes all the possible parameters and then have the other two functions call that function, passing in default parameters where appropriate. I'm not sure that saves us much since the function body is effectively one line, though.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -30,6 +30,9 @@
 #include "common/SharedString.h"
 #include "StorageManagerAssetHandle.h"
 #include "AssetSerializer.h"
+#include "AssetFactory.h"
```
tst-lsavoie: Are these lines necessary? Since you moved Find and ValidateRefForInput, I don't think they're used.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -30,6 +30,9 @@
...
+
+using namespace AssetFactory;
```
tst-lsavoie: Generally speaking it's not a good idea to have a `using namespace` statement in a header file. I can understand why it makes life easier here, though.

earth_enterprise/src/fusion/autoingest/sysman/AssetD.h
```diff
@@ -56,4 +62,26 @@ class AssetImplD : public virtual AssetImpl
 typedef DerivedAssetHandle_<Asset, AssetImplD> AssetD;
 typedef MutableAssetHandleD_<AssetD> MutableAssetD;
 
+template<typename MutableAssetHandleType, typename ConfigType, typename MutableVersionHandleType>
```
tst-lsavoie: Asset versions already have an AssetType field that specifies their corresponding asset type. You can use that to eliminate the need to pass both the asset and asset version type in these two functions.

earth_enterprise/src/fusion/autoingest/AssetFactory.h
tst-sbayard: > Also, there's an opportunity to combine these 3 functions.
I initially went that route, but I ran into a seg fault that seemed to have to do with having `const std::vector<SharedString>& inputs_ = std::vector<SharedString>()` and then passing `inputs_` into `AssetStorageType::MakeStorage`.
Maybe I'm remembering that incorrectly. I can try it again to see if that works.

tst-lsavoie: It's not enough duplication that I'm very concerned about it. I'm OK either way.

tst-lsavoie:
Unit tests, integration tests, and tutorial tests have passed for me, so aside from my previous comments everything looks good.

earth_enterprise/src/fusion/autoingest/AssetFactory.h
tst-sbayard: I started looking again at trying to consolidate them. It'd take a little reworking of the auto-generated function arguments since the default parameters would need to be at the end. And some of those auto-gen arguments include commas, some don't. And they're used other places besides these functions. It looks likely to be error-prone and more work than I'd hope.
At this point, I think it's best to leave it as is.

earth_enterprise/src/fusion/autoingest/sysman/AssetD.h
tst-sbayard: Not sure that helps in this situation since one of the parameters is an asset type, so we can't have a parameter that's `MutableVersionHandleType::AssetType`.

earth_enterprise/src/fusion/autoingest/plugins.src/Database.src
```diff
@@ -263,7 +263,7 @@ DatabaseAssetImplD::HandleEditRequest(const DatabaseEditRequest &req)
     db->Modify(inputs, req.meta, dbConfig);
   } else {
     dbConfig.indexVersion = 1;
-    db = DatabaseFactory::Make(dbname, inputs, req.meta, dbConfig);
+    db = AssetFactory::Make<MutableDatabaseAssetD, DatabaseConfig>(dbname, inputs, req.meta, dbConfig);
```
tst-lsavoie: Very minor: I think you can get rid of the second template parameter here and several other places.

earth_enterprise/src/fusion/autoingest/sysman/AssetD.h
```diff
@@ -56,4 +62,26 @@ class AssetImplD : public virtual AssetImpl
...
 typedef MutableAssetHandleD_<AssetD> MutableAssetD;
 
+template<typename MutableVersionHandleType, typename MutableAssetHandleType, typename ConfigType>
+MutableVersionHandleType AssetImplD::MakeNewVersion(MutableAssetHandleType &asset, const ConfigType &config)
+{
+    typedef typename MutableVersionHandleType::Impl VerImplType;
+    MutableVersionHandleType newver(std::make_shared<VerImplType>
+                                      (asset.operator->(), config));
+
+    asset->AddVersionRef(newver->GetRef());
```
tst-lsavoie: I guess there's a concern here where MutableVersionHandleType and MutableAssetHandleType don't necessarily have to match and it would still compile. I can live with that, though.

PR #1440, opened 2019-09-10
New rst folders

tst-swanson:
This should have the following folders of rst files:
geeServerAdmin
installGEE
portable
troubleshooting
releaseNotes
legal


tst-swanson:
files in wrong location

PR #1461, opened 2019-09-23
Used sphinxprettysearchresults plugin for better search preview

tst-hrajamoney:
files to review : docs/geedocs/docsrc/BUILD_DOCS.md
docs/geedocs/docsrc/conf.py
.gitignore


docs/geedocs/docsrc/BUILD_DOCS.md
```diff
@@ -0,0 +1,175 @@
...
+      ```
+
+      Keep the change, which has |release| instead of 5.3.3. This will help
+      update the documentation based on the current opengee release.
+
+      Merges may become tricky if there are conflicts in the generated html.
+      if for example, we are bring 5.3.2 changes in to master we will be having 5.3.2
+      html changes. you can keep the change in 5.3.2 branch while resolving conflicts.
+      Resolve conflicts in rst and generate html for master which will update 5.3.3
+      html.
```
tst-rwildes: Would we not want to just blindly accept the html generated results?  If so I'm not sure how things become tricky?

tst-hrajamoney: It should be straight forward... unless some master changes accidentally got in to release_5.3.2 folder like last time. We should never marge master changes in to release_5.3.2 folder after we branched. If we need to fix something after we branched, we need to push to 5.3.2 branch first and then merge those changes in to master.

PR #1468, opened 2019-09-25
WIP: Using quadtree paths to calculate overlap

tst-jkent:
Wrote an experimental algorithm to calculate the quadtree paths for a test set of images and then determine the overlap with another image.
Wrote a unit test that will compare the results of the current algorithm for calculating overlap to that of the experimental algorithm.


tst-lsavoie:
Closing since this is experimental code and not intended to be checked in. We can always reopen this PR if needed.

PR #1470, opened 2019-09-25
I1469 glc assembly should check kml for validity

tst-eclamar:


tst-eclamar:
Erroneous PR.  Only 2 files changed, not 858.

PR #1474, opened 2019-09-27
Issue 1472: Fix broken maps.json and others in Portable Files

tst-jlarocco:
Use 'source' from input form instead of DATABASE_HOST
Added HttpGet method for making HTTP requests in GlobeBuilder
Replaced urllib.urlopen calls with calls to HttpGet

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
@@ -30,7 +30,9 @@ import re
 import shutil
 import sys
 import time
+import ssl
 import urllib
+import urllib2
 from lxml import etree
 
 from common import form_wrap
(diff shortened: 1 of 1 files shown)


tst-rwildes:
This also needs to be applied to 5.3.1.1 (you don't need to apply to 5.3.1 as that did not get the SSL fix)

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
```diff
@@ -344,6 +346,38 @@ def AppendInfoFile(self, message):
...
+        context.load_cert_chain(cert_file, keyfile=key_file)
+      else:
+        context = ssl.create_default_context()
+        context.check_hostname = False
+        context.verify_mode = ssl.CERT_NONE
+
+      fp = urllib2.urlopen(url, context=context)
+      http_status_code = fp.getcode()
+      response_data = fp.read()
+      fp.close()
```
tst-rwildes: Is it possible to use the `with` statement to auto close this?  If the read raised an exception would we leak the open?

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
```diff
@@ -438,11 +473,13 @@ def AddJsonIcons(self, source, json):
...
-        fp = urllib.urlopen(url)
-        fpw = open("%s/%s" % (self.icons_dir, icon), "w")
-        fpw.write(fp.read())
-        fp.close()
-        fpw.close()
+        data, code = self.HttpGet(url)
+        if code == 200:
+          fpw = open("%s/%s" % (self.icons_dir, icon), "w")
+          fpw.write(data)
+          fpw.close()
```
tst-rwildes: Also here can we use the `with` clause to ensure close is called?

https://www.pythonforbeginners.com/files/with-statement-in-python

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
```diff
@@ -512,12 +547,11 @@ def BuildSearchDb(self, source, polygon):
...
-          fp = urllib.urlopen(url)
-          http_status_code = fp.getcode()
+          data, http_status_code = self.HttpGet(url)
           if http_status_code == 200:
             self.Status("Copying search poi data: gepoi_%s to globe" % poi_id)
             fpw = open(search_file, "w")
-            fpw.write(fp.read().strip())
+            fpw.write(data.strip())
             fpw.write("\n")
             fpw.close()
```
tst-rwildes: I know this isn't part of your change but since you are changing things close but see if we can add a `with` here too.

tst-jlarocco:
Closing this also to retarget to 5.3.1.1

PR #1480, opened 2019-10-02
Add more dependencies to generated code

### LINKED ISSUE #1447: Compiler error: "FieldGenerator.h not found" when compiling using scons cache
**Before opening an issue**

Questions should go to the GEE mailing list (https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!forum/google-earth-enterprise)
or Slack Open GEE channel (http://slack.opengee.org/) GitHub issues are for bug reports and suggestions for new features.

**Describe the bug**

When building using scons cache a dependency bug is sometimes encountered.  This error does not happen every time as it depends on how fast things are extracted from cache but `geautoingest` library depends on generated header files but does not declare an explicit dependency to these files.  Implicit Dependencies 'should' catch this but there is a known issue in scons that prevents Implicit Dependencies from working well with generated files.  The work around is to use `Depends()` and explicitly add the generated files as dependencies to `geautoingest`

**To Reproduce**

With scons cache enabled build and re-build until this error happens.

**Expected behavior**

The compiler error should not happen

**Screenshots**

N/A

**Desktop**

 Build error.

**Additional context**

tst-rwildes:
Fix #1447

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/SConscript
+++ b/earth_enterprise/src/fusion/autoingest/SConscript
@@ -189,6 +189,12 @@ for idlfile in idlfiles:
 
 commonobj = map(lambda src: env.SharedObject(src), commoncpp)
 
+for obj in commonobj:
+    env.Depends(obj, commonh)
+
+for cppfile in commoncpp:
+    env.Depends(cppfile[:-3] + '.obj', commonh)
+
 
 geautoingest = env.sharedLib('geautoingest',
                              commonobj + [ 'ReadOnlyExtra.cpp',
--- a/earth_enterprise/src/fusion/autoingest/SConscript
+++ b/earth_enterprise/src/fusion/autoingest/SConscript
@@ -18,6 +18,7 @@
 
 Import('env')
 import string
+import os 
 origenv = env;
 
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/fusion/autoingest/SConscript
```diff
@@ -189,6 +189,12 @@ for idlfile in idlfiles:
 
 commonobj = map(lambda src: env.SharedObject(src), commoncpp)
 
+for obj in commonobj:
+    env.Depends(obj, commonh)
+
+for cppfile in commoncpp:
+    env.Depends(cppfile[:-3] + '.obj', commonh)
```
tst-blaird: I believe the slice and append of '.obj' is not removing the existing '.' from the cppfile string.  From a quick test:

```>>> s = "filename.cpp"
>>> s[:-3]
'filename.'```

It might be better to use `os.path.splitext(cppfile)[0] + '.obj'`.

tst-jlarocco (approved):
Looks good to me!

tst-blaird (approved):
Looks good.  Approved.

PR #1481, opened 2019-10-02
Include config fields for memory use calculation

tst-jkent:
Added GetSize methods to the various config classes related to the derived asset/assetversion objects.
Added GetSize method to derived asset/assetversion class to account for previous missed members.
Updated memory use calculation to utilize new GetSize methods.
Updated existing GetSize methods to account for padding.


earth_enterprise/src/common/khInsetCoverage.h
```diff
@@ -177,6 +178,15 @@ class khInsetCoverage {
                   uint stepOutSize,
                   uint paddingSize);
 
+  uint64 GetSize() {
+    uint64 extentsVecSize = 0;
```
tst-lsavoie: Isn't there a GetObjectSize method for this?

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -19,29 +19,44 @@
...
+#include "notify.h"
+#include <qstring.h>
 
 // determine amount of memory used by generic objects
 template<class T>
 inline uint64 GetObjectSize(const T &obj) {
     return sizeof(obj);
 }
-// determine amount of memory used by a vector and all of it's contents
+// determine amount of memory used by a vector's contents
```
tst-lsavoie: It just occurred to me that you could have a generic method to get the size of objects in any container, which would work for vectors, sets, lists, etc. I think the only thing it wouldn't work for is maps. Something like:
```
template<class Container>
inline uint64 GetObjectSize(const Container &container) {
  uint64 total = 0;
  for (const auto & elem : container) {
    total += GetObjectSize(elem);
  }
  return total;
}
```

I guess you  might run into trouble where the compiler tried to use that for everything, though, so maybe it wouldn't work.

earth_enterprise/src/common/khInsetCoverage.h
tst-jkent: There is a method for calculating the size of objects in a vector as well as other collections. The difference here is that CacheSizeCalculations currently does not have access to khExtents, as well as other classes that I've implemented GetSize mthods for. I did not include them in CacheSizeCalculations to avoid dependency problems.

earth_enterprise/src/common/CacheSizeCalculations.h
tst-jkent: The same issue here with the dependency problem that I was unsure of how to get around. I had definitely considered multiple options for addressing this, but this seemed like the simplest and most straightforward.

tst-jkent: If I include these other classes, I can add specific GetObjectSize methods that just call their respective GetSize method.

tst-lsavoie:
I want to do a more thorough review, but on a first pass the code seems reasonable. I have a couple of general thoughts:
- There's no obvious difference between the names `GetSize` and `GetObjectSize`. To make things worse, one of them calls `sizeof(*this)` and the other doesn't. I think the code would be clearer if you renamed one or both of them to make the difference clear. So maybe `GetSize` stays the same and `GetObjectSize` becomes `GetHeapUsage` to indicate that it only counts what's dynamically allocated (i.e., on the heap). Also, if there are cases where `GetObjectSize` includes `sizeof(*this)` and cases where it doesn't, that's extra confusing and we need to clean that up, maybe with a third function name.
- Along the same lines, it bugs me that in some cases you have to call `object.GetSize()` and in other cases you call `GetObjectSize(object)` to get the size of the object. I'd like to be able to call the same function for everything.
- I know this is a bit late, but I think in the AssetGen*.pl files (or maybe khidl.pl, I don't remember which) you have access to a list of member variables defined in the config file and their types. You could use that information to auto-generate the GetSize functions, which could save you from modifying every one of these files. Since you've already made the changes I'm not sure that helps, but it may be worth looking into because it could make future development easier.

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -19,29 +19,44 @@
...
 #include <memory>
 #include <vector>
+#include <set>
+#include "notify.h"
+#include <qstring.h>
 
 // determine amount of memory used by generic objects
 template<class T>
 inline uint64 GetObjectSize(const T &obj) {
     return sizeof(obj);
```
tst-lsavoie: I think you may want to return 0 here since you will already have a `sizeof(*this)` call in whatever calls this function.

earth_enterprise/src/common/khCache.h
```diff
@@ -205,7 +205,7 @@ class khCache {
   }
   // calculates the current size of a cache item
   uint64 calculateCacheItemSize(Item *item) {
-    return khCacheItemSize + GetObjectSize(item->key) + GetObjectSize(item->val);
+    return khCacheItemSize + GetHeapUsage(item->key) + GetHeapUsage(item->val);
```
tst-lsavoie: I think you want to call `GetSize` here instead of `GetHeapUsage`.

earth_enterprise/src/common/khxml/khidl.pl
```diff
@@ -1147,6 +1151,32 @@ sub DumpClass
...
+            if ($curr == $size) {
+                if ($haveFirst) {
+                    print $fh $pad, $indent x2, "+ ::GetHeapUsage($member->{name});\n";
+                }else {
+                    print $fh $pad, $indent x2, "return ::GetHeapUsage($member->{name});\n";
+                }
+                last;
+            }
+	        if ($haveFirst) {
+            print $fh $pad, $indent x2, "+ ::GetHeapUsage($member->{name})\n";
```
tst-lsavoie: Minor: there's some funny indentation going on here and in the `else` clause.

earth_enterprise/src/fusion/autoingest/Asset.h
```diff
@@ -78,15 +78,13 @@ class AssetImpl : public AssetStorage, public StorageManaged {
...
-    return(GetObjectSize(name)
-    + GetObjectSize(type)
-    + GetObjectSize(subtype)
-    + GetObjectSize(inputs)
-    + meta.GetSize()
-    + GetObjectSize(versions)
-    + GetObjectSize(timestamp)
-    + GetObjectSize(filesize));
+  virtual uint64 GetHeapUsage() const {
+    return ::GetHeapUsage(name)
```
tst-lsavoie: Any reason some of the members are missing here (i.e., `type`, `filesize`)? Even if they don't use the heap you can still call GetHeapUsage and it will return 0.

earth_enterprise/src/fusion/autoingest/AssetVersion.h
```diff
@@ -140,25 +140,19 @@ class AssetVersionImpl : public AssetVersionStorage, public StorageManaged {
...
-    + GetObjectSize(outfiles)
-    + meta.GetSize()
-    + GetObjectSize(beginTime)
-    + GetObjectSize(progressTime)
-    + GetObjectSize(endTime)
-    + GetObjectSize(taskid)
-    + GetObjectSize(timestamp)
-    + GetObjectSize(filesize));
+  virtual uint64 GetHeapUsage() const {
+    return ::GetHeapUsage(name)
```
tst-lsavoie: Same question: any reason some fields are missing?

earth_enterprise/src/fusion/autoingest/storage/CombinedRPConfig.idl
```diff
@@ -19,6 +19,7 @@
 #include <autoingest/.idl/storage/MaskgenConfig.h>
 #include <autoingest/.idl/storage/MosaicConfig.h>
 #include <autoingest/.idl/storage/KRPConfig.h>
+#include "CacheSizeCalculations.h"
```
tst-lsavoie: Minor: you could save yourself a little pain by adding this line as part of the autogen code as well.

earth_enterprise/src/fusion/autoingest/storage/UnifiedIndexConfig.idl
```diff
@@ -33,7 +34,7 @@ class UnifiedIndexEntry {
   uint32 channel_id_ = load:uint32(0);
   std::string date_string_ = std::string();  // Format expected: "YYYY-MM-DD"
 #pragma GenerateIsUpToDate
-}
+};
```
tst-lsavoie: Very minor: is this semicolon required? I didn't expect to see it.

earth_enterprise/src/common/khCache.h
tst-jkent: GetHeapSize has a specification for pointers/shared pointers that calls GetSize on the object pointed to. If I were to call GetSize with the current structure, it would call sizeof on the pointer/sharedpointer itself and then call GetHeapUsage on the pointer/sharedpointer. That would then detect the specific method to call and then call sizeof on the object pointed to and then GetHeapUsage again on the object pointed to. Since we calculate a cache items size excluding key and value, I figured this might not only be unnecessary but inaccurate. Not sure if I have the right idea.
We should call sizeof on the pointer itself right? Only the object pointed to?

earth_enterprise/src/fusion/khraster/khRasterProductLevel.h
```diff
@@ -135,7 +135,14 @@ class khRasterProductLevel
                              (uint64)tileExtents().endCol() *
                              (uint64)RasterProductTileResolution);
   }
-
+//TODO: finish def
```
tst-lsavoie: Reminder that you've got a TODO here

earth_enterprise/src/fusion/searchtabs/SearchTabs.idl
```diff
@@ -64,4 +64,4 @@ class SearchTabLabelOverrides {
       (SearchTabDefinition::DefaultPOIDefinition());
   }
 #/hquote
-}
+};
```
tst-lsavoie: Minor: again, not sure if the semicolon is necessary or if it matters

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -245,9 +245,9 @@ main(int argc, char *argv[])
         outline("Number of cached assets: %u", taskLists.num_assets_cached);
         outline("Number of cached asset versions: %u",
                 taskLists.num_assetversions_cached);
-        outline("Total memory used by cached assets: %s",
+        outline("Approx. memory used by cached assets: %s",
```
tst-lsavoie: Very minor: it occurred to me that we could save a couple of lines in the getop output by putting the number of items and the size on the same line, like this: `Number of cached assets: <x>, approx. memory used: <y> B`. That way we could save 2 lines and print out more information about running and waiting tasks. Of course, this has nothing to do with this change, but I thought I'd mention it in case you're feeling ambitious.

earth_enterprise/src/common/khCache.h
tst-lsavoie: I think you're missing the actual size of the key and value here. This line:
`const uint64 khCacheItemSize = sizeof(khCacheItem<Key, Value>) - sizeof(Key) - sizeof(Value);`
takes out the size of the Key and Value, and this line:
`return khCacheItemSize + GetHeapUsage(item->key) + GetHeapUsage(item->val);`
doesn't add them back in. So for the Value in particular, which is a shared pointer to something, you have the size of the object it points to but not the size of the pointer itself. You could fix that on either line, either by not subtracting the key and value size when calculating khCacheItemSize or by calling GetSize here.

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -19,29 +19,77 @@
...
-    return sizeof(obj);
+inline uint64 GetHeapUsage(const T* &ptr) {
+    uint64 ptrSize = (ptr ? GetSize(*ptr) : 0);
+    return ptrSize;
 }
-// determine amount of memory used by a vector and all of it's contents
+//determine amount of memory used by an object pointed to by a shared_ptr
 template<class T>
-inline uint64 GetObjectSize(const std::vector<T> &vec) {
+inline uint64 GetHeapUsage(const std::shared_ptr<T> &shptr) {
```
tst-lsavoie: Minor: for completeness you might also want a specialization for khRefGuard that's similar to the ptr and shared_ptr specializations. I don't think it matters because we're not using khRefGuards in the asset and asset version caches, though.

tst-jkent: I had one previously, but removed it since we're not using them, but I can re-implement it.

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -19,29 +19,77 @@
...
+#include <type_traits>
+#include "SharedString.h"
 
-// determine amount of memory used by generic objects
+//determine amount of memory used by an object pointed to by a pointer
 template<class T>
-inline uint64 GetObjectSize(const T &obj) {
-    return sizeof(obj);
+inline uint64 GetHeapUsage(const T* &ptr) {
+    uint64 ptrSize = (ptr ? GetSize(*ptr) : 0);
```
tst-lsavoie: This apparently works because it compiles, but I'm confused at how you're able to call GetSize in this function when GetSize is defined below. It seems like you would need a function declaration above this. Is there a chance that it's somehow calling the wrong function, or have you already verified that?

earth_enterprise/src/common/CacheSizeCalculations.h
tst-lsavoie: I'm happy either way. On the one hand, it doesn't make sense to include an unnecessary function. On the other, what if we do end up calling GetHeapSize on a khRefGuard at some point in the future and don't notice that it's not implemented? I think I'd come down on the size of leaving it out, but it's up to you. I probably shouldn't have said anything to begin with.

earth_enterprise/src/common/CacheSizeCalculations.h
tst-jkent: I hadn't thought of that, but I have verified it works. I'll move it for peace of mind.

earth_enterprise/src/common/khCache.h
tst-lsavoie: On a different subject, I was thinking about your concern about the performance cost of always calculating memory usage. We can get around that pretty easily by checking `limitCacheMemory` in this function. If it's true, we calculate sizes like we're doing now, and if it's false, we return 0. That's a simple enough fix that I think it's worth doing.

tst-jkent: Might be a good idea to not include that info in getop as well or else it will report no memory used if I'm not mistaken and that would be confusing.

earth_enterprise/src/common/CacheSizeCalculations.h
tst-lsavoie: If you move it you have the opposite problem because GetSize calls GetHeapUsage. I think the only "safe" path is to put just a function declaration at the top for GetSize - then everything is covered.

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -19,29 +19,77 @@
...
+// determine amount of memory used by a qstring
+inline uint64 GetHeapUsage(const QString &qstr) {
+    uint64 qstrSize = (qstr.capacity() * sizeof(char16_t));
+    return qstrSize;
+}
+// determine amount of memory used by a given object
+// returns 0 for all non specified objects
+template<class T>
+inline uint64 GetHeapUsage(const T &obj) {
+    notify(NFY_WARN, "\tGENERAL of %s", typeid(obj).name());
```
tst-lsavoie: Minor: I think we'll want to remove this before checking in the code. This logging statement would show up a lot, and in many cases nothing would be wrong.

earth_enterprise/src/common/khCache.h
tst-lsavoie: Good point.

earth_enterprise/src/common/CacheSizeCalculations.h
```diff
@@ -19,29 +19,77 @@
...
+    }
+    return total;
 }
-// determine amount of memory used by a shared_ptr and the object it points to
+// determine amount of memory used by a given object
 template<class T>
-inline uint64 GetObjectSize(const std::shared_ptr<T> &guard) {
-    return sizeof(guard) + (guard ? guard->GetSize() : 0);
+inline uint64 GetSize(const T &obj) {
+    notify(NFY_WARN, "GetSize of %s:", typeid(obj).name());
```
tst-lsavoie: Also should remove these two notify statements. I suspect you were planning to do that anyway.

earth_enterprise/src/fusion/autoingest/Asset.h
tst-lsavoie: Looking at some of the other classes it looks like you're just not calling GetHeapSize for primitives. I can live with that.

earth_enterprise/src/common/khCache.h
```diff
@@ -250,7 +251,7 @@ class khCache {
     item = new Item(key, val);
     Link(item);
     map[key] = item;
-    item->size = calculateCacheItemSize(item);
+    if (limitCacheMemory) item->size = calculateCacheItemSize(item);
```
tst-lsavoie: I think you need to initialize `item->size` before this call as well.

earth_enterprise/src/fusion/autoingest/StorageManager_unittest.cpp
```diff
@@ -56,10 +56,10 @@ class TestItem : public StorageManaged, public TestItemStorage {
 
   // determine amount of memory used by TestItem
   uint64 GetSize() {
-    return (GetObjectSize(val)
-    + GetObjectSize(type)
-    + GetObjectSize(savename)
-    + GetObjectSize(saveSucceeds));
+    return GetHeapUsage(val)
```
tst-lsavoie: Technically there's some things missing here (`loadname` and the stuff in StorageManaged) but I'm not sure it matters for unit tests.

tst-lsavoie:
Except for the one missed initialization in khCache I'm ready to approve. Everything looks solid.

Keep in mind that I didn't run any tests. We still need to make sure that the reported cache size is reasonable compared to the actual memory usage and it would be a good idea to characterize the performance cost of calculating cache sizes.

tst-lsavoie:
Oh, and as usual, I forgot that we should include a note in the release notes about how the cache size calculation was improved.

PR #1484, opened 2019-10-03
Removed timedLock before getting task list.

### LINKED ISSUE #1483: Allow getop to retrieve status even while system manager is busy
Right now, `gesystemmanager` won't respond to the request to get current tasks while it is busy doing processing like overlap calculation. As a result, `getop` times out waiting for `gesystemmanager` and won't display any information.

It would be nice to allow `getop` to retrieve and show its information even when `gesystemmanager` is "busy".

tst-sbayard:
Resolves #1483 

Since `GetCurrTasks` is just reading information, and since it doesn't really matter if the task list is slightly out of date for a few seconds, it's not really necessary to acquire a lock before before reading. We just need to make sure we catch any exceptions that could occur.

**Verification steps**

- Create a project or database that requires at least a few minutes of "load and think" time initially.
- Start the build and run `getop` while the system manager is busy
- Verify that `getop` is able to display information
- After system manager "unlocks" and starts creating tasks, verify that `getop` is able to show the running tasks without causing any errors during the build


earth_enterprise/src/fusion/autoingest/sysman/khResourceManager.cpp
```diff
@@ -925,20 +925,26 @@ void
...
-                              task->submitTime(),
-                              task->activationError));
-  }
-
-  for (Providers::iterator p = providers.begin();
-       p != providers.end(); ++p) {
-    p->second->AddToTaskLists(ret);
+  try {
+    for (TaskWaitingQueue::iterator t = taskWaitingQueue.begin();
+        t != taskWaitingQueue.end(); ++t) {
```
tst-jkent: Could this be replaced with a range loop?

earth_enterprise/src/fusion/autoingest/sysman/khResourceManager.cpp
```diff
@@ -925,20 +925,26 @@ void
...
+      ret.waitingTasks.push_back
+        (TaskLists::WaitingTask(task->verref(),
+                                task->taskid(),
+                                task->priority(),
+                                task->submitTime(),
+                                task->activationError));
+    }
+
+    for (Providers::iterator p = providers.begin();
+        p != providers.end(); ++p) {
```
tst-jkent: Same here.

earth_enterprise/src/fusion/autoingest/sysman/khResourceManager.cpp
```diff
@@ -925,20 +925,26 @@ void
 khResourceManager::GetCurrTasks(TaskLists &ret)
 {
   assert(!mutex.TryLock());
```
tst-jkent: This assert now fails when getop is run after installing fusion.

earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp
```diff
@@ -1236,11 +1233,6 @@ std::string khAssetManager::RetrieveTasking(const FusionConnection::RecvPacket&
 	      .arg(ToQString(GET_TASKS)));
 	  }
   }
-  catch (khTimedMutexException e) {
-    // Replying with a string beginning "ERROR:" passes an exception message back to the caller
-    // alternatively we could throw an exception but that could flood fusion logs with warnings
-    replyPayload = sysManBusyMsg;
```
tst-lsavoie: Minor: probably should get rid of `sysManBusyMsg`. I doubt it's used anywhere else.

earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp
```diff
@@ -990,7 +990,6 @@ khAssetManager::MercatorMapDatabaseModify(const MapDatabaseEditRequest &req) {
 
 void
 khAssetManager::GetCurrTasks(const std::string &dummy, TaskLists &ret) {
-  assert(!mutex.TryLock());
   notify(NFY_INFO, "GetCurrTasks");
   khLockGuard lock(theResourceManager.mutex);
```
tst-lsavoie: We should move the timedLock logic here so that we can still report a sensible error back to system manager during the times that this lock is held for a long time.

earth_enterprise/src/fusion/autoingest/sysman/khResourceManager.cpp
tst-sbayard: Turns out we can't without changing the data structure (it's `std::set` right now). But at this point we're planning to keep the lock around this section of code, so we won't have to worry about mulit-threaded access.

earth_enterprise/src/fusion/autoingest/sysman/khAssetManager.cpp
```diff
@@ -1236,9 +1235,9 @@ std::string khAssetManager::RetrieveTasking(const FusionConnection::RecvPacket&
 	      .arg(ToQString(GET_TASKS)));
 	  }
   }
-  catch (khTimedMutexException e) {
-    // Replying with a string beginning "ERROR:" passes an exception message back to the caller
-    // alternatively we could throw an exception but that could flood fusion logs with warnings
+  catch (khTimedMutexException e) {
```
tst-lsavoie: Minor: it looks like there's extra whitespace at the end of these 3 lines

tst-lsavoie:
Minor: there is a call to GetCurrTasks in khAssetManagerDispatch.cpp that I think is never called. It would be nice to remove that if possible.

tst-lsavoie:
I always forget this part. We need to put something in the release notes about this.

PR #1490, opened 2019-10-04
I1202 cannot select kip as source

tst-eclamar:
Remove '/header.xml' from string returned from file selection dialog.


tst-eclamar:
An error occurred during the creation of the branch - closing.

Issue #1498, opened 2019-10-09
Automated Fusion fails when republishing databases


tst-nrichman:
**Describe the bug**

Automated Fusion fails to republish automated fusion databases.

**Steps to reproduce the behavior:**

1. Use Automated Fusion to publish a database
2. Use Automated Fusion to un-publish a database
3. When trying to re-publish the database, Automated Fusion fails

tst-rwildes:
Issue fixed

PR #1499, opened 2019-10-09
Remove unused field from query (#1498)

tst-nrichman:
The default_db values are missing from the preceding result check. However, they're primarily unused and should simply be removed from the overall SELECT query.

### DIFF
--- a/earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
+++ b/earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
@@ -617,14 +617,13 @@ class PublishManagerHelper(stream_manager.StreamManager):
 
     target_details = {}
     query_string = ("""SELECT db_table.host_name, db_table.db_name,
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
```diff
@@ -617,14 +617,13 @@ def _QueryTargetDbDetailsByPath(self, target_path):
 
     target_details = {}
     query_string = ("""SELECT db_table.host_name, db_table.db_name,
-              virtual_host_table.virtual_host_name, target_table.serve_wms, publish_context_table.ec_default_db
-              FROM target_table, target_db_table, db_table, virtual_host_table, publish_context_table
+              virtual_host_table.virtual_host_name, target_table.serve_wms
```
tst-rwildes: Just change the select clause to not bring back that column.  The from clause and the where clause can stay the same.

PR #1508, opened 2019-10-18
Include config fields for memory use calculation

tst-jkent:
Updated methods in CacheSizeCalculations.
Updated config files to auto generate GetHeapUsage methods.
Updated release notes.
Changed default memory limit.


earth_enterprise/src/fusion/autoingest/MiscConfigStorage.idl
```diff
@@ -28,8 +28,8 @@ class MiscConfigStorage {
   uint        VersionCacheSize = uint(128000);
 
   // The maximum amount of memory allocated for each cache in bytes.
-  uint64      MaxAssetCacheMemorySize = uint64(1073741824);
-  uint64      MaxVersionCacheMemorySize = uint64(1073741824);
+  uint64      MaxAssetCacheMemorySize = uint64(10737418240);
```
tst-lsavoie: We may need to think about what the default memory size should be. I'm not sure it's reasonable to assume that everyone has 20 GB of RAM. Our [system requirements](http://www.opengee.org/geedocs/5.3.2/answer/installGEE/sysReqGEEServer.html) state users need 4 GB of RAM and recommend at least 8 for Fusion, so maybe we should default to using 3 or 7 GB combined.

Along the same lines, assets tend to be much smaller than asset versions, so it might be wise to have an unequal distribution of memory between them. I could see giving 4x or 8x as much memory to the asset versions as to the assets.

I'm happy to hear other thoughts on this as well. There's not a clear right answer.

docs/geedocs/docsrc/answer/releaseNotes/relNotesGEE5_3_3.rst
```diff
@@ -23,6 +23,14 @@ Release notes: Open GEE 5.3.3
       removed for most calculations, such as when system manager is determining
       which resources overlap. 
 
+      **Reported memory used by the caches is more accurate**.
```
tst-lsavoie: Documentation is tricky. I'd suggest rewording this to make us sound less bad.  Maybe something like this:

```
**Improved memory usage reporting for caches**.
The memory usage estimation for the asset and
asset version caches has been improved and is
much more accurate.
```

docs/geedocs/docsrc/answer/releaseNotes/relNotesGEE5_3_3.rst
```diff
@@ -23,6 +23,14 @@ Release notes: Open GEE 5.3.3
       removed for most calculations, such as when system manager is determining
       which resources overlap. 
 
+      **Reported memory used by the caches is more accurate**.
+      Previously, the calculations for memory usage were
+      missing data and the reported usage in ``getop`` was
+      underestimated. Now, the reported usage is closer
+      to the actual memory used.
+
+=======
```
tst-sbayard: This looks like it was left over from a merge. Should be removed (and maybe the blank line after it).

earth_enterprise/src/fusion/autoingest/tools/getop.cpp
```diff
@@ -242,13 +243,17 @@ main(int argc, char *argv[])
...
 
         outline("");
-        outline("Number of cached assets: %u", taskLists.num_assets_cached);
-        outline("Number of cached asset versions: %u",
-                taskLists.num_assetversions_cached);
-        outline("Total memory used by cached assets: %s",
-                readableMemorySize(taskLists.asset_cache_memory).c_str());
-        outline("Total memory used by cached asset versions: %s",
-                readableMemorySize(taskLists.version_cache_memory).c_str());
+        if (MiscConfig::Instance().LimitMemoryUtilization) {
```
tst-sbayard: Adding a comment here to reflect our prior conversation.
As discussed, pending performance testing, it would be nice to have the memory usage calculated and displayed if the performance cost is relatively negligible.

PR #1516, opened 2019-10-29
1515 fix always skipped test

tst-cjeffries:
Removes what an old unused sample test. This test was always skipped in geecheck. Now it will no longer incorrectly report that fusion is not installed.

To test:

1. Build and install OpenGEE.
2. Navigate to the geecheck page.
3. Verify that testWithSampleCondition no longer appears.


tst-blaird:
Hi @tst-cjeffries -- I would have thought the resolution for this issue would be to change the "Fusion is not installed" text to something like "Sample test condition not met".  Is there a reason that removing the sample test is a better solution?

tst-ccamp:
> @tst-blaird That is a good point. The primary reason I thought to remove it is that geecheck is a (potentially) user facing page and I'm not sure we want a sample test that serves no function on that page. However, I am fine with just changing the string. Perhaps @tst-ccamp can point me in the right direction here?

@tst-cjeffries I agree with you. We don't need the sample test on the production page. I'm not entirely sure we need a sample test at all. I think the other existing tests should be more than enough info for developers to go by. So, yeah, go ahead and delete.

tst-cjeffries:
Updated the doc to refer users to the existing tests as a guide rather than the sample.

tst-jzmolek:
Reviewed and tested. Looks good.

PR #1523, opened 2019-10-30
Improve handling of TaskDone messages

### LINKED ISSUE #1522: Optimize TaskDone message handling
Resource providers send TaskDone messages to system manager when they complete a task. These messages trigger state changes that can take a long time to process, especially for very large builds. In extreme cases, system manager takes so long processing these messages that it can't assign tasks to resource providers quickly enough and they end up waiting on system manager to process the TaskDone messages and assign more tasks. We should optimize this processing so that Fusion works efficiently for large builds.

We previously fixed a similar issue in #1510. This builds on that work to handle more resource provider messages efficiently.

tst-lsavoie:
This PR improves the handling for TaskDone messages to make them faster and easier to understand. It also adds tests for the AssetOperation code (and fixes some bugs in that code found through the tests).

No release notes updates are included in this PR because it fits within the release notes updates included in PR #1509.

Fixes #1522

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/StorageManagerAssetHandle_unittest.cpp
+++ b/earth_enterprise/src/fusion/autoingest/StorageManagerAssetHandle_unittest.cpp
@@ -64,14 +64,14 @@ TEST(AssetHandleTest, Finalize) {
   ASSERT_TRUE(finalized);
 }
 
-TEST(AssetHandleTest, ConvertConstToMutable) {
+TEST(AssetHandleTest, ConvertMutableToConst) {
   AssetHandle<TestItem> mutableHandle;
   AssetHandle<const TestItem> constHandle;
   constHandle = mutableHandle;
   ASSERT_EQ(mutableHandle.operator->(), constHandle.operator->());
 }
 
-TEST(AssetHandleTest, ConvertMutableToConst) {
+TEST(AssetHandleTest, ConvertConstToMutable) {
   // We should not be able to convert a mutable handle to a const handle.
   if(std::is_assignable<AssetHandle<TestItem>, AssetHandle<const TestItem>>::value) {
     // If the above statement is wrapped in ASSERT_FALSE it will fail to compile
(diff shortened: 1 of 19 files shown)


earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
```diff
@@ -87,9 +87,57 @@ void HandleTaskProgress(const TaskProgressMsg & msg) {
...
+    const SharedString & ref,
+    AssetDefs::State oldState,
+    uint32 numInputsWaitingFor,
+    uint32 numChildrenWaitingFor,
+    bool graphOps) {
+  if (graphOps) {
+    auto version = assetOpStorageManager->Get(ref);
+    if (version) {
+      // Update the lists of waiting assets
+      stateUpdater->UpdateWaitingAssets(version, oldState, {numInputsWaitingFor, numChildrenWaitingFor});
```
tst-sbayard: I think UpdateWaitingAssets may get called twice for a single update. Once here in response to a version's `OnStateChange` getting called, and once before this in `StateUpdater::RunStateChangeHandlers`. Is that right? I don't suppose it would cause an error, but could be extra work.

tst-lsavoie: Good catch; you're right about that. I moved the call in AssetVersionD from SetState to OnStateChange mostly because the child waiting count is a member of CompositeAssetVersionImplD, and SetState is defined in a parent class, so it doesn't have access to that variable. It's going to take some thinking to decide on the best way to handle that.

tst-lsavoie: Fixed.

earth_enterprise/src/fusion/autoingest/StorageManagerAssetHandle_unittest.cpp
```diff
@@ -63,6 +64,22 @@ TEST(AssetHandleTest, Finalize) {
   ASSERT_TRUE(finalized);
 }
 
+TEST(AssetHandleTest, ConvertConstToMutable) {
```
tst-sbayard: I'm a little confused on this function. `ConvertConstToMutable` sounds like we want to start with a const handle and create a mutable handle to the same resource. But this test looks like it's doing it the other way around: creating a const handle from a mutable handle.

tst-lsavoie: You're right, I got the name backwards.

Issue #1241, opened 2019-03-26
geglxinfo is only processing the first mesh in a terrain packet


tst-blaird:
When the geglxinfo utility encounters a terrain packet it is only extracting the first geterrain::Mesh from the packet.  Terrain packets contain multiple meshes.  The utility should process them all.

https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/fusion/portableglobe/geglxinfo.cpp#L234

```c++
        else if (index_item.packet_type == kTerrainPacket) {
          LittleEndianReadBuffer decompressed;
          etEncoder::DecodeWithDefaultKey(&buffer[0], buffer.size());
          if (KhPktDecompress(buffer.data(), buffer.size(), &decompressed)) {
            geterrain::Mesh m;
            m.Pull(decompressed);
            std::ostringstream s;
            m.PrintMesh(s);
            writePacketToFile(index_item, s.str(), false, "globetiles", "_terrain");
          }
        }
```

tst-blaird:
Pull request with code fix and release notes update was merged.  Closing.

PR #1531, opened 2019-11-05
1241 geglxinfo will process all meshes in a terrain packet

tst-blaird:
#1241 These changes update geglxinfo to process all meshes in a terrain packet.  Prior to this PR only the first mesh in a terrain packet would be processed.

Testing:
1. Build and run geglxinfo against a portable globe that contains terrain data with the --extract_packets parameter.  The terrain packets will be written under the directory globetiles/1/.
2. Verify that the terrain files written contain all the mesh entries for each packet.


tst-jlarocco:
Also tested and saw terrain packets with all meshes.

PR #1533, opened 2019-11-05
Escaped all supplemental search fields from the server for HTML compatibility

### LINKED ISSUE #1532: Supplemental Search fields should be escaped for HTML when viewing on an HTML page
There are fields on the supplemental search page that come from the server in JSON format, but they are not escaped for HTML which could cause problems if the fields have HTML characters in them. These fields should be escaped to allow HTML characters.

Steps to reproduce the behavior: Modify the search database to have HTML tags, for example in a suggested search string, and see that Google Earth Client renders the supplemental search page with the modified HTML tags as HTML, and not as plain text.

tst-jzmolek:
Adds HTML escaping to fields used on the supplemental search page in Google Earth client.

To test:
1) Modify the search_def_table to use a search suggestion with added HTML tags.
    i.e.
`set search_def_content = '{"additional_config_param": null, "label": "Coordinates", "additional_query_param": "flyToFirstElement=true&displayKeys=location", "service_url": "/gesearch/CoordinateSearch", "result_type": "KML", "html_transform_url": "about:blank", "kml_transform_url":  "about:blank", "suggest_server": "about:blank", "fields": [{"key": "q", "suggestion": "<h1>Injected HTML</h1><h2>Latitude, longitude (e.g., 23.47, -132.67)</h2>", "label": null}]}' where search_def_id = 3`
2) Publish a globe database selecting the Search Tab and Supplemental Search Tab for the search type that was modified in step 1) (i.e. Coordinate). Both tabs have to be added because the Supplemental Search page only appears if the corresponding Search exists.
3) Open the published globe in EC. The main search suggestion appears rendered in HTML (This is believed to happen in EC which we can't change), but after clicking on the label (button) for the supplemental search page, the suggestion appears as plain text showing the HTML tags. This is the result of escaping the HTML characters on the supplemental search page.

Tested on CentOS7 with Win 10 client.
Fixes #1532

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/htdocs/earth/search_supplemental_ui.html
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/htdocs/earth/search_supplemental_ui.html
@@ -113,7 +113,7 @@ function createSearchList() {
           '<h1  onclick="showHideDiv(\'' + divId + '\')"' +
           ' id="div_' + divId +'">' +
           '<span id="arrow_' + divId + '" class="right_arrow">' +
-          '<label>' + label + '</label></span>' + '</h1>' +
+          '<label>' + escapeHTML(label) + '</label></span>' + '</h1>' +
           '<span id="span_' + divId + '" style="display:none;"></span>';
     } else {
       targetDiv.innerHTML +=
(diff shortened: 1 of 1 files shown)


earth_enterprise/src/fusion/portableglobe/cutter/htdocs/earth/search_supplemental_ui.html
```diff
@@ -186,6 +186,13 @@
   return retrieveDatabases.responseText;
 }
 
+// Escapes HTML characters so that the given string can be used within HTML
+function escapeHTML(str) {
+  var regex = /[&<>"']/g
```
tst-dsugar: So this escapes the standard html markers...

tst-mswartz (approved):
Reviewed (but not tested), looks good to me.

tst-dsugar (approved):

Issue #1535, opened 2019-11-07
Release_5.3.2.1 Cutter is not using Python2.7


tst-cjeffries:
Cutter (and related scripts) are not using Python 2.7 on EL6. This results in errors when they are used.

**To Reproduce**

Attempt to use cutter on EL6.

**Expected behavior**

Cutter works error free, and uses Python 2.7.

PR #1536, opened 2019-11-07
#1535: Switch shebangs to python 2.7

tst-cjeffries:
Switches shebangs to explicitly use python 2.7. This forces all scripts to use the correct version of python regardless of system python version.

To test:
1. Install OpenGEE.
2. Enable cutter (`/opt/google/bin/gecutter enable`)
3. Cut a portable globe.
4. Assemble a glc file.
5. Disassemble a glc file.


earth_enterprise/src/server/wsgi/wms/ogc/xml/v111/capabilities.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python2.7
```
tst-dsugar: Just taking one of these as a point, and question.  I see you had decided to go with env.

Issue #1538, opened 2019-11-11
coverage builds fail when scons cache is enabled


tst-rwildes:
**Describe the bug**

If you enable scons cache when building and attempt to do a coverage build to enable the collecting of code coverage during unit test runs the build will fail if you do a clean and re-build.

**To Reproduce**

1. Do build a coverage build while scons cache is enabled
2. Run a clean (delete the build directory to be sure everything is cleaned)
3. Rebuild again with scons cache still enabled
4. Build fails

**Expected behavior**

Second clean build should succeed

**Additional context**

This is happening because scons is unaware about the `.gcno` files used by coverage builds.  Because scons is unaware of these files it does not add them to the cache and when rebuilding from that cache these files are not created.

tst-rwildes:
Fix merged in

PR #1539, opened 2019-11-11
changes to make coverage builds compatible with scons cache

tst-rwildes:
This fixes issue #1538 
These changes will register the coverage builds `.gcno` files with scons as additional targets of compiling c files.


tst-blaird (approved):
Changes look good to me.  Only reviewed -- did not test.

PR #1542, opened 2019-11-11
WIP - Prototype quadtree filtering for overlaps

tst-sbayard:


earth_enterprise/src/common/quadtreepath.h
```diff
@@ -78,6 +78,10 @@ class QuadtreePath {
       && (path2.IsAncestorOf(path1) || path2 > path1);
   }
 
+  static inline bool OverlapsAtLevel(const QuadtreePath &path1, const QuadtreePath &path2, uint level) {
```
tst-lsavoie: I think we now know that this function is wrong, or at least misses some important edge cases. It may be wise to leave it in for posterity, but we could comment it out or add a comment indicating that it's not fully correct.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.cpp
```diff
@@ -370,6 +371,25 @@ void FindNeededImageryInsets(
     std::vector<uint>            &neededIndexes,
     uint beginMinifyLevel,
     uint endMinifyLevel) {
+
+  std::vector<khExtents<uint32> > extentsVec;
```
tst-lsavoie: This could explain some of the missing performance improvement - it's creating a duplicate vector here. That can be fixed in a real implementation.

earth_enterprise/src/fusion/autoingest/sysman/InsetTilespaceIndex.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// khExtents<double> getExtents(const InsetInfoAndIndex &info){
+//     return info.ptrInsetInfo->degExtents;
+// }
+
+// InsetTilespaceIndex methods
+template <class ExtentContainer>
+QuadtreePath InsetTilespaceIndex<ExtentContainer>::add(const ExtentContainer &toAdd){
+    int level;
```
tst-lsavoie: Minor: looks like this is not used.

earth_enterprise/src/fusion/autoingest/sysman/InsetTilespaceIndex_test.cpp
```diff
@@ -0,0 +1,214 @@
+//
```
tst-lsavoie: Minor: it looks like this test still uses the "old" version of the "new" algorithm, i.e., assume images overlap if their quad-tree paths are completely the same. I'm sure you're aware of that; just thought I'd point it out for completeness.

earth_enterprise/src/fusion/autoingest/sysman/InsetTilespaceIndex.h
```diff
@@ -0,0 +1,190 @@
...
+
+    std::vector<const T*> GetElementsAtQuadTreePath(QuadtreePath path, uint startLevel) 
+    {
+        std::vector<const T*> vec;
+
+        const uint pathLength = path.Level();
+        uint pathIndex = 0;
+        std::shared_ptr<QuadKeyTreeNode<T>> nodeptr = levelZeroNode;
+
+        while (pathIndex < pathLength && pathIndex < startLevel && nodeptr != nullptr){
```
tst-lsavoie: Minor: it occurred to me that we might be able to save some time by returning two lists - a list of items "above" this one in the tree and a list of items "below" this one in the tree. I think the ones below are guaranteed to overlap while the ones above may or may not overlap. We could use that to reduce the number of comparisons even further.

I haven't thought this through all the way, so it might not work. In any case, it's something to worry about later.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -111,17 +130,42 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
-        extra.insetInfo,
-        numInsets,
+        qualifiedInsetInfoPtrs,
+        qualifiedInsetInfoPtrs.size(),
         config.beginMinifyLevel,
         config.endMinifyLevel,
         level);
     extra.calc->setEnvelope(env);
-    neededIndexes = std::move(extra.calc->GetOverlapForLevel(gencov));
+    std::vector<uint> tempNeededIndexes = std::move(extra.calc->GetOverlapForLevel(gencov));
```
tst-lsavoie: Minor: I think you could save some time here by storing the index directly in the QuadKeyTree rather than a pointer to the inset info and then using a separate map to look up the index. I haven't thought it through completely, but if we can make that work it would improve performance.

earth_enterprise/src/fusion/autoingest/sysman/QuadKeyTree_unittest.cpp
```diff
@@ -0,0 +1,123 @@
...
+};
+
+// TEST DATA
+// std::array<std::string, 9> data {{ "this", "is", "some", "string", "data", 
+//                                    "one fish", "two fish" , "red fish", "blue fish"}};
+QuadtreePath qtp01("01");
+QuadtreePath qtp011("011");
+QuadtreePath qtp012("012");
+QuadtreePath qtp0123("0123");
+QuadtreePath qtp1("1");
```
tst-lsavoie: Does QTP 1 even make sense? I thought QTPs always started with 0.

earth_enterprise/src/fusion/autoingest/sysman/QuadKeyTree_unittest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    tree.AddElementAtQuadTreePath(qtpEmpty, &qtpEmpty);
+    tree.AddElementAtQuadTreePath(qtp1, &qtp1);
+    tree.AddElementAtQuadTreePath(qtp01, &qtp01);
+    tree.AddElementAtQuadTreePath(qtp011, &qtp011);
+    tree.AddElementAtQuadTreePath(qtp012, &qtp012);
+    tree.AddElementAtQuadTreePath(qtp0123, &qtp0123);
+    
+    std::vector<const QuadtreePath*> vec = tree.GetElementsAtQuadTreePath(qtp012, 3);
+
+    // Should have matched everything except qtp011
```
tst-lsavoie: Minor: it also doesn't match qtp1.

earth_enterprise/src/common/quadtreepath.h
tst-sbayard: Oops. I guess this was created as part of the previous investigation but isn't used now. I completely missed that. I'll go ahead and take it out.

earth_enterprise/src/fusion/autoingest/sysman/InsetTilespaceIndex_test.cpp
tst-sbayard: Yeah, good catch. I originally removed this file from the SCons build just to kind of get it out of the way while I worked on the new algorithm. I mostly forgot about it after that. Honestly at this point, I'm not even sure this is necessary. The guts of this are being tested by `QuadKeyTree_unittest.cpp `.
Any objection to just removing this file?

earth_enterprise/src/fusion/autoingest/sysman/QuadKeyTree_unittest.cpp
tst-sbayard: My understanding is that level 0 is implied but not represented in these objects. That's the sense I got while looking through the code in quadtreepath_unittest.cpp.
i.e., the first digit in the paths shown here represent the tile at level 1.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
tst-sbayard: I'm not sure if you mean to store the index *instead of* the inset pointer or in addition to. There's probably a bit of a wrinkle either way since we need a vector of `InsetInfo*` to pass into `GetOverlapForLevel`. Either way there would be some copying involved, I think.
But to your point, this is pretty brute force, and there's almost certainly a way to make it more efficient.

earth_enterprise/src/fusion/autoingest/sysman/InsetTilespaceIndex_test.cpp
tst-sbayard: Removed it.

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
tst-lsavoie: Yeah, that's basically what I'm getting at - this is a place where we can optimize.

earth_enterprise/src/fusion/autoingest/sysman/InsetInfo.cpp
tst-sbayard: The `FindNeededImageryInsets` functions were modified to make testing easier in the last investigation. I didn't change it back here, but the prior version could be used. I think that would eliminate this issue.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -181,7 +181,8 @@ class RasterProjectAssetVersionImplD {
   uint32 BuildIndex(
     const khTilespace &tilespace,
     const std::vector<InsetInfo<RasterProductAssetVersion> >& inset_infos,
-    const std::string& date);
+    const std::string& date/*,
+    InsetTilespaceIndex<InsetInfo<RasterProductAssetVersion>> &tilespaceIndex*/);
```
tst-lsavoie: Was this change meant to be checked in? It's just commented out.

tst-sbayard: Oops. That was left over from the initial stab I took at it. I'll take that out.

earth_enterprise/src/fusion/autoingest/plugins.src/RasterProject.src
```diff
@@ -1085,15 +1097,20 @@ uint32 RasterProjectAssetVersionImplD::BuildIndex(
 
     notify(NFY_PROGRESS, "BuildIndex finding intersections for inset %d of %lu",
            i+1, genInfo.size());
-
+    int maxQuadTreeLevel = 0;
+    QuadtreePath coverageMBR = localIndex.getQuadtreeMBR(genInfo[i].coverage.degreeExtents(), maxQuadTreeLevel, endMinifyLevel);
+    std::vector<const InsetInfo<RasterProductAssetVersion>*> qualifiedInsetInfoPtrs = localIndex.intersectingExtents(coverageMBR, genInfo[i].coverage.beginLevel(), endMinifyLevel);
```
tst-lsavoie: Just checking - was the removal of the call to `PreprocessForInset` here intentional? I think it's fine, I just want to make sure it's what was intended.

tst-sbayard: Yeah, that was on purpose. I initially called `PreprocessForInset` after `intersectingExtents`, then I went a step further to take out the call to `PreprocessForInset`.
My thinking was that since we're theoretically just trimming out things we know we won't need anyway, `intersectingExtents` might be effective enough on its own. That seems to be the case so far.

tst-lsavoie: Makes sense.

tst-lsavoie:
Closing this since it's intended as a prototype. We can reopen it in the future if needed.

PR #1545, opened 2019-11-12
Quick fix for Python Shebangs

tst-cjeffries:
getversion.py needs to use the system python.


tst-cjeffries:
Source branch is causing problems

PR #1546, opened 2019-11-12
WIP - Periodically write modified assets to disk during packgen creation

tst-jkent:


earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -355,5 +362,35 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
...
 
+template<class AssetType>
+void StorageManager<AssetType>::WriteDirty() {
+  if (writeDirty) {
+    std::lock_guard<std::recursive_mutex> lock(storageMutex);
+    khFilesTransaction savetrans(".new");
+    typename std::map<AssetKey, PointerType>::iterator entry = dirtyMap.begin();
+    notify(NFY_WARN, "WriteDirty: %s", ToString(typeid(AssetType).name()).c_str());
+    while (entry != dirtyMap.end()) {
+      if (entry->second.use_count() == 2) {
```
tst-lsavoie: Eventually we really ought to call WriteDirtyToDotNew here rather than duplicating most of the code.

earth_enterprise/src/fusion/autoingest/plugins.src/MercatorRasterProject.src
```diff
@@ -1025,6 +1025,8 @@ uint32 MercatorRasterProjectAssetVersionImplD::BuildIndex(
     packgenkids.push_back(packgen->GetRef());
     packgenassets.push_back(packgen->GetAssetRef());
     packgenvers.push_back(packgen);
```
tst-lsavoie: This is part of the reason things aren't getting written out. We're saving handles to all the packgens in this vector, so none of them will be written to disk or removed from the cache. That said, there's one packgen per image (I think) and multiple packlevels, and we should be writing those out to disk, so there should still be a big improvement.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -355,5 +362,35 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
...
+        std::string filename = entry->second->XMLFilename() + ".new";
+
+        if (serializer->Save(entry->second, filename)) {
+          savetrans.AddNewPath(filename);
+          notify(NFY_WARN, "\tRemoved: %s", entry->first.toString().c_str());
+          entry = dirtyMap.erase(entry);
+        }
+        else {
+          notify(NFY_WARN, "\tNot Saved: %s", entry->first.toString().c_str());
```
tst-lsavoie: Probably need a `++entry` call here as well.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -338,6 +338,10 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
   std::lock_guard<std::recursive_mutex> lock(storageMutex);
   typename std::map<AssetKey, PointerType>::iterator entry = dirtyMap.begin();
   while (entry != dirtyMap.end()) {
+    if ((entry->second.use_count() > 2) && writeDirty)  {
```
tst-lsavoie: I think there's a small bug here - if we're calling SaveDirtyToDotNew from ApplyPending, we want to always write out everything in the dirty map, even if `writeDirty` is true. That said, I don't think it affected your testing in any way.

tst-jkent: I don't believe this condition is met when SaveDirtyToDotNew is called in ApplyPending. Running the same build resulted in no assets having a use count greater than 2 when called from ApplyPending before this condition was introduced.

PR #1553, opened 2019-11-19
Better xml memory management

### LINKED ISSUE #1552: XML objects are not properly cleaned up
The objects in khxml.cpp/.h, notably GEXMLObject, GEDocument, and its derivatives, should have virtual destructors but do not, which causes resource leaks because Xerces objects are never cleaned up. If we make these destructors virtual we may also be able to remove the constant Xerces re-initialization and memory tracking allocator that currently clean up memory leaks, which should give us a performance boost.

tst-lsavoie:
Some of the objects in khxml.cpp/.h were missing virtual destructors, which prevented some XML cleanup routines from running. We worked around this by using a custom memory allocator that tracked leaked memory and periodically terminating Xerces and clearing out any leaked memory. This PR fixes the missing virtual destructors. Testing indicates that with the destructor fixed there are no more memory leaks, so this PR also removes the periodic Xerces termination and custom memory allocator. In all tests so far, this significantly improves performance and memory usage is also slightly lower.

Fixes #1552


docs/geedocs/docsrc/answer/releaseNotes/relNotesGEE5_3_3.rst
```diff
@@ -71,6 +71,11 @@ Release notes: Open GEE 5.3.3
          * - 1241
            - geglxinfo should parse all meshes in terrain packets
            - geglxinfo will parse all meshes in terrain packets
+         * - 1552
+           - Make XML object destructors virtual
```
tst-sbayard: Talking about virtual destructors seems a bit technical for release notes. Could we make it more generic like "Properly cleaning up XML objects" or something like that?

tst-lsavoie: Fair point. I was going with the title from the issue on GitHub, but I guess that can be changed too. It's fixed now.

PR #1557, opened 2019-11-22
Don't run state change handlers unnecessarily

### LINKED ISSUE #1556: Resume should bypass handlers when setting states to New
A resume operation works by setting the state of each affected asset to New and then calculating what its state should be based on the states of its inputs and children. When setting the state to New, it should do so directly, without running any handlers, as can be seen in CompositeAssetVersionImplD::Rebuild in AssetVersionD.cpp:

```
  state = AssetDefs::New; // low-level to avoid callbacks
  SyncState(notifier);
```

The new resume function in the state updater doesn't follow this pattern however; instead, it runs callbacks when setting states to new (though it does avoid calling notifications at least). There are no known negative implications of this other than possibly slowing things down a bit, but we should fix it to be consistent with the old code.

tst-lsavoie:
Fixes a difference between the new StateUpdater code and the legacy AssetVersionD code in which OnStateChange was called unnecessarily during the first phase of a resume.

Fixes #1556

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
@@ -374,9 +374,7 @@ void StateUpdater::UpdateWaitingAssets(
   const SharedString ref = version->GetRef();
   const AssetDefs::State newState = version->state;
   waitingListeners.Update(ref, newState, oldState, waitingFor.inputs);
-  if (IsParent(version)) {
-    inProgressParents.Update(ref, newState, oldState, waitingFor.children);
-  }
+  inProgressParents.Update(ref, newState, oldState, waitingFor.children);
 }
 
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -372,9 +374,7 @@ void StateUpdater::UpdateWaitingAssets(
   const SharedString ref = version->GetRef();
   const AssetDefs::State newState = version->state;
   waitingListeners.Update(ref, newState, oldState, waitingFor.inputs);
-  if (IsParent(version)) {
```
tst-lsavoie: FYI, this is a little code cleanup. The update function won't add the ref to the waiting list unless its waiting for at least one child, so non-parents will never be added to the waiting list anyway.

tst-lsavoie:
I've decided not to include this in the release notes because users won't really notice a difference. I'm open to other opinions on that, though.

PR #1559, opened 2019-12-02
Delete methods

tst-basilhuffman:
Running cppcheck on the codebase has shown a list of errors and/or vulnerabilities, one of which is to address a lack of copy/move constructors and assignments. It was determined by deleting the methods, that they aren't strictly necessary.

### DIFF
--- a/earth_enterprise/src/common/compressor.h
+++ b/earth_enterprise/src/common/compressor.h
@@ -187,7 +187,7 @@ struct ReadBuffer : public jpeg_source_mgr {
     ReadBuffer* src = static_cast<ReadBuffer*>(cinfo->src);
     src->next_input_byte = &src->buf[0];
     src->bytes_in_buffer = src->bufSize;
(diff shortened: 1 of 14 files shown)


earth_enterprise/src/common/compressor.h
```diff
@@ -187,7 +187,7 @@ struct ReadBuffer : public jpeg_source_mgr {
     ReadBuffer* src = (ReadBuffer*)cinfo->src;
```
tst-dsugar: "boolean" may not necessarily be the same as the true C "bool" type, and hence return true may cast to something like (int)-1, where TRUE may be (int)1???   So I am wondering if we want to either keep the original TRUE/FALSE macros, or make sure we also use "bool".  I would have to dig a bit deeper to see how these were defined...

tst-basilhuffman: addressed

PR #1560, opened 2019-12-02
Security updates 

tst-mswartz:
Change str functions to strn functions to reduce possibility of overflows or improper usage.

### DIFF
--- a/earth_enterprise/src/common/khFileUtils.cpp
+++ b/earth_enterprise/src/common/khFileUtils.cpp
@@ -793,7 +793,7 @@ khTmpFilename(const std::string &basename, mode_t mode)
   // must make a writable copy for mkstemp to play with
   size_t bufsize = fname.size() + 7;
   char tmpname[bufsize];
-  snprintf(tmpname, bufsize - 1, "%sXXXXXX", fname.c_str());
+  snprintf(tmpname, bufsize, "%sXXXXXX", fname.c_str());
   int fd = mkstemp(tmpname);
 
   if (fd == -1) {
@@ -814,7 +814,7 @@ khCreateTmpDir(const std::string &prefix)
   // must make a writable copy for mkstemp to play with
   size_t bufsize = dirname.size() + 7;
   char tmpname[bufsize];
-  snprintf(tmpname, bufsize - 1, "%sXXXXXX", dirname.c_str());
+  snprintf(tmpname, bufsize, "%sXXXXXX", dirname.c_str());
   char *result = mkdtemp(tmpname);
 
   if (result == NULL) {
--- a/earth_enterprise/src/common/khFileUtils.cpp
+++ b/earth_enterprise/src/common/khFileUtils.cpp
@@ -119,7 +119,7 @@ std::string khComposeTimePath(const struct tm& time) {
 std::string
 khBasename(const std::string &filename)
 {
-  int buflen = filename.size()+1;
+  size_t buflen = filename.size()+1;
   char buf[buflen];
   strncpy(buf, filename.c_str(), buflen);
   return basename(buf);
--- a/earth_enterprise/src/common/khFileUtils.cpp
+++ b/earth_enterprise/src/common/khFileUtils.cpp
@@ -791,9 +791,9 @@ khTmpFilename(const std::string &basename, mode_t mode)
   khEnsureParentDir(fname);
 
   // must make a writable copy for mkstemp to play with
-  int bufsize = fname.size() + 7;
+  size_t bufsize = fname.size() + 7;
   char tmpname[bufsize];
-  snprintf(tmpname, bufsize, "%sXXXXXX", fname.c_str());
+  snprintf(tmpname, bufsize - 1, "%sXXXXXX", fname.c_str());
   int fd = mkstemp(tmpname);
 
   if (fd == -1) {
@@ -812,9 +812,9 @@ khCreateTmpDir(const std::string &prefix)
   std::string dirname = khTmpDirPath() + "/" + prefix;
 
   // must make a writable copy for mkstemp to play with
-  int bufsize = dirname.size() + 7;
+  size_t bufsize = dirname.size() + 7;
   char tmpname[bufsize];
-  snprintf(tmpname, bufsize, "%sXXXXXX", dirname.c_str());
+  snprintf(tmpname, bufsize - 1, "%sXXXXXX", dirname.c_str());
   char *result = mkdtemp(tmpname);
 
   if (result == NULL) {
--- a/earth_enterprise/src/common/khFileUtils.cpp
+++ b/earth_enterprise/src/common/khFileUtils.cpp
@@ -119,16 +119,18 @@ std::string khComposeTimePath(const struct tm& time) {
 std::string
 khBasename(const std::string &filename)
 {
-  char buf[filename.size()+1];
-  strcpy(buf, filename.c_str());
+  int buflen = filename.size()+1;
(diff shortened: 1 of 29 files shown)


earth_enterprise/src/common/khFileUtils.cpp
```diff
@@ -789,9 +791,9 @@ khTmpFilename(const std::string &basename, mode_t mode)
   khEnsureParentDir(fname);
 
   // must make a writable copy for mkstemp to play with
-  char tmpname[fname.size()+7];
-  strcpy(tmpname, fname.c_str());
-  strcat(tmpname, "XXXXXX");
+  int bufsize = fname.size() + 7;
```
tst-basilhuffman: Did we determine where this magic number is coming from? Might be better to make it a const variable (maybe `uint8_t`?)

earth_enterprise/src/common/khFileUtils.cpp
```diff
@@ -810,9 +812,9 @@ khCreateTmpDir(const std::string &prefix)
   std::string dirname = khTmpDirPath() + "/" + prefix;
 
   // must make a writable copy for mkstemp to play with
-  char tmpname[dirname.size()+7];
-  strcpy(tmpname, dirname.c_str());
-  strcat(tmpname, "XXXXXX");
+  int bufsize = dirname.size() + 7;
```
tst-basilhuffman: Same as above

earth_enterprise/src/common/notify.cpp
```diff
@@ -132,12 +132,12 @@ std::string GetNotifyPrefixValue(khNotifyLevel severity, const std::string& noti
...
         (strftime(buf, sizeof(buf), "%F %T", &ts) > 0)) {
-      sprintf(prefix_buf, "[%s] %s: ", buf, notifyLevels[severity]);
+      snprintf(prefix_buf, 1024, "[%s] %s: ", buf, notifyLevels[severity]);
     } else {
-      sprintf(prefix_buf, "[time] %s: ", notifyLevels[severity]);
+      snprintf(prefix_buf, 1024, "[time] %s: ", notifyLevels[severity]);
     }
   } else {
-    sprintf(prefix_buf, "%s %s:\t",
+    snprintf(prefix_buf, 1024, "%s %s:\t",
```
tst-basilhuffman: Prefer const

earth_enterprise/src/common/packet.cpp
```diff
@@ -325,7 +325,7 @@ int etDataPacket::setFlatFileCount(int num)
   FILE *fp;
   char fname[500];
 
-  sprintf(fname, "%s.header", flatbasename);
+  snprintf(fname, 500, "%s.header", flatbasename);
```
tst-basilhuffman: same as above. in this instance, the const would be used when fname is declared also

also, applies to the other instances in this file

earth_enterprise/src/fusion/ffio/ffioReader.cpp
```diff
@@ -34,7 +34,7 @@ ffio::Reader::Reader(const std::string &ffdir)
 
   for (uint i = 0; ; ++i) {
     char fnum[64];
-    sprintf(fnum, "%02d", i);
+    snprintf(fnum, 64, "%02d", i);
```
tst-basilhuffman: same as above

earth_enterprise/src/fusion/ffio/ffioWriter.cpp
```diff
@@ -34,7 +34,7 @@ void
 ffio::Writer::OpenNextFile(void)
 {
   char fnum[64];
-  sprintf(fnum, "%02d", nextfilenum++);
+  snprintf(fnum, 64, "%02d", nextfilenum++);
```
tst-basilhuffman: same as above

earth_enterprise/src/fusion/mtjtools/histogram/histogram.cpp
```diff
@@ -89,7 +79,7 @@ int main(int argc, char* argv[])
 
   // process remaining argument, if any
   if (cursor < argc && input[0] == '\0')
-    strcpy(input, argv[cursor++]);
+    strncpy(input, argv[cursor++], 1024);
```
tst-basilhuffman: same as above

earth_enterprise/src/fusion/mtjtools/scroll/imageview.cpp
```diff
@@ -584,12 +586,12 @@ void ImageView::setFilename(char *name) {
   }
 
   // remember new file name
-  strcpy(_filename, name);
+  strncpy(_filename, name, 512);
```
tst-basilhuffman: same as above, applies throughout file

earth_enterprise/src/common/khFileUtils.cpp
tst-mswartz: I'm assuming it's for the 6 Xs and a null term.

earth_enterprise/src/common/notify.cpp
tst-mswartz: I used numeric literals because the same things are used to define all the buffers, I guess we could change everything to BUFLEN or something.

earth_enterprise/src/common/khFileUtils.cpp
tst-dsugar: technically bufsize probably should be a size_t actually.

earth_enterprise/src/fusion/tools/gevectorquery.cpp
```diff
@@ -53,8 +53,9 @@ int main(int argc, char** argv) {
         argn++;
         if (argn >= argc)
           usage();
-        output = reinterpret_cast<char*>(alloca(strlen(argv[argn]) + 1));
-        strcpy(output, argv[argn]);
+        int outputlen = strlen(argv[argn]);
```
tst-dsugar: For similar reasons this is probably also a size_t, which I think is what auto would give us.  We could now use auto since we are c++11...

earth_enterprise/src/common/khFileUtils.cpp
```diff
@@ -119,16 +119,18 @@ std::string khComposeTimePath(const struct tm& time) {
 std::string
 khBasename(const std::string &filename)
 {
-  char buf[filename.size()+1];
-  strcpy(buf, filename.c_str());
+  int buflen = filename.size()+1;
```
tst-dsugar: I still think thats a size_t, but its not that important

earth_enterprise/src/common/packet.cpp
```diff
@@ -296,7 +296,7 @@ int etDataPacket::getFlatFileCount()
   char fname[500];
   int count;
 
-  sprintf(fname, "%s.header", flatbasename);
+  snprintf(fname, 500, "%s.header", flatbasename);
```
tst-dsugar: probably should be sizeof(fname) in case anyone ever changes it, or have a size_t bufsize = 500, and use that for the array size and the snprintf.

earth_enterprise/src/common/packet.cpp
```diff
@@ -467,9 +465,9 @@ int etDataPacket::saveFlat(char * fname, char* buffer, int buffersize)
   //parsebcode(fname);
 
   //Enter File into Directory
-  sprintf(text, "%s", fname);
+  snprintf(text, 100, "%s", fname);
```
tst-dsugar: maybe sizeof(text)....

Issue #1564, opened 2019-12-05
Apache builds are missing crypto modules


tst-nrichman:
When trying to enable mod_session_crypto in Apache configuration files, the built module file was missing from gehttpd. Being able to use this module in configurations would be useful for encrypting user session data.

**Expected behavior**
After building apache, the file:
`earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party/apache2/install/opt/google/gehttpd/modules/mod_session_crypto` should exist.

PR #1566, opened 2019-12-05
Build Apache with crypto enabled (#1564)

tst-nrichman:
Adds crypto flags to Apache's optional build configurations to ensure Apache is building mod_session_crypto.  Fixes issue #1564 

**Verification**
1. Clone and build earthenterprise as normal.
2. Ensure the output module exists:
`earthenterprise/earth_enterprise/src/NATIVE-REL-x86_64/third_party/apache2/install/opt/google/gehttpd/modules/mod_session_crypto.so`


tst-blaird:
The latest changes still look fine.  Approval stands.

PR #1567, opened 2019-12-05
WIP: Checking pipeline ubuntu 16. Do not review

tst-basilhuffman:


earth_enterprise/src/fusion/portableglobe/shared/file_unpacker.h
```diff
@@ -35,7 +35,10 @@ namespace fusion_portableglobe {
 class FileUnpacker {
  public:
   explicit FileUnpacker(const char* package_file);
-
+  FileUnpacker(FileUnpacker&&) = delete;
```
tst-rwildes: Yes that is the better way to making things not copy/move.

PR #1571, opened 2019-12-05
Remove obsolete directory

tst-lsavoie:
Removes an obsolete directory and the unused code it contained.


tst-lsavoie:
Replaced by PR #1572

PR #1574, opened 2019-12-09
Fix glfs pointer

tst-rwildes:
This is to create a proper GLFS pointer for `PyYAML-3.10.tar.gz` that seems to of been added without one.


tst-lsavoie (approved):
Reviewed and tested a build and simple GLC assembly (where this is used). Everything looks good.

tst-cjeffries:
Looks good to me

tst-cjeffries (approved):

PR #1577, opened 2019-12-12
Use defusedxml in server code

tst-jkent:
Added defusedxml as a dependency.
Replaced xml library with defused equivalent for server.


tst-lsavoie:
There's another change that's needed here to mark defusedxml as a dependency for the RPM, which will make sure defusedxml is installed before the Open GEE server RPM. We need to add the line `requires('python-defusedxml')` in the `openGeeServerRpm` section in `earth_enterprise/rpms/build.gradle`. See https://github.com/google/earthenterprise/compare/master...tst-lsavoie:defusedxml. We probably ought to add the same line in the `openGeeServerDeb` section as well for completeness.

earth_enterprise/src/server/pywms/ogc/wms_111/wmsxml/capabilities.py
```diff
@@ -31,22 +31,22 @@
     ) = range(3)
 XMLParser_import_library = None
 try:
-  # lxml
-  from lxml import etree as etree_
-  XMLParser_import_library = XMLParser_import_lxml
+  import defusedxml.ElementTree as etree_
```
tst-lsavoie: This code seems to be trying to find an XML library that works to import. Since we're forcing it to use defusedxml, we can probably remove lines 27-68 and replace it with a single `import defusedxml.ElementTree as etree_`. There's a couple other places where the code handles different libraries that we can likely clean up as well - basically, anything that refers to `XMLParser_import_*` is no longer needed.

The same applies to several of the other files on this PR.

tst-lsavoie:
There are still several references to `lxml`. We should clean these up as well (except maybe the ones in cutter since we haven't touched those files yet):
```
rpms/build.gradle
459:    requires('python-lxml')

BUILD_Ubuntu.md
43:    python-setuptools python2.7 python2.7-dev python-lxml python-git \

BUILD_RHEL_CentOS.md
83:sudo yum install -y ant bzip2 doxygen gcc-c++ patch python-argparse python-lxml python-setuptools \
90:sudo yum install -y ant bzip2 doxygen gcc-c++ patch python-argparse python27-lxml python27-setuptools \

src/server/pywms/ogc/wms_130/wmsxml/exceptions_wms.py
247:    if XMLParser_import_library == XMLParser_import_lxml:

src/server/pywms/ogc/wmts/xml/exception.py
228:  if XMLParser_import_library == XMLParser_import_lxml:

src/server/pywms/ogc/wms_130/wmsxml/capabilities_wms.py
254:  if XMLParser_import_library == XMLParser_import_lxml:

src/server/pywms/ogc/wms_111/wmsxml/exception.py
227:  if XMLParser_import_library == XMLParser_import_lxml:

src/server/pywms/ogc/wmts/xml/capabilities.py
227:  if XMLParser_import_library == XMLParser_import_lxml:

src/server/pywms/ogc/wms_111/wmsxml/capabilities.py
227:  if XMLParser_import_library == XMLParser_import_lxml:

src/server/wsgi/wms/ogc/xml/v111/exception.py
228:  if XMLParser_import_library == XMLParser_import_lxml:

src/server/wsgi/wms/ogc/xml/v130/exceptions_wms.py
246:    if XMLParser_import_library == XMLParser_import_lxml:

src/server/wsgi/wms/ogc/xml/v111/capabilities.py
227:  if XMLParser_import_library == XMLParser_import_lxml:

src/server/wsgi/wms/ogc/xml/v130/capabilities_wms.py
254:    if XMLParser_import_library == XMLParser_import_lxml:

src/fusion/portableglobe/cutter/cgi-bin/core/glc_assembler.py
33:from lxml import etree

src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
38:from lxml import etree
```

earth_enterprise/src/server/wsgi/common/utils.py
```diff
@@ -23,7 +23,7 @@
 import re
 import socket
 import urllib2
-import xml.sax.saxutils as saxutils
+import defusedxml.sax.saxutils as saxutils
```
tst-lsavoie: The server is choking on this line. Apparently defusedxml doesn't implement saxutils. It might be best to create some kind of "wrapper package" that pulls the parsing functions from defusedxml and the other functions from other libraries. I haven't tested this at all but I think a file similar to the below would work:
```
import defusedxml
from xml.sax import saxutils

defusedxml.sax.saxutils = saxutils
# Other substitutions as needed
```
Then we could import the above file instead of defusedxml and use it like defusedxml.

earth_enterprise/src/server/wsgi/search/common/utils.py
```diff
@@ -22,7 +22,7 @@
 import logging.config
 import re
 from string import Template
-from xml.sax.saxutils import escape
+from defusedxml.sax.saxutils import escape
```
tst-lsavoie: Same as above: defusedxml doesn't implement saxutils

earth_enterprise/src/server/wsgi/common/utils.py
tst-lsavoie: I guess in this particular case we can just go back to the previous import. A wrapper might be useful if we have to make several replacements, though.

earth_enterprise/src/server/wsgi/search/plugin/custom_POI_search_handler.py
```diff
@@ -21,7 +21,7 @@
 import logging.config
 from string import Template
 import urllib2
-import xml.etree.cElementTree as ET
+import defusedxml.cElementTree as ET
```
tst-lsavoie: We're going to run into trouble in this file where we call `ET.SubElement` and `ET.tostring`, neither of which exist in defusedxml.

earth_enterprise/src/server/wsgi/common/utils.py
tst-lsavoie: There's an example of this here: https://wsgidav.readthedocs.io/en/latest/_modules/wsgidav/xml_tools.html. Basically, we could put something like the following code in a file and import it instead of defusedxml.
```
    from defusedxml import ElementTree as etree

    # defusedxml doesn't define these non-parsing related objects
    from xml.etree.ElementTree import Element, SubElement, tostring

    etree.Element = _ElementType = Element
    etree.SubElement = SubElement
    etree.tostring = tostring
```
I haven't tried this at all - it's just an idea at this point. I have no idea what the problems or pitfalls might be.

tst-lsavoie:
There are two implementations of WMS and one of them is unused. In addition to that there is an unused WMTS implementation. We can fix both issues by removing the entire pywms folder. That will also make reviewing and testing this easier as we can verify that WMS still works and have fewer files to review.

earth_enterprise/src/server/wsgi/common/utils.py
```diff
@@ -23,9 +23,12 @@
 import re
 import socket
 import urllib2
-import xml.sax.saxutils as saxutils
+import 
+import defusedxml.sax as DefusedSax
+from xml.sax import saxutils
 from geAbstractionFetcher import GetHostName
 
+DefusedSax.saxutils = saxutils
```
tst-sbayard: Is the purpose of this to prevent incorrect error reporting by Bandit?

tst-jkent: This is because defusedxml.sax doesn't contain saxutils. This was an issue reported by Bandit though.

tst-lsavoie: Yeah, my wrapper suggestion doesn't make as much sense here. It's most helpful when we're using some functions from defusedxml and some from the regular xml package. Here it might be simpler to just leave the old import.

If you want to get fancy, you could have a completely separate .py file that creates a package that combines the pieces we need from defusedxml and python xml and just have a single import statement here (and everywhere else we use xml) that imports that package.

earth_enterprise/src/server/wsgi/serve/push/search/util/search_schema_parser.py
```diff
@@ -123,9 +123,9 @@ def Parse(self, search_file, table_name, file_prefix=None):
       logger.info("File prefix is '%s'", file_prefix)
     self.__StartDocument()
     try:
-      context = xml.etree.cElementTree.iterparse(search_file,
+      context = ET.iterparse(search_file,
                                                  SearchSchemaParser.EVENTS)
-    except xml.etree.cElementTree.ParseError, e:
+    except ET.ParseError, e:
```
tst-lsavoie: This may be an issue. I don't think `cElementTree` includes `ParseError`, though I haven't had the chance to check that yet. For some reason, `ElementTree` does include it, but I don't know if there are problems switching between the two versions of `*ElementTree`.

earth_enterprise/src/server/wsgi/serve/push/search/util/search_schema_parser.py
```diff
@@ -123,9 +123,9 @@ def Parse(self, search_file, table_name, file_prefix=None):
       logger.info("File prefix is '%s'", file_prefix)
     self.__StartDocument()
     try:
-      context = xml.etree.cElementTree.iterparse(search_file,
+      context = ET.iterparse(search_file,
```
tst-lsavoie: Minor: maybe ought to clean up formatting here. With the shorter function call you can just put `SearchSchemaParser.EVENTS` on the same line.

tst-lsavoie:
I'm getting failures when I install from RPMs on CentOS 6 (CentOS 7 works fine). The issue seems to be that when we install `python-defusedxml` it installs it for Python 2.6 but not Python 2.7, and Open GEE uses 2.7. The commands to install the correct version of defusedxml are
```
sudo yum -y install python27-pip
sudo pip2.7 install defusedxml
```
That doesn't work well with RPMs because RPMs can't depend on pip modules. I think the only way to fix it will be to change the `requires` for the RPMs from `python-defusedxml` to `python27-pip` and then add a line in `earth_enterprise/rpms/opengee-server/snippets/pre-install.sh` to run the pip command. I think that will work for all platforms we support (except Ubuntu, which doesn't use RPMs), but we'll need to test that.

tst-lsavoie:
Oh, and don't forget to update the BUILD*.md files with the new commands as well.

earth_enterprise/BUILD_Ubuntu.md
```diff
@@ -39,7 +39,7 @@ sudo apt-get install \
     libx11-dev libxcursor-dev libxerces-c-dev libxft-dev libxinerama-dev \
     libxml2-dev libxml2-utils libxmu-dev libxrandr-dev libyaml-cpp-dev \
     openssl libpcre3 libpcre3-dev \
-    python-dev python-imaging python-psycopg2 python-setuptools python2.7 \
-    python2.7-dev python-lxml python-git \
+    python-dev python-imaging python-psycopg2 \
```
tst-lsavoie: I don't think you need to change this file to install python27-pip. I think Ubuntu works fine if you just install python-defusedxml.

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -147,12 +147,13 @@ sudo yum install -y \
   libcap-devel libmng-devel libpng-devel libX11-devel libXcursor-devel \
   libXft-devel libXinerama-devel libxml2-devel libXmu-devel libXrandr-devel \
   ogdi-devel openjpeg-devel openjpeg2-devel openssl-devel pcre pcre-devel \
-  perl-Alien-Packages perl-Perl4-CoreLibs proj-devel python-devel python27-devel python-unittest2 \
+  perl-Alien-Packages perl-Perl4-CoreLibs proj-devel \
+  python-devel python27-devel python27-pip python-unittest2 \
   rpm-build rpmrebuild rsync scons shunit2 \
   xerces-c xerces-c-devel xorg-x11-server-devel yaml-cpp-devel zlib-devel
```
tst-lsavoie: We need another step here to run `sudo pip2.7 install defusedxml` after running the yum command.

Let me summarize everything so we're on the same page:

- We need to install defusedxml for the proper version of python on:
    - All platforms (CentOS 6, 7, and Ubuntu 16)
    - All methods for installing GEEP (manual instructions and RPMs/DEBs)
- For Ubuntu, installing `python-defusedxml` works fine. So we need to:
    - Add `python-defusedxml` to the manual instructions
    - Add `python-defusedxml` as a DEB requirement (so that it will be installed automatically when we install the DEB)
- For CentOS 6, installing `python-defusedxml` will install defusedxml for the wrong version of Python, so we have to fall back to a two-step process: 1) install pip2.7 and 2) use pip2.7 to install defusedxml.
    - For the manual instructions, we simply add both commands.
    - For the RPMs, we can add `python-pip27` as a requirement, but there's no way to add a pip package as a requirement. Thus, we have to do the second step (`pip install defusedxml`) in the pre-install script.
- For CentOS 7, installing `python-defusedxml` works fine, so we can either go with that method or use the same two-step process as on CentOS 6 for consistency.

Hopefully that gives you a clearer picture of what needs to be changed and where.

earth_enterprise/src/server/wsgi/search/plugin/custom_POI_search_handler.py
```diff
@@ -21,10 +21,14 @@
 import logging.config
 from string import Template
 import urllib2
-import xml.etree.cElementTree as ET
+from xml.cElementTree import SubElement, tostring
```
tst-lsavoie: There's an error here. It should be `xml.etree.cElementTree`.

tst-lsavoie:
FYI, I tested WMS compliance for 1.1.1 and 1.3.0 using http://cite.opengeospatial.org/teamengine. This code produces the exact same results as the previous code. Unfortunately, neither version passes fully, but that's a different problem. I also verified that if I install this version of the PR and then uninstall defusedxml the test results are much worse, which implies that we're using defusedxml correctly.

tst-lsavoie:
I have tested everything on this PR except for BUILD*.md, build.gradle, and pre-install.sh -- basically everything to do with installing the defusedxml dependency because we haven't decided how we're going to handle that yet. So far all tests pass and everything looks great.

tst-ccamp:
I reviewed the changes and they look good. I really like how much it simplifies a lot of the import code. The Travis build was failing, which is not surprising since the third_party libraries have changed. So I cleared the cache and started a new one. Hopefully this one succeeds.

earth_enterprise/src/third_party/defusedxml/SConscript
```diff
@@ -0,0 +1,90 @@
+Import('third_party_env')
```
tst-lsavoie: We need to include the copyright header in this file. The Copyright should read "Copyright 2019 The Open GEE Contributors".

earth_enterprise/src/third_party/defusedxml/SConscript
```diff
@@ -0,0 +1,90 @@
...
+defusedxml_target = '%s/.install' % current_dir
+defusedxml_install = defusedxml_env.Command(
+    defusedxml_target, defusedxml_build,
+    [defusedxml_env.MultiCommand('\n'.join([
+        'cd %s' % build_root,
+        'rm -rf %s' % install_root_lib,
+        'rm -rf %s' % install_root_doc,
+        'mkdir -p %s' % install_root_lib,
+        'mkdir -p %s' % install_root_doc,
+        'rsync -pvr build/lib.linux-x86_64-%s/ %s' % (
```
tst-lsavoie: I think this line should be changed to `'rsync -pvr build/lib/defusedxml/ %s' % (install_root_lib),` but I'm still getting build errors further down.

earth_enterprise/src/third_party/SConscript
```diff
@@ -27,6 +27,7 @@ third_party_env = env.Clone()
 if third_party_env['bundle_3rd_libs']:
   third_party_libs = [
       'date',
+      'defusedxml',
```
tst-lsavoie: I believe you can fix the build issues by moving defusedxml down in this list so it's near pyyaml (in both places). I'm testing that right now.

tst-lsavoie: My test build seems to have passed the third_party build now, so this seems to work.

earth_enterprise/src/third_party/defusedxml/SConscript
```diff
@@ -0,0 +1,90 @@
...
+defusedxml_build = defusedxml_env.Command(
+    defusedxml_target, defusedxml_extract,
+    [defusedxml_env.MultiCommand(
+        'cd %s\n'
+        '%s%s %s setup.py build\n'
+        'touch %s' % (build_root, defusedxml_env['ENV']['mod_env'],
+                      env_opt, python_cmd, defusedxml_target))])
+
+# [3] Create defusedxml master installer.
+install_root_lib = '%s/lib/python%s/site-packages/yaml' % (
```
tst-sbayard: Minor: don't need `/yaml` in this path. `pyyaml` is the only other third-party build that adds that part.

tst-lsavoie: The build is now passing for me but trying to push a database fails because it can't find defusedxml. This may fix that.

earth_enterprise/src/third_party/SConscript
```diff
@@ -59,11 +59,13 @@ if third_party_env['bundle_3rd_libs']:
...
       'PIL',
       'skia',
-      'pyyaml'
+      'pyyaml',
+      'defusedxml'
   ]
 else:
   third_party_libs = [
       'date',
+      'defusedxml',
```
tst-lsavoie: This one should also be moved down.

earth_enterprise/src/third_party/defusedxml/SConscript
tst-lsavoie: The end goal is to have the defusedxml files end up under `/opt/google/gepython/Python-2.7.1/lib/python2.7/site-packages/defusedxml` after GEE is installed. If I manually make that change my tests pass.

earth_enterprise/src/third_party/defusedxml/SConscript
```diff
@@ -0,0 +1,89 @@
...
+defusedxml_target = '%s/.install' % current_dir
+defusedxml_install = defusedxml_env.Command(
+    defusedxml_target, defusedxml_build,
+    [defusedxml_env.MultiCommand('\n'.join([
+        'cd %s' % build_root,
+        'rm -rf %s' % install_root_lib,
+        'rm -rf %s' % install_root_doc,
+        'mkdir -p %s' % install_root_lib,
+        'mkdir -p %s' % install_root_doc,
+        'rsync -pvr build/lib.linux-x86_64-%s/ %s' % (install_root_lib),
```
tst-lsavoie: The source directory is still wrong on this line. It should be `rsync -pvr build/lib/defusedxml %s' % (install_root_lib)`

tst-lsavoie: Also, don't forget to add the copyright comment at the top of the file (see above).

PR #1581, opened 2019-12-14
1579: Add UseMinification option to misc.xml.

### LINKED ISSUE #1579: Add an option to misc.xml for disabling minification
**Describe the bug**

OpenGEE performs "minification" to generate imagery for different zoom levels.

In the current implementation the imagery at level N is generated by "minifying" the imagery at level N+1, introducing a dependency between minification tasks.
 
Disabling minification and building imagery packets using the source imagery (instead of the results from the previous level) reduces dependencies between tasks, which in turn improves concurrency and overall project build time in some situations.  The cost of the extra performance is slightly increased memory usage, because the (potentially large) source image must be read into memory more often.

**To Reproduce**

Build OpenGEE with minification (this will be the case with a fresh clone) and measure the time and memory required to build a large imagery project.

Disable minification by hard coding beginMinifyLevel and endMinifyLevel to 0, rebuilding OpenGEE, and measure the time and memory required to build the project.

**Expected behavior**

There should be a configuration option in the asset root's misc.xml to enable or disable minification so that end users can tweak performance for their particular workload.

tst-jlarocco:
Fixes #1579


earth_enterprise/integration_test/specs/minification.spec
```diff
@@ -0,0 +1,50 @@
...
+' * Example steps
+' -----------------------------------------
+' POST-STEPs for this SCENARIO go here.
+
+'##########################################
+'# DO NOT DELETE ANYTHING ABOVE THIS LINE #
+'##########################################
+
+
+Open GEE New Imagery Resource Command Test
```
tst-lsavoie: Should change the title to refer to minification

earth_enterprise/src/fusion/autoingest/plugins.src/Database.src
```diff
@@ -881,8 +885,12 @@ DatabaseAssetVersionImplD::CreateSubsampledTerrain(
       FindNeededTerrainInsets(ssGenInfo[i].coverage,
                               terrainInsetPtrs, numTerrain,
                               neededIndexes,
-                              beginMinifyLevel,
-                              endMinifyLevel);
+                              (MiscConfig::Instance().UseMinification
```
tst-lsavoie: Unless I'm missing something it doesn't look like you need these changes. Since you set begin/endMinifyLevel correctly above, you shouldn't need to check the flag again here.

earth_enterprise/src/fusion/autoingest/plugins.src/Database.src
```diff
@@ -771,8 +771,12 @@ DatabaseAssetVersionImplD::CreateSubsampledTerrain(
     terrain_project->LoadInsetInfo(ClientTmeshTilespaceFlat,
                                   beginSSLevel, endSSLevel,
                                   terrainInsets);
-    const uint beginMinifyLevel  = StartTmeshLevel;
-    const uint endMinifyLevel    = terrainInsets[0].effectiveMaxLevel;
+    const uint beginMinifyLevel  = (MiscConfig::Instance().UseMinification
```
tst-lsavoie: Minor: this logic is repeated a number of times. It might be nice to put it in a function somehow. It's not that much logic, though, so maybe it's not useful.

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,57 @@
...
+import shutil
+
+# Re-use some of the functionalities
+import assets
+
+
+MISC_XML_PATH = os.path.join(assets.ASSET_ROOT, ".config", "misc.xml")
+
+
+# now = datetime.datetime.now()
```
tst-lsavoie: Should remove commented out code.

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,57 @@
...
+# outputDirectoryForTest = "minification.output." + now.strftime("%Y.%m.%d.%H.%M.%S")
+# outputROOT = "/gevol/assets"
+
+
+@step("Turn minification <status>")
+def turn_minification(status):
+    "Turn off minification in misc.xml."
+
+    fromState = None
```
tst-lsavoie: Looks like these two variables are unused.

earth_enterprise/integration_test/specs/minification.spec
```diff
@@ -0,0 +1,50 @@
...
+* Turn minification "off"
+* Restart fusion
+* Create imagery resource "blue_marble_no_minify" from "Imagery/bluemarble_4km.tif"
+* Create imagery resource "sf_hires_no_minify" from "Imagery/usgsSFHiRes.tif"
+* Create imagery project "without_minification"
+* Add imagery resource "blue_marble_no_minify" to project "without_minification"
+* Add imagery resource "sf_hires_no_minify" to project "without_minification"
+* Build imagery project "without_minification"
+* Wait for imagery project "without_minification" to reach state "Succeeded"
+* Verify project "with_minification" has more dependencies than project "without_minification"
```
tst-lsavoie: Minor: it might be nice to look for a specific number of dependencies here, but I can live without that.

earth_enterprise/src/fusion/autoingest/MiscConfigStorage.idl
```diff
@@ -40,6 +40,9 @@ class MiscConfigStorage {
   uint         PackGenAnalysisThreads = uint(1);
   GraphOpsType GraphOperations = MiscConfigStorage::NO_GRAPH_OPS;
 
+  // True to minify previous imagery level, false to minify source image
```
tst-lsavoie: A few of points about this comment:
- "True" will minify previous imagery and terrain levels, correct (not just imagery)?
- The "false" text is a bit confusing to me. I would say something like "false to build from source imagery"

earth_enterprise/integration_test/specs/minification.spec
tst-lsavoie: I was actually referring to the line above this, which refers to the "New Imagery Resource Command", which was copied from another test. It should mention minification instead.

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,51 @@
+from getgauge.python import step, before_scenario, Messages
```
tst-lsavoie: Minor: there may be some unused imports here. Might be nice to clean those up.

earth_enterprise/src/fusion/autoingest/MiscConfigStorage.idl
```diff
@@ -40,6 +40,9 @@ class MiscConfigStorage {
   uint         PackGenAnalysisThreads = uint(1);
   GraphOpsType GraphOperations = MiscConfigStorage::NO_GRAPH_OPS;
 
+  // True to minify previous imagery and terrain levels, false to build from source imagery
```
tst-lsavoie: I should have realized this with my last suggestion, but at the end of this sentence it should probably refer to imagery and terrain.

tst-lsavoie:
One question I have that should be tested - if we set begin and endMinifyLevel both to 0, does that mean no levels will be minified or only level 0 will be minified? We should double check that we're not accidentally minifying one level.

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,51 @@
...
+
+# Re-use some of the functionalities
+import assets
+
+
+MISC_XML_PATH = os.path.join(assets.ASSET_ROOT, ".config", "misc.xml")
+
+
+@step("Turn minification <status>")
+def turn_minification(status):
```
tst-lsavoie: Minor point: this won't work if `UseMinification` isn't in the misc.xml file to begin with, which is allowed by Fusion. I'd suggest checking if the line is there, using sed to add `<UseMinification>0<\/UseMinification>` right before `</MiscConfigStorage>` if it's not there, then running the existing sed commands as they are.

tst-lsavoie:
> One question I have that should be tested - if we set begin and endMinifyLevel both to 0, does that mean no levels will be minified or only level 0 will be minified? We should double check that we're not accidentally minifying one level.

FYI, I checked on this, and setting both to 0 does disable minification at level 0, so we seem to be ok there.

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,59 @@
...
+    # "Fusion Fatal: socket connect '127.0.1.1:13031': Connection refused"
+    run_and_wait("sudo service gefusion status")
+    
+
+@step("Verify project <first> has more dependencies than project <second>")
+def verify_minification(first, second):
+    firstDepCount = len(subprocess.check_output(["/opt/google/bin/gequery", "--dependencies", os.path.join(assets.IMAGERY_PROJECT_PATH, first)]).splitlines())
+    secondDepCount = len(subprocess.check_output(["/opt/google/bin/gequery", "--dependencies", os.path.join(assets.IMAGERY_PROJECT_PATH, second)]).splitlines())
+    print "firstDepCount = {} secondDepCount = {}".format(firstDepCount, secondDepCount)
+    assert(firstDepCount == 119)
```
tst-lsavoie: Minor: this makes this function less general. It might be a good idea to pass in the number of expected dependencies for each resource. I'm fine with it as is, though.

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,59 @@
...
+    assert(secondDepCount == 101)
+    assert(firstDepCount > secondDepCount)
```
tst-lsavoie: Minor: this assert might not be necessary any more.

tst-lsavoie:
I always forget this, but we need to add this option to [this documentation page](https://github.com/google/earthenterprise/blob/master/docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst) and the release notes.

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,59 @@
...
+
+    newStatus = {"off": "0",
+                 "on":  "1"}[status]
+
+    miscTree = parse(MISC_XML_PATH)
+
+    for config in miscTree.iter('MiscConfigStorage'):
+        minificationCount = len(list(config.iter('UseMinification')))
+        if minificationCount == 0:
+            minification = ElementTree.SubElement(config, 'UseMinification')
```
tst-lsavoie: I get an error here: "global name 'ElementTree' is not defined"

earth_enterprise/integration_test/step_impl/minification.py
```diff
@@ -0,0 +1,59 @@
...
+            minification.text = newStatus
+        else:
+            for useMin in config.iter('UseMinification'):
+                useMin.text = newStatus
+
+    miscTree.write(MISC_XML_PATH)
```
tst-lsavoie: This may be a problem in some setups. The user that runs the tests may not have permissions to change the misc.xml file. I'm not really sure how to handle that.

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -410,6 +410,21 @@ Configure Fusion performance
       - ``ALL_GRAPH_OPS`` or ``2``: use all new operations regardless of
         performance impact
 
+      .. rubric:: Use Minification
```
tst-lsavoie: We need to build the docs and check in the generated files.

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -410,6 +410,21 @@ Configure Fusion performance
...
+
+      Included in ``/gevol/assets/.config/misc.xml`` is a
+      ``UseMinification`` option that controls whether terrain and imagery 
+      levels are generated from the previous level or from the original source
+      file.  This allows reducing dependencies between levels in favor of
```
tst-lsavoie: My understanding of this change is slightly different. Not using minification reduces dependencies between levels which reduces memory usage, but it may reduce the performance of some tasks. Basically, system manager does less work but packgen tasks do more work. However, that is based on code inspection, not testing, so your description may be closer to the truth.

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -410,6 +410,21 @@ Configure Fusion performance
...
+      .. rubric:: Use Minification
+
+      Included in ``/gevol/assets/.config/misc.xml`` is a
+      ``UseMinification`` option that controls whether terrain and imagery 
+      levels are generated from the previous level or from the original source
+      file.  This allows reducing dependencies between levels in favor of 
+      increased memory use, and can increase performance in some situations by
+      increasing parallelism.
+      ``UseMinification`` can be set to the following values:
+      - ``0``: Each level is minified from the original source.  This increases
```
tst-jkent: The description of the UseMinification option says it controls how levels are generated but here it says it is how it is minified. This may be confusing since the option is called UseMinification, yet both settings say they set how the level is minified instead of disabling or enabling minification.

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -413,16 +413,14 @@ Configure Fusion performance
       .. rubric:: Use Minification
 
       Included in ``/gevol/assets/.config/misc.xml`` is a
-      ``UseMinification`` option that controls whether terrain and imagery 
-      levels are generated from the previous level or from the original source
-      file.  This allows reducing dependencies between levels in favor of 
-      increased memory use, and can increase performance in some situations by
-      increasing parallelism.
+      ``UseMinification`` option that controls how data for imagery and terrain
+      zoom levels is generated.
```
tst-jkent: Minor: Should be **are** generated

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -413,16 +413,14 @@ Configure Fusion performance
...
       ``UseMinification`` can be set to the following values:
-      - ``0``: Each level is minified from the original source.  This increases
-        memory use, but decreases dependencies between resources.
-      - ``1``: Each level is minified from the previous level.  This increases
-        the number of dependencies between resources, but decreases memory use.
+      - ``1``: Imagery and terrain for each zoom level is generated by
```
tst-jkent: Minor: I believe the same applies for both option descriptions here.

docs/geedocs/5.3.3/answer/fusionAdministration/confTaskRulesForFusionPerf.html
```diff
@@ -460,11 +461,23 @@ <h1>Configure Fusion performance<a class="headerlink" href="#configure-fusion-pe
 <li><code class="docutils literal notranslate"><span class="pre">ALL_GRAPH_OPS</span></code> or <code class="docutils literal notranslate"><span class="pre">2</span></code>: use all new operations regardless of
 performance impact</li>
 </ul>
+<p class="rubric" id="use-minification">Use Minification</p>
```
tst-jkent: Same as before. **is** should be **are**

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -410,10 +410,24 @@ Configure Fusion performance
...
 
+      .. _Use_Minification:
+      .. rubric:: Use Minification
+
+      Included in ``/gevol/assets/.config/misc.xml`` is a ``UseMinification``
+      option that controls how data for imagery and terrain zoom levels are
+      generated.  ``UseMinification`` can be set to the following values:
+
+      - ``1``: Imagery and terrain for each zoom level are generated by
+        minifying the level below it.  This is the default.
```
tst-lsavoie: Might be nice to add "where possible" to the end of the first sentence. Even if "UseMinification" is on, it won't minify for levels higher than the lowest zoom level of any image in the project.

docs/geedocs/5.3.3/answer/fusionAdministration/confTaskRulesForFusionPerf.html
```diff
@@ -495,4 +495,4 @@ <h3>Navigation</h3>
       Created using <a href="http://sphinx-doc.org/">Sphinx</a> 1.7.5.
```
tst-lsavoie: We should back out these 5.3.3 changes fully if we can. It doesn't look like these changes matter, but we shouldn't be changing any 5.3.3 docs anyway.

I may be able to do that from the web interface.

earth_enterprise/integration_test/step_impl/minification.py
tst-lsavoie: I think we may need to change the test_setup.sh script to change the permissions on misc.xml so that anyone can modify it. I think that's fine for running tests.

docs/geedocs/5.3.3/answer/fusionAdministration/confTaskRulesForFusionPerf.html
tst-jlarocco: Oops, I had misconfigured something in Emacs and it added newlines at the end of these when I backed out the other changes...

docs/geedocs/5.3.4/sources/answer/releaseNotes/relNotesGEE5_3_4.rst.txt
```diff
@@ -28,8 +28,9 @@ new PostgreSQL version used by Open GEE 5.3.4.
 
 -[ Resolved Issues ]-
 
-      Number  Description  Resolution  Issue number  Issue Description
-Issue Resolution
+      Number  Description  Resolution  1604  WMS capability doesn't
```
tst-lsavoie: Is this change intended to be on this PR? It doesn't look relevant.

Also, are there release notes changes for the minify feature?

tst-jlarocco: It's not, but I'm not sure what to do about it.  Looks like a pull request merged a few days ago updated the release notes RST but forgot to regenerate the HTML and .rst.txt 

https://github.com/google/earthenterprise/commit/1985695bc580c15546640232cd993fdc43565440

earth_enterprise/integration_test/specs/minification.spec
```diff
@@ -0,0 +1,50 @@
...
+* Turn minification "off"
+* Restart fusion
+* Create imagery resource "blue_marble_no_minify" from "Imagery/bluemarble_4km.tif"
+* Create imagery resource "sf_hires_no_minify" from "Imagery/usgsSFHiRes.tif"
+* Create imagery project "without_minification"
+* Add imagery resource "blue_marble_no_minify" to project "without_minification"
+* Add imagery resource "sf_hires_no_minify" to project "without_minification"
+* Build imagery project "without_minification"
+* Wait for imagery project "without_minification" to reach state "Succeeded"
+* Verify project "with_minification" has "119" dependencies and project "without_minification" has "101"
```
tst-lsavoie: Minor: I'd suggest turning minification back on at the end of this test since that's the default. I'm running some tests right now and it seems to take longer because minification is off for the rest of the tests.

My approval stands with or without this change.

earth_enterprise/integration_test/specs/minification.spec
```diff
@@ -0,0 +1,51 @@
...
+* Turn minification "on"
```
tst-lsavoie: There's a missing step here to restart fusion after turning minification on.

Also, while you're at it, it would be handy to add some tags to this test. Something like `Tags: minify` just under the `----` line would be great.

Neither of those needs to hold up the PR, so my approval stands.

PR #1584, opened 2019-12-18
Integrate 5.3.2.1

tst-dsugar:
I only think one change may be needed from for master since last merge from 5.3.2.1.  There was other work @tst-lsavoie did to kill some endpoint code in 5.3.2.1, but I was not initially certain if we needed to do that for master.  There was also some work @tst-rwildes  did with relocating files to git lfs.  Just need to confirm if these changes are needed on master, if so I will add them to this branch before we merge.  I saw some work on scons/builds, but since master builds currently, I assume its either dup of master work or not applicable.  Fortunately not a lot of work was done in 5.3.2.1 after python 2.7.


tst-lsavoie:
We should include my work to remove the unused endpoint. That's intended to be permanent.

Is there any reason not to just merge the whole branch? If we don't we risk losing things.

tst-dsugar:
> We should include my work to remove the unused endpoint. That's intended to be permanent.
> 
> Is there any reason not to just merge the whole branch? If we don't we risk losing things.

If you do the whole branch merge you get 60 files changes, and a dozen merge conflicts.  Most of them are about reverting docs one version back.  So clearly that is wrong.  I simply am going to cherrypick the few changes as needed after colby merged python 2.7 from 5.3.2.1, they really have diverged a lot.

tst-dsugar:
> We should include my work to remove the unused endpoint. That's intended to be permanent.
> 
> Is there any reason not to just merge the whole branch? If we don't we risk losing things.

And yes, I will add your changes then...

tst-ccamp:
> If you do the whole branch merge you get 60 files changes, and a dozen merge conflicts. Most of them are about reverting docs one version back. So clearly that is wrong. I simply am going to cherrypick the few changes as needed after colby merged python 2.7 from 5.3.2.1, they really have diverged a lot.

I'm generally opposed to cherry-picking on principle. It pollutes the git history. I think we can easily remove the automatically generated doc changes from this branch. Once the source changes are merged in, then we can run a new doc gen on the master branch. That should get rid of most of the conflicts.

tst-lsavoie:
I think this can be closed because it was superseded by PRs #1588 and #1589.

PR #1591, opened 2019-12-20
Add license file for defusedxml

tst-jkent:


tst-lsavoie:
Update: I ran a build with this file included and everything was fine (which is expected since this file doesn't affect the build).

Issue #1593, opened 2019-12-27
Python Dependency Issues on EL6


tst-cjeffries:
Portable documentation does not indicate that dependencies need to be installed with python 2.7 rather than the system python.

1. Install portable following the instructions.
2. Attempt to run portable (via directly running the python file).
3. Notice that dependencies are missing from python2.7.

This only affects EL6.

___________________________________________________________________

Python27-lxml does not get installed during RPM installations. 

1. Install OpenGEE from rpms built on a different platform.
2. Attempt to make a cut from a globe.
3. The cut will fail due to import errors.

Only affects EL6.

___________________________________________________________________

pyyaml does not get detected by opengee server. It is installed incorrectly.

1. Build and install OpenGEE.
2. Attempt to assemble a GLC.
3. The assembly will fail due to an import error.

Only affects EL6.

tst-cjeffries:
PR has been merged.

PR #1597, opened 2019-12-31
#1593: Resolve Assorted Python Dependency Issues

tst-cjeffries:
Updates to fix various python dependency issues:

- Portable shebang fixed and documentation changed.
- Pyyaml now installs to the correct location.
- OpenGEE rpms now require python27-lxml (to ensure that it gets installed automatically).

To test:

- Build portable on el6 and verify that it builds with and works with python 2.7.
- Build and install on el6 and verify that glc_assembler works.
- Take built RPMs and install on a different el6 machine. Verify that cutter works.

These changes were made in conjunction with documentation changes on the [wiki](https://github.com/google/earthenterprise/wiki/Portable-Server#run-time-prerequisites).

### DIFF
--- a/earth_enterprise/src/portableserver/build.py
+++ b/earth_enterprise/src/portableserver/build.py
@@ -1,4 +1,4 @@
-#! /usr/bin/python
+#! /usr/bin/env python2.7
 #-*- Python -*-
 #
 # Copyright 2017 GEE Open Source Team <github.com/google/earthenterprise>
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -385,6 +385,7 @@ task openGeeCommonRpm(type: GeeRpm, dependsOn: openGeeCommonInitScripts) {
         'OpenLDAP and ' +
         'OpenSSL and ' +
         'Skia 5.1.2'
+ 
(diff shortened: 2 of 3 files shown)


earth_enterprise/src/portableserver/build.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/python
+#! /usr/bin/env python2.7
```
tst-dsugar: This probably makes sense to be fully consistent between portable and everything else now....

tst-dsugar:
Should this target 5.3.3?

earth_enterprise/src/portableserver/build.py
tst-cjeffries: Yes, plus it was causing portable to be using python 2.6.

tst-cjeffries:
Originally I had it for 5.3.3, but Rick wanted these changes applied to 5.3.2.1, and then merged to 5.3.3. It makes sense to do it that way because the problems here were introduced with the python dependency changes so they need to be in 5.3.2.1.

tst-jzmolek:
Reviewed and tested. Looks good to me. Although shouldn't this target release_5.3.3?

PR #1599, opened 2020-01-03
Dont load unneeded children and inputs

tst-sbayard:
Using a new DFS visitor that avoids unnecessarily loading children and inputs into the cache when propagating Canceled and Failed states.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/AssetVersion.h
+++ b/earth_enterprise/src/fusion/autoingest/AssetVersion.h
@@ -138,7 +138,6 @@ class AssetVersionImpl : public AssetVersionStorage, public StorageManaged {
   }
 
   virtual bool InputStatesAffectMyState(AssetDefs::State stateByInputs, bool blockedByOfflineInputs) const {assert(false); return true;}
-  virtual bool ChildStatesAffectMyState() const {assert(false); return true;}
   const SharedString & GetRef(void) const { return name; }
   std::string GetAssetRef(void) const {
     AssetVersionRef verref(name);
--- a/earth_enterprise/src/fusion/autoingest/AssetVersion.h
+++ b/earth_enterprise/src/fusion/autoingest/AssetVersion.h
@@ -137,6 +137,8 @@ class AssetVersionImpl : public AssetVersionStorage, public StorageManaged {
     return (state == AssetDefs::Bad);
   }
 
+  virtual bool InputStatesAffectMyState(AssetDefs::State stateByInputs, bool blockedByOfflineInputs) const {assert(false); return true;}
+  virtual bool ChildStatesAffectMyState() const {assert(false); return true;}
   const SharedString & GetRef(void) const { return name; }
   std::string GetAssetRef(void) const {
     AssetVersionRef verref(name);
--- a/earth_enterprise/src/fusion/autoingest/storage/AssetDefs.idl
+++ b/earth_enterprise/src/fusion/autoingest/storage/AssetDefs.idl
@@ -41,6 +41,9 @@ class AssetDefs {
   static bool NotFinished(State state) {
     return !Finished(state);
   }
+  static bool IsBlocking(State state) {
+    return state & (AssetDefs::Blocked | AssetDefs::Canceled | AssetDefs::Failed);
+  }
   static bool CanRebuild(State state) {
     return state & (AssetDefs::Canceled | AssetDefs::Failed);
   }
--- a/earth_enterprise/src/fusion/autoingest/storage/AssetDefs.idl
+++ b/earth_enterprise/src/fusion/autoingest/storage/AssetDefs.idl
@@ -42,10 +42,10 @@ class AssetDefs {
     return !Finished(state);
   }
   static bool IsBlocking(State state) {
-    return state & (AssetDefs::Blocked | AssetDefs::Canceled | AssetDefs::Failed);
+    return state & (Blocked | Canceled | Failed | Offline | Bad);
   }
   static bool CanRebuild(State state) {
-    return state & (AssetDefs::Canceled | AssetDefs::Failed);
+    return state & (Canceled | Failed);
   }
   static bool CanOffline(State state) {
     return ((state != Offline) &&
--- a/earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
@@ -129,7 +129,7 @@ void HandleTaskDone(const TaskDoneMsg & msg, MiscConfig::GraphOpsType graphOps) 
         stateUpdater->SetSucceeded(version);
       }
       else {
-        stateUpdater->SetFailed(version);
+        stateUpdater->SetStateForRefAndDependents(msg.verref, AssetDefs::Failed, AssetDefs::Finished);
       }
     }
     else if (!version) {
--- a/earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
@@ -129,7 +129,7 @@ void HandleTaskDone(const TaskDoneMsg & msg, MiscConfig::GraphOpsType graphOps) 
         stateUpdater->SetSucceeded(version);
       }
       else {
-        stateUpdater->SetStateForRefAndDependents(msg.verref, AssetDefs::Failed, AssetDefs::Finished);
+        stateUpdater->SetFailed(version);
       }
     }
     else if (!version) {
--- a/earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
@@ -50,7 +50,7 @@ void RebuildVersion(const SharedString & ref, MiscConfig::GraphOpsType graphOps)
       }
     }
 
-    stateUpdater->SetStateForRefAndDependents(ref, AssetDefs::New, AssetDefs::CanRebuild);
+    stateUpdater->SetAndPropagateState(ref, AssetDefs::New, AssetDefs::CanRebuild);
   }
   else {
     MutableAssetVersionD version(ref);
@@ -80,7 +80,7 @@ void CancelVersion(const SharedString & ref, MiscConfig::GraphOpsType graphOps) 
       }
     }
 
-    stateUpdater->SetStateForRefAndDependents(ref, AssetDefs::Canceled, AssetDefs::NotFinished);
+    stateUpdater->SetAndPropagateState(ref, AssetDefs::Canceled, AssetDefs::NotFinished);
   }
   else {
     MutableAssetVersionD version(ref);
--- a/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
@@ -168,7 +168,7 @@ void DependentStateTreeFactory::FillInVertex(
   if (vertices[name].includeConnections) {
     // In some cases, assets in the dependent tree don't need their inputs and
     // children to calculate their states.
-    if (includeDepDescendents || !tree[myVertex].inDepTree) {
+    if (includeDepDescendents) {
       for (const auto &child : version->children)
       {
         auto childVertex = AddOrUpdateVertex(child, false, false);
--- a/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
@@ -169,8 +169,8 @@ void DependentStateTreeFactory::FillInVertex(
   // state. If I don't need to recalculate my state, I don't need to add any of
   // my connections. I'm only used to calculate someone else's state.
   if (vertices[name].includeConnections) {
-    // In some cases, assets in the dependent tree don't need their inputs and
-    // children to calculate their states.
+    // In some cases, assets don't need all their inputs and children to
+    // calculate their states.
     if (includeAllChildrenAndInputs) {
       for (const auto &child : version->children)
       {
--- a/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
@@ -153,8 +153,8 @@ void DependentStateTreeFactory::FillInVertex(
     tree[myVertex].inDepTree = false;
     vertices[name].includeConnections = false;
   }
-  // If I'm in the dependency tree I need to add my dependents because they are
-  // also in the dependency tree.
+  // If I'm in the dependency tree, and if the new state propagates to dependents,
+  // then I need to add my dependents to the graph and the dependency tree.
   if (tree[myVertex].inDepTree && includeDependentChildren) {
     vector<SharedString> dependents;
     version->DependentChildren(dependents);
--- a/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
@@ -33,7 +33,8 @@ class DependentStateTreeFactory
 
     SharedString ref;
     std::function<bool(AssetDefs::State)> includePredicate;
-    bool includeDepDescendents;
+    bool includeDependentChildren;
+    bool includeAllChildrenAndInputs;
     StorageManagerInterface<AssetVersionImpl> * const storageManager;
     map<SharedString, VertexData> vertices;
     size_t index;
@@ -56,10 +57,12 @@ class DependentStateTreeFactory
     DependentStateTreeFactory(
         const SharedString &ref,
         std::function<bool(AssetDefs::State)> includePredicate,
-        bool includeDepDescendents,
+        bool includeDependentChildren,
+        bool includeAllChildrenAndInputs,
         StorageManagerInterface<AssetVersionImpl> *sm) :
-      ref(ref), includePredicate(includePredicate),
-      includeDepDescendents(includeDepDescendents), storageManager(sm), index(0) {}
+      ref(ref), includePredicate(includePredicate), 
+      includeDependentChildren(includeDependentChildren),
+      includeAllChildrenAndInputs(includeAllChildrenAndInputs), storageManager(sm), index(0) {}
     DependentStateTree BuildTree();
 };
 
@@ -152,7 +155,7 @@ void DependentStateTreeFactory::FillInVertex(
   }
   // If I'm in the dependency tree I need to add my dependents because they are
   // also in the dependency tree.
-  if (tree[myVertex].inDepTree) {
+  if (tree[myVertex].inDepTree && includeDependentChildren) {
     vector<SharedString> dependents;
     version->DependentChildren(dependents);
     for (const auto & dep : dependents) {
@@ -168,7 +171,7 @@ void DependentStateTreeFactory::FillInVertex(
   if (vertices[name].includeConnections) {
     // In some cases, assets in the dependent tree don't need their inputs and
     // children to calculate their states.
-    if (includeDepDescendents) {
+    if (includeAllChildrenAndInputs) {
       for (const auto &child : version->children)
       {
         auto childVertex = AddOrUpdateVertex(child, false, false);
@@ -213,8 +216,9 @@ void DependentStateTreeFactory::AddEdge(
 DependentStateTree BuildDependentStateTree(
     const SharedString & ref,
     std::function<bool(AssetDefs::State)> includePredicate,
-    bool includeDepDescendents,
+    bool includeDependentChildren,
+    bool includeAllChildrenAndInputs,
     StorageManagerInterface<AssetVersionImpl> * sm) {
-  DependentStateTreeFactory factory(ref, includePredicate, includeDepDescendents, sm);
+  DependentStateTreeFactory factory(ref, includePredicate, includeDependentChildren, includeAllChildrenAndInputs, sm);
   return factory.BuildTree();
 }
--- a/earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
+++ b/earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
@@ -347,10 +347,7 @@ class StateUpdater::SetBlockingStateVisitor : public default_dfs_visitor {
 
       // Set the state for assets in the dependent tree.
       if (data.inDepTree) {
-        // Check if we're going to recalculate the state below. If not, we need
-        // to run the handlers now.
-        bool runHandlers = true;//UserActionRequired(newState);
-        SetState(vertex, newState, {0, 0}, runHandlers);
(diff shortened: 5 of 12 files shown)


earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
```diff
@@ -152,7 +155,7 @@ void DependentStateTreeFactory::FillInVertex(
   }
   // If I'm in the dependency tree I need to add my dependents because they are
```
tst-lsavoie: Might want to update this comment, though I'm not sure quite how to word it.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +278,105 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+        AssetDefs::State newState,
+        const WaitingFor & waitingFor) const {
+      SharedString name = tree[vertex].name;
+      AssetDefs::State oldState = tree[vertex].state;
+      if (newState != oldState) {
+        notify(NFY_PROGRESS, "Setting state of '%s' from '%s' to '%s'",
+              name.toString().c_str(), ToString(oldState).c_str(), ToString(newState).c_str());
+        auto version = updater.storageManager->GetMutable(name);
+        if (version) {
+            updater.SetVersionStateAndRunHandlers(version, newState, waitingFor);
```
tst-lsavoie: Inconsistent indentation

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +278,105 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+
+          // Get the new state directly from the asset version since it may be
+          // different from the passed-in state
+          auto & data = tree[vertex];
+          data.state = version->state;
+
+          if (AssetDefs::Canceled == newState || AssetDefs::Blocked == newState || AssetDefs::Failed == newState){
```
tst-lsavoie: There's a fair amount of duplication between the two SetState functions. It would be nice if we could combine them somehow. I think this if statement is the only significant difference. I guess you could change the other SetState function to have different sets for children and inputs and then check both of them in the corresponding finish_vertex function. Then I think you could pull the common functionality into a parent class and remove the duplicate function.

Just thinking out loud. I always try to get rid of as much duplication as possible.

earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
```diff
@@ -129,7 +129,7 @@ void HandleTaskDone(const TaskDoneMsg & msg, MiscConfig::GraphOpsType graphOps)
         stateUpdater->SetSucceeded(version);
       }
       else {
-        stateUpdater->SetFailed(version);
+        stateUpdater->SetStateForRefAndDependents(msg.verref, AssetDefs::Failed, AssetDefs::Finished);
```
tst-lsavoie: Either this should still use SetFailed or we should remove SetFailed from the StateUpdater.

tst-lsavoie: Also, I think this will only set the state to Failed if the state is Finished (which means Failed, Succeeded, Canceled, Offline, or Bad). At this point the asset should be InProgress, so nothing will ever get set to failed.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +278,105 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+      const AssetVertex & data = tree[vertex];
+      notify(NFY_PROGRESS, "Calculating state for '%s'", data.name.toString().c_str());
+
+      // Set the state for assets in the dependent tree.
+      if (data.inDepTree) {
+        SetState(vertex, newState, {0, 0});
+      }
+      else {
+        // Set the state to blocked if needed, otherwise do nothing
+        auto version = updater.storageManager->Get(data.name);
```
tst-lsavoie: This code looks up the version here and inside the SetState function. It would avoid the double lookup if you just pass in the version. Of course, that destroys my previous comment about combining the two SetState functions.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +278,105 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+          SetState(vertex, AssetDefs::Blocked, {0,0});
+        }
+      }
+    }
+};
+
+/**********************************************
+* StateUpdater
+**********************************************/
 void StateUpdater::SetStateForRefAndDependents(
```
tst-lsavoie: This function name isn't quite right any more since you may or may not set the state for the dependents.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater_unittest.cpp
```diff
@@ -977,6 +989,90 @@ TEST_F(StateUpdaterTest, Cancel) {
   assertStateNotSet(sm, "ci3");
 }
 
+TEST_F(StateUpdaterTest, FailedInput) {
+  SetVersions(sm, {
+                    MockVersion("l"),
```
tst-lsavoie: I get that one's an input and one's a listener, but `l` and `i` are kind of hard to tell apart.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater_unittest.cpp
```diff
@@ -977,6 +989,90 @@ TEST_F(StateUpdaterTest, Cancel) {
...
+  GetMutableVersion(sm, "i")->state = AssetDefs::InProgress;
+
+  // MockVersion::InputStatesAffectMyState will return true, so setting this input
+  // to a Failed state should cause the listener to become Blocked
+  updater.SetStateForRefAndDependents(fix("i"), AssetDefs::Failed, [](AssetDefs::State) { return true; });
+
+  assertStateSet(sm, "l");
+  assertStateSet(sm, "i");
+  assert(AssetDefs::Blocked == GetMutableVersion(sm, "l")->state);
+  assert(AssetDefs::Failed == GetMutableVersion(sm, "i")->state );
```
tst-lsavoie: Minor: extra space before the close paren.

tst-lsavoie:
I made some comments and I still need to test, but overall this looks right.

tst-lsavoie:
This might need an entry in the release notes for improving cancel performance.

earth_enterprise/src/fusion/autoingest/AssetVersion.h
```diff
@@ -137,6 +137,8 @@ class AssetVersionImpl : public AssetVersionStorage, public StorageManaged {
     return (state == AssetDefs::Bad);
   }
 
+  virtual bool InputStatesAffectMyState(AssetDefs::State stateByInputs, bool blockedByOfflineInputs) const {assert(false); return true;}
+  virtual bool ChildStatesAffectMyState() const {assert(false); return true;}
```
tst-lsavoie: Minor: you could get rid of the `assert(false)` here and then you wouldn't have to define the function in LeafAssetVersionImplD.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +278,105 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+      }
+      else {
+        // Set the state to blocked if needed, otherwise do nothing
+        auto version = updater.storageManager->Get(data.name);
+        if (hasBlockingInputs->find(data.name) != hasBlockingInputs->end() && 
+            version->InputStatesAffectMyState(AssetDefs::Blocked, true)) {
+          SetState(vertex, AssetDefs::Blocked, {0,0});
+        }
+        else if (hasBlockingChildren->find(data.name) != hasBlockingChildren->end() && 
+            version->ChildStatesAffectMyState()) {
```
tst-lsavoie: Minor: I think technically you could get rid of the call to `version->ChildStatesAffectMyState()`. Since leaf assets have no children, then the only way for it to be in the hasBlockingChildren list is for it to be a composite asset, and then ChildStatesAffectMyState will always return true. So we get the same result if we just check hasBlockingChildren and not ChildStatesAffectMyState.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +278,105 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
-    bool includeDepDescendents = !UserActionRequired(newState);
+    bool includeAllChildrenAndInputs = !UserActionRequired(newState);
+    // For all states except Failed, we will load dependent children into the tree
+    bool includeDependentChildren = newState != AssetDefs::Failed;
     DependentStateTree tree = BuildDependentStateTree(
-        verref, updateStatePredicate, includeDepDescendents, storageManager);
-    depth_first_search(tree, visitor(SetStateVisitor(*this, tree, newState)));
+        verref, updateStatePredicate, includeDependentChildren, includeAllChildrenAndInputs, storageManager);
+      
+    if (AssetDefs::Canceled == newState || AssetDefs::Failed == newState)
```
tst-lsavoie: Minor: I feel like there should be an "IsBlocking" function you can pass a state to. There isn't one, but it might come in handy in the future.

earth_enterprise/src/fusion/autoingest/sysman/AssetOperation.cpp
tst-lsavoie: > Also, I think this will only set the state to Failed if the state is Finished (which means Failed, Succeeded, Canceled, Offline, or Bad). At this point the asset should be InProgress, so nothing will ever get set to failed.

I've now verified this through testing. I inserted code to randomly fail tasks in the resource provider, and Failed tasks are not marked appropriately. I think instead of `AssetDefs::Finished` we need to unconditionally return true. The third argument's purpose is to determine which assets in the dependency tree will be marked as Failed. Since none of them will return true for AssetDefs::Finished, none are marked as failed.

tst-sbayard: Yeah, that was my fault for not looking closely at what the predicate was being used for. I'll update that.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: Are you talking about pulling both `SetState` and `finish_vertex` into a base class, or just `SetState`?

tst-lsavoie: Just SetState. It would be great if we could get the common parts of finish_vertex into a base class as well, but that seems harder.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: Yeah, that struck me. I thought about `SetStateForRef`, but that's awfully close to `SetState`. `SetAndPropagateState` would be perfect if it wasn't already used in `AssetVersionImpl`.
Would it be too confusing to re-use that name here as well?

tst-lsavoie: I think that would work well, actually. Then the two `SetAndPropagateState` functions would do the same thing in different ways, and the one in `AssetVersionImpl` will go away eventually.

earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
tst-sbayard: Updated. I'm open to suggestions on the wording.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: There's `CanRebuild` which looks like the same thing:
```
static bool CanRebuild(State state) {
   return state & (AssetDefs::Canceled | AssetDefs::Failed);
 }
```
Could rename that or just add `IsBlocking` and have `CanRebuild` call that.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: Removed `ChildStatesAffectMyState` altogether since it ended up serving no purpose.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: I created the base class and pulled SetState into it. I opted to make it a template method and let each derived class keep its separate sets. They could have used sets in the base class, but I like that the names are more explicit and a littler clearer this way. I could be talked out of that if there's a strong feeling to the contrary.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: Nevermind, I guess `IsBlocking` would be a bit different. `Blocked` is also a blocking state.

earth_enterprise/src/fusion/autoingest/AssetVersion.h
tst-sbayard: Removed `ChildStatesAffectMyState`.

earth_enterprise/src/fusion/autoingest/storage/AssetDefs.idl
```diff
@@ -41,6 +41,9 @@ class AssetDefs {
   static bool NotFinished(State state) {
     return !Finished(state);
   }
+  static bool IsBlocking(State state) {
+    return state & (AssetDefs::Blocked | AssetDefs::Canceled | AssetDefs::Failed);
```
tst-lsavoie: `Bad` and `Offline` are technically blocking states as well.

earth_enterprise/src/fusion/autoingest/sysman/DependentStateTree.cpp
```diff
@@ -168,7 +171,7 @@ void DependentStateTreeFactory::FillInVertex(
   if (vertices[name].includeConnections) {
     // In some cases, assets in the dependent tree don't need their inputs and
```
tst-lsavoie: Minor: this comment is slightly out of date. Something like "In some cases, assets don't need all their inputs and children to calculate their states.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +291,76 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+  private:
+    using BlockedSet = std::unordered_set<SharedString>;
+
+    // This is a shared_ptr because the visitor will be copied several times
+    // during the course of the traversal.
+    const std::shared_ptr<BlockedSet> hasBlockingInputs;
+    const std::shared_ptr<BlockedSet> hasBlockingChildren;
+
+    void MarkParentsForLaterProcessing(AssetHandle<AssetVersionImpl> & version) const override {
+      if (AssetDefs::IsBlocking(newState)){
```
tst-lsavoie: Minor: hopefully you wouldn't use this class at all unless newState is Blocking. Maybe make this an assert in the constructor?

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +291,76 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+      // Set the state for assets in the dependent tree.
+      if (data.inDepTree) {
+        SetState(vertex, newState, {0, 0}, true);
+      }
+      else {
+        // Set the state to blocked if needed, otherwise do nothing
+        auto version = updater.storageManager->Get(data.name);
+        if (hasBlockingInputs->find(data.name) != hasBlockingInputs->end() && 
+            version->InputStatesAffectMyState(AssetDefs::Blocked, true)) {
+          SetState(vertex, AssetDefs::Blocked, {0,0}, true);
```
tst-lsavoie: Minor: after this point you can remove `data.name` from `hasBlockingInputs`. I don't know if that will make it faster, though - it's extra work at this point to remove the name, but it's fewer names to search through for future vertices. Same applies to the `hasBlockingChildren` statement.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +291,76 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+* StateUpdater
+**********************************************/
+void StateUpdater::SetAndPropagateState(
     const SharedString & ref,
     AssetDefs::State newState,
     function<bool(AssetDefs::State)> updateStatePredicate) {
   try {
     SharedString verref = AssetVersionImpl::Key(ref);
-    bool includeDepDescendents = !UserActionRequired(newState);
+    bool includeAllChildrenAndInputs = !UserActionRequired(newState);
```
tst-lsavoie: There's a small bug here: if `newState` is `Failed`, we want `includeAllChildrenAndInputs` to be false. Basically, any time we use a blocking state visitor we want includeAllChildrenAndInputs to be false. The way it is now isn't wrong, but it does unnecessary work.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: I actually had that initially, but I took it out, since `erase` would have to iterate through the list to find it. I wasn't sure if that would be worth the trade-off. But it looks like maybe I can `erase` using the iterator returned from `find` to do it much faster.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: Good catch. I'll update that.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-lsavoie: Yeah, I agree. I don't know if it will make things faster or not. It's up to you if you want to implement it.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater_unittest.cpp
```diff
@@ -977,6 +985,68 @@ TEST_F(StateUpdaterTest, Cancel) {
...
+                  });
+  SetListenerInput(sm, "a", "b");
+
+  GetMutableVersion(sm, "a")->state = AssetDefs::Waiting;
+  GetMutableVersion(sm, "b")->state = AssetDefs::InProgress;
+
+  GetMutableVersion(sm, "a")->inputStatesAffectMyState = false;
+
+  // MockVersion::InputStatesAffectMyState will return true, so setting this input
+  // to a Failed state should cause the listener to become Blocked
```
tst-jkent: Since we have set **inputStatesAffectMyState = false**, should this comment not state that the listener should not become Blocked?

tst-sbayard: Good catch. Copy/paste strikes again ;)

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: Both `includeAllChildrenAndInputs` and `includeDependentChildren` were going to have to check for `newState` being `Failed`, so I just decided to make it a separate if statement. Not sure which way is cleaner.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: 1. Since the documentation for `erase` says that the version that takes a position is constant time, I figured it was probably a little faster to do that.
2. It's now always checking to see if `data.name` is in `hasBlockingChildren` and `hasBlockingInputs`. The rationale is that, since I'm removing `data.name` from the sets, I should check both since it could theoretically be in both (I've no idea if that's possible in actuality). But beyond that, I think it makes the logic a bit cleaner, and I suspect it makes the unit test coverage a little better since there are fewer "hidden" branches. Also, the impact of sometimes calling `SetState` a second time should be minimal since it will return right away if the state isn't changing. It's probably a bit of a wash on the whole, performance-wise. But this seems good to me. 
As always, open to discussion.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
```diff
@@ -278,16 +291,86 @@ class StateUpdater::SetStateVisitor : public default_dfs_visitor {
...
+        SetState(vertex, newState, {0, 0}, true);
+      }
+      else {
+        // Set the state to blocked if needed, otherwise do nothing
+        auto version = updater.storageManager->Get(data.name);
+        auto inputIt = hasBlockingInputs->find(data.name);
+        if (inputIt != hasBlockingInputs->end() && 
+            version->InputStatesAffectMyState(AssetDefs::Blocked, true)) {
+          SetState(vertex, AssetDefs::Blocked, {0,0}, true);
+          hasBlockingInputs->erase(inputIt);
```
tst-lsavoie: Minor: if the asset is in `hasBlockingInputs` but `InputStatesAffectMyState` returns false then this line will never be called, so some stuff may not get removed from the `hasBlockingInputs`. Not a huge deal, but not hard to fix either.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater_unittest.cpp
```diff
@@ -977,6 +985,68 @@ TEST_F(StateUpdaterTest, Cancel) {
...
+                  });
+  SetListenerInput(sm, "a", "b");
+
+  GetMutableVersion(sm, "a")->state = AssetDefs::Waiting;
+  GetMutableVersion(sm, "b")->state = AssetDefs::InProgress;
+
+  GetMutableVersion(sm, "a")->inputStatesAffectMyState = false;
+
+  // MockVersion::InputStatesAffectMyState will return false, so the listener's
+  // state should be uanffected.
```
tst-lsavoie: Minor: "uanffected" is spelled wrong.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater.cpp
tst-sbayard: Good catch. I completely missed that.

earth_enterprise/src/fusion/autoingest/sysman/StateUpdater_unittest.cpp
tst-sbayard: I was using the French spelling, but I'll change it if you insist on using English ;)

PR #1605, opened 2020-01-07
1604: Fix WMS to work with reverse proxies

### LINKED ISSUE #1604: WMS with reverse proxies doesn't work
When using a reverse proxy for WMS requests, the GetCapabilitiesRequest returns OnlineResource with a link to the backend server, not the reverse proxy. However the link should be to the reverse proxy since that is where the client connects.

Set up a reverse proxy to a GEE Server with a WMS-served map:

1. Install Apache 2.4 on a separate machine from GEE (non-bundled with GEE since the GEE-specific version doesn't include all the reverse proxy capabilities)
2. Modify httpd.conf to set it up as a proxy:
```
LoadModule proxy_module modules/mod_proxy.so
LoadModule proxy_fcgi_module modules/mod_proxy_fcgi.so
LoadModule proxy_scgi_module modules/mod_proxy_scgi.so
LoadModule proxy_fdpass_module modules/mod_proxy_fdpass.so
LoadModule proxy_wstunnel_module modules/mod_proxy_wstunnel.so
LoadModule proxy_ajp_module modules/mod_proxy_ajp.so
LoadModule proxy_express_module modules/mod_proxy_express.so
LoadModule proxy_http_module modules/mod_proxy_http.so
LoadModule ssl_module modules/mod_ssl.so

# To proxy on the http port use this, otherwise modify ssl.conf for https
ProxyPass "/" "youractualgeeserver.net"
ProxyPassReverse"/" "youractualgeeserver.net" # Probably not needed
```
3. If testing https, modify ssl.conf
```
SSLEngine on

SSLProxyEngine on
# For my simple testing I disabled some checks
SSLProxyVerify none
SSLProxyCheckPeerCN off
SSLProxyCheckPeerName off

ProxyPass "/" "https://youractualgeeserver.net:443/"
ProxyPassReverse "/" "https://youractualgeeserver.net:443/"

# I also created my own certificate file and updated SSLCertificateFile and SSLCertificateKeyFile, but that probably wasn't needed since it came with some defaults already
```
4. Restart the proxy server: `sudo systemctl restart httpd`
5. Do a GetCapabilitiesRequest to the proxy server, and see in the response that OnlineResource points to the backend GEE server, not the proxy server.

tst-jzmolek:
This will allow WMS to work with reverse proxies by returning the correct link to the proxy server instead of the backend GEE server.

I have tested this on CentOS7 with ATAK and QGIS clients.

Note that this won't handle cases where the protocol (http vs https) changes between the proxy server and GEE server. It is a valid case, but not as important right now. It might be a little more work to set that up too since that info wasn't seen in the header from the proxy server.

To test:
1. Set up a reverse proxy to a GEE Server with a WMS-served map
2. Do a GetCapabilitiesRequest to the proxy server, and see in the response that OnlineResource points to the proxy server.
3. Viewing a WMS source should work wether using a reverse proxy or not.

Fixes #1604

### DIFF
--- a/docs/geedocs/5.3.4/answer/releaseNotes/relNotesGEE5_3_4.html
+++ b/docs/geedocs/5.3.4/answer/releaseNotes/relNotesGEE5_3_4.html
@@ -108,10 +108,6 @@ correctly to the new PostgreSQL version used by Open GEE 5.3.4.</p>
 <td>Issue Description</td>
 <td>Issue Resolution</td>
 </tr>
-<tr class="row-odd"><td>1604</td>
-<td>WMS capability doesn't work with reverse proxies</td>
-<td>It now works, as long as the protocol (http vs https) matches</td>
-</tr>
 </tbody>
 </table>
 <p class="rubric">Known Issues</p>
--- a/docs/geedocs/5.3.4/answer/releaseNotes/relNotesGEE5_3_4.html
+++ b/docs/geedocs/5.3.4/answer/releaseNotes/relNotesGEE5_3_4.html
@@ -108,6 +108,10 @@ correctly to the new PostgreSQL version used by Open GEE 5.3.4.</p>
 <td>Issue Description</td>
 <td>Issue Resolution</td>
 </tr>
+<tr class="row-odd"><td>1604</td>
+<td>WMS capability doesn't work with reverse proxies</td>
+<td>It now works, as long as the protocol (http vs https) matches</td>
+</tr>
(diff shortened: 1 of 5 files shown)


tst-blaird (approved):
The changes look good to me.  I only code reviewed and did not test the changes.

tst-dsugar (commented):
This should also be in the release notes for master/5.3.4.

tst-dsugar (approved):

PR #1607, opened 2020-01-08
Created make.bat for sphinx html creation for windows users

tst-hrajamoney:
To test in Windows:

1) Open a cmd prompt 
2) add autocrlf = input in earthenterprise/.git/config as shown below   

[core]
    repositoryformatversion = 0
    filemode = false
    bare = false
    logallrefupdates = true
    symlinks = false
    ignorecase = true
    autocrlf = input

3) cd docsrc folder 
    make html 

4) Do a git status and no unwanted files are modified 

5) do a dummy rst change and make html and verify changes 
    undo the dummy rst change and make html - to reverse the prev step

Note: Verify in linux that make html still works  good.


tst-swanson:
Got some errors:
image

Issue #1608, opened 2020-01-09
Cutter fails when polygon level is 24


tst-mbiancalana:
**Description**

When cutting using a polygon, if the Polygon Level is set to 24, the cut fails.

**To Reproduce**

1. Publish a map
1. Open cutter
1. Select a region to cut using a polygon
1. Add a name and descriptions
1. Set World Level to 4 and Region Level to 10 (these do not matter, but I've chosen these for a hopefully quick cut)
1. Expand the Advanced options and set Polygon Level to 24
1. Proceed to create the cut

Cutter will fail with an error similar to this in the log on the right:
```
Description: asdfsadfs
Added globe directory: /opt/google/gehttpd/htdocs/cutter/globes/.globe_builder/asdf_5481_1578608620.891026/asdf Ok

Saved polygon to /opt/google/gehttpd/htdocs/cutter/globes/.globe_builder/asdf_5481_1578608620.891026/asdf/earth/polygon.kml
Convert polygon to quadtree nodes ... Executing: /opt/google/bin/gepolygontoqtnodes --qt_nodes_file="/opt/google/gehttpd/htdocs/cutter/globes/.globe_builder/asdf_5481_1578608620.891026/qt_nodes.txt" --kml_polygon_file="/opt/google/gehttpd/htdocs/cutter/globes/.globe_builder/asdf_5481_1578608620.891026/asdf/earth/polygon.kml" --max_level=24 --mercator Raising error: Fusion Notice: Added polygon of 5 vertices. Fusion Notice: Found 1 polygons. Fusion Notice: Added polygon of 5 vertices. Fusion Notice: Found 1 polygons. Fusion Fatal: vector::_M_default_append (return code 255) FAILED:
FAILED:
FAILED
```

**Expected behavior**

Cutter succeeds and a glm file is produced.

**Desktop**

 - Server OS: Any
 - Client OS: Windows 10
 - Browser Any
 - Version 5.3.2

**Additional context**

As the logs point out the failure occurs inside `gepolygontoqtnodes`. Specifically, the error happens here: https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/fusion/portableglobe/polygontoqtnodes.cpp#L245

Because the std::vector `level_dim_sizes_` has only 24 elements (making the last _index_ 23), but `max_level` in this case is 24.

[...]

PR #1610, opened 2020-01-10
I1608: Expand level_dim_sizes_ to handle level 24

tst-mbiancalana:
Addresses #1608 

To verify:
Build and publish a map or globe
Use cutter to create a polygon-based cut with Polygon Level (under Advanced) set to 24.

Note: If your polygon is not very small, then the cut will take a very long time. Zoom all the way in before making the polygon (or use a very small polygon from a KML file).

I performed my testing for this on CentOS 6.10.


tst-blaird (approved):
The change looks good to me.  I tested it on CentOS 7.

tst-ccamp (approved):
Code looks good to me

PR #1617, opened 2020-01-16
Do not purge the cache unnecessarily when there are large numbers of dirty assets

tst-jkent:
Added an option to config for setting the maximum allowed percentage difference between the size of the cache and size of the dirty map.
Implemented check for whether or not to prune the cache when Add or Remove is called determined by whether or not the percentage difference of cache and dirty map size exceeds the configured limit.


earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -355,5 +368,30 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
...
+bool StorageManager<AssetType>::DetermineIfPrune() {
+  std::lock_guard<std::recursive_mutex> lock(storageMutex);
+  uint32 cacheSize = CacheSize();
+  uint32 dirtySize = DirtySize();
+  uint32 oriNum;
+  uint32 newNum;
+  uint32 diff;
+  float percent;
+
+  if (cacheSize > dirtySize) {
```
tst-lsavoie: `cacheSize` will always be >= `dirtySize` - if it's in the dirty map, it's also in the cache.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -355,5 +368,30 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
...
+    oriNum = cacheSize;
+    newNum = dirtySize;
+  }
+  else {
+    oriNum = dirtySize;
+    newNum = cacheSize;
+  }
+  if (oriNum == newNum || oriNum == 0 || newNum == 0) { return false; }
```
tst-lsavoie: I would return `true` if either of the sizes is 0. If the cache size is 0, it's small enough that it doesn't matter because it won't take long to prune anyway. If the dirty map size is zero, there's no downside to pruning the cache because the dirty map isn't forcing a bunch of stuff to stay in the cache. However, if the two are equal we don't want to prune because it will be wasted effort. However, I think that case will be caught by your calculations below.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -355,5 +368,30 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
...
+  diff = oriNum - newNum;
+  percent = ((float)diff / (float)oriNum) * 100;
+  notify(NFY_WARN, "Percent Diff: %f", percent);
+  return (percent > maxSizeDiff);
```
tst-lsavoie: I think this code is correct but it's confusing to me to write it this way. I think you can replace most of the body of this function with `return (dirtySize < cacheSize * maxSizeDiff);` Then you only prune if the dirty map isn't taking up a large portion of the cache. You could equivalently use `return (dirtySize < cacheSize * (1 - maxSizeDiff));` if you prefer to specify the sizes differently.

Also, I'd recommend changing `maxSizeDiff` to `maxPurgePercent` or something a little more descriptive.

earth_enterprise/src/common/khCache.h
```diff
@@ -217,6 +218,7 @@ class khCache {
     limitCacheMemory = enabled;
     maxCacheMemory = maxMemory;
   }
+  void setPruneConditionVal(bool val) { pruneConditionMet = val; }
```
tst-lsavoie: You shouldn't need a new function and variable to implement this. The `Add` and `Remove` function already have an option that lets you decide whether to prune.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -164,6 +169,7 @@ template<class AssetType>
 inline void
 StorageManager<AssetType>::AddNew(const AssetKey & key, const PointerType & value) {
   std::lock_guard<std::recursive_mutex> lock(storageMutex);
+  cache.setPruneConditionVal(DetermineIfPrune());
```
tst-lsavoie: Related to my above comment, you can remove this call and just change the below line to `cache.Add(key, value, DetermineIfPrune());`

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -229,6 +237,7 @@ StorageManager<AssetType>::Get(
       // The file has changed on disk.
 
       // Drop the current entry from the cache.
+      cache.setPruneConditionVal(DetermineIfPrune());
```
tst-lsavoie: In this case we don't want to call `DetermineIfPrune()` at all. We're already passing `false` for the `prune` option - we can just leave that. What happens here is that we're removing something from the cache and then adding something else a little lower. There's no point in pruning twice, so we skip pruning for the remove and only prune during the add.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -355,5 +360,22 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
...
+  //uint32 diff;
+  //float percent;
+
+  //if (cacheSize == 0) { return true; }
+  //diff = cacheSize - dirtySize;
+  //percent = ((float)diff / (float)cacheSize) * 100;
+  //notify(NFY_WARN, "Percent Diff: %f", percent);
+  //return (percent > maxPurgePercent);
+  notify(NFY_WARN, "CacheSize: %f, DirtySize: %f, Percent: %f", cacheSize, dirtySize, (cacheSize * maxPurgePercent));
+  return (dirtySize < (cacheSize * maxPurgePercent));
```
tst-lsavoie: Just so this is clear, this changes the meaning of maxPurgePercent compared to your old code. If maxPurgePercent was 0.75, the old code would stop purging when the dirty map got to be 25% of the cache (if I'm reading the code correctly). With the updated code, it will stop purging when the dirty map gets to be 75% of the cache. Either is fine; we just have to make sure we document it correctly when we get to that point.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -355,5 +360,11 @@ bool StorageManager<AssetType>::SaveDirtyToDotNew(
   return true;
 }
 
+template<class AssetType>
+bool StorageManager<AssetType>::DetermineIfPrune() {
+  std::lock_guard<std::recursive_mutex> lock(storageMutex);
+  return ((float)DirtySize() < ((float)CacheSize() * (1 - maxPurgePercent)));
```
tst-lsavoie: A few small comments:

I'd recommend changing this to `<=`. That way, if we set maxPurgePercent to 0, this function will always return true (which allows us to get the behavior we get without these changes).

I'm not sure that casting to `float` is necessary. Sure, integer math won't be as exact, but it's probably close enough for this purpose.

The lock is not strictly necessary in this function. Everywhere it is called the code will have already acquired the lock.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -92,6 +95,7 @@ class StorageManager : public StorageManagerInterface<AssetType> {
     inline void AddNew(const AssetKey &, const PointerType &);
     inline void AddExisting(const AssetKey &, const PointerType &);
     inline void NoLongerNeeded(const AssetKey &, bool = true);
+    bool DetermineIfPrune();
```
tst-lsavoie: I would suggest making this a private method.

earth_enterprise/src/fusion/autoingest/MiscConfigStorage.idl
```diff
@@ -34,6 +34,9 @@ class MiscConfigStorage {
   uint64       MaxVersionCacheMemorySize = uint64(2415919104);
 
   bool         LimitMemoryUtilization = false;
+
+  // The maximum percentage difference between cache size and dirty map size.
+  float         MaxPurgePercentage = float(.75);
```
tst-lsavoie: I think we should set this to a number that makes the behavior identical to previous behavior by default - so 0. That's up for discussion, though.

tst-lsavoie: Also, since the meaning of this number has changed we may want to think of a better name. MaxPurgePercent made more sense before you added the `1 - ` in front of it. Now, this number means that at least this percentage of the cache has to not be dirty for us to prune. So maybe `NonDirtyPercentForPrune`? `MinNonDirtyFractionForPrune`? If you want to go crazy you could make it a whole sentence: `MinFractionOfNonDirtyCacheItemsForPruning`. I don't really like any of those options but I think the code will be clearer if the variable name indicates that this is the minimum fraction of the cache that has to be non-dirty for us to prune. You could go with something like `PruneFraction` or `PruneThreshold`, which aren't awful but aren't as descriptive because they don't tell you what it's a threshold of. It also might be easier to name if you reverse the meaning again by taking out the `1 - `, and then it becomes the maxiumum fraction of the cache that can be dirty before we stop pruning, so maybe `MaxCacheDirtyPercentForPruning`?

tst-lsavoie:
Just a reminder that we will need release notes and documentation updates eventually.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-jkent: In a previous comment you mentioned that we do not want to prune the cache if the dirty map and cache size are the same. Changing this condition to <= would cause the prune condition to be met if maxPurgePercent is also set to 0.

tst-lsavoie: The reasoning behind this suggestion is that we're adding a feature that may make Fusion faster, but it may not. So, I think there should be a way to turn it off completely. Users will have the option to set maxPurgePercent to 0 and get the old behavior, or they have the option to set it to a different number and possibly get better performance (or possibly get worse performance if it doesn't work as well as I think it will). So, overall, I think it will work fastest if we don't purge when the dirty map and the cache are the same size, but I want to give users that option in case I'm wrong.

tst-jkent: I'm in the process of changing this to **CacheSize() * maxPurgePercent**. With that in mind, **PurgePercent** now has the opposite affect. If we keep the condition `<` as opposed to `<=`and set the **PurgePercent** to 1, the cache will be pruned as long as the sizes are not the same.

tst-jkent: I hadn't seen your comment, I am now changing the condition to `<=`.

earth_enterprise/src/fusion/autoingest/StorageManager.h
```diff
@@ -173,14 +179,14 @@ template<class AssetType>
...
-  cache.Add(key, value);
+  cache.Add(key, value, DetermineIfPrune());
 }
 
 template<class AssetType>
 inline void
 StorageManager<AssetType>::NoLongerNeeded(const AssetKey & key, bool prune) {
   std::lock_guard<std::recursive_mutex> lock(storageMutex);
-  cache.Remove(key, prune);
+  cache.Remove(key, (DetermineIfPrune() && prune));
```
tst-lsavoie: It looks like the `prune` parameter is never used in the `NoLongerNeeded` function. I'd suggest removing it and just passing `DetermineIfPrune()` to `cache.Remove`.

tst-lsavoie:
We need release notes updates and documentation changes for this update. You can put that on a separate PR if you like.

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-jkent: This is needed for various unit tests in cases where we specifically do not want to prune the cache. If false is passed to NoLongerNeeded, then the cache will not be pruned despite the result of DetermineIfPrune.

earth_enterprise/src/fusion/autoingest/MiscConfigStorage.idl
```diff
@@ -34,6 +34,9 @@ class MiscConfigStorage {
   uint64       MaxVersionCacheMemorySize = uint64(2415919104);
 
   bool         LimitMemoryUtilization = false;
+
+  // The percentage of cache size that the dirty map size must meet for cache to not be pruned.
```
tst-lsavoie: This comment is worded a bit awkwardly. I'd suggest something like "The cache will not be pruned if the dirty map is more than this fraction of the cache."

earth_enterprise/src/fusion/autoingest/StorageManager_unittest.cpp
```diff
@@ -580,6 +580,36 @@ TEST_F(StorageManagerTest, InvalidRef) {
...
+  storageManager.SetPrunePercent(.5);
+
+  PointerType newItem = make_shared<TestItem>();
+  ASSERT_EQ(storageManager.CacheSize(), 0) << "Storage manager has unexpected item in cache";
+  ASSERT_EQ(storageManager.DirtySize(), 0) << "Storage manager has unexpected item in dirty map";
+
+  storageManager.AddNew("newItem", newItem);
+  ASSERT_EQ(storageManager.CacheSize(), 1) << "Storage manager has wrong number of items in cache";
+  ASSERT_EQ(storageManager.DirtySize(), 1) << "Storage manager has wrong number of items in dirty map";
+  ASSERT_EQ(storageManager.DetermineIfPrune(), false) << "Storage manager has determined wrong choice for pruning cache";
```
tst-lsavoie: Minor: there is an `ASSERT_FALSE` (and an `ASSERT_TRUE` for the other `DetermineIfPrune` check).

earth_enterprise/src/fusion/autoingest/StorageManager.h
tst-lsavoie: If it's needed for the unit tests I'm fine with leaving it in.

tst-jkent: There are various places in StorageManager.h where it is specifically stated that the cache should not be pruned after a remove, but to wait until after an item is added. With this in mind, I think we should leave this argument as is.

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -344,12 +344,20 @@ Configure Fusion performance
...
-      the number of objects in the cache. It should be noted that as of
-      now the reported memory usage is very low and that this option is
-      not recommended for production environments.
+      ``MaxAssetCacheMemorySize`` and ``MaxVersionCacheMemorySize`` and
+      is in terms of bytes. While this option is enabled, the caches will
+      not be limited by the number of objects in the cache.
+
+      .. _prune-percent:
+
+      Another experimental option in ``misc.xml`` allows the user to limit
```
tst-lsavoie: Might be a good idea to note that this can improve performance.

docs/geedocs/docsrc/answer/releaseNotes/relNotesGEE5_3_4.rst
```diff
@@ -17,6 +17,10 @@ Release notes: Open GEE 5.3.4
       tasks has been modified to use less memory. This change is experimental
       and can be enabled using the :ref:`GraphOperations flag <graph-operations>`.
 
+      **Option to limit cache pruning**.
+      Users can now limit how often the cache is pruned by use of an experimental
```
tst-lsavoie: Maybe specify which cache ("...the asset and asset version caches are pruned...").

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -344,12 +344,20 @@ Configure Fusion performance
...
+      how often the asset and asset version caches are pruned, which can
+      improve performance. This option is controlled by ``PrunePercent``
+      and is set to 100 by default. Assets that have been moddified can not
```
tst-lsavoie: Minor: it just occurred to me that it might be good to say that 100 causes the old behavior of pruning at every opportunity.

Also, there's a spelling error in "moddified".

docs/geedocs/docsrc/answer/fusionAdministration/confTaskRulesForFusionPerf.rst
```diff
@@ -344,12 +344,23 @@ Configure Fusion performance
...
+      .. _prune-percent:
+
+      Another experimental option in ``misc.xml`` allows the user to limit
+      how often the asset and asset version caches are pruned, which can
+      improve performance. Assets that have been modified can not be 
+      removed from the cache until they are written to disk. To limit
+      pruning, the amount of unwritten modified assets is compared to a 
+      percentage of the current size of each cache. This percentage is
+      controlled by ``PrunePercent`` and is set to 100 by default, which
+      retains current behavior. The value for ``PrunePercent`` can be any
```
tst-lsavoie: Minor: "retains current behavior" reads a bit funny to me since the current behavior is to do whatever `PrunePercent` says. You could say "retains previous behavior", but I think it would be better to be more explicit and say something like "prunes the caches as often as possible."

I don't think it's a huge deal, though, so I'm going to go ahead and approve.

PR #1618, opened 2020-01-16
Fix intermittent failure in integration tests

tst-lsavoie:
Fixes an intermittent failure in the integration tests where an asset finishes building before the test can check that it's in progress.


tst-jkent:
I ran the integration tests and they all passed.

PR #1620, opened 2020-01-17
(WIP) Add public function for finding all files in specific globe directory

tst-nrichman:
Adds a function that, give a specific directory, can return all file location information and names of files in that directory.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/glc_unpacker.cpp
+++ b/earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/glc_unpacker.cpp
@@ -449,6 +449,30 @@ bool GlcUnpacker::FindFile(const char* file_name,
 }
 
 /**
+ * Find each file in a provided directory on a layer and retrieve their names,
+ * offsets, and sizes.
+ */
+bool GlcUnpacker::FindLayerDirectoryFiles(const char* directory_name,
+                                     int layer,
+                                     std::map<const char* , PackageFileLoc*>* file_locs) {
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/glc_unpacker.cpp
```diff
@@ -448,6 +448,30 @@ bool GlcUnpacker::FindFile(const char* file_name,
...
+    FileUnpacker* unpacker = GetUnpacker(layer);
+
+    for (int i = 0; i < unpacker->IndexSize(); ++i) {
+      const char *package_file = unpacker->IndexFile(i);
+
+      PackageFileLoc file_loc;
+
+      if (FindFile(package_file, &file_loc)) {
+        if (!strncmp(directory_name, package_file, strlen(directory_name))) {
+          (*file_locs).emplace(package_file, &file_loc);
```
tst-rwildes: you are passing the caller a pointer to a stack allocated variable that will likely be invalid after this call (or even loop iteration).  I might think a walker would be more flexible.  Then the caller can choose to either use the data right way (avoiding the copy) or save the data for use later (then likely needing to do a copy) or a mix there of.

tst-rwildes: See `MapDataPacketWalker` in this same class.  Maybe have a file walker where you can walk the embedded files of a portable file and do things with them as they are walked.

PR #1624, opened 2020-01-20
someone did not generate doc change

tst-rwildes:


tst-ccamp:
Since this WMS change has now gone into 5.3.3, I don't think this necessary any longer. Plus, we should remove the reference to 1604

tst-dsugar:
Indeed, this ended up delivered in 5.3.3.  It should not be in the 5.3.4 release notes.

Issue #1622, opened 2020-01-19
Make public unpacker function for accessing globe directories


tst-nrichman:
The FindFile function provides public access to find specific files within a globe. However, there is currently no function to do the same for directories. Functionality should be added allowing developers to access a listing of the files, allowing more flexibility in how they're accessed.

tst-blaird:
I don't think this needs to be added to release notes.

PR #1626, opened 2020-01-21
Add file walker for retrieving globe file list

tst-nrichman:
#1622 

Adds a walker that allows callers to retrieve file listings on each layer. A string callback is used to pass the name of each file.


earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/callbacks.h
```diff
@@ -8,4 +8,6 @@ class IndexItem;
 
 using map_packet_walker = std::function<bool(int layer, const IndexItem&)>;
 
+using map_file_walker = std::function<bool(const std::string&)>;
```
tst-rwildes: Add `int layer` to the callback so those walking all layers know what layer the file.  It is redundant for those calling the single layer walker.  If that bothers you you could have two walker callback types one for each of the file walking methods.  Or if it doesn't bother you I think it is fine to be redundant and use the same walker callback for both.

tst-nrichman: That doesn't bother me, seems right that you'd want to easily know what layer something is on and the redundancy created is pretty minor and only for one use case.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/glc_unpacker.cpp
```diff
@@ -333,6 +333,31 @@ void GlcUnpacker::MapDataPacketWalker(int layer, const map_packet_walker& walker
   iter->second->MapDataPacketWalker(layer, walker);
 }
 
+/**
+ * Find the names of all files in each layer of a globe.
+ */
+void GlcUnpacker::MapFileWalker(const map_file_walker& walker)
+{
+    for( auto& index_item : unpacker_index_) {
+        MapFileWalker(index_item.first, walker);
```
tst-rwildes: you need a way to break out of this if the call returned false in the inner loop.  The packet walker handles this by having the layer version of the walker implemented in `FileUnpacker` and that method returns a boolean which signals if the callback indicated a stop in the inner loop.

tst-rwildes:
Over all just had minor feedback.  You should add some unit tests to this.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/filewalker_unittest.cpp
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+    * Test that the contents of the test GLB have expected size and values
+    */
+  void TestGlbContents() {
+    PortableGlcReader reader(glb_file.c_str());
+    GlcUnpacker unpacker(reader, false, false);
+    std::list<std::string> files;
+    unpacker.MapFileWalker(0, [&] (int layer, const std::string& file_name) {
```
tst-rwildes: The second parameter should be `const char *` .  I'm surprised this even compiled but it did.  I think it must be a bug in the compiler because it should of failed.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/filewalker_unittest.cpp
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+    * Test that the contents of the test GLC layers have expected size and values
+    */
+  void TestGlcContents() {
+    PortableGlcReader reader(glc_file.c_str());
+    GlcUnpacker unpacker(reader, true, false);
+    std::map<int, std::list<std::string>> file_map;
+    unpacker.MapFileWalker([&] (int layer, const std::string& file_name) {
```
tst-rwildes: Same here. `file_name` should be type `const char*`

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/filewalker_unittest.cpp
```diff
@@ -0,0 +1,144 @@
...
+    EXPECT_TRUE(FileExists(glb_file.c_str()));
+
+    if (!FileExists(glb_file.c_str())) {
+      return;
+    }
+
+    PortableGlcReader reader(glb_file.c_str());
+    GlcUnpacker unpacker(reader, false, false);
+    std::list<std::string> files;
+    unpacker.MapFileWalker(0, [&] (int layer, const char* file_name) {
```
tst-rwildes: Should add a check to ensure the return values on all these `MapFileWalker` calls are returning what is expected.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/filewalker_unittest.cpp
```diff
@@ -0,0 +1,144 @@
...
+    EXPECT_EQ(files.size(), 59);
+
+    EXPECT_FALSE(Contains(files, std::string("data")));
+
+    EXPECT_TRUE(Contains(files, std::string("data/index")));
+
+    EXPECT_TRUE(Contains(files, std::string("icons/773_l.png")));
+
+    files.clear();
+    unpacker.MapFileWalker(0, [&] (int layer, const char* file_name) {
```
tst-rwildes: Can this be its own test?  Just to break things up a bit.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/filewalker_unittest.cpp
```diff
@@ -0,0 +1,144 @@
...
+    EXPECT_TRUE(Contains(file_map[2], std::string("icons/773_l.png")));
+
+    EXPECT_EQ(file_map[1].size(), 59);
+
+    EXPECT_EQ(file_map[2].size(), 52);
+  }
+};
+
+// Tests GLB and GLC directory contents
+TEST_F(FileWalkerTest, TestFiles) {
```
tst-rwildes: Can we have three test and move these calls into there own test.  It just makes things more granular when things fail we have a more isolated place to look.  I was thinking three tests: `TestFilesGlobe`, `TestFilesComposite`, `TestFilesCompositeEarlyExit`.  Or something like that.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/filewalker_unittest.cpp
```diff
@@ -0,0 +1,152 @@
...
+  {
+	  auto it = std::find(element_list.begin(), element_list.end(), element);
+	  return it != element_list.end();
+  }
+
+  // Helper function to check if a file exists
+  bool FileExists(const char *filename)
+  {
+    std::ifstream ifile(filename);
+    return (bool)ifile;
```
tst-blaird: I had to hunt down how this works.  Found it at http://www.cplusplus.com/reference/ios/ios/operator_bool/.  👍

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/filewalker_unittest.cpp
```diff
@@ -0,0 +1,152 @@
+// Copyright 2020 Google Inc.
```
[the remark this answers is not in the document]
tst-blaird: That copyright line should probably also be added to the glc_unpacker.* files, below the Google copyright.

PR #1628, opened 2020-01-23
Retry failed tasks

### LINKED ISSUE #1629: Retry failed tasks
Fusion should have the option to retry failed tasks so that it can recover from transient errors.

tst-lsavoie:
Adds code to automatically retry failed tasks to make it easier to recover from transient failures.

Fixes #1629


earth_enterprise/src/fusion/autoingest/sysman/khResourceProvider.cpp
```diff
@@ -645,12 +648,34 @@ khResourceProvider::JobLoop(StartJobMsg start)
...
+            sleep(sleepBetweenTriesSec);
+          }
+          // Once we have the lock again, check if someone deleted the job while
+          // we were sleeping
+          job = FindJobById(jobid);
+          if (!Valid(job)) return;
+        }
+      }
+      success = RunCmd(job, jobid, start.commands[cmdnum], cmdtime, endtime, progressSent);
+      if (!Valid(job)) return;  // check if somebody already asked for me to go away
```
tst-sbayard: What would cause `job` to be invalid right after `RunCmd`? Being canceled by a user?

tst-lsavoie: I think a user canceling it would be an example of how this could happen. Based on the old code, if we release the lock we first have to check if the job is still valid after re-acquiring the lock. This particular instance is actually a continuation of the check on line 729 - we release the lock to wait for the command to run, then on line 729 we check to see if the job is still valid after getting the lock again. If it's not, we return to this point in the code, so we check again if it's valid to distinguish between cases where the job completed unsuccessfully and the job was canceled.

docs/geedocs/docsrc/answer/releaseNotes/relNotesGEE5_3_4.rst
```diff
@@ -17,6 +17,11 @@ Release notes: Open GEE 5.3.4
       tasks has been modified to use less memory. This change is experimental
       and can be enabled using the :ref:`GraphOperations flag <graph-operations>`.
 
+      **Option to retry failed tasks**.
+      Users can now configure Fusion to
+      :ref:`retry tasks that fail <retry-failed-tasks>`, which allows
+      Fusion to deal with intermittent failures gracefully.
```
tst-sbayard: Very minor: "deal with intermittent failures gracefully" almost sounds like "fail gracefully" to me. I think something that "which allows Fusion to recover from intermittent failures" may give a more clear description.
But if this is the only comment on the PR, I wouldn't mess with it.

tst-sbayard: I should note clearly, other than this, everything looks great. And again, I wouldn't hold up the PR over this. So I still approve it.
Reviewed only, not tested.

Issue #1598, opened 2020-01-03
Incorrect RedHat 7 Install Instructions


tst-jlarocco:
**Describe the bug**

The instructions for building RPMs and installing on RHEL7 are missing calls to yum-config-manager that are required to install the perl-WWW-curl dependency.

Specifically, [this wiki page](https://github.com/google/earthenterprise/wiki/Install-Fusion-or-Earth-Server), [BUILD_RPMS.md](https://github.com/google/earthenterprise/blob/master/earth_enterprise/BUILD_RPMS.md), and [BUILD_RHEL_CentOS.md](https://github.com/google/earthenterprise/blob/master/earth_enterprise/BUILD_RHEL_CentOS.md) are missing these commands in the sections detailing RHEL7 setup:

```
sudo yum-config-manager --enable rhel-7-server-rhui-optional-rpms
sudo yum-config-manager --enable rhel-7-server-rhui-extras-rpms
```

**To Reproduce**
Follow the instructions to build RPMs on a RHEL7 machine.
Attempt to install the RPMs using yum:

```
sudo yum install opengee-*.rpm
```

See that yum fails with an error about the missing perl-WWW-Curl dependency.

**Expected behavior**

OpenGEE installs correctly.

**Desktop**

 - Server OS: RedHat Enterprise Linux 7
 - Version 5.3.3

tst-lsavoie:
Closed via PR #1630

PR #1630, opened 2020-01-24
1598 update rhel instructions

tst-jlarocco:
#1598 

Some updates to the RedHat and Centos build instructions.

Some of the packages and software repositories being used have changed recently.


earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -14,13 +14,14 @@ sudo yum install -y epel-release
 ### RHEL 7
 
 ```bash
+sudo subscription-manager repos --enable=rhel-7-server-optional-rpms
```
tst-lsavoie: I think this command will work on regular RHEL 7 instances, but it throws an error if you run it on AWS instances. For AWS this worked for me: `sudo yum-config-manager --enable rhel-7-server-rhui-optional-rpms`. We'll need to include both.

earth_enterprise/BUILD.md
```diff
@@ -31,7 +31,7 @@ platforms on how to setup the dependencies, tools, and compilers.
         more error prone):
 
         ```bash
-        git clone git@github.com:google/earthenterprise.git
+        git clone https://github.com/google/earthenterprise.git
```
tst-lsavoie: We should change the git command a few lines down to use https as well to be consistent.

earth_enterprise/BUILD_RHEL_CentOS.md
tst-jlarocco: I had to register with RedHat, using "sudo subscription-manager register", before I could install anything on RHEL 6 and 7 in AWS.  What error are you getting?

I went with subscription-manager because yum-config-manager wasn't installed by default on RHEL6 and 7 minimal install, so using it instead of yum-config-manager let the instructions be the same in and out of AWS.

tst-lsavoie: The rub is that registering with RedHat costs money on a production system, and AWS covers the cost for you if you use yum-config-manager instead of subscription-manager. That's why I think it's best to have both, even if you have to add a step to install yum-config-manager.

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -43,29 +44,28 @@ sudo yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-6.
...
-Alternatively, install the latest version from the IUS repository (recommended).
-See the [Getting Started](https://ius.io/GettingStarted/) page to find the RPM
-URL for your system. Use this URL below:
+# To Install Git 2.16 on RHEL 6 and Centos 6
+sudo yum install -y https://repo.ius.io/ius-release-el6.rpm
+sudo yum install -y git216
 
-```bash
-sudo yum install -y wget
-cd /tmp
```
tst-lsavoie: There's an extra space here. It looks kind of funny.

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -43,29 +44,28 @@ sudo yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-6.
...
 ```bash
+
+# To the system default version:
 sudo yum install -y git
-```
 
-Alternatively, install the latest version from the IUS repository (recommended).
-See the [Getting Started](https://ius.io/GettingStarted/) page to find the RPM
-URL for your system. Use this URL below:
+# To Install Git 2.16 on RHEL 6 and Centos 6
```
tst-ccamp: Since we are editing this section,  please move the recommended method above the instructions for system default

tst-ccamp:
@tst-jlarocco I already fixed it on this PR, but please start putting the link to the Issue number in the description instead of just the title. That will add a link from the Issue to the PR. Doing the title alone doesn't do this.

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -14,13 +14,21 @@ sudo yum install -y epel-release
 ### RHEL 7
 
 ```bash
+# Setup subscriptions for RHEL 7
```
tst-lsavoie: Minor: it would be nice to have some indication that this only applies to non-AWS RHEL 7 instances. You could include "non-AWS" in this comment, or maybe put the AWS version first and then change the comment to something like "For standard RHEL 7".

tst-lsavoie:
@tst-jlarocco looks like you've got some conflicts to resolve.

tst-lsavoie:
This is working for me on all systems except that on RHEL 6 I had to combine these instructions with some of the changes that have been made on master.

tst-lsavoie:
I ran into a problem with libpng12:
```
Checking for C library gtest... yes
Checking for C header file libpng12/png.h... no
libpng12 must be installed!
scons: *** [build] Error 1
```

tst-lsavoie:
I'm seeing this problem on CentOS 6 now
```
checking Python.h presence... no
checking for Python.h... no
configure: error: header file <Python.h> is required for Python
scons: *** Error 1
scons: *** [NATIVE-REL-x86_64/third_party/postgresql/.configure] Error 1
scons: *** [build] Error 2
```
I'll double check this and the RHEL 7 problem.

tst-jlarocco:
> I'm seeing this problem on CentOS 6 now
> 
> ```
> checking Python.h presence... no
> checking for Python.h... no
> configure: error: header file <Python.h> is required for Python
> scons: *** Error 1
> scons: *** [NATIVE-REL-x86_64/third_party/postgresql/.configure] Error 1
> scons: building terminated because of errors.
> scons: *** [build] Error 2
> scons: building terminated because of errors.
> ```
> 
> Update: I was able to reproduce this problem on CentOS 6 and on RHEL 6. It seems the problem is that `python-devel` was removed from the list of CentOS 6/RHEL 6 dependencies.

Okay, looks like Postgres still requires the Python 2.6 headers.  I added python-devel back to the install list and it's building for me now.

PR #1635, opened 2020-01-28
Fix portable server freeze on Windows

### LINKED ISSUE #1634: Portable server on Windows stops working if user clicks inside terminal window
## Description
When the user clicks inside the command window, it enters "Select" mode and blocks all output. The causes the Python interpreter to get stuck waiting for a `print` command to complete, and it does no work until the user leaves "Select" mode.

## Reproduction steps
1. Run portable server on Windows
1. Click inside the command prompt so that it enters "Select" mode
1. In a browser, go to http://localhost:9335; nothing should happen
1. Return to the command prompt and press the ESC key to exit "Select" mode
1. Try again to reach http://localhost:9335; this time it will load properly

tst-sbayard:
Fixes #1634 

## Description
On Windows, portable server will attempt to disable Quick Edit Mode when it starts up, just prior to starting the I/O loop.

## Implementation
The portable server builder already copied files from platform-specific folders into the main folder, overwriting generic versions. This functionality was used to add Windows-specific functionality to change the console mode.

A generic version of a new file, `platform_specific_functions.py` was added under `earthenterprise/earth_enterprise/src/fusion/portableglobe/servers/`. It contains one function that does nothing. This function is imported and called in `portable_server.py`. This allows `portable_server.py` to be exactly the same across all platforms. 

For a Windows portable build, a different version of `platform_specific_functions.py` is used that provides the necessary functionality.


tst-blaird:
Hi @tst-sbayard -- I'd like to review and test this one when it's ready.  I see the WIP tag so I'm assuming I shouldn't test yet, correct?

tst-sbayard:
@tst-blaird Awesome, thanks! I'm still testing, but I'll let you know as soon as it's ready.

tst-blaird:
@tst-sbayard I would think this should be tested on both Windows 7 32-bit and Windows 10 64-bit so that's what I plan to do.  Let me know if there's another combination that should also be tested.

tst-sbayard:
> @tst-sbayard I would think this should be tested on both Windows 7 32-bit and Windows 10 64-bit so that's what I plan to do. Let me know if there's another combination that should also be tested.

@tst-blaird I think that sounds good. The way the portable server build works, you may want to test on one Linux distro just to verify that the right versions of the platform_specific_functions.py is used.

I tested on Windows 7 and 10 (both 32-bit, though) and on Cent OS 7.

tst-blaird:
Update:  I've successfully tested on Windows 10 64-bit and am working on Windows 7 32-bit.

tst-blaird:
Update:  Successfully tested on Windows 7 32-bit but I did notice one thing that I want to compare back to Windows 10.

tst-blaird:
@tst-sbayard I think this is a minor issue but I did notice the following:  With this change if you press CTRL+C in the command prompt window to try to quit the portable server it does not appear to immediately quit the script as there is no reaction by the script in the command prompt window.  However, if in your browser you now try to use the portable server you get the traceback with the KeyboardInterrupt exception.  It's just a bit confusing to not get that traceback immediately.  At first I hit CTRL+C several times trying to get it to quit.

tst-sbayard:
@tst-blaird I think that's actually a pre-existing issue with how the
tornado I/O loop works. I'll double check next week, but I'm fairly sure
that's the same behavior as before, at least on Windows.

tst-blaird:
I think the changes look good.  I might have added a couple of logging statements to the new code in platform_specific_functions.py but I'm not worried about it.  Waiting for Stephen's response to the one behavior I noticed and then I'll finish out the review.

tst-blaird:
> I think that's actually a pre-existing issue with how the tornado I/O loop works.
@tst-sbayard you are absolutely correct.  I guess I never noticed it before.  😃 Ok, I'm good with the changes.  Approved!

tst-blaird:
This probably needs to be added to the release notes.

PR #1644, opened 2020-02-07
Rework gehttpd_init to remove initscripts.

tst-cjeffries:
Reworks gehttpd_init to exclusively use apachectl. This allows us to remove the dependency for initscripts.

To test:

- Install on all supported platforms
- Verify that all of the geserver scripts work (stop, start, restart, etc.)
- Run through the tutorial to verify geserver functionality.

### DIFF
--- a/earth_enterprise/src/server/gehttpd_init
+++ b/earth_enterprise/src/server/gehttpd_init
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2017 Google Inc.
+# Copyright 2017-2020 Google Inc.
 #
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/server/gehttpd_init
```diff
@@ -48,6 +48,11 @@ if test -f /opt/google/gehttpd/bin/envvars; then
   . /opt/google/gehttpd/bin/envvars
 fi
 
+if [ "$EUID" != "0" ] ; then
```
tst-dsugar: good point....

PR #1646, opened 2020-02-10
Move XML load into AssetSerializer

### LINKED ISSUE #1649: Consolidate XML loading code into AssetSerializer
To continue consolidating functionality and reducing dependence on auto-generated C++ code, it would be good to move the code that loads assets from XML into `AssetSerializer`.

tst-sbayard:
Fixes #1649 

This change adds the concept of a plugin registry where each asset type and asset version type will register itself as a plugin. This allows client code to simply query for the plugin interface based on the asset name.

As a result, the XML loading code for assets has been moved to `AssetSerializerLocalXML`, and LoadAnyGen.pl has been completely removed.


earth_enterprise/src/fusion/autoingest/AssetFactory.h
```diff
@@ -304,13 +327,13 @@ namespace AssetFactory
...
 
-  template <class MutableDerivedVersionHandleType, class ConfigType, class VersionDType>
+  template <class MutableDerivedVersionHandleType, class ConfigType, class VersionDType, class CachedInputType>
   MutableDerivedVersionHandleType ReuseOrMakeAndUpdate(const std::string& ref_,
                                                        AssetDefs::Type type_,
                                                        const std::vector<SharedString>& inputs_,
                                                        const khMetaData& meta_,
                                                        const ConfigType& config_,
-                                                      const std::vector<AssetVersion>& cachedinputs_)
+                                                      const std::vector<CachedInputType>& cachedinputs_)
```
tst-lsavoie: Minor, and not your change, but this line doesn't have the same indentation as the others. Would be nice to fix it.

earth_enterprise/src/fusion/autoingest/AssetGenPublicCpp.pl
```diff
@@ -85,6 +86,26 @@ ()
     void GetConfig(DOMElement *elem, $config &config);
 }
 
+
+
+namespace{
```
tst-lsavoie: Just thinking out loud: it would sure be nice if we could combine some of this duplicated code for assets and asset versions. That's true here and in several other places. Perhaps we could have a common base class for both? If we did that we could even go to a single Storage Manager, which would make managing memory usage easier.

To be clear, I don't want you do make this change on this PR. It would be too much. It's an interesting idea, though.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  static void RegisterPlugin(
+    std::string const & assetTypeName, 
+    std::unique_ptr<AssetPluginInterface> assetPlugin)
+  {
+    PluginRegistry()[assetTypeName] = std::move(assetPlugin);
+  }
+
+  static void UnregisterPlugin(
```
tst-lsavoie: Looks like this function is only used in the unit tests. I'm in favor of removing the function if possible. Less code is better.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
```diff
@@ -5,6 +5,10 @@
 #include "common/khxml/khxml.h"
 #include "common/khxml/khdom.h"
 #include "autoingest/.idl/storage/AssetDefs.h"
+#include <khxml/khdom.h>
+using namespace khxml;
```
tst-lsavoie: It's generally a bad idea to include a `using namespace` declaration in a header file. We should remove that and just use `khxml::` as needed.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
```diff
@@ -14,13 +18,81 @@ class AssetSerializerInterface {
     virtual ~AssetSerializerInterface() = default;
 };
 
+/*
+* GetFileInfo and ReadXMLDocument were added as function templates to allow
+* them to be overridden in unit tests. The versions here should be used for
+* all asset types during normal operation.
+*/
+template<class AssetType>
```
tst-lsavoie: It looks like these don't need to be templates. If you remove the template header and mark them `inline` I think they will work fine.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
```diff
@@ -14,13 +18,81 @@ class AssetSerializerInterface {
...
+              throw khException(kh::tr("No document element"));
+            std::string tagname = FromXMLStr(top->getTagName());
+            result = AssetFactory::CreateNewFromDOM<AssetType>(tagname,top);
+            if (nullptr == result) {
+              AssetThrowPolicy::WarnOrThrow
+                (kh::tr("Unknown asset type '%1' while parsing %2")
+                .arg(ToQString(tagname), ToQString(filename)));
+            }
+          } catch (const std::exception &e) {
+          AssetThrowPolicy::WarnOrThrow
```
tst-lsavoie: Indentation looks off here.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
```diff
@@ -14,13 +18,81 @@ class AssetSerializerInterface {
...
+                  + filename);
+        }
+      } else {
+          AssetThrowPolicy::WarnOrThrow(kh::tr("No such file: ") + filename);
+      }
+
+      if (!result) {
+        notify(NFY_DEBUG, "Creating placeholder for bad asset %s",
+              boundref.c_str());
+        // Use SourceAssetImpl since AssetImpl is pure virtual
```
tst-lsavoie: I'm not sure this comment makes sense anymore because there's no reference to SourceAssetImpl here.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
tst-lsavoie: Update: maybe they need to be templates so you can override them in the unit tests. I guess I missed that in the comment.

earth_enterprise/src/fusion/autoingest/AssetSerializer_unittest.cpp
```diff
@@ -83,14 +116,155 @@ class AssetSerializerTest : public testing::Test {
...
+  geDocPtr = std::unique_ptr<GEDocument>(new GECreatedDocument(""));
+  ASSERT_FALSE(geDocPtr == nullptr);
+  string errorMsg = RunLoadForException(serializer);
+  ASSERT_EQ(errorMsg, "Error loading filename.xml: No document element");
+}
+
+TEST_F(AssetSerializerTest, Load_UnknownAssetType) {
+  string docElement = "DummyAsset";
+  geDocPtr = CreateEmptyDocument(docElement);
+  ASSERT_FALSE(geDocPtr == nullptr);
```
tst-lsavoie: Minor: there's an `ASSERT_NE` for not equal tests.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
```diff
@@ -14,13 +18,81 @@ class AssetSerializerInterface {
...
+        result = AssetFactory::CreateNewInvalid<AssetType>(
+          AssetType::GetPlaceholderAssetRegistryKey(), boundref
+        );
+
+        // Leave timestamp alone if it failed but there was a valid timestamp
```
tst-lsavoie: I'm not sure what this comment means. The code doesn't leave the timestamp alone unless there's a failure creating the placeholder asset, which seems pretty unlikely. Maybe just remove the comment?

earth_enterprise/src/fusion/autoingest/AssetGenPublicCpp.pl
tst-sbayard: I'm with you there. I'd love to have a common base for them.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
tst-sbayard: Yeah, if not for the unit tests, those wouldn't need to be separate functions at all, template or otherwise. I pulled them out like that so that I wouldn't need to create temporary files for `GetFileInfo` and `ReadDocument` to act on.
It does add some otherwise unnecessary complexity, so if there's a better way to go about it, I'm certainly open to it.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
tst-sbayard: Good catch. This code came from C++-inside-perl and the indentation was all off. That one slipped by.
There was also an inconsistency between the Load and Save methods here (2 vs 4 spaces). I'm making it all two-space indentation.

earth_enterprise/src/fusion/autoingest/AssetSerializer_unittest.cpp
tst-sbayard: I think that was brain fart, because I'm using  `ASSERT_NE` other places. Also, I meant to take this assert out anyway. I was using that to make sure that `CreateEmptyDocument` worked like I expected. 
I'll remove this and the other, similar one.

earth_enterprise/src/fusion/autoingest/AssetSerializer.h
tst-sbayard: Yeah, that's an odd one. I almost pulled it out before. I _think_ what it's meant for is the case where the call to `GetFileInfo` succeeds but the call to `ReadXMLDocument` returns null.
i.e., the file got created at some point, but the contents are invalid, so report the timestamp of the original file.
I kind of like the idea of making a note of that here, but I'll re-do the comment to be more clear.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
tst-sbayard: It was really helpful for making sure each test started in a clean state, but you're right, it's otherwise it's not necessary.
So, I derived from `AssetRegistry` in the unit tests and added the `UnregisterPlugin` method to that class instead. I think that's cleaner.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
```diff
@@ -0,0 +1,85 @@
...
+
+  static AssetPluginInterface * GetPlugin(const std::string & assetTypeName) {
+    auto it = PluginRegistry().find(assetTypeName);
+    if (it != PluginRegistry().end())
+      return it->second.get();
+    return nullptr;
+  }
+
+protected:
+  static std::unordered_map<std::string, std::unique_ptr<AssetRegistry::AssetPluginInterface>> & PluginRegistry();
```
tst-lsavoie: Minor: This is a long return type. A `using` statement would be handy.

earth_enterprise/src/fusion/autoingest/AssetSerializer_unittest.cpp
```diff
@@ -27,28 +27,43 @@ using namespace khxml;
...
 class TestItem : public TestItemStorage {
   public:
     using Base = TestItemStorage;
     static int nextValue;
     static string loadFile;
+    static string xmlFilename;
+    static bool newInvalidCalled;
     int val;
+    uint64 timestamp; 
+    time_t filesize;
```
tst-jkent: I think these types need to be switched for **timestamp** and **filesize**

tst-sbayard: D'oh! Good catch.

tst-lsavoie:
Code looks fine and tests pass. We just need to resolve some of the performance differences we're seeing.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
```diff
@@ -0,0 +1,87 @@
...
+  static void RegisterPlugin(
+    std::string const & assetTypeName, 
+    std::unique_ptr<AssetPluginInterface> assetPlugin)
+  {
+    PluginRegistry()[assetTypeName] = std::move(assetPlugin);
+  }
+
+  static AssetPluginInterface * GetPlugin(const std::string & assetTypeName) {
+    auto it = PluginRegistry().find(assetTypeName);
+    if (it != PluginRegistry().end())
```
tst-lsavoie: If this if statement fails it means something is wrong with the code, correct? We should never have a plugin that hasn't registered its interface. Maybe we can change this statement into an assert and always return `it->second.get()`. That way, code that calls this function doesn't have to check for `nullptr`.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  static void RegisterPlugin(
+    std::string const & assetTypeName, 
+    std::unique_ptr<AssetPluginInterface> assetPlugin)
+  {
+    PluginRegistry()[assetTypeName] = std::move(assetPlugin);
+  }
+
+  static AssetPluginInterface * GetPlugin(const std::string & assetTypeName) {
```
tst-lsavoie: Minor: maybe should return `const AssetPluginInterface *`. I don't know how much practical difference it will make, but we don't want callers modifying the interface.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
```diff
@@ -0,0 +1,87 @@
...
+    auto it = PluginRegistry().find(assetTypeName);
+    if (it != PluginRegistry().end())
+      return it->second.get();
+    return nullptr;
+  }
+
+  using PluginMap = std::unordered_map<std::string, std::unique_ptr<AssetRegistry<AssetType>::AssetPluginInterface>>;
```
tst-lsavoie: I'm not sure a `unique_ptr` is necessary here. We might be able to get away with allowing copying for `AssetPluginInterface`s and just having a `std::unordered_map<std::string, AssetRegistry<AssetType>::AssetPluginInterface>`. That would save us some indirection and possibly make the code faster.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
```diff
@@ -0,0 +1,87 @@
...
+    std::string const & assetTypeName, 
+    std::unique_ptr<AssetPluginInterface> assetPlugin)
+  {
+    PluginRegistry()[assetTypeName] = std::move(assetPlugin);
+  }
+
+  static AssetPluginInterface * GetPlugin(const std::string & assetTypeName) {
+    auto it = PluginRegistry().find(assetTypeName);
+    if (it != PluginRegistry().end())
+      return it->second.get();
```
tst-lsavoie: Returning a pointer here results in double indirection - you have a pointer to an `AssetPluginInterface` that includes a pointer to a `NewFromDOM` method. If we could eliminate one of those pointers we could get a nice speed boost. You could try returning a reference instead of a pointer. I'm not sure if that's handled as a pointer under the covers or not, but it's worth a try. You could also try returning a copy instead of a pointer, but then you've got copy costs to worry about.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
tst-lsavoie: I take it back. You need the if/else statement in case there's a malformed XML file. It may be worth considering if there's a way to avoid returning nullptr.

tst-sbayard: Malformed XML could cause a failure in `AssetFactory::CreateNewFromDOM`, but I don't think it comes into play here. 
I can't think of any condition where this could/should return `nullptr`. Unless an asset type was removed from one version of Fusion to another, but I can't imagine that happening.

tst-lsavoie: Are you guaranteed to have a valid `assetTypeName` at this point? I thought it was read straight from the XML. So maybe I'm not so much concerned about malformed XML as well-formed XML with garbage data.

tst-lsavoie:
FYI, I put together some sample code that builds an unordered_map and a large if/else statement with the asset names we're using and compared the speed. The unordered_map wins by a large margin:
```
If/Else
  Init time:    3.89e-07
  Look Up time: 3.92026
Map
  Init time:    4.3408e-05
  Look Up time: 0.531324
Overall time: 5.48761
```
So that's likely not the performance difference. At this point it seems the most likely culprit is the extra indirection with pointers and iterators all over the place.

earth_enterprise/src/fusion/autoingest/AssetRegistry.h
tst-sbayard: "Guaranteed" is probably a strong word. It's coming from an XML file that Fusion created, so it's probably _relatively_ safe except in the case of a corrupted file, but I'm all for leaving the check in there as good practice.

Issue #1650, opened 2020-02-14
WMS https doesn't work behind reverse proxies


tst-cjeffries:
**Describe the bug**

Reverse proxies that use https do not work with WMS. WMS returns http urls instead of https.

**To Reproduce**

Setup OpenGEE behind a reverse proxy using https. Notice that returned WMS payloads incorrectly use http.

**Expected behavior**

WMS payloads use https when behind a reverse proxy that use https

PR #1651, opened 2020-02-14
#1650: WMS now uses forwarded protocol

tst-cjeffries:
WMS payloads now use the X-Forwarded-Proto header to set the url protocol in the payload. This allows WMS to work through reverse proxies that use a different protocol.

To test:
- Build and install OpenGEE.
- Build and publish a 2D database with WMS enabled.
- Setup a reverse proxy that accepts https and forwards to the OpenGEE server on http.
- Verify that the WMS payloads have the proper url and protocol for the proxy:
    - `curl --insecure  "https://<proxy_fqdn>/<db_name>/wms?SERVICE=WMS&REQUEST=GetCapabilities"`

### DIFF
--- a/earth_enterprise/src/server/wsgi/wms/ogc/common/utils.py
+++ b/earth_enterprise/src/server/wsgi/wms/ogc/common/utils.py
@@ -143,6 +143,10 @@ def GetServerURL(environ):
   complete_url += urllib.quote(environ.get("PATH_INFO", ""))
 
   if environ.get("HTTP_X_FORWARDED_HOST"):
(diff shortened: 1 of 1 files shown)


tst-jzmolek:
Looks good to me. I had tested this prior.

tst-blaird:
The changes look good to me as well.  @tst-jzmolek so no need to test this change?

tst-blaird:
Does this need a release notes change?

tst-cjeffries:
@tst-blaird Thats a good point.

PR #1655, opened 2020-02-27
Add more Depends calls for generated code

tst-rwildes:
Fixes issue #1657

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/SConscript
+++ b/earth_enterprise/src/fusion/autoingest/SConscript
@@ -149,6 +149,8 @@ for plugin in plugins:
     env.Command('plugins/'+plugin+'Asset.cpp',
                 [ 'AssetGenPublicCpp.pl', 'AssetGen.pm', src ],
                 [ "${SOURCES[0]} ${SOURCES[2]} $TARGET" ])
+    env.Depends('plugins/'+plugin+'Asset.cpp', 'plugins/'+plugin+'Asset.h')
+    env.Depends('plugins/'+plugin+'Asset.os', 'plugins/'+plugin+'Asset.cpp')
 
 env.Command('ReadOnlyFromStorage.cpp',
             [ 'FromStorageGen.pl' ],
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/fusion/autoingest/SConscript
```diff
@@ -149,6 +149,8 @@ for plugin in plugins:
     env.Command('plugins/'+plugin+'Asset.cpp',
                 [ 'AssetGenPublicCpp.pl', 'AssetGen.pm', src ],
                 [ "${SOURCES[0]} ${SOURCES[2]} $TARGET" ])
+    env.Depends('plugins/'+plugin+'Asset.cpp', 'plugins/'+plugin+'Asset.h')
```
tst-lsavoie: I don't think the .cpp needs to depend on the .h. More likely the .os needs to depend on both the .cpp and .h. Same comment in the other file.

earth_enterprise/src/fusion/autoingest/sysman/SConscript
```diff
@@ -45,6 +45,10 @@ for plugin in plugins:
     env.Command('plugins/'+plugin+'AssetD.cpp',
                 [ '../AssetGenDaemonCpp.pl', '../AssetGen.pm', src ],
                 [ "${SOURCES[0]} ${SOURCES[2]} $TARGET" ])
+    env.Depends('plugins/'+plugin+'AssetD.cpp','plugins/'+plugin+'AssetD.h')
+    env.Depends('plugins/'+plugin+'AssetD.os', 'plugins/'+plugin+'AssetD.cpp')
+
+env.Depends('../plugins/ProjectedMapProjectAsset.cpp', 'plugins/MapLayerAssetD.h')
```
tst-lsavoie: I think this should be a .os depending on a .h as well.

earth_enterprise/src/fusion/autoingest/SConscript
tst-rwildes: yep that makes sense

earth_enterprise/src/fusion/autoingest/sysman/SConscript
tst-rwildes: will do

PR #1663, opened 2020-03-11
move to gradle 6.2.2

tst-rwildes:
addresses #1675


tst-blaird (approved):
Changes look good to me.  Built successfully on EL6 and EL7.  You probably should add some description text to the PR.  😃

tst-nrichman (approved):

PR #1664, opened 2020-03-11
1640: Update and bundle SWIG

tst-cjeffries:
Update and bundle a newer version swig to allow full migration of C++11.

To test (on EL6, and Ubuntu):
- Build portable: You should see swig build in the output at this step.
- Install portable
- Verify that it is working.
- Build and install open.
- Verify that it works.

Also review changes to the documentation and the GitHub portable wiki page.


.gitattributes
```diff
@@ -6,3 +6,4 @@
 *.glc filter=lfs diff=lfs merge=lfs -text
 *.glm filter=lfs diff=lfs merge=lfs -text
 earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/test_data/** binary
+earth_enterprise/third_party/swig/swig-4.0.1.tar.gz filter=lfs diff=lfs merge=lfs -text
```
tst-rwildes, line 9: The `*.gz` above doesn't cover this?

earth_enterprise/src/fusion/gepublish/SConscript
```diff
@@ -46,7 +47,7 @@ swig_env['CPPFLAGS'] += ['-fPIC']  # Needed for 64-bit compilation.
 swig_env['CPPFLAGS'] += ['-I%s' % swig_env['python_include']]
 swig_env.Append(SWIGFLAGS = ['-python', '-c++'])
 
-libgepublishmanagerhelper = swig_env.sharedLib('gepublishmanagerhelper',
+libgepublishmanagerhelper = swig_env.SharedLibrary('gepublishmanagerhelper',
```
tst-rwildes, line 50: What is this change doing?

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/file_unpacker.h
```diff
@@ -35,16 +35,6 @@ class GlcReader;
  * Class for unpacking files from a single package file.
  */
 class FileUnpacker {
-  /**
```
tst-rwildes, line 38: You should add these back in but as deleted and public.  Also should do the same for move operations

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
```diff
@@ -91,6 +99,8 @@ def BuildLibrary(os_dir, ignore_results):
     archData = platform.architecture(pathToLib)
     if archData[0] == "64bit":
       specialDefs = "-DMS_WIN64"
+  elif os_dir == "Linux":
+    build_swig(source_dir)
```
tst-rwildes, line 103: We only need to do this on Linux?  What about Mac or Windows?

.gitattributes
tst-cjeffries: I bet the git lfs add added this. Shall I remove then?

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
tst-cjeffries: I honestly don't know about mac, but windows we have the newer version of swig available to us, so we don't need to build it.

earth_enterprise/src/fusion/gepublish/SConscript
tst-cjeffries: Ah. Yeah that needs changed back to sharedlib. Sharedlibrary is the scons builder, and sharedLib appears to be our kh version.

tst-cjeffries:
Its also worth noting that the wiki page will need to be updated for portable. I will do this once the PR is merged in, as it isn't a static file that can be edited as part of the PR.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/shared/file_unpacker.h
tst-cjeffries: Misunderstood the TODO. Thanks.

tst-cjeffries: Added back is a deleted and public.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
tst-rwildes: Looks like according to docs Portable is not supported on Mac so we can forget about that.  I do thing the docs need to be updated to reflect these changes: https://www.opengee.org/geedocs/5.3.4/answer/portable/portableUserGuideWinLinux.html

tst-cjeffries: Ok. That seems logical to me. Ill make reviewing the changes to the wiki part of reviewing the PR.

earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/build_and_test.py
```diff
@@ -45,6 +45,14 @@
 import opengee.version
 
 
+def build_swig(source_dir):
```
tst-rwildes, line 48: IDK...  At the end of the day should we just say this script is for Windows building of Portable only.  And give a scons target for building portable when in Linux?

tst-cjeffries: I agree, but that would be quite a major refactor. For the time being I believe it makes the most sense to make those changes as a separate issue.

PR #1665, opened 2020-03-13
changes to portable server while loading 3D globes

### LINKED ISSUE #1666: User sees "plugin failed to load" alerts when browsing to portable server page
Customer is complaining about the user experience of this.
Since we know the plugin is going to fail every time, we really should avoid trying to load it in the first place. 
Instead of attempting to load the plugin, show a message:
“3D previews not available, please use the GEEC to view your globe"

tst-hrajamoney:
Fixes #1666  Changed the msg to : 3D previews not available, please use the GEEC to view your globe
while loading 3d globes in portable server.
Change the default globe loaded to tutorial_sf.glm, so that on localhost:9335, 2D map is displayed.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/servers/local/preview/setup.html
+++ b/earth_enterprise/src/fusion/portableglobe/servers/local/preview/setup.html
@@ -44,9 +44,7 @@
   <div id="earth"></div>
   <div id="NoMap" style="display: none;">
     <div class="GlobeImgText">
-      <span>There was a problem loading the Google Earth Plugin.</span>
-      <span>The Google Earth Plugin is currently only available on <b>Windows</b> and <b>Mac OS X 10.6+</b>.</span>
-      <span><a href="http://earth.google.com/intl/en/plugin/">Click here to learn more.</a></span>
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/fusion/portableglobe/servers/local/preview/setup.html
```diff
@@ -44,9 +44,7 @@
   <div id="earth"></div>
   <div id="NoMap" style="display: none;">
     <div class="GlobeImgText">
-      <span>There was a problem loading the Google Earth Plugin.</span>
-      <span>The Google Earth Plugin is currently only available on <b>Windows</b> and <b>Mac OS X 10.6+</b>.</span>
-      <span><a href="http://earth.google.com/intl/en/plugin/">Click here to learn more.</a></span>
+      <span>3D previews are not available, please use the <b>GEEC</b> to view your globe.</span>
```
tst-sbayard: I might change the wording slightly, like instead of "not available", maybe specify "not available in a web browser".
And I think it might be good to spell out "Google Earth Enterprise Client" instead of "GEEC".
Otherwise, I think that's a good message.

Issue #1698, opened 2020-04-08
Standardize the use of unsigned integers


tst-nrichman:
KhTypes.h defines a set of type definitions for unsigned integers that are used throughout OpenGee. This makes the declarations of unsigned integers less consistent and more reliant on our header file. All unsigned integers should be replaced with standard library declarations.

tst-rwildes:
Fix merged

PR #1699, opened 2020-04-08
1698 Standardize Integers

tst-nrichman:
Removes the integer typedefs from khTypes.h and makes use of the standard library definitions.
Fixes issue #1698


earth_enterprise/src/fusion/portableglobe/servers/fileunpacker/Windows/khTypes.h
```diff
@@ -22,11 +22,8 @@
 #include <assert.h>
 #include <sys/types.h>
 
-typedef unsigned short                  uint16;
 typedef short                           int16;
```
tst-rwildes, line 25: should you remove these too?  It would be more portable if the code didn't use these anymore and used `std` versions instead.

PR #1701, opened 2020-04-09
Qt4 migration

tst-basilhuffman:
migrating qt3 to qt4


tst-rwildes:
Can this commit history be cleaned up a bit before merging?  You don't have to squash it into a single commit but try and merge some of these commits together into a single logical change set for each commit record that has a meaningful message describing the change.

PR #1709, opened 2020-04-14
Prevent encoding buffer overflows

### LINKED ISSUE #1723: Overflows in Encode function
**Describe the bug**

The Encode function in etencoder.cc has buffer overflows if the key is shorter than 24 bytes or not a multiple of 8 bytes.

**To Reproduce**

The following mini program illustrates the issue:

```
#include <cstring>
#include <iostream>
#include <memory>
#include <set>
#include "common/etencoder.h"

using namespace std;

unique_ptr<uint8_t[]> GetByteArray(uint32_t length) {
  unique_ptr<uint8_t[]> data(new uint8_t[length]);
  memset(data.get(), 0, length);
  return data;
}

int main() {
  const uint32_t KEY_LENGTHS = 100;
  const uint32_t DATA_LENGTHS = 100;
  for(uint32_t klen = 0; klen < KEY_LENGTHS; ++klen) {
    auto key = GetByteArray(klen);
    for(uint32_t dlen = 0; dlen < DATA_LENGTHS; ++dlen) {
      cout << "Testing key length: " << klen << ", data length: " << dlen << endl;
      auto data = GetByteArray(dlen);
      etEncoder::Encode(data.get(), dlen, key.get(), klen);
    }
  }

  return 0;
}
```

Compile it with address sanitizer and run it to see the buffer overflows. Here's a sample Makefile

```
SOURCES=$(wildcard *.cpp)
OBJECTS=$(SOURCES:.cpp=.o)
CPPFLAGS=-g -std=c++11 -fsanitize=address -fno-omit-frame-pointer -fno-common
EXE=test

.PHONY: clean

$(EXE): $(OBJECTS)
	$(CXX) $(CPPFLAGS) -o $@ $^

clean:
	rm -rf $(EXE) *.o
```

**Expected behavior**

Address sanitizer does not find any buffer overflows in the above test.

tst-lsavoie:
Prevents some additional buffer overflows when the encoding key is not a multiple of 8 bytes. Also improves testing and error handling.

The dbroot is used in lots of places in the code, so here are some ideas to test:
- Build and publish a database and view it in Earth Client
- Run the unit tests
- Run earthenterprise/earth_enterprise/src/NATIVE-DBG-x86_64/bin/gedbrootgoogleurlremover on a dbroot with an invalid key and make sure it doesn't crash. This will be hard because GEE won't let you create a dbroot with an invalid key. You could modify the code so that the Encode function always throws an exception and make sure it's caught.

Fixes #1723 (the part that wasn't fixed in PR #1707)


earth_enterprise/src/common/gedbroot/eta_dbroot.cc
```diff
@@ -77,22 +77,21 @@ bool EtaDbroot::DecodeBinary(const std::string &binary,
...
+    throw khSimpleException("Corrupted dbroot: Key has unexpected value.");
   }
 
   // Read the header and see if it is a valid binary dbroot file.
   std::string header_buffer;
   header_buffer.append(binary, 0, sizeof(etDbaseRootHeader));
   // Decrypt the entire header
-  etEncoder::Decode(&header_buffer[0], header_buffer.size(),
-                    key_buffer.data(), key_buffer.size());
+  etEncoder::DecodeWithDefaultKey(&header_buffer[0], header_buffer.size());
```
tst-lsavoie, line 87: I switched this (and the Decode call below) to use the default key because the code verifies that the key matches the default key above.

earth_enterprise/src/common/gedbroot/proto_dbroot_unittest.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <gtest/gtest.h>
+
+#include "proto_dbroot.h"
+#include "common/khSimpleException.h"
+
+// Test various error and non-error conditions when loading proto dbroots.
+// Since we're validating the logic for loading the dbroots, we don't worry too
+// much about the contents.
+
+TEST(ProtoDbroot, InvalidFile) {
```
tst-lsavoie, line 24: I started writing unit tests for the Eta dbroot as well, but it would take significant effort to find an Eta dbroot and it's not worth the effort for something that's never going to change.

earth_enterprise/src/common/gedbroot/gedbroot_google_url_remover.cpp
```diff
@@ -97,26 +97,32 @@ int main(int argc, char* argv[]) {
...
-      dbroot.Write(fname_dbroot + ".before.txt", geProtoDbroot::kTextFormat);
-    }
-
-    dbroot_is_valid = RemoveUnwantedGoogleUrls(&dbroot);
-    dbroot.Write(cleaned_output_fname, geProtoDbroot::kEncodedFormat);
-
-    if (want_debug) {
-      dbroot.Write(cleaned_output_fname + ".txt", geProtoDbroot::kTextFormat);
+  bool dbroot_is_valid = false;
+  try {
```
tst-sbayard, line 101: Was there an uncaught exception that necessitated a try/catch? Does this end up calling `EtaDbroot::DecodeBinary` where the "new" throws are?

tst-lsavoie: This calls all the way down to Encode, but I think it goes through proto_dbroot.cc, not eta_dbroot.cc. There was already the possibility that an exception would be thrown and it wasn't catching it. I figured I'd fix it since I added another exception.

earth_enterprise/src/common/etencoder.cc
```diff
@@ -109,6 +109,10 @@ const std::string kDefaultKey(
 
 
 void Encode(void* data, uint32 datalen, const void* key, uint32 keylen) {
+  if (keylen % 8 != 0) {
+    throw khSimpleException("Encryption keys must be a multiple of 8 bytes long.");
```
tst-jlarocco, line 113: Should this have an empty return like the if block on lines 126-130?

tst-lsavoie: I interpreted the return on line 129 as being a part of the standard processing when the encoded string is the same as the input string, so it's not really an error condition. That's why I opted to throw an exception here to make it clear that there was a problem, not normal processing. I guess you could argue line 129 is also error checking since it checks for null pointers as well, but I think it's safe to say that when this condition occurs we want processing to stop, not go on like normal.

Issue #1715, opened 2020-04-16
Documentation currently references old/unsupported configurations


tst-cjeffries:
Currently multiple files reference old and unsupported operating systems (SUSE, Ubuntu 10/12). These references need removed, as these operating systems are no longer supported.

**To Reproduce**

See `docs/README.md`, `earth_enterprise/src/README.build`, etc.

**Expected behavior**

All references to old/unsupported operating systems are removed.

PR #1716, opened 2020-04-16
#1715: Remove unsupported configurations from documentation

tst-cjeffries:
Removes references to unsupported operating systems in the documentation.

No testing required.


tst-basilhuffman:
i think this one might be a whackamole situation if any other distro refrences are found. lgtm though

Issue #1640, opened 2020-01-31
Swig 2 is preventing migration to C++11 


tst-cjeffries:
Currently, OpenGEE uses Swig 2 which does not allow for C++11 constructs. This primarily affects portable server, however, in order to to fully migrate to C++11, we must upgrade to a newer compatible version of swig (3 or 4). Due to limited support in system repositories, we suggest bundling it into our third_party code base.

**To Reproduce**

Was found when migrating some portable server code to C++11.

**Expected behavior**

OpenGEE uses as new version of swig, and bundles it to avoid issues with availability.

PR #1721, opened 2020-04-17
#1640: Bundle and Upgrade Swig

tst-cjeffries:
Bundles a newer version of swig to complete the C++11 migration. 

To Test:
1. Build portable.
2. Verify that it builds and works.
3. Build OpenGEE.
4. Verify that the build completes and fusion and server are both working.


tst-dsugar:
still looks good

PR #1732, opened 2020-05-04
Fix issue with extra versions

### LINKED ISSUE #1731: Historical builds create extra versions of assets
In some cases, when building a historical imagery project, extra versions of assets are created. The build is still successful, but the resulting database is larger than necessary.

tst-lsavoie:
This fixes two issues that cause extra versions to be produced unnecessarily during a build. The resulting database is correct, but the size of the database on disk is much larger than necessary.

The actual fixes are fairly small, but I did quite a bit of debugging and refactoring to track down the issue. Instead of throwing away the refactoring I am including it, so the changes are larger than one might expect. I tried to organize the changes into good commits, so it may be easier to review one commit at a time.

Here is a list of the major things I changed. I'll also provide comments on the PR to explain some of the changes.
- Got rid of the "NoLongerNeeded" call that sometimes removed asset versions from the cache too early.
- Re-copied the ReuseOrMakeAndUpdate and ReuseOrMakeAndUpdateSubasset code from previous versions into the Asset Factory. I'm not sure exactly what the difference is, but this fixes an issue that causes extra versions to be created.
- Changed the template parameters for the functions in Asset Factory. The majority of lines changed relates to this. I created a new struct for every asset type that combines all the relevant related types for that asset and pass that as the template parameter to Asset Factory functions. This simplifies the code and makes it easier to get right, but it requires a lot of changes.
- The above change required making significant changes to the unit tests. I also modified the tests to eliminate redundancies and achieve 100% coverage of Asset Factory.
- Added some error checking to Storage Manager.
- Changed how the Asset Serializer calculates filenames to match how the Storage Manager calculates filenames. I believe both methods were effectively identical, but this ensures there are no issues.

As a side effect of this fix, builds should be somewhat faster (due to building fewer versions of assets) and memory usage may be much less (due to extra versions of assets not hanging around in memory after they are removed from the cache). I've seen memory usage reductions of as much as 2/3 in system manager.

Fixes #1731


earth_enterprise/src/fusion/autoingest/AssetFactory.h
```diff
@@ -261,274 +238,189 @@ namespace AssetFactory
...
               inputs_,
               meta_,
               config,
               cachedinputs_);
   }
 
-
-  template <class MutableDerivedVersionHandleType,
-            class ConfigType, class VersionDType>
-  MutableDerivedVersionHandleType ReuseOrMakeAndUpdate(const std::string& ref_,
```
tst-lsavoie, line 286: I copied over all the ReuseOrMakeAndUpdate* functions from earlier versions, which was the fix to one of the issues. Unfortunately it makes the diff hard to read.

earth_enterprise/src/fusion/autoingest/AssetFactory.h
```diff
@@ -261,274 +238,189 @@ namespace AssetFactory
...
-                  if ((version->state != AssetDefs::Offline) &&
-                      (version->inputs.size() == 0) &&
-                      config_.IsUpToDate(version->config))
-                  {
-                      version.MakePermanent();
-                      return version;
-                  }
-                  else
-                  {
-                      version.NoLongerNeeded();
```
tst-lsavoie, line 311: Removing the NoLongerNeeded calls was a fix to one of the bugs.

earth_enterprise/src/fusion/autoingest/AssetFactory.h
```diff
@@ -49,210 +55,181 @@ namespace AssetFactory
       return nullptr;
   }
 
-  template<class AssetType>
```
tst-lsavoie, line 52: Most of the changes in this file are to use new template parameters that make the code clearer and make it easier to use the right template parameters.

earth_enterprise/src/fusion/autoingest/AssetFactory_unittest.cpp
```diff
@@ -27,49 +28,13 @@
 using namespace std;
 using namespace AssetFactory;
 
-// forward declare classes
```
tst-lsavoie, line 30: I essentially rewrote the MockAsset class hierarchy from scratch. I tried to keep parts from the original version but that probably won't show up in the diff. The tests themselves should be more similar, but I did make some changes, including getting rid of duplicate tests and adding new tests to get better coverage.

earth_enterprise/src/fusion/autoingest/AssetFactory.h
```diff
@@ -49,210 +55,181 @@ namespace AssetFactory
...
   template<class AssetType>
-  AssetType Find(const std::string & ref) {
-    return Find<AssetType>(ref, AssetType::Impl::EXPECTED_TYPE);
+  inline typename AssetType::AssetD Find(const std::string & ref) {
+    return Find<typename AssetType::AssetD>(ref, AssetType::TYPE, AssetType::SUBTYPE);
   }
 
-  template<class VersionType>
+  template<class AssetType>
+  inline typename AssetType::VersionD FindVersion(const std::string & ref, const AssetDefs::Type & type) {
```
tst-lsavoie, line 86: The `Find` function is a special case since it can find assets or asset versions, but the new template parameter includes both assets and versions in the same struct. I got around that by creating a `FindVersion` function so the regular `Find` function can always return an asset. The `FindVersion` function is only used below in this file.

earth_enterprise/src/fusion/autoingest/AssetGenDaemonH.pl
```diff
@@ -283,6 +275,20 @@ ()
 
 print $fh <<EOF;
 
+// Convenience class that ties together related types
```
tst-lsavoie, line 278: This struct puts all the classes that relate to a specific asset type into one place where it's easy to reference several of them at once. It makes the Asset Factory functions much cleaner, but it requires a lot of changes to use it.

tst-jlarocco: This makes the template parameters a LOT cleaner.

tst-jlarocco:
I didn't test, but the code looks good to me.

I don't see where the bug was in ReuseOrMakeAndUpdate... It looks nearly identical.

PR #1740, opened 2020-05-11
Consolidate all file system accesses in mod_fdb to use a single class

tst-jkent:


earth_enterprise/src/common/geFilePool.cpp
```diff
@@ -20,17 +20,65 @@
...
+
+bool POSIXFileAccessor::PwriteAll(AbstractFileIdentifier* pID, const void* buffer, size_t size, off64_t offset) {
+  return khPwriteAll(pID->getAsFD(), buffer, size, offset);
+}
+
+// ****************************************************************************
+// ***  FileIdentifierFactory
+// ****************************************************************************
+namespace FileIdentifierFactory {
+  static AbstractFileIdentifier* getIdentifier(int i) {
```
tst-sbayard, line 50: Probably best to have this return `shared_ptr<AbstractFileIdentifier>` instead of `AbstractFileIdentifier*`. Then return `std::make_shared<POSIXIdentifier>(i)`.

earth_enterprise/src/common/geFilePool.cpp
```diff
@@ -20,17 +20,65 @@
...
+    return new POSIXIdentifier{i};
+  }
+}
+
+// ****************************************************************************
+// ***  FileAccessorFactory
+// ****************************************************************************
+namespace FileAccessorFactory {
+  static AbstractFileAccessor* getAccessor(const std::string &fname) {
```
tst-sbayard, line 59: Similar to the suggestion above, it would be better to have the return type be `shared_ptr<AbstractFileAccessor>` nd return `std::make_shared<POSIXFileAccessor>()`.

Same for the `getAccessor` function below.

earth_enterprise/src/common/geFilePool.cpp
```diff
@@ -20,17 +20,65 @@
...
+    return new POSIXFileAccessor();
+  }
+
+  static AbstractFileAccessor* getAccessor(AbstractFileIdentifier* aID) {
+    if ( typeid(*aID) == typeid(POSIXIdentifier) ) {
+      return new POSIXFileAccessor();
+    }
+    return NULL;
```
tst-sbayard, line 67: Better to return `nullptr` instead of `NULL`.

earth_enterprise/src/common/geFilePoolImpl.h
```diff
@@ -20,18 +20,34 @@
...
+  POSIXIdentifier(int i = -1) : fd{ i } {}
+  bool isValid() override { return fd != -1; }
+  void invalidate() override { fd = -1; }
+  int getAsFD() override { return fd; }
+};
 
 class FileReservationImpl : public khMTRefCounter {
   bool isWriter;
-  int fd;
+  AbstractFileIdentifier* fid;
```
tst-sbayard, line 45: In coordination with the `shared_ptr` changes suggested above, this should be `std::shared_ptr<AbstractFileIdentifier> fid;`

tst-lsavoie:
I'm surprised that these are the only changes required when serving to clients. There are plenty of files under published_dbs that aren't part of a file bundle, but maybe they don't go to clients? I would think the dbroot files do - maybe I'm wrong about that.

earth_enterprise/src/common/geFilePool.cpp
```diff
@@ -20,17 +20,65 @@
...
+    return std::make_shared<POSIXFileIdentifier>(i);
+  }
+}
+
+// ****************************************************************************
+// ***  FileAccessorFactory
+// ****************************************************************************
+namespace FileAccessorFactory {
+  static std::shared_ptr<AbstractFileAccessor> getAccessor(const std::string &fname) {
+    return std::make_shared<POSIXFileAccessor>();
```
tst-lsavoie, line 60: I'm not sure if creating a new file accessor every time is the right approach. That could result in a lot of memory allocation and clean up. It might be cleaner to treat this more like a singleton where it always returns the same file accessor every time.

earth_enterprise/src/common/geFilePool.cpp
```diff
@@ -20,17 +20,65 @@
...
 bool FileReservationImpl::UnlockAndClose_(geFilePool &pool) {
   // run when pool.mutex IS locked
-  if (fd != -1) {
+  if (fid->isValid()) {
     int result = 0;
     {
       khUnlockGuard unlock(pool.mutex);
-      result = isWriter ? khFsyncAndClose(fd) : khClose(fd);
-      fd = -1;
+      result = isWriter ? FileAccessorFactory::getAccessor(Fid())->FsyncAndClose(Fid()) : FileAccessorFactory::getAccessor(Fid())->Close(Fid());
```
tst-lsavoie, line 80: Minor: you could make this cleaner by calling `getAccessor` once (maybe in the constructor) and storing a pointer to it.

earth_enterprise/src/common/geFilePoolImpl.h
```diff
@@ -20,18 +20,35 @@
...
+};
 
 class FileReservationImpl : public khMTRefCounter {
   bool isWriter;
-  int fd;
+  std::shared_ptr<AbstractFileIdentifier> fid;
  public:
-  int Fd(void) const { return fd; }
-  FileReservationImpl(void) : isWriter(false), fd(-1) { }
+  std::shared_ptr<AbstractFileIdentifier> Fid(void) const { return fid; }
```
tst-lsavoie, line 48: You could probably use a `unique_ptr` here instead of a `shared_ptr` since `unique_ptr`s are cheaper. The different types of pointers represent different types of ownership (shared or unique/single, unsurprisingly). Here, ownership means who is supposed to delete the pointer.

- `shared_ptrs` are for when multiple objects own the same pointer. We use this in the cache - an item cannot be deleted until the cache and everything that has pulled the item out of the cache are done with it. Whoever has the last pointer is the one that deletes it.
- `unique_ptrs` are for when a single object owns the pointer and will eventually delete it. It looks like this example fits that case. Other objects might use the `fid` but this object owns it.

You can use `shared_ptrs` where `unique_ptrs` would work but it makes the code less efficient because it has to track how many references there are to the pointer.

In this case, I think you could change it to a `unique_ptr`. You'd also change the `Fid` function to return a raw pointer (`fid.get`) rather than the `unique_ptr`. Returning a `unique_ptr` means "I'm giving up ownership of this object and you now own it". Returning a raw pointer means "I still own this object but you can use it for a little while". It seems like the second case is what's happening here.

Actually, all of the `shared_ptrs` in this PR can possibly be changed to `unique_ptrs`, though that might require more refactoring.

There's more discussion about smart pointers and ownership here: https://www.modernescpp.com/index.php/c-core-guidelines-rules-to-smart-pointers.

earth_enterprise/src/common/geFilePool.cpp
```diff
@@ -20,17 +20,65 @@
...
+bool POSIXFileAccessor::PwriteAll(std::shared_ptr<AbstractFileIdentifier> pID, const void* buffer, size_t size, off64_t offset) {
+  return khPwriteAll(pID->getAsFD(), buffer, size, offset);
+}
+
+// ****************************************************************************
+// ***  FileIdentifierFactory
+// ****************************************************************************
+namespace FileIdentifierFactory {
+  static std::shared_ptr<AbstractFileIdentifier> getIdentifier(int i) {
+    return std::make_shared<POSIXFileIdentifier>(i);
```
tst-lsavoie, line 51: Minor: it might be more efficient to simply return the POSIXFileIdentifier rather than a pointer to it, i.e.,
```
return POSIXFileIdentifier(i);
```
That trades a copy for a bunch of pointer accesses, which I suspect is more efficient. The compiler may even be able to optimize out the copy.

tst-lsavoie:
Here are the things I think are missing:

- fusiondbservice.cpp: call to khReadStringFromFile (as you pointed out)
- serverdbReader.cpp, line 212: khExists
- serverdb.cpp: several calls to fopen, khFILECloser, etc

That said, I like what you've got so far. I'd be OK with finishing this PR and then creating a new one to handle the other areas.

earth_enterprise/src/common/geFilePool.cpp
tst-sbayard: We discussed that briefly. We weren't sure about the implications if multiple files were being accessed by the system at one time.

earth_enterprise/src/common/geFilePool.cpp
```diff
@@ -49,16 +97,17 @@ bool FileReservationImpl::UnlockAndOpen_(geFilePool &pool,
                                  // our invariant numFdsUsed_ < maxNumFds
   {
     khUnlockGuard unlock(pool.mutex);
-    fd = khOpen(fname, flags, createMask);
+    pool.aFA = FileAccessorFactory::getAccessor(fname);
```
tst-sbayard, line 100: I'm not sure about having `aFA` as a member of `geFilePool`. Each time a file is opened (via the call to `FileReservationImpl::UnlockAndOpen_`, `pool.aFA` will get reassigned. Which _may_ not cause an error, but isn't really good practice. 
I think it would be better to either:
1. Keep `aFA` as a member of the file pool and only assign it if it isn't currently assigned to anything.
2. Get rid of `aFA` in `geFilePool` and just locally use the return value of `FileAccessorFactory::getAccessor` and not keep it around.

This also gets into the topic of whether to try to re-use a single `AbstractFileAccessor` or return a new one each time. That probably warrants further discussion.

earth_enterprise/src/common/geFilePoolImpl.h
```diff
@@ -20,18 +20,49 @@
...
 
 class FileReservationImpl : public khMTRefCounter {
   bool isWriter;
-  int fd;
+  std::unique_ptr<AbstractFileAccessor> aFA;
  public:
-  int Fd(void) const { return fd; }
-  FileReservationImpl(void) : isWriter(false), fd(-1) { }
+  AbstractFileAccessor *AFA(void) const { return aFA.get(); }
+  FileReservationImpl(void) : isWriter(false) { }
```
tst-lsavoie, line 63: It would be a good idea to create an empty POSIXFileAccessor here that is marked invalid and doesn't point to a file. That way you are guaranteed that you can dereference `aFA` without causing a memory error, even before you open the file.

tst-jkent: I'm not sure I'm following. This is meant to return the abstractfileaccessor member of filereservationimpl, which should already be instantiated as one of the derived classes.

tst-sbayard: aFA doesn't get assigned until `FileReservationImpl::UnlockAndOpen_` is called, so the pointer returned by `aFA.get()` would be null, I believe. I think he's saying that you may want to set it to an initial value in the constructor. 
But to do that, it would have to be knowledgeable of `POSIXFileAccessor`, not just `AbstractFileAccessor`. Just brainstorming some options:
- Go ahead and create an empty `POSIXFileAccessor`, and just don't worry about it for now.
- Have a new type of accessor called `InvalidFileAccessor` that gets returned by a call to `getAccessor` if you pass in an empty file path

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,35 @@
...
+  return khPreadAll(this->getAsFD(), buffer, size, offset);
+}
+
+bool POSIXFileAccessor::PwriteAll(const void* buffer, size_t size, off64_t offset) {
+  return khPwriteAll(this->getAsFD(), buffer, size, offset);
+}
+
+std::unique_ptr<AbstractFileAccessor> AbstractFileAccessor::getAccessor(const std::string &fname, int flags, mode_t createMask) {
+  //notify(NFY_WARN, "Filename: %s", fname.c_str());
+  if (fname.rfind("/gevol", 0) == 0 || fname.rfind("/tmp", 0) == 0) {
```
tst-sbayard, line 28: I think this will miss cases where the asset root or publish root are not under `/gevol`. That's the default, but users can specify different asset and publish roots. I _think_ it might be enough just to see if the path starts with `/`.
@tst-lsavoie Thoughts?

tst-lsavoie: I'd just use a `POSIXFileAccessor` here by default. We can worry about picking the right accessor when we implement the plugin capability.

earth_enterprise/src/common/serverdb/serverdb.cpp
```diff
@@ -93,35 +93,35 @@ bool ServerdbConfig::Load(std::string config_file) {
...
+  aFA->Open(config_file.c_str(), "w");
+  if (!aFA->isValid()) {
     notify(NFY_WARN, "Unable to save ServerDB config file %s",
            config_file.c_str());
     return false;
   }
-  khFILECloser closer(fp);
 
-  fprintf(fp, "%s%s%d\n", kDbType.c_str(), kDelim.c_str(), db_type);
+  fprintf(aFA->getFP(), "%s%s%d\n", kDbType.c_str(), kDelim.c_str(), db_type);
```
tst-sbayard, line 104: The `fprintf` calls are directly modifying disk files, so we probably want to make an equivalent method in `FileAccessor` and call that instead.  That would probably let us get rid of the `getFP()` method entirely as well.

earth_enterprise/src/common/FileAccessor.h
```diff
@@ -0,0 +1,54 @@
...
+  virtual bool PwriteAll(const void* buffer, size_t size, off64_t offset) = 0;
+  virtual bool ReadStringFromFile(const std::string &filename, std::string &str, std::uint64_t limit = 0) = 0;
+  virtual bool Exists(const std::string &filename) = 0;
+  virtual int feof() = 0;
+  virtual void fgets(char *buf, int bufsize) = 0;
+  virtual void fprintf(const char * __restrict format, ...) = 0;
+};
+
+class POSIXFileAccessor: public AbstractFileAccessor {
+  int fileDescriptor;
```
tst-lsavoie, line 25: I don't love that there are two different ways to refer to a file. This is likely to cause trouble in the future if someone opens a file with a file descriptor and then calls a function that uses a file pointer. Is there any way we can combine them. Normally I'd suggest switching to all `FILE *`, but with the kh* functions taking file descriptors maybe it would be easier to convert the function that use a file pointer to use a file descriptor instead.

tst-lsavoie: On the flip side, I'm not sure if file descriptors support eof. Still, it would be best to avoid this duplication, or at least add a bunch of asserts that the user is calling the right function (so at the beginning of a function that uses a file descriptor, assert that the file descriptor is valid. Similar for file pointers).

earth_enterprise/src/common/FileAccessor.h
```diff
@@ -0,0 +1,54 @@
...
+  FILE* fp;
+public:
+  POSIXFileAccessor(int i = -1, FILE* f = NULL) : fileDescriptor{ i }, fp{ f } {}
+  ~POSIXFileAccessor(void) {
```
tst-lsavoie, line 29: You need to add a virtual destructor in the parent class (`virtual ~AbstractFileAccessor() = default;`). Otherwise, you could run into a problem like this:
```
AbstractFileAccessor * fa = new POSIXFileAccessor();
fa.Open(filename);
// do stuff
delete fa;
```
The last line will call the AbstractFileAccessor destructor because that's the kind of pointer you have. If the destructor is not virtual, it will NOT call the POSIXFileAccessor destructor, so the file will never be closed.

earth_enterprise/src/common/FileAccessor.h
tst-sbayard: I was thinking about this as well. One thought would be to only use the file descriptor by default. But if a client calls one of the methods like `fgets` or `fprintf` that need a file pointer, you could call `fdopen` internally to get a `FILE*` from the file descriptor. In that case, you could store the `FILE*` in a member variable that normally doesn't get used. And closing one will close the other (or at least closing the `FILE*` will close the file descriptor as well).

tst-sbayard: Also, in that implementation, you may be able to get rid of the second `Open` method: `virtual void Open(const char *fname, const char *mode) = 0;`

tst-jkent: As of now, fileDescriptor is just an int. Are you suggesting making it it's own class?

tst-sbayard: No, I'm just talking about making POSIXFileAccessor a bit smarter. 

1. Get rid of the second `Open` method and just add an optional parameter, `mode`, to the first one.
2. Always initialize the `fp` member to `nullptr` in the constructor
3. In the member methods that use the `fp` member, just make sure it's initialized as follows.
`void POSIXFileAccessor::fgets(char *buf, int bufsize) {
 
  VerifyFileStream();
 ::fgets(buf, bufsize, fp);
}`

With `VerifyFileStream` defined something like
`inline void POSIXFileAccessor::VerifyFileStream() {
	if (!fp)
		fp = fdopen(fileDescriptor, mode);
}`

I might also modify the `Close` method just slightly to check the `fp` first. If you call `fclose`, I know it closes the file descriptor as well. I don't know if the reverse is true for `close`.

`int POSIXFileAccessor::Close() {
  int result = 0;
  if (fp) {
    result = ::fclose(fp);
  }
  else {
    result = khClose(fileDescriptor);
  }
  this->invalidate();
  return result;
}`

tst-sbayard: I _think_ that should work, but I haven't tried to build it with those changes.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,64 @@
...
+bool POSIXFileAccessor::ReadStringFromFile(const std::string &filename, std::string &str, std::uint64_t limit) {
+  return khReadStringFromFile(filename, str, limit);
+}
+
+bool POSIXFileAccessor::Exists(const std::string &filename) {
+  return khExists(filename);
+}
+
+int POSIXFileAccessor::fgets(void *buf, size_t bufsize) {
+  return read(fileDescriptor, buf, bufsize);
```
tst-sbayard, line 56: I don't think this is going to do exactly the same thing as `fgets`. `fgets` stops reading if it encounters a newline character. `read` will just read `bufsize` bytes, stopping short only if it hits the end of the file.

tst-lsavoie: I can't find a similar function that works on file descriptors. Your best bet may be to read the entire file into a string, create a file pointer from the stream, and then run fgets on that pointer. The downside is that we'll be holding the whole file in memory, but from my investigation this file is typically only a few lines. You've already got ReadStringFromFile to read the file into a string, and `fmemopen` will create a file pointer from the string. Then you can use `fgets` on the string as if it was a file.

tst-sbayard: The only place right now that uses this `fgets` method is reading the whole file and processing each line. What if we get rid of the `fgets` method here, and just replace it with a method that reads the entire contents of a file and returns a vector of strings containing all the lines?

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,82 @@
...
+bool POSIXFileAccessor::Exists(const std::string &filename) {
+  return khExists(filename);
+}
+
+bool POSIXFileAccessor::GetLinesFromFile(std::vector<std::string> &lines, const size_t bufsize) {
+  char buf[bufsize];
+  buf[0] = 0;
+
+  try {
+    read(fileDescriptor, buf, bufsize);
```
tst-sbayard, line 60: Doing a single read like this will only work if the file length is less than `bufsize`, which isn't guaranteed. You should probably replace this cal with a loop, something like this:
```
std::string content;
ssize_t bytesRead = read(fileDescriptor, buf, bufsize -1);
while (bytesRead > 0) {
	buf[bytesRead] = 0;
	content += buf;	
	bytesRead = read(fileDescriptor, buf, bufsize -1);
}
```
The rest of the function can stay like it is, I think. Except `std::string content(buf);` can be remove below since it's declared in this snippet.

Disclaimer: I haven't tested that, so there could be errors in it.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+    else {
+      result = khOpenForRead(fname);
+    }
+  }
+  else {
+    result = khOpen(fname, flags, createMask);
+  }
+  fileDescriptor = result;
+  return result;
```
tst-lsavoie, line 24: Minor: do we need to return the file descriptor? It would be nice to keep it encapsulated in this class as much as possible.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,88 @@
...
+  char buf[bufsize];
+  buf[0] = 0;
+  std::string content;
+
+  try {
+    ssize_t bytesRead = read(fileDescriptor, buf, bufsize -1);
+
+    while (bytesRead > 0) {
+      buf[bytesRead] = 0;
+	    content += buf;
```
tst-lsavoie, line 65: Inconsistent indentation.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  catch (...) {
+    return false;
+  }
+}
+
+void POSIXFileAccessor::fprintf(const char *format, ...) {
+  va_list arguments;
+  va_start(arguments, format);
+  ::dprintf(fileDescriptor, format, arguments);
```
tst-lsavoie, line 85: I think you need `vdprintf` here since you're passing a `va_list` instead of a series of arguments.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,88 @@
...
+bool POSIXFileAccessor::ReadStringFromFile(const std::string &filename, std::string &str, std::uint64_t limit) {
+  return khReadStringFromFile(filename, str, limit);
+}
+
+bool POSIXFileAccessor::Exists(const std::string &filename) {
+  return khExists(filename);
+}
+
+bool POSIXFileAccessor::GetLinesFromFile(std::vector<std::string> &lines, const size_t bufsize) {
+  char buf[bufsize];
```
tst-lsavoie, line 55: This function would be a good candidate for a unit test.

tst-jkent: I was thinking the same. I'll write something up.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,87 @@
...
+  buf[0] = 0;
+  std::string content;
```
tst-lsavoie, line 57: Minor: it would probably be more efficient to use a `stringstream` here because they're designed for strings that change a lot. I doubt it will make much difference, though, since this is reading fairly short files.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  try {
+    ssize_t bytesRead = read(fileDescriptor, buf, bufsize -1);
+
+    while (bytesRead > 0) {
```
tst-lsavoie, line 62: It just occurred to me that you could replace this whole loop with `ReadStringFromFile`. That would make things simpler.

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,87 @@
...
+      buf[bytesRead] = 0;
+      content += buf;
+      bytesRead = read(fileDescriptor, buf, bufsize -1);
+    }
+
+    TokenizeString(content, lines, "\n");
+
+    for (auto it = std::begin(lines); it != std::end(lines); ++it) {
```
tst-lsavoie, line 70: I think you want `auto & it` here so that you're cleaning the string, not a copy of it.

earth_enterprise/src/common/FileAccessor.cpp
tst-jkent: khReadStringFromFile does have open call, which brings back the additional open call. Maybe we just call ReadStringFromFile here and remove the FileAccessor::Open call before this?

tst-lsavoie: Good point. I'd be ok with bringing back the open here and removing it from `ServerdbConfig::Load`. You'll have to make sure you check the return value of `ReadStringFromFile` and handle errors appropriately. @tst-sbayard might have a different opinion, though.

earth_enterprise/src/common/fileaccessor_unittest.cpp
```diff
@@ -0,0 +1,22 @@
+#include <gtest/gtest.h>
+#include "FileAccessor.h"
+
+using namespace std;
+using namespace testing;
+
+TEST(FileAccessorTest, validate_lines_from_file) {
+    string filename = "/tmp/TestFile.txt";
```
tst-sbayard, line 8: Is `/tmp/TestFile.txt` getting created as part of the test?

earth_enterprise/src/common/fileaccessor_unittest.cpp
```diff
@@ -0,0 +1,22 @@
...
+    vector<string> lines;
+    unique_ptr<AbstractFileAccessor> aFA = AbstractFileAccessor::getAccessor(filename);
+
+    aFA->Open(filename, "w");
+    assert (aFA->getFD() != -1);
+    aFA->Close();
```
tst-sbayard, line 14: It looks like this test is simply opening and closing the file. Is it supposed to be calling `GetLinesFromFile`?

earth_enterprise/src/common/geFilePoolImpl.h
```diff
@@ -20,18 +20,18 @@
...
-
 class FileReservationImpl : public khMTRefCounter {
   bool isWriter;
-  int fd;
+  std::unique_ptr<AbstractFileAccessor> aFA;
  public:
-  int Fd(void) const { return fd; }
-  FileReservationImpl(void) : isWriter(false), fd(-1) { }
+  AbstractFileAccessor *AFA(void) const { return aFA.get(); }
+  FileReservationImpl(void) : isWriter(false), aFA(std::unique_ptr<POSIXFileAccessor>(new POSIXFileAccessor())) { }
```
tst-lsavoie, line 32: Minor: I think you can simplify the initialization of `aFA` to just:
```
aFA(new POSIXFileAccessor())
```

earth_enterprise/src/common/FileAccessor.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "FileAccessor.h"
+
+std::unique_ptr<AbstractFileAccessor> AbstractFileAccessor::getAccessor(const std::string &fname) {
+  return std::unique_ptr<POSIXFileAccessor>(new POSIXFileAccessor());
+}
+
+// ****************************************************************************
+// ***  POSIXFileAccessor
+// ****************************************************************************
+void POSIXFileAccessor::Open(const std::string &fname, const char *mode, int flags, mode_t createMask) {
```
tst-lsavoie, line 24: Minor: I suspect there's a way to combine `mode` and `flags`, but we can save that for another time.

fwg3 (commented):
#

PR #1748, opened 2020-05-26
WIP: migrations from master to qt4 migration branch

tst-dsugar:
This is a rolling update for bringing greater concurrence between qt4 migration and master, and further commits will certainly be added.  To make it easier to track with master, this perhaps should not be squashed when finally merged to the migration branch.


earth_enterprise/src/common/mttypes/Semaphore.h
```diff
@@ -25,15 +26,15 @@ namespace mttypes {
...
-  void Release(uint32 num = 1);
+  void Acquire(std::uint32_t num = 1);
+  void Release(std::uint32_t num = 1);
 
  private:
   bool LockedOKToAcquire(void) const;
 
-  uint32 avail_;
+  std::uint32_t avail_;
   bool   done_;
```
tst-basilhuffman, line 38: unrelated, but is `done_` used anywhere?

earth_enterprise/src/common/packet.cpp
```diff
@@ -78,14 +79,14 @@ etPointerTranslator * etDrawablePacket::packetPointerTranslator = NULL;
 int etPattern::hash(const char* pattern)
 {
   const int M = 26;
```
tst-basilhuffman, line 81: also unrelated, but do we know what this is?

would it be worthwhile to make primes a std::array?

earth_enterprise/src/common/packet.h
```diff
@@ -425,18 +427,18 @@ struct etPointerTranslator
 {
   etPointerCypher * pointerList;
   int pointerListSize;
-    
+
```
tst-basilhuffman, line 430: unrelated: a big chunk of this code needs to be ported over to c++11, at some point we should think about creating an issue for this

earth_enterprise/src/common/packet.h
```diff
@@ -449,14 +451,14 @@ struct etPointerTranslator
...
   int          texId_deprecated;
   ushort       numPt;
   ushort       bitFlags;
 #ifdef VECTOR_COMPRESSION
-  uint32 localPt;
+  std::uint32_t localPt;
 #else
-  uint32 localPt;
+  std::uint32_t localPt;
 #endif
```
tst-basilhuffman, line 462: can probably just remove the guards here since it looks like `localPt` will always be declared

earth_enterprise/src/common/packet.h
```diff
@@ -477,9 +479,9 @@ struct etStreetPacketData
   ushort       bitFlags;
 #ifdef FORCE_BASE32
 #ifdef VECTOR_COMPRESSION
-  uint32 localPt_OFFSET;
+  std::uint32_t localPt_OFFSET;
 #else
-  uint32 localPt_OFFSET;
+  std::uint32_t localPt_OFFSET;
```
tst-basilhuffman, line 484: same, `localPt_OFFSET` looks to be always declared the same type

earth_enterprise/src/common/packet.h
```diff
@@ -632,15 +634,15 @@ struct etStreetPacket : public etDataPacket
...
   ushort       numPt;
   ushort       bitFlags;
-    
+
 #ifdef VECTOR_COMPRESSION
-  uint32 localPt;
+  std::uint32_t localPt;
 #else
-  uint32 localPt;
+  std::uint32_t localPt;
```
tst-basilhuffman, line 645: same, can remove guards here

earth_enterprise/src/common/packet.h
```diff
@@ -652,19 +654,19 @@ struct etPolyLinePacketData_BASE32
...
   ushort       bitFlags;
-    
+
 #ifdef FORCE_BASE32
 #ifdef VECTOR_COMPRESSION
-  uint32 localPt_OFFSET;
+  std::uint32_t localPt_OFFSET;
 #else
-  uint32 localPt_OFFSET;
+  std::uint32_t localPt_OFFSET;
```
tst-basilhuffman, line 669: same, guards can be removed within `FORCE_BASE32` block

earth_enterprise/src/common/packet.h
```diff
@@ -812,15 +814,15 @@ struct etPolyLinePacket : public etDataPacket
...
   int          texId;
   ushort       numPt;
   ushort       bitFlags;
 
 #ifdef VECTOR_COMPRESSION
-  uint32 localPt;
+  std::uint32_t localPt;
 #else
-  uint32 localPt;
+  std::uint32_t localPt;
```
tst-basilhuffman, line 825: last comment here, it looks like in just about every situation where the `FORCE_BASE32` and `VECTOR_COMPRESSION` macros are defined, `localPt` will always be defined as a 32 bit integer and the guards can be removed. unrelated to this pr, but i feel this file should be looked at, at some point down the road

PR #1750, opened 2020-05-29
1749 Enable Serving WMS for 3D Databases

### LINKED ISSUE #1749: Allow users to serve WMS of 3D files from server UI
It was previously believed that serving WMS was unusable to obtain tiles from 3D portable files. However, it was recently found that serving WMS on a 3D portable file will still send 2D tile information to WMS clients.

3D Databases can be published with --serve_wms, but there should be an option in the UI for doing it as well.

tst-nrichman:
All of the back-end work to publish databases with the proper WMS serve is already in place. This change simply makes the button visible in the geserver user interface.

Fixes #1749

### DIFF
--- a/earth_enterprise/src/server/wsgi/wms/ogc/xml/v130/capabilities_wms.py
+++ b/earth_enterprise/src/server/wsgi/wms/ogc/xml/v130/capabilities_wms.py
@@ -2335,12 +2335,12 @@ class Layer(GeneratedsSuper):
     subclass = None
     superclass = None
     def __init__(self, opaque=False, cascaded=None, fixedHeight=None, fixedWidth=None, noSubsets=False, queryable=False, Name=None, Title=None, Abstract=None, KeywordList=None, CRS=None, EX_GeographicBoundingBox=None, BoundingBox=None, Dimension=None, Attribution=None, AuthorityURL=None, Identifier=None, MetadataURL=None, DataURL=None, FeatureListURL=None, Style=None, MinScaleDenominator=None, MaxScaleDenominator=None, Layer=None):
-        self.opaque = _cast(bool, opaque)
+        self.opaque = _cast(int, opaque)
         self.cascaded = _cast(int, cascaded)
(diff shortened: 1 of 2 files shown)


earth_enterprise/src/server/wsgi/wms/ogc/xml/v130/capabilities_wms.py
```diff
@@ -2335,12 +2335,12 @@ class Layer(GeneratedsSuper):
     subclass = None
     superclass = None
     def __init__(self, opaque=False, cascaded=None, fixedHeight=None, fixedWidth=None, noSubsets=False, queryable=False, Name=None, Title=None, Abstract=None, KeywordList=None, CRS=None, EX_GeographicBoundingBox=None, BoundingBox=None, Dimension=None, Attribution=None, AuthorityURL=None, Identifier=None, MetadataURL=None, DataURL=None, FeatureListURL=None, Style=None, MinScaleDenominator=None, MaxScaleDenominator=None, Layer=None):
-        self.opaque = _cast(bool, opaque)
+        self.opaque = _cast(int, opaque)
```
tst-blaird, line 2338: What's the reason for changing the cast of opaque, noSubsets, and queryable to int?  Looking at the file they are set to bool values of True/False.

tst-nrichman: Trying to integrate our WMS serving with OWSLib. The library expects our responses to return integers. So it's fine to set true/false throughout, but the actual value will need the cast.

tst-blaird: ok, that works.  Thanks for the explanation.

tst-ccamp:
Make sure to test this with portable globes. WMS will not work for them.

tst-hrajamoney:
Tested WMS on 3d globes and 2d maps and 2d portables. Code change looks good.

PR #1752, opened 2020-06-01
EL8 Changes

tst-cjeffries:
Contains changes needed to get fusion working on el8


earth_enterprise/src/common/compressor.cpp
```diff
@@ -15,12 +15,19 @@
...
 #include <common/khStringUtils.h>
 
+#ifdef USING_LIBPNG16
+#include <libpng16/png.h>
+#endif
+
+#ifdef USING_LIBPNG15
+#include <libpng15/png.h>
+#endif
+
```
tst-basilhuffman, line 30: even though i guess we aren't supporting it, it might be better to have something like 
`#if defined USING_LIBPNG16`
`#include <libpng16/png.h>`
`#elif defined USING_LIBPNG15`
`#include <libpng15/png.h>`
`#else`
`#include <libpng12/png.h>`
`#endif`

just something to consider, not a showstopper

earth_enterprise/src/fusion/rasterfuse/vipm/procmesh.cpp
```diff
@@ -22,11 +22,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 
-
-#ifndef min
-#define min(a,b) ((a) < (b) ? (a) : (b))
-#endif
-
```
thisisbasil, line 29: do we know where or if this was used?

earth_enterprise/src/SConstruct
```diff
@@ -176,7 +176,10 @@ el_ver = ''
 if os.path.isfile('/etc/redhat-release') and not os.path.isfile('/etc/os-release'):
   el_ver = 'el6'
 else:
-  el_ver = 'el7'
+  if 'el8' in os.uname()[2]:
+    el_ver = 'el8'
+  else:
+    el_ver = 'el7'
```
thisisbasil, line 182: might be worthwhile to make `png_ver` an env variable so we can account for Ubuntu later on

earth_enterprise/src/common/compressor.cpp
tst-cjeffries: Thats a fair point. Im thinking we will be unable to support 1.2 due to qt4, but using it as a fail safe isn't a bad idea.

PR #1753, opened 2020-06-02
QT4 EL8 Changes

tst-cjeffries:
Same as previous PR, but with squashed commits.


earth_enterprise/src/SConstruct
```diff
@@ -277,12 +280,29 @@ else:
...
+                '-Wno-error=class-memaccess',
+                '-Wno-error=memset-elt-size',
+                '-Wno-error=address',
+                '-Wno-error=unused-local-typedefs']
+
+# Determine libpng version based on el version
+if el_ver == "el8":
+  cppflags += ['-DUSING_LIBPNG16']
+else:
+  cppflags += ['-DUSING_LIBPNG15']
```
dyfet, line 304: So this would include Ubuntu....

earth_enterprise/src/common/gedbroot/SConscript
```diff
@@ -38,6 +38,11 @@ gedbroot = env.sharedLib('gedbroot',
                                'QtCore'
                                ])
 
+if env['el_ver'] == 'el8':
+  png_ver = 'png16'
+else:
+  png_ver = 'png15'
```
tst-dsugar, line 44: So this would cover ubuntu 18.04/16.04 too...?

tst-cjeffries: Assuming that it uses png15. We may need a slight adjustment in logic if we go with 16. Something along the lines of `if not env['el_ver'] or env['el_ver'] == 'el8':`.

tst-basilhuffman: i think it would be cleaner to make `png_ver` an environment variable instead of `el_ver`

tst-basilhuffman:
looks good, just had that one comment

earth_enterprise/src/common/gedbroot/SConscript
tst-cjeffries: I agree with that. I thought of that after I had finished up.

PR #1757, opened 2020-06-04
Update readme to point to the download tutorial file

thomric2:
The readme had an out of date link. Instead of duplicating the amount of places a link could be just point people to the download script that contains the correct link.


thomric2:
@googlebot I signed it!

tst-lsavoie:
Good catch. Thanks for the PR.

earth_enterprise/tutorial/README.md
```diff
@@ -1 +1 @@
-The files in this directory are used to package and install the fusion tutorial source files. Due to size constraints they are not included in the repo. In order to build, first download the files [here](http://data.opengee.org/FusionTutorial-Full.tar.gz) and extract them directly into the `FusionTutorial` directory.
\ No newline at end of file
+The files in this directory are used to package and install the fusion tutorial source files. Due to size constraints they are not included in the repo. In order to build, first download the files using download_tutorial.sh and place them directly into the `FusionTutorial` directory.
```
tst-lsavoie, line 1: Minor request: would you add tick marks around `download_tutorial.sh`:
```
download the files using `download_tutorial.sh`
```
That will designate it as a command to run.

thomric2: Sounds good. Also since the script was one level deeper than this readme I added the folder before the script.

PR #1760, opened 2020-06-08
Fix Asset adding/saving paths in dialogs

tst-cjeffries:
Allows assets from outside the asset root to be selected by allowing navigation of other folders. Saving outside the asset root is is fixed as well. Also the QEvent ID used by the change directory event is now properly selected (and reserved), rather than just using the min custom event ID (1000). 

To test:

- Build and install.
- Save a resource in a subdir of the asset root.
- Add said resource to a project.

### DIFF
--- a/earth_enterprise/src/fusion/fusionui/AssetChooser.h
+++ b/earth_enterprise/src/fusion/fusionui/AssetChooser.h
@@ -72,7 +72,7 @@ class AssetChooser : public AssetChooserBase {
 
   // from QWidget
   virtual void keyPressEvent(QKeyEvent* e);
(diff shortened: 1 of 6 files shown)


earth_enterprise/src/fusion/fusionui/AssetChooser.h
```diff
@@ -72,7 +72,7 @@ class AssetChooser : public AssetChooserBase {
 
   // from QWidget
   virtual void keyPressEvent(QKeyEvent* e);
```
tst-dsugar, line 74: Since they are victuals, they probably should be made "final" to assure they match a base class virtual.

PR #1767, opened 2020-06-12
Qt4 push bug fix

tst-mswartz:


tst-cjeffries:
Tested on EL8. Verified that it no longer crashes.

PR #1770, opened 2020-06-17
fixed el8 flags bleeding into el7 builds

tst-dsugar:
This change broke el7 builds, because the flags do not apply to gcc 4.8.  I suspect this was done for ubuntu 18.04, but we really need a clean way to specify ubuntu 18.04 flags separate from 16.04 and earlier centos7/6.  Really these flags are by gcc version, and this is what should be tested for rather than the os version...


tst-basilhuffman:
i believe this has become OBE by 1808

PR #1771, opened 2020-06-17
removing index entries for GEE

tst-swanson:
Removing the few index entries for Open so they don't interfere with GEEP


tst-lsavoie:
This needs to be rebased on release_5.3.5. There are some changes intended for 5.3.6 in the PR.

tst-hrajamoney:
The rebase was done yesterday and here is the new PR: https://github.com/google/earthenterprise/pull/1776

PR #1773, opened 2020-06-22
Fix server existence check

### LINKED ISSUE #1772: Geecheck reports server is not installed in dual configuration
**Describe the bug**

If GEE Server is installed by itself on a machine (i.e., Fusion is not installed), geecheck will report that Server is not installed and will skip several tests.

**To Reproduce**

1. Install Open GEE Server on a test machine. Do NOT install Open GEE Fusion.
2. Log into the admin console, click the settings button, and click diagnostics.
3. Note that most of the diagnostics are skipped and the reason for skipping for several of them is "GEE Server is not installed"

**Expected behavior**

geecheck runs tests for GEE Server when GEE Server is installed.

**Desktop**

 - Server OS: Ubuntu 16
 - Client OS: Ubuntu 16
 - Browser: Chrome
 - Version: 5.3.5

**Additional context**

The issue is the following function in earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/geecheck_tests/common.py:
```
def IsGeeServerInstalled():
  """Check if GEE Server is installed."""
  gee_server_start_script = '/etc/init.d/gefusion'
  return os.path.exists(gee_server_start_script)
```
It determines whether GEE Server is installed by looking at the GEE Fusion init script.

tst-lsavoie:
Fixes #1772 

To test:
1. Install only Server on a machine (not Fusion)
1. Run the diagnostics
1. Verify that the only reason listed for skipping tests is because Fusion is not installed. No tests should be skipped because Server is not installed.

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/geecheck_tests/common.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/geecheck_tests/common.py
@@ -61,7 +61,7 @@ def IsFusionInstalled():
 
 def IsGeeServerInstalled():
   """Check if GEE Server is installed."""
-  gee_server_start_script = '/etc/init.d/gefusion'
+  gee_server_start_script = '/etc/init.d/geserver'
   return os.path.exists(gee_server_start_script)
 
 


tst-jkent:
Verified fix works.

PR #1779, opened 2020-07-01
1778 Fix Cutter SSL Crash

### LINKED ISSUE #1778: Databases published with SSL virtual hosts fail to cut
**Describe the bug**

When SSL is enabled and an SSL compliant virtual host is added, users are able to perform cuts on the published database. However, the cutting process fails around the end with an invalid certificate error:

 <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:618)>

**To Reproduce**

1. Install geserver, enable SSL, and add an SSL virtual host.
2. Publish a database to the SSL VH.
3. Enable cutter, and perform a cut on the DB.

**Expected behavior**

The cut should succeed, and the completed file should be ready for use.

tst-nrichman:
Fixes an issue where cutter runs into a certificate error when it's almost done. This is due to not passing the proper context in one location, and not setting a default context in another.

Fixes #1778

### DIFF
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/globe_cutter.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/globe_cutter.py
@@ -18,7 +18,8 @@
 """Module implementing core cutting functionality."""
 
 import json
-import urllib
+import ssl
+import urllib2
 
 from core import search_tab_template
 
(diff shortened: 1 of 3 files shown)


earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/core/globe_cutter.py
```diff
@@ -41,10 +43,13 @@ def GetSearchTabs(source):
...
-    if fp.getcode() == 200:
-      search_tabs = fp.read()
-    fp.close()
+    try:
+      fp = urllib2.urlopen(url, context=ssl._create_unverified_context())
+      if fp.getcode() == 200:
+          search_tabs = fp.read()
+      fp.close()
+    except:
+      pass
```
tst-blaird, line 52: We should print something here if an exception happens.

tst-nfarah:
do we have a github issue for this PR ? if yes please link it against the PR. If not i would create it and link it.

PR #1780, opened 2020-07-02
fix cutter definition problem

tst-basilhuffman:
not 100% sure why defusedxml is having issues but may need to drop back to standard xml parser


tst-jzmolek:
Don't we want this as part of the release? It should target release_5.3.5

PR #1787, opened 2020-07-08
new branch to retarget

tst-basilhuffman:


earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/globe_cutter_app.py
```diff
@@ -206,7 +209,7 @@ def SavePolygon(self, polygon):
       if polygon:
         # Check XML validity and standardize representation
         xml = etree2.ElementTree(etree.fromstring(polygon))
-        xml.write(fp, xml_declaration=True, encoding='UTF-8')
+        xml.write(fp, xml_declaration=True, encoding='UTF-8', default_namespace=DEFAULT_KML_NAMESPACE)
```
tst-dsugar, line 212: Where is the change to read the default namespace from the kml if it has one?

PR #1794, opened 2020-07-15
Fix imagery project transparency

### LINKED ISSUE #1785: Imagery Projects - Transparent areas appears as black
When I add a layer that is partially transparent (or its nodata values are masked), the built resource looks good in the resource preview but not in the final result.

**To Reproduce**

Build an imagery resource with nodata areas, define a mask, add it as a single resource to an imagery project (without blue marble), add to db (without any other project), build and publish.

**Expected behavior**

Areas of nodata should appear as transparent areas.

**Screenshots**

### With bug:
before-fix-cropped

### Without bug:

after-fix-cropped

after-fix-cropped2

**Desktop**

 - Server OS: Ubuntu 16.04 (on docker)
 - Client OS:  Ubuntu 18.04 
 - Browser: chrome
 - Version: 5.3.4

**Additional context**

I have fixed this in [this branch](https://github.com/google/earthenterprise/compare/master...duvifn:fix-alpha). If this looks good to you I would be happy to open a PR.

duvifn:
Fixes #1785.
I'm also a bit busy now, but I checked the `Allow edits by maintainers` checkbox, so if I'm unresponsive feel free to change whatever you want before merging.
Thank you very much

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
@@ -260,12 +260,10 @@ void PacketGenAssetVersionImplD::UpdateChildren(
            // if blendVer->GetOutputFilename(0) is empty, this Asset still 
             // doesn't has an output file, so the path doesn't exist and it's going 
             // to be built. So we mark it as exists
-            // TODO: ensure this is true
           if (blendVer->GetOutputFilename(0).empty() 
                   || khExists(cache_alpha_path)) {
             cached_blend_alpha = blendVer.Ref();
           }
-          cached_blend_alpha = blendVer.Ref();
           cached_blend = blendVer.Ref();
         }
         packetLevelConfig.insets[merge_idx] =
@@ -299,7 +297,6 @@ void PacketGenAssetVersionImplD::UpdateChildren(
         // if prevLevelVer->GetOutputFilename(0) is empty, this Asset still 
         // doesn't has an output file, so the path doesn't exist and it's going 
         // to be built. So we mark it as exists
-        // TODO: ensure this is true
         const bool cache_alpha_exist = 
             prevLevelVer->GetOutputFilename(0).empty() 
                 || khExists(cache_alpha_path);
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
@@ -255,12 +255,15 @@ void PacketGenAssetVersionImplD::UpdateChildren(
           // assets) that have no alpha blend cached.
           levInputs[inputs_idx+1] = blendVer.Ref();
           levInputVers[inputs_idx+1] = blendVer;
+          const std::string output_file_name = blendVer->GetOutputFilename(0);
           const std::string cache_alpha_path =
-              blendVer->GetOutputFilename(0) + "/cache_alpha.pack";
-           // if blendVer->GetOutputFilename(0) is empty, this Asset still 
-            // doesn't has an output file, so the path doesn't exist but it's going 
-            // to be built. So we mark it as exists
-          if (blendVer->GetOutputFilename(0).empty() 
+              output_file_name + "/cache_alpha.pack";
+          // if output_file_name is empty, this Asset still 
+          // doesn't has an output file (this is a new asset), so the path 
+          // doesn't exist. But since in our version alpha cache is going
+          // to be used (according to above comment), we assume that it's 
+          // going to be there when it would be needed.
+          if (output_file_name.empty() 
                   || khExists(cache_alpha_path)) {
             cached_blend_alpha = blendVer.Ref();
           }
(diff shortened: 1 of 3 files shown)


duvifn:
@googlebot I signed it!

earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
```diff
@@ -255,9 +255,16 @@ void PacketGenAssetVersionImplD::UpdateChildren(
...
           const std::string cache_alpha_path =
-              blendVer->GetOutputFilename(0) + "/cache_alpha.pack";
-          if (khExists(cache_alpha_path)) {
+              output_file_name + "/cache_alpha.pack";
+          // if output_file_name is empty, this Asset still 
+          // doesn't has an output file (this is a new asset), so the path 
+          // doesn't exist. But since in our version alpha cache is going
+          // to be used (according to above comment), we assume that it's 
+          // going to be there when it would be needed.
+          if (output_file_name.empty()
```
tst-lsavoie, line 266: I'm a bit concerned about this change. At this point, all the resources should be built, so if there is an alpha mask it should already exist. I don't know of a situation in which the alpha mask would appear after this point. I'm also worried that this will break some other situation, such as upgrades from 4.x. Can you provide a more detailed set of steps for reproducing this issue? I haven't been able to reproduce it yet, but if I could I might be able to suggest a safer fix.

duvifn: @tst-lsavoie thanks.
I hope I recall correctly, since the last time I touched this was 2 months ago.
I just want to be aligned on the terminology- this is an asset version of packetlevel asset (not resource), and if I recall correctly it's produced every time this resource or resources below it in the project are changed.
Unfortunately I didnt find the code for the 4.x version (the first version I managed to install from the git was 5.2 and it had this bug).
It's does seem that at this point the packetgen/packetlevel should already be built, but in fact the packetlevel (not only the alpha cache) doesn't exist at this stage on the disk and the asset doesn't has an output path. Maybe the task order for building process should be fixed.
Since this is a new asset version (otherwise the packetlevel  would have already been built and have an output path), which is now being built in version 5.x envireonment, it looked to me safe to continue even with 4.x projects, but this is the first time I'm getting into fusion source code so you probably will have a better fix.

I wrote more details about how to reproduce this issue on the issue itself.

tst-lsavoie:
I'm following the steps you posted in the issue and I'm still not able to reproduce this. Every time I add the clipped image as a new resource and build it it ends up being masked correctly. Is there any else you're doing that might be different? I've uploaded the image I'm using to the link below in case that's helpful.

https://drive.google.com/file/d/1IWjQvGDgGNRlGRVuRNImyfuAazGXN6xm/view?usp=sharing

The reason I'm being so careful with this is because we've been bitten by changes like this before - we'll fix something and then find out a couple of releases later that we also broke something else. This code is more or less unchanged since GEE was open sourced, so there's no one working on it that understands it completely.

duvifn:
The resource does masked properly for me, too. The built imagery project, however, doesn't reflect this, at least in my configuration.

duvifn:
Perhaps the following use case would be clearer:
Build two flat DBs:
Layer 1- Only with blue-marble
Layer 2- As I mentioned in the issue.
Then publish them with WMS support and load them one on top of the other in wms client (like qgis).
With this issue- Layer 1 can't be seen when layer 2 is on top of it (transparent areas appear as black ).
Without this issue transparent areas appear as transparent so you can see through them.

tst-lsavoie:
Ok, with those instructions I can reproduce it. I've never seen that use case before. I'll have to think about how to handle it.

duvifn:
> The reason I'm being so careful with this is because we've been bitten by changes like this before - we'll fix something and then find out a couple of releases later that we also broke something else. This code is more or less unchanged since GEE was open sourced, so there's no one working on it that understands it completely.

Ah, of course.
What I know (hope I understood & recall correctly) from debugging this:
Apparently alpha tiles got into the code in version 5.x. Until then imagery project contained only jpeg tiles, so no data areas appeared as black. The alpha channel is currently stored  as a png buffer in a different field in the protobuf, separately from the `rgb`  which are stored as jpeg. When client ask the server for a tile the server check if alpha channel exists and if so it composites and returns png file. Otherwise it returns jpeg.
I think it's done this way to preserve back compatibility but it also save a lot of storage space (jpeg is far more efficient than png).

Since we need to merge layers we have to take into account all insets (resources) that below current inset in the project. Calculating the exact  alpha value for all of the levels (X all the insets below) is a waste of time if we can use calculations from previous levels and layers, so it's get cached.
So, for one inset- we compute only the max level (based on resource's kip/kmp and layers below it) and loop over lower levels and define if they has cached alpha from higher levels ([and if so we put its path](https://github.com/google/earthenterprise/blob/665301c6510a4fde9ce0010d77df32c1bb55f90d/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src#L266-L278)). The problem is that at this point the packlevel assetver is not yet processed so it's actually doesn't exist on the disk, and also, its paths are not populated yet ([`output_file_name`](https://github.com/google/earthenterprise/blob/665301c6510a4fde9ce0010d77df32c1bb55f90d/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src#L266-278) is empty).
As a result, the maximum level's alpha is computed properly (since it's using resource's [kmp](https://github.com/google/earthenterprise/blob/665301c6510a4fde9ce0010d77df32c1bb55f90d/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src#L154) as the source of the alpha) but in lower levels it does't:  [this if](https://github.com/duvifn/earthenterprise/blob/665301c6510a4fde9ce0010d77df32c1bb55f90d/earth_enterprise/src/fusion/rasterfuse/RasterMerger.cpp#L206) is always `false` (since (`inset->cached_blend_reader` is null pointer) so fusion decides this is an opaque tile and doesn't compute its alpha.

tst-lsavoie:
Based on your comments and my code inspection it seems safe. Let me suggest an alternate implementation that seems a bit cleaner to me.

1. In [PacketGen.src](https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src#L262), on the two checks for cache_alpha.pack (currently lines 262 and 293), we don't check for the existence of the file. We just use `blendVer.Ref()` (or `prevLevelVer.Ref()`) unconditionally whether or not it exists.
1. In [RasterMerger.cpp](https://github.com/google/earthenterprise/blob/master/earth_enterprise/src/fusion/rasterfuse/RasterMerger.cpp#L163) we add an existence check before creating the `cached_blend_alpha_reader` on line 163. We'd also need to move the comment about 4.x to this point so devs know why the file might not exist, and we should add a log statement any time we decide not to create an alpha reader because the file doesn't exist, probably at the INFO level.

What do you think? Do you want to make those changes? I can do it as well, but I don't want to take your chance to get some commits in the codebase.

duvifn:
This indeed does look cleaner to me.

>  Do you want to make those changes? I can do it as well, but I don't want to take your chance to get some commits in the codebase.
Thanks but this is your solution and I don't want to take credit for it (but I appreciate it!).
I hope I will contribute in the future (I currently have some features, which I hope to contribute when I have the time to bring them to desired level).

tst-lsavoie:
Replaced by PR #1836 based on the discussion above.

PR #1803, opened 2020-07-21
Fix "geserver.service changed on disk" message after uninstalling Fusion or Server

tst-jkent:
Added systemctl restart calls in fusion and server uninstall scripts.

### DIFF
--- a/earth_enterprise/rpms/opengee-fusion/snippets/post-uninstall.sh
+++ b/earth_enterprise/rpms/opengee-fusion/snippets/post-uninstall.sh
@@ -32,5 +32,5 @@ remove_users_and_groups()
 # (See <https://wiki.debian.org/MaintainerScripts>.)
 if [ "$1" = "0" ] || [ "$1" = "purge" ] ; then
     remove_users_and_groups
-    if [ command -v systemctl ]; then systemctl daemon-reexec; fi
+    if [ `command -v systemctl` ]; then systemctl daemon-reexec; fi
 fi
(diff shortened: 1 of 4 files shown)


tst-lsavoie:
We should write an issue for this bug if there isn't one already.

earth_enterprise/src/installer/uninstall_fusion.sh
```diff
@@ -112,6 +112,7 @@ main_uninstall()
     remove_user
     remove_group
     show_final_success_message
+    systemctl daemon-reexec
```
tst-lsavoie, line 115: Minor: this should maybe come before the final success message in both cases.

tst-lsavoie:
I think we need to make a similar change for RPMs

tst-lsavoie:
CentOS 6/RHEL 6 don't have `systemctl`. You'll need to add an if statement around all 4 of these lines to make sure they aren't called when `systemctl` is not available. A simple `if [ command -v systemctl ]; then` should work. If systemctl is not available you don't need to do anything - this problem only exists on machines with `systemctl`.

earth_enterprise/rpms/opengee-fusion/snippets/post-uninstall.sh
```diff
@@ -32,4 +32,5 @@ remove_users_and_groups()
 # (See <https://wiki.debian.org/MaintainerScripts>.)
 if [ "$1" = "0" ] || [ "$1" = "purge" ] ; then
     remove_users_and_groups
+    if [ command -v systemctl ]; then systemctl daemon-reexec; fi
```
tst-lsavoie, line 35: Turns out you need backticks around `command -v systemctl`:

```
if [ `command -v systemctl` ]; then systemctl daemon-reexec; fi
```

Issue #1800, opened 2020-07-20
Upgrade from 5.3.4 to 5.3.5 leaves a non-working ExampleSearch


tst-eclamar:
During the 5.3.5 release we moved ExampleSearch to the opengee-extra RPM. However, in previous versions it was in the opengee-server RPM. Each search tab involves both an entry in the postgres database and python files on disk that implement the search tab. When we upgrade from 5.3.4 to 5.3.5 but don't install opengee-extra, the entry for ExampleSearch is not removed from the postgres database, but the python files that implement it are removed from disk, so the search shows up in the admin console but doesn't work.

The easiest way to implement this is probably to add something to the opengee-server pre-install script that removes ExampleSearch from the postgres database. This is somewhat awkward since it may be immediately re-added afterwards if the user is also installing opengee-extra, but it's probably the smoothest option.

It seems that users should be able to install opengee-extra and then uninstall it as a workaround, but that didn't work in my testing.

Update opengee-server RPM to remove ExampleSearch from Posgres databases, so it can no longer be selected when publishing databases or configuring searches through the Admin Console

Add a thorough help page explaining how to upgrade to 5.3.5+ and what will happen to ExampleSearch

Make sure that Databases that previously had used ExampleSearch are not broken (i.e., you can still view the globe even if example search doesn't work)

Make sure that installing the opengee-extras RPM will allow pre-published databases to work correctly, i.e. search results are returned.

Steps to Reproduce
Install Open GEE 5.3.4

Upgrade to Open GEE 5.3.5, but do not install opengee-extra

Open the Server admin page, click on "Search Tabs", and note that "ExampleSearch" is included in the list (it should not be there)

Publish a database using the ExampleSearch search tab.

Open the database in EC and try to do an example search (such as for "pacific heights"). No results are returned.

PR #1807, opened 2020-07-22
I1800 - Non-working examplesearch, take 2

tst-eclamar:
Fix the problem (https://github.com/google/earthenterprise/issues/1800) with ExampleSearch not being available when upgrading to 5.3.5.
1.  Add a second .sql file so that "Serve" and "Extra" own their own copy of the delete script.
2.  Add a segment to the server/post-install.sh script to detect if Extra is installed and run delete script if not.
3.  Modify webpages to document changes.


tst-nfarah:
Let's link the github issue in this PR if existent. If not let's create it first.

earth_enterprise/src/server/examplesearch_2delete.sql
```diff
@@ -0,0 +1,19 @@
+--
```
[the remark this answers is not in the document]
tst-mbiancalana: I see in the initial description why there is an additional duplicated file- one for each RPM.

tst-lsavoie, line 1: I don't love the idea of having a duplicate file but I see the need for it and I can't think of anything better. It would be nice to eliminate the duplication, though.

earth_enterprise/rpms/opengee-server/snippets/post-install.sh
```diff
@@ -69,6 +69,14 @@ main_postinstall()
 
     #11) done!
     service geserver start
+
+    #12) Turn off examplesearch (will be turned on by Extra, if installed).
```
tst-lsavoie, line 73: This should probably be moved up to before #6, which is where the examplesearch database is currently installed. In any case it should come before starting the server in #11. Thinking about it more, I don't think the if statement is necessary because if you install extra and then server, server was the one to actually install example search, opengee-extra just laid down the files for it. We can leave the if statement in to be safe, though.

tst-mbiancalana:
In testing this, everything works as it should, installing 5.3.6 opengee-server RPM without also installing opengee-extra removes ExampleSearch from the search tabs options.

Previously published globes that included the ExampleSearch tab continue to behave as before. That includes globes published with 5.3.5 that have the broken ExampleSearch as mentioned in the original issue (I1800). I don't think this is something that should be fixed in code, but users will need to republish any globes that are broken in this way.

tst-tcalder:
Agreeing with what the others said, this looks good to me.

earth_enterprise/rpms/opengee-server/snippets/post-install.sh
tst-eclamar: Postgresql needs to be running in order for the 'psql' command to execute;  thus, it is run after the geserver is started.  Moving it up after #6 causes the script to error, being unable to run 'psql'.

tst-lsavoie: Ah, OK. That makes sense.

PR #1818, opened 2020-08-11
Move implementation of GetLinesFromFile into AbstractFileAccessor.

tst-jlarocco:


earth_enterprise/src/common/FileAccessor.h
```diff
@@ -35,7 +36,7 @@ class AbstractFileAccessor {
   virtual bool PwriteAll(const void* buffer, size_t size, off64_t offset) = 0;
   virtual bool ReadStringFromFile(const std::string &filename, std::string &str, std::uint64_t limit = 0) = 0;
   virtual bool Exists(const std::string &filename) = 0;
-  virtual bool GetLinesFromFile(std::vector<std::string> &lines, const std::string &filename) = 0;
+  virtual bool GetLinesFromFile(std::vector<std::string> &lines, const std::string &filename);
```
tst-lsavoie, line 39: Minor: this function doesn't need to be virtual if you're never going to override it.

tst-jlarocco: Yeah, I wasn't sure whether to take that out, and ended up leaving it in case a third party plugin wanted to override it for some reason.  That does seem unlikely, though, especially with a default implementation, and it can always be added back if it's needed.

PR #1826, opened 2020-08-21
Qt4 merge1

tst-dsugar:
This is a partial merge that runs by itself and brings closer correspondence with master.


tst-cjeffries:
@googlebot I consent.

PR #1823, opened 2020-08-20
Add AbstractFileAccessor::GetFileInfo() for use in geFilePool

tst-jlarocco:
Created AbstractFileAccessor::GetFileInfo(), and switched calls to khGetFileInfo inside geFilePool.cpp use it instead.

PR #1827, opened 2020-08-25
Build libcurl with bundled OpenSSL and with CA bundle enabled.

tst-jlarocco:
These changes allow the bundled libcurl to access HTTPS.


tst-jlarocco:
These changes were already included in: https://github.com/google/earthenterprise/pull/1823

PR #1829, opened 2020-08-26
Bundle Qt4 and build it in Scons

tst-cjeffries:
Bundles Qt4 (so we have it for all platforms) and builds it in scons. Qt4 shared objects are also installed for runtime.

To test (on El6, El7, EL8, and Ubuntu):

1. Build and install
2. Verify that fusion works, and that Qt shared object files are installed under `/opt/google/lib`


tst-dsugar:
With qt4 bundled I might have expected this to build in ci, but travis has ubuntu 14...so this should be looked into.

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -128,8 +128,8 @@ Execute:
 ```bash
 sudo yum install -y \
   bison-devel boost-devel cmake daemonize freeglut-devel \
-  gdbm-devel geos-devel giflib-devel GitPython \
-  libcap-devel libmng-devel libpng12-devel libX11-devel libXcursor-devel \
+  gdbm-devel geos-devel giflib-devel glib2-devel GitPython gtk2-devel \
+  libcap-devel libicu-devel libmng-devel libpng12-devel libtiff-devel libX11-devel libXcursor-devel \
```
tst-basilhuffman, line 132: libpng12 is only for el6, correct?

tst-cjeffries: Id think so.

PR #1836, opened 2020-09-02
Issue #1785: Fix imagery project transparency

### LINKED ISSUE #1785: Imagery Projects - Transparent areas appears as black

tst-lsavoie:
Fixes #1785 

Thanks to @duvifn for discovering the problem and suggesting a solution.

The problem:

Fusion caches alpha bands to make building multiple layers more efficient. However, it checks for the existence of the alpha band before using it because 4.x did not cache alpha. This is a problem because sometimes the check for the cached alpha happens before it has been built, so Fusion incorrectly decides that there is no alpha and does not apply transparency correctly.

The fix:

This PR moves the check for the cached alpha folder to just before it is used so that it has time to build it if it needs to.

To verify:

1. Create a flat 2D database containing only blue marble and publish it with WMS enabled.
1. Create a flat 2D database containing only this file: https://drive.google.com/file/d/1IWjQvGDgGNRlGRVuRNImyfuAazGXN6xm/view?usp=sharing and no blue marble. It is important that the imagery resource and project are brand new and have not been built before.
1. View both databases in a tool like QGIS with the clipped file on top of blue marble. You should see a small amount of high resolution imagery in San Francisco and blue marble everywhere else. Without this fix, the clipped image would have an opaque background and you would not see blue marble at all.

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
+++ b/earth_enterprise/src/fusion/autoingest/plugins.src/PacketGen.src
@@ -249,18 +249,9 @@ void PacketGenAssetVersionImplD::UpdateChildren(
           // In these cases we pull directly from the product. It
           // will also happen for the last thing on the stack who
           // hasn't had his blendvers filled in yet.
-
-          // Note: Before creating a merge inset, we check if the cached alpha
-          // blend directory exists for the asset since in GEE-5.x we need to
-          // pick up the GEE-4.x Imagery/Terrain Projects (the PacketLevel
-          // assets) that have no alpha blend cached.
           levInputs[inputs_idx+1] = blendVer.Ref();
           levInputVers[inputs_idx+1] = blendVer;
-          const std::string cache_alpha_path =
-              blendVer->GetOutputFilename(0) + "/cache_alpha.pack";
-          if (khExists(cache_alpha_path)) {
-            cached_blend_alpha = blendVer.Ref();
-          }
+          cached_blend_alpha = blendVer.Ref();
           cached_blend = blendVer.Ref();
         }
         packetLevelConfig.insets[merge_idx] =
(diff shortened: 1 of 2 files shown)


tst-lsavoie:
@duvifn do you have time to review and test this PR? If you are happy with the code and it passes your tests, you can click on the "Files Changed" tab above, then on "Review Changes" at the top right and approve the PR.

FYI, I'm going to be unavailable for the next few days but will come back to this mid next week.

duvifn:
@tst-lsavoie Thangs! 
This looks good to me and also tested with the given  use case.

duvifn (approved):

duvifn:
I also checked this with multi layer project (bluemarble + clipped raster).

tst-lsavoie:
I've run a number of tests on this and everything works as expected. No problems found.

tst-jlarocco (approved):
My apologies for taking so long to look at this.

Code looks good, and I did a quick test in QGIS and everything looked good.

Thanks again to @duvifn for pointing this out and the initial code for fixing it.

PR #1837, opened 2020-09-02
Enable cutter on systems with restrictive umasks

tst-lsavoie:


tst-sbayard (approved):
Reviewed and tested. Verified that cutter functionality is working when enabled and inaccessible when disabled. Also inspected file permissions for cutter-related files to verify they were correct after both `gecutter enable` and `gecutter disable`.

tst-jlarocco (approved):
Looks good to me.

PR #1853, opened 2020-10-08
Added 'host' option to PostgreSQL connection settings

tst-dsheremata:
Added 'host' option to PostgreSQL connection settings, defined in postgres.properties.  Default setting is still '/tmp', for unix-domain sockets, but changing it can support remote PostgreSQL connections, more suitable for clustered server deployments.  

To test: 
* Validate that the `geserveradmin --adddb ...` command works, as well as `--pushdb`, `--publishdb`, etc.  with the default configurations (i.e., no config changes) on a single GEE Server instance. 
* Validate functionality of the web admin console on the GEE Server.
* Configure a remote PostgreSQL database (Note: PostgreSQL config files (`/var/opt/google/pgsql/data/postgresql.conf` and `/var/opt/google/pgsql/data/pg_hba.conf` must be configured to allow the remote connections.  Also ensure that firewalls/security groups/VPC etc. allow remote connections.  )
* Configure a separate GEE Server to connect to that database, e.g. by adding `host=your.postgressqlhost.fqdn.net` to the `/opt/google/gehttpd/wsgi-bin/conf/postgres.properties` file. 
* Validate that the `geserveradmin --adddb ...` command works, as well as `--pushdb`, `--publishdb`, etc. work with the new configuration
* Validate functionality of the web admin console on the GEE Server works with the new configuration.

### DIFF
--- a/earth_enterprise/src/common/config/gePrompt.cpp
+++ b/earth_enterprise/src/common/config/gePrompt.cpp
@@ -116,11 +116,11 @@ enterDirname(const std::string msg, const std::string &dflt,
 "%1 does not exist.\n"
 "Should this tool create it now")
                   .arg(dir), 'Y')) {
-        geCapabilitiesGuard cap_guard(
-            CAP_DAC_OVERRIDE,     // let me read all files
-            CAP_DAC_READ_SEARCH,  // let me traverse all dirs
-            CAP_CHOWN,            // let me chown files
-            CAP_FOWNER);          // let me chmod files I dont own
+        // geCapabilitiesGuard cap_guard(
+        //     CAP_DAC_OVERRIDE,     // let me read all files
+        //     CAP_DAC_READ_SEARCH,  // let me traverse all dirs
+        //     CAP_CHOWN,            // let me chown files
+        //     CAP_FOWNER);          // let me chmod files I dont own
         if (khMakeDir(dir)) {
           done = true;
         }
--- a/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/postgres_manager_wrap.py
+++ b/earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/postgres_manager_wrap.py
@@ -35,6 +35,7 @@ import postgres_properties
 
 class PostgresManagerWrap(object):
   DB_PORT = postgres_properties.PostgresProperties().GetPortNumber()
+  DB_HOST = postgres_properties.PostgresProperties().GetHost()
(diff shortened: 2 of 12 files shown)


earth_enterprise/src/common/config/gePrompt.cpp
```diff
@@ -116,11 +116,11 @@ enterDirname(const std::string msg, const std::string &dflt,
 "%1 does not exist.\n"
 "Should this tool create it now")
                   .arg(dir), 'Y')) {
-        geCapabilitiesGuard cap_guard(
-            CAP_DAC_OVERRIDE,     // let me read all files
-            CAP_DAC_READ_SEARCH,  // let me traverse all dirs
-            CAP_CHOWN,            // let me chown files
-            CAP_FOWNER);          // let me chmod files I dont own
+        // geCapabilitiesGuard cap_guard(
```
tst-ccamp, line 119: I saw in the commit description this was done so it would work in a non-privileged container. Do we want to keep it removed permanently?

earth_enterprise/src/fusion/portableglobe/cutter/cgi-bin/common/postgres_manager_wrap.py
```diff
@@ -35,6 +35,7 @@
 
 class PostgresManagerWrap(object):
   DB_PORT = postgres_properties.PostgresProperties().GetPortNumber()
+  DB_HOST = postgres_properties.PostgresProperties().GetHost()
   DB_USER = "geuser"
```
tst-dsugar, line 39: Hooking up to an arbitrary and authenticating to a remote database will normally require both a user and password, and both should be set from properties or somewhere...

earth_enterprise/src/server/wsgi/common/postgres_properties.py
```diff
@@ -26,6 +29,10 @@ def __init__(self):
     self.port_number = utils.GetPostgresPortNumber()
     if not self.port_number:
       self.port_number = "5432"
+    self.host = utils.GetPostgresHost()
```
tst-dsugar, line 32: We need to add utils.GetPostgressUser() and password, and have default handling for empty/unset.

earth_enterprise/src/common/config/gePrompt.cpp
tst-dsugar: Is this an entirely secondary issue? I think it should be conditional, even if we temporary have it initially just under some compile-time flag.

PR #1854, opened 2020-10-16
opengee-extra RPM fails to register plugin on when using peer authentication

tst-sbayard:
If peer authentication is configured for PostgreSQL, the ExampleSearch will not be registered properly if the *NIX user is not in the user map. This change runs the command as `gepguser` to ensure it executes properly.

### DIFF
--- a/earth_enterprise/rpms/opengee-extra/snippets/post-install.sh
+++ b/earth_enterprise/rpms/opengee-extra/snippets/post-install.sh
@@ -24,7 +24,7 @@ main_postinstall()
         # Set up the ExampleSearch plugin. The opengee-server-core RPM does
         # this by calling geresetpgdb, but there's no need to do a full reset
         # here.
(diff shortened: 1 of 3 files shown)


tst-lsavoie:
There's also code that modifies examplesearch in opengee-server/snippets/post-install.sh. You should fix that one too.

tst-lsavoie:
@googlebot I consent.

tst-lsavoie:
Turns out I have permissions on your fork, so I made the change to the opengee-server script.

tst-sbayard:
Well, I was testing the change locally before pushing, but there you go ;P

tst-lsavoie:
@googlebot I consent.

tst-jlarocco (approved):
LGTM

tst-lsavoie (approved):
I've tested with various methods of installing, uninstalling, and upgrading, and everything works as expected.

PR #1862, opened 2020-11-03
Updated variables (#1861)

### LINKED ISSUE #1861: Invalid variables when reporting postgres errors
There are several variable naming errors in the function `_GetAllAssets` in `earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py`. Here is a part of the function:

```
    except psycopg2.Warning as w:
      logger.error(w)
      http_io.ResponseWriter.AddJsonFailureBody(response, str(e))
    except psycopg2.Error as e:
      logger.error(e)
      http_io.ResponseWriter.AddJsonFailureBody(response, str(w))
```
Note that in the first `except` statement the variable is called `w`, but in the Json response it is using the variable `e`. In the second `except` statement it has the same problem in reverse. We need to clean this up so that it always uses the correct variable.

augustocamaral2003:
Fixes #1861

### DIFF
--- a/earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
+++ b/earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
@@ -1128,10 +1128,10 @@ class PublishManagerHelper(stream_manager.StreamManager):
       http_io.ResponseWriter.AddJsonFailureBody(response, str(e))
     except psycopg2.Warning as w:
       logger.error(w)
-      http_io.ResponseWriter.AddJsonFailureBody(response, str(e))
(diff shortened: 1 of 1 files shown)


augustocamaral2003:
@googlebot I signed it!

tst-lsavoie:
@augustocamaral2003 thanks for the PR. I'll work on getting it reviewed.

tst-lsavoie (approved):
The code changes look fine. I haven't tested it.

tst-jlarocco:
@augustocamaral2003  Looks like there's one more spot (lines 1102 and 1105) where it has the 'w's and 'e's swapped.  Would you be able to grab that one also?

PR #1864, opened 2020-11-11
Updated variables (#1861)

### LINKED ISSUE #1861: Invalid variables when reporting postgres errors

augustocamaral2003:
Fixes #1861

### DIFF
--- a/earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
+++ b/earth_enterprise/src/server/wsgi/serve/publish/publish_manager_helper.py
@@ -1099,10 +1099,10 @@ class PublishManagerHelper(stream_manager.StreamManager):
       http_io.ResponseWriter.AddJsonFailureBody(response, str(e))
     except psycopg2.Warning as w:
       logger.error(w)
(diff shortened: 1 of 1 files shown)


tst-lsavoie (approved):
Code looks good. I haven't tested it.

tst-jlarocco (approved):
Code looks good, and I verified unit tests pass and ran some manual tests to verify these exceptions are being handled.

tst-lsavoie:
Thanks for the help @augustocamaral2003

PR #1874, opened 2020-12-10
Update to scons 4.x

tst-cjeffries:
Updates our code to scons 4.x and updates our scons files to work with python 3.

To test:
1. Build and install
2. Verify that these things complete successfully
3. Verify with the tutorial data

### DIFF
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -224,7 +224,7 @@ try:
   minor=int(arr[2])
   el_ver = ''
   if arr[1] < 16:
-    print('Unknown Ubuntu version, minimum required version is 16.04.')
+    print('Minimum required Ubuntu version is 16.04.')
     Exit(1)
     is_min_ubuntu = False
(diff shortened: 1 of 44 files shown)


earth_enterprise/src/SConstruct
```diff
@@ -193,25 +193,19 @@ else:
   else:
     el_ver = 'el7'
 
-if bundle_3rd_libs:
-  python_version = '2.7.5'
+
+if el_ver == 'el6':
```
tst-dsugar, line 197: I wonder if we need to consider el6 in the timeframe of this branch...

tst-basilhuffman (approved):
lgtm

earth_enterprise/src/SConstruct
tst-cjeffries: I don't disagree. I just simplified the logic that was already there. When we drop python 2, we shouldnt need to do this sort of thing anymore, as everything should be running the same version of python.

tst-dsugar (approved):

PR #1879, opened 2020-12-23
Qt4 database push bug fix

tst-mswartz:
Commenting out the two connect() lines that cause the segfaults. They don't seem to be required for cleanup, and databases are pushed successfully to the server.  It seems as though the push and publish threads are attempting to schedule themselves for deletion by themselves (?) but according to the Qt documentation it seems as though this should happen automatically when the dialog boxes are closed.


tst-basilhuffman:
does not crash, just have a question re: keeping the lines in with comment vs removing them

tst-dsugar (commented):
Retarget to master, because source branch already on 5.3.8. Would suggest cherry-picking into the release branch.

tst-basilhuffman (approved):

tst-dsugar (approved):

PR #1880, opened 2021-01-04
Fix issues with external sites loading google maps

### LINKED ISSUE #1902: Fix issues with applications using Fusion Maps API when they aren't hosted on the GEE Server
Fusion maps reloads jQuery, which wipes out the version loaded by application including plugins
Empty tile image grabbed from the application's server rather than GEE
Google metrics being sent to application server rather than to GEE

cwgrant:
Fix issues with applications using Fusion Maps API when they aren't hosted on the GEE Server.

- Fusion maps reloads jQuery, which wipes out the version loaded by application including plugins
- Empty tile image grabbed from the application's server rather than GEE
- Google metrics being sent to application server rather than to GEE

Fixes #1902

### DIFF
--- a/earth_enterprise/src/maps/api/bootstrap.js
+++ b/earth_enterprise/src/maps/api/bootstrap.js
@@ -1,11 +1,11 @@
 //  Copyright 2017 Google Inc.
-// 
+//
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
 //  You may obtain a copy of the License at
-// 
+//
 //       http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 //  Unless required by applicable law or agreed to in writing, software
 //  distributed under the License is distributed on an "AS IS" BASIS,
 //  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


tst-jlarocco:
I've tested using the tutorial data and maps/example_local_v*.html pages, and also tested pre-loading JQuery 3.5 before loading bootstrap.js, and everything seems to work fine.

I'll see if I can get a second person to review this today and get this merged.

tst-jlarocco (approved):

earth_enterprise/src/maps/api/bootstrap.js
```diff
@@ -22,21 +22,28 @@ google.maps = google.maps || {};
...
+                                                        ,
+                                                        GEE_BASE_URL /* GE metrics */],
+                                                        [GEE_BASE_URL + GEE_API_PATH,'internal'],[0],0.0,,,,,], loadScriptTime);
   };
 
   var loadScriptTime = (new Date).getTime();
 
-  getScript(GEE_BASE_URL + '/js/jquery.js');
+  // Only load jQuery if it isn't loaded already.
+  if (!window.jQuery) {
```
tst-lsavoie, line 40: Do we need this change in the other versions of bootstrap.js?

tst-lsavoie: Update: none of the other files load jQuery, so this behavior is correct.

tst-lsavoie (approved):
Reviewed the code changes; everything looks good.

earth_enterprise/src/maps/api/bootstrap.js
tst-jlarocco: I think the idea was not to change the behavior of the old versions.  The maps API seems to work fine with the newer jQuery, but it's possible people are depending on maps loading the older 1.5 version over the version they have loaded.

Now that I'm thinking about it, though, we should mention this in the release notes.

PR #1881, opened 2021-01-08
Add fusion and server shutdowns to gee-common preinstall.

tst-cjeffries:
Update the opengee-common rpm pre-installs script to stop both fusion and server.


tst-jzmolek (approved):
Looks good

tst-rwildes (approved):
Looks good

tst-lsavoie (approved):
I've tested an upgrade and verified that this fixes the issue.

PR #1883, opened 2021-01-13
Difficult read

tst-basilhuffman:
after the upgrade to qt4 the level view text has no contrast and becomes difficult to read. this changes the background color, allowing for readability


tst-mswartz (approved):
Looks good to me, approving

tst-dsugar (approved):

PR #1893, opened 2021-01-26
1891 update libcurl

tst-jlarocco:
These are a few changes related to the upcoming OpenSSL 1.1.1 upgrade.

* The most significant is an upgrade to the bundled libcurl to the latest upstream version, 4.74.0
* Disabling SSLv3 in Apache
* A small fix to the mod_wsgi configuration that fixes a bug on EL 6 platforms.

### DIFF
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -214,7 +214,6 @@ else:
     python_home = '/usr'
 
 # setup platform specific operations
-is_os_ubuntu = False
 is_min_ubuntu = True
 # /etc/issue was the previous method of checking Ubuntu release version
 # however, some flavors of Ubuntu e.g. MATE, have used varying formats
--- a/earth_enterprise/src/SConstruct
+++ b/earth_enterprise/src/SConstruct
@@ -218,23 +218,17 @@ is_os_ubuntu = False
 is_min_ubuntu = True
 # /etc/issue was the previous method of checking Ubuntu release version
 # however, some flavors of Ubuntu e.g. MATE, have used varying formats
-# using ilsb_release, subprocess and re is more readable than awk parsing also
-try:
-  vstring=subprocess.check_output('lsb_release -i',shell=True)
-  system=re.split('[\t\.\n]',vstring)
-  if system[1] == "Ubuntu":
-    is_os_ubuntu = True
+# using lsb_release, subprocess and re is more readable than awk parsing also
+if not os.path.isfile('/etc/redhat-release'):
   vstring=subprocess.check_output('lsb_release -r',shell=True)
   arr=re.split('[\t\.\n]',vstring)
   major=int(arr[1])
   minor=int(arr[2])
   el_ver = ''
-  if arr[1] < 16:
+  if major < 16:
     print 'Unknown Ubuntu version, minimum required version is 16.04.'
     Exit(1)
-    is_min_ubuntu = False
-# lsb_release is not present, must be RHEL or CentOS, move on
-except:
+else:
     print "OS is not Ubuntu"
 
 extragccflags = []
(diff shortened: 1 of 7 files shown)


earth_enterprise/src/third_party/mod_wsgi/SConscript
```diff
@@ -95,6 +95,9 @@ else:
                     mod_wsgi_env['ENV']['mod_env'], env_opt, config_opt, root_dir,
                     mod_wsgi_target))
 
+print('AAAA> el_ver: ' + mod_wsgi_env['el_ver'])
```
tst-lsavoie, line 98: This looks like temporary debugging statements that should be removed.

earth_enterprise/src/SConstruct
```diff
@@ -218,23 +218,17 @@ is_os_ubuntu = False
...
 # /etc/issue was the previous method of checking Ubuntu release version
 # however, some flavors of Ubuntu e.g. MATE, have used varying formats
-# using ilsb_release, subprocess and re is more readable than awk parsing also
-try:
-  vstring=subprocess.check_output('lsb_release -i',shell=True)
-  system=re.split('[\t\.\n]',vstring)
-  if system[1] == "Ubuntu":
-    is_os_ubuntu = True
+# using lsb_release, subprocess and re is more readable than awk parsing also
+if not os.path.isfile('/etc/redhat-release'):
```
tst-lsavoie, line 222: It looks like we might need to set `is_os_ubuntu` to `True` in here.

earth_enterprise/src/third_party/libcurl/SConscript
```diff
@@ -17,7 +18,7 @@
 
 
 Import('third_party_env')
-libcurl_version = 'curl-7.45.0'
+libcurl_version = 'curl-7.74.0'
```
tst-lsavoie, line 21: We should update `README.google` in this same directory because it also has the curl version number in it.

earth_enterprise/src/SConstruct
tst-lsavoie: Actually, it looks like it's not used, so we could just delete it on line 217, but that's fairly minor.

earth_enterprise/src/third_party/apache2/apache-ge-httpd-ssl.conf.in.patch
```diff
@@ -18,7 +18,7 @@
 -SSLProtocol all -SSLv3
 +#SSLProtocol all -SSLv3
 +# Limit SSL protocols.
-+SSLProtocol -ALL +TLSv1.2 +SSLv3 +TLSv1
++SSLProtocol -ALL +TLSv1.2 +TLSv1
```
tst-ccamp, line 21: Maybe this is beyond the scope of the ticket, but should we also consider disabling TLSv1? Even the comment above this line says that it shouldn't be used past 2016. We kept it longer because older versions of EC didn't support TLSv1.2, but that should have been taken care of for a long time now, and if a customer still has a reason to use it, that should be a deliberate choice to use an unsecure protocol rather than the default.

tst-lsavoie: That's a good point. We only removed SSLv3 because that's the one we're forced to remove to upgrade to Open SSL 1.1.1. I could see taking out TLSv1 as well if we're certain it's not going to cause problems.

tst-ccamp: Hmmm, I'm not sure I'd say I'm certain. I can look into whether we have any statistics we can refer to.

tst-lsavoie (approved):
Code looks good and my tests passed.

tst-nfarah:
I built it and tested it. Looks good.

tst-nfarah (approved):

PR #1894, opened 2021-01-28
Limit / correct assetmanager window size

tst-mswartz:
Verification steps:
1. Cannot stretch the assetmanager window to be wider or taller than 3000 pixels
2. While fusionui is not running, edit ~/.fusion/assetmanager.layout to have a width of some enormous size such as 100000
3. Relaunch fusionui. It does not segfault. Open assetmanager. It is a normal size. 
4. Click on a couple of things and maybe resize the assetmanager window and then close it. Open assetmanager.layout again. The incorrect size is set to something more normal


tst-cjeffries (approved):
Looks good to me. Tested and approve.

tst-basilhuffman (approved):

tst-basilhuffman (approved):

PR #1897, opened 2021-01-29
Upgrade to OpenSSL 1.1.1

### LINKED ISSUE #1895: Upgrade to Open SSL 1.1.1
Upgrade the bundled Open SSL to the latest version of 1.1.1.

tst-lsavoie:
This PR upgrades our bundled version of Open SSL to the latest version. It also fixes some problems with the RPMs that were discovered during this upgrade. Specifically, some of the post-install scripts rely on Open GEE binaries, but in the middle of a yum upgrade you might be using old binaries that are trying to link to libraries that have already been upgraded, resulting in errors and crashes. Thus, this PR also includes code to force upgrades to happen in a safe order.

Fixes #1895

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
       libstdc++6 libtool libgif-dev libtiff-dev libgtk2.0-dev libglib2.0-dev
       libx11-dev libxcursor-dev libxerces-c-dev libxft-dev libxinerama-dev
       libxml2-dev libxml2-utils libxmu-dev libxrandr-dev libyaml-cpp-dev libpcre3-dev
-    openssl shunit2 git git-lfs libqt4-dev qt4-default libqt4-opengl-dev
+    shunit2 git git-lfs libqt4-dev qt4-default libqt4-opengl-dev
     python-dev python-git python-imaging python-setuptools python-psycopg2 python-lxml
     swig xorg-dev zlib1g-dev
 script: echo "Running build..."
--- a/earth_enterprise/BUILD_Ubuntu.md
+++ b/earth_enterprise/BUILD_Ubuntu.md
@@ -38,7 +38,7 @@ sudo apt-get install \
     libstdc++6 libtool libgif-dev libtiff-dev libgtk2.0-dev libglib2.0-dev \
     libx11-dev libxcursor-dev libxerces-c-dev libxft-dev libxinerama-dev \
     libxml2-dev libxml2-utils libxmu-dev libxrandr-dev libyaml-cpp-dev \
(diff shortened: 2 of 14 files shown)


.travis.yml
```diff
@@ -17,7 +17,7 @@ before_install:
       libstdc++6 libtool libgif-dev libtiff-dev libgtk2.0-dev libglib2.0-dev
       libx11-dev libxcursor-dev libxerces-c-dev libxft-dev libxinerama-dev
       libxml2-dev libxml2-utils libxmu-dev libxrandr-dev libyaml-cpp-dev libpcre3-dev
-    openssl shunit2 git git-lfs libqt4-dev qt4-default libqt4-opengl-dev
```
tst-dsugar, line 20: I don't think it's really possible to avoid installing the system ssl libraries because so many other packages reference them.

tst-lsavoie: The system libraries should be installed by default. This is just indicating that we don't require openssl for Open GEE to run.

earth_enterprise/BUILD_Ubuntu.md
```diff
@@ -35,10 +35,10 @@ sudo apt-get install \
     libgdbm-dev libgeos-dev libgeos++-dev libgif-dev libgtest-dev \
     libjpeg-dev libjpeg8-dev libmng-dev libogdi3.2-dev \
     libperl4-corelibs-perl libpng12-0 libpng12-dev libpq-dev libproj-dev \
-    libstdc++6 libtool libgif-dev libtiff-dev libgtk2.0-dev libglib2.0-dev \
+    libstdc++6 libtool libtiff-dev libgtk2.0-dev libglib2.0-dev \
```
tst-nfarah, line 38: Do we need to add `libicu-dev` as well ?

tst-nfarah (changes requested):

earth_enterprise/rpms/build.gradle
```diff
@@ -258,7 +258,16 @@ task openGeePostGisRpm(type: GeeRpm) {
     autoFindProvides = true
```
tst-nfarah, line 258: Update copyright year to include 2021

earth_enterprise/BUILD_Ubuntu.md
tst-lsavoie: I don't know of a reason to add it. It's not a direct requirement of Open GEE that I'm aware of, and if any of these packages require it, apt should automatically bring it in.

tst-lsavoie:
We're running into some problems testing and need to put some thought into where we're going with OpenSSL long term. Closing this for now.

PR #1898, opened 2021-02-01
Crash-causing OBOB fixed that resulted from faulty iterator logic

tst-sbayard:
Title says it. In the upgrade to QT4, it looks like the value returned by `*it` after the last element is no longer zero as the original logic was assuming. The result was an attempt to delete one more item than was in the list.

Replaced with a simple `for` loop which also resulted in trimming down the code.


tst-dsugar (approved):

tst-basilhuffman (approved):
lgtm

PR #1901, opened 2021-02-08
Vector resource dialog fixes

tst-mswartz:
Verification steps:   
1. All fields in  the vector resource dialog are visible 
2. Date and provider are able to be set and modified


tst-basilhuffman (approved):
lgtm

tst-sbayard (approved):
Verified that the fields are now visible in the "Vector Resource" dialog.

PR #1906, opened 2021-02-17
Fix segfault caused by malformed URLs

### LINKED ISSUE #1905: Improperly formatted icon requests to portables can crash the server
**To Reproduce**

1. Publish a database
2. Enable cutter and cut a portable map or globe from the database. You may need to include a vector layer.
3. From the admin console, click "Manage Portables", then "Register" next to a portable map
4. Publish the newly registered portable
5. Access the following URL: `http://server-url/portable-name/flatfile?lf-0-=icons/773_l.png`

The server will stop serving tiles and there will be a segfault in the error log.

**Expected behavior**

The server returns a 404 when it receives a malformed URL

**Additional context**

The correct URL is `http://server-url/portable-name/flatfile?lf-0-icons/773_l.png` (without the `=`).

tst-lsavoie:
Fixes #1905

### DIFF
--- a/earth_enterprise/src/server/mod_fdb/SConscript
+++ b/earth_enterprise/src/server/mod_fdb/SConscript
@@ -30,6 +30,7 @@ fdb_module = env.sharedLib('mod_fdb',
                 'fdbservice.cpp', 'portableservice.cpp', 'fusiondbservice.cpp',
                 'tileservice.cpp', 'motf_generator.cpp'],
                LIBS=mod_fdb_libs,
+               SONAME='mod_fdb.so',
                LINK=['g++'])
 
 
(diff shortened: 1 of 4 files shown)


tst-hrajamoney:
@googlebot I consent

earth_enterprise/src/server/mod_fdb/SConscript
```diff
@@ -21,21 +22,27 @@ import os
...
+               LINKFLAGS=env['LINKFLAGS'] + ['-Wl,-soname,mod_fdb.so'])
+
 
 # Install /opt/google/gehttpd/auth directory only
 auth_dir = env.installdirs['httpd'].entry_abspath('auth')
 if (not os.path.exists(auth_dir)):
     os.makedirs(auth_dir)
 
-env.install('httpdmodules', [module])
+#env.install('server_lib', [fdb_module])
```
tst-hrajamoney, line 44: should we delete the line instead of commenting?

tst-lsavoie: Yes, good catch

earth_enterprise/src/server/mod_fdb/portableservice_unittest.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+TEST_F(PortableServiceTest, FlatfileQueryNoArgs) {
+  TestBadRequest(r, "flatfile", TARGET, GLOBE, 0, false, args, "query", nullptr);
+}
+
+TEST_F(PortableServiceTest, FlatfileQuery) {
+  args.emplace("blist", "0301324");
+  args.emplace("request", "QTPacket");
+  args.emplace("version", "1");
+  TestGoodRequest(r, "flatfile", TARGET, GLOBE, 0, false, args, "query", nullptr, 0x656c626174726f50);
```
tst-jlarocco, line 137: Could this parameter be changed to take a string or array of characters as the final parameter?  Or  maybe use a function to convert a string to an uint64_t?

tst-jlarocco: Alright, I didn't realize that.  It's a little confusing but I don't have a better solution.

tst-jlarocco (approved):
Code looks good!

PR #1908, opened 2021-02-19
Force RPM install order

### LINKED ISSUE #1912: RPMs can install in the wrong order when they are installed in a single transaction
**Describe the bug**

RPMs can install out of order when they are installed as part of a single transaction. For example, if you have a previous version of Open GEE installed and then run:

```
yum install opengee-common opengee-postgis opengee-server opengee-fusion
```

yum will only ensure that the state is consistent at the end of the transaction, not during the transaction. Thus, the RPMs could install in the order opengee-common, opengee-server, opengee-postgis. This is a problem if common upgrades a library that's not compatible with the older version of postgis because server relies on postgis in its post-install script, and postgis will not be in a working state (due to the incompatibility with the newer version of common).

This problem also occurs with opengee-extra, but it is harder to fix because extra can be installed in any order relative to the other RPMs, but it relies on Fusion and Server working if they have been installed previously. Thus, if you have an old version of Open GEE installed and opengee-common upgrades a library that is incompatible with an older version of opengee-postgis, you will have a problem in this scenario:

```
yum install opengee-common
yum install opengee-extra
```

The install will fail because postgis (and therefore opengee-server) will be in a non-working state after the opengee-common upgrade, causing the opengee-extra post-install script to fail.

**To Reproduce**

There are two related problems here and both are difficult to reproduce. The first depends on the behavior of yum, and both depend on upgrading to a new version of a library. For an example of code that upgrades a library, see https://github.com/google/earthenterprise/pull/1897/files.

Problem 1:

# Install Open GEE 5.3.7 or before from RPMs
# Create Open GEE RPMs that upgrade a library (such as in the PR above)
# Install the new version of Open GEE using `yum install opengee-common opengee-postgis opengee-server opengee-fusion`
# If yum installs the RPMs in the order opengee-common, opengee-server, opengee-postgis (which is not forbidden), the opengee-server post-install script will fail.

Problem 2:

# Install Open GEE 5.3.7 or before from RPMs
# Create Open GEE RPMs that upgrade a library (such as in the PR above)
# Install the new opengee-common RPM
# Install the new opengee-extra RPM. Note that the post-install script fails

**Expected behavior**

# RPMs are forced to install in a supported order, even if they are installed/upgraded in a single transaction
# The opengee-extra RPM installs without failing no matter the install/upgrade order

tst-lsavoie:
Forces RPMs to install in the required order so that pre- and post- install scripts are guaranteed to work. Also includes other minor fixes.

Fixes #1912

### DIFF
--- a/earth_enterprise/BUILD_Ubuntu.md
+++ b/earth_enterprise/BUILD_Ubuntu.md
@@ -35,7 +35,7 @@ sudo apt-get install \
     libgdbm-dev libgeos-dev libgeos++-dev libgif-dev libgtest-dev \
     libjpeg-dev libjpeg8-dev libmng-dev libogdi3.2-dev \
     libperl4-corelibs-perl libpng12-0 libpng12-dev libpq-dev libproj-dev \
-    libstdc++6 libtool libgif-dev libtiff-dev libgtk2.0-dev libglib2.0-dev \
+    libstdc++6 libtool libtiff-dev libgtk2.0-dev libglib2.0-dev \
     libx11-dev libxcursor-dev libxerces-c-dev libxft-dev libxinerama-dev \
     libxml2-dev libxml2-utils libxmu-dev libxrandr-dev libyaml-cpp-dev \
     openssl libpcre3 libpcre3-dev \
--- a/earth_enterprise/rpms/build.gradle
+++ b/earth_enterprise/rpms/build.gradle
@@ -1,4 +1,4 @@
-// Copyright 2017 - 2020 the Open GEE Contributors
+// Copyright 2017 - 2021 the Open GEE Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


tst-nfarah:
What issue this fixes ? If not filed yet, let's file and link it in this PR.

earth_enterprise/BUILD_Ubuntu.md
```diff
@@ -35,7 +35,7 @@ sudo apt-get install \
     libgdbm-dev libgeos-dev libgeos++-dev libgif-dev libgtest-dev \
     libjpeg-dev libjpeg8-dev libmng-dev libogdi3.2-dev \
     libperl4-corelibs-perl libpng12-0 libpng12-dev libpq-dev libproj-dev \
-    libstdc++6 libtool libgif-dev libtiff-dev libgtk2.0-dev libglib2.0-dev \
+    libstdc++6 libtool libtiff-dev libgtk2.0-dev libglib2.0-dev \
```
tst-lsavoie, line 38: This is a minor fix that removes a duplicate library.

tst-lsavoie:
@tst-nfarah I have added a link to an issue that includes reproduction steps. Unfortunately the reproduction is non-trivial.

Also, these fixes were pulled from https://github.com/google/earthenterprise/pull/1897/files, and they went through fairly significant testing on that PR.

earth_enterprise/rpms/build.gradle
```diff
@@ -463,7 +472,6 @@ task openGeeCommonDeb (type: GeeDeb, dependsOn: openGeeCommonInitScripts) {
     requires('libstdc++6')
     requires('libtool')
     requires('libxml2-utils')
-    requires('openssl')
```
tst-nfarah, line 466: this removal is done because PostGIS does not need SSL ? I can see the case where a client connecting to Postgres needs SSL on client and server side. Unless this is a dup since Open SSL is already deployed by GEE server.

tst-lsavoie: This is just some cleanup. Notice that this is the section of code that builds the debs, not the rpms. The rpm section doesn't have this requirement, so I think it was left here by mistake.

tst-nfarah (approved):

tst-jlarocco (approved):
Looks good!

PR #1914, opened 2021-03-01
Fix user home directory permissions

### LINKED ISSUE #1913: Fix user home directory permissions
Because of this improper ownership some things that should succeed seem to fail.  The issue I found was with the fact that gepguser could not access its own .profile file in it's own home directory.  This caused some STIG testing scripts to fail and the work around was to do: sudo chown -R gepguser:gegroup /opt/google/.users/gepguser for those scripts to work.  IDK of any other bad effects this might have but it's really hard to say this won't cause other issues we don't know about yet.  When we create these users and use a custom home directory like this we should ensure proper ownership is given.  This can be a trap for other issues to spring up unexpectantly in the future so it would be a good thing (and probably simple thing) to fix.

tst-ccamp:
Fixes #1913 

**Verification Steps**
1. Build and install for Ubuntu 16
2. Verify the /opt/google/.users/* directories are owned by each user (gepguser, geapacheuser, gefusionuser) and assigned to the group gegroup
3. Build RPMs for el7 and install
4. Verify the /opt/google/.users/* directories are owned by each user (gepguser, geapacheuser, gefusionuser) and assigned to the group gegroup
5. For users gepguser and geapacheuser, verify that the .profile file also has the same ownership info

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,11 @@ The application suite consists of three core components:
 
 ## Releases
 
+[Release 5.3.7](https://github.com/google/earthenterprise/releases/tag/5.3.7-1700.3)
+
+* [Release Notes](https://www.opengee.org/geedocs/5.3.7/answer/releaseNotes/relNotesGEE5_3_7.html)
+* [Release Documentation](https://www.opengee.org/geedocs/5.3.7/)
+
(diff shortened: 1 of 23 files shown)


tst-basilhuffman (approved):
tested, lgtm

tst-cjeffries (approved):
lgtm

PR #1917, opened 2021-03-10
Remove publish root ownership change from RPM installer, add --nochown to geconfigurepublishroot

### LINKED ISSUE #1910: Slow publish root configuration during upgrades
**Describe the bug**

Installing OpenGEE RPMs with a lot (terrabytes) of data in the publish root (/gevol/published_dbs) causes upgrades to take a very long time.

**To Reproduce**

1. Install OpenGEE using RPMs created with the Gradle script in [/earth_enterprise/rpms](https://github.com/google/earthenterprise/tree/master/earth_enterprise/rpms).
2. Create a project and publish a large database.
3. Create new RPMs
4. Upgrade OpenGEE using the new RPMs
5. While the upgrade is running use `top` or `ps -faux` to observe the running processes.  At some point `geconfigurepublishroot`, `chown -R`, and `chmod -R` will run for significant amounts of time.  The larger the publish root, the longer it will run.
6. Manually run `geconfigurepublishroot` against the publish root using `sudo /opt/google/bin/geconfigurepublishroot --noprompt --path=/gevol/published_dbs/`.  This will take a long time for a large publish root, even though there should be no changes happening.

**Expected behavior**

Upgrading or configuring an existing publish root should be fast if nothing is changing.

**Desktop**

 - Server OS: Any supported by OpenGEE
 - Version: 5.3.7.0, 5.3.7.1, and most recent releases.

**Additional context**

The root of the problem is that both `geconfigurepublishroot` and the installer are changing ownership and permissions on a large number of files, even if it's not necessary to do so.  To make matters worse, the installer changes the permissions, calls `geconfigurepublishroot`, and then changes the permissions again.

One scenario where these changes would be necessary during the install would be to handle the scenario where OpenGEE is installed, a database published, OpenGEE is removed and the `geapacheuser` and `gefusionuser` users removed and the UIDs re-assigned.  At a later time OpenGEE is re-installed and pointed at the old publish root, but it would now have a different owner and potentially different permissions, so the `chown` and `chmod` would be necessary.

tst-jlarocco:
Fixes #1910

The RPM installer no longer changes ownership or permissions of the publish root.  If the user needs to change publish root ownership they can do it after installation with `geconfigurepublishroot`.

For large publish roots, it's recommended to use `geconfigurepublishroot --nochown ...` unless the ownership and permissions need to be fixed.


tst-lsavoie:
IIRC, @tst-ccamp had suggested making `nochown` the default and requiring users to pass a flag if they want to run the chown. What do you think about that? You'd need to make the same change to geconfigureassetroot.

Also, we should update the documentation with the new flag.

tst-lsavoie (changes requested):
Suggested changes:

- Make nochown the default
- Update the documentation with the new flag

tst-jlarocco:
> IIRC, @tst-ccamp had suggested making `nochown` the default and requiring users to pass a flag if they want to run the chown. What do you think about that? You'd need to make the same change to geconfigureassetroot.
> 
> Also, we should update the documentation with the new flag.

Yeah, that makes sense because it's what people would want most of the time.

earth_enterprise/rpms/opengee-fusion/snippets/post-install.sh
```diff
@@ -150,10 +150,10 @@ install_or_upgrade_asset_root()
         if [ "$IS_SLAVE" = "false" ]; then
             OWNERSHIP=`find "$ASSET_ROOT" -maxdepth 0 -printf "%g:%u"`
             if [ "$OWNERSHIP" != "$GEGROUP:$GEFUSIONUSER" ] ; then
-                NOCHOWN=""
+                DOCHOWN="--chown"
```
tst-lsavoie, line 153: Just a thought: perhaps in this case we should just print a message and not do the chown. I'm not sure which would be a better user experience - a possible surprise that the install takes a long time or a possible surprise that the permissions are not right. I'm good either way.

docs/geedocs/docsrc/answer/fusionAdministration/relocateAssetRoot.rst
```diff
@@ -48,7 +48,7 @@ Relocate your asset root
          command, the tool auto-detects the problems that need to be
          repaired and fixes them.
 
-         ``sudo /opt/google/bin/geconfigureassetroot --repair new_path``
+         ``sudo /opt/google/bin/geconfigureassetroot --repair --chown new_path``
```
tst-lsavoie, line 51: There is a call to `geconfigurepublishroot` below on line 69. Do we need to add the `--chown` option there as well?

docs/geedocs/docsrc/answer/fusionAdministration/commandReference.rst
```diff
@@ -179,10 +179,9 @@ Command reference
 
       -\\-**chown**
 
-      *Optional*. Prevents attempts by this command to fix
-      file/directory privileges. You may consider setting this option
-      when you do not want any prvilege settings to change as a result
-      of reconfiguring your asset root.
+      *Optional*. Attempt to change file owner and permissions.  This is
+      only needed when the asset root has been created or modified by users
+      other than the defualt fusion user.
```
tst-lsavoie, line 184: Spelling: defualt

docs/geedocs/docsrc/answer/fusionAdministration/commandReference.rst
```diff
@@ -278,6 +277,18 @@ Command reference
       or the configuration fails, the program will return -1 (0 is
       returned on success).
 
+      -\\-chown
+
+      *Optional*. Correct permissions and ownership of the publish root.
+      This is needed when the publish root was created or modified by a user
+      other than the default fusion user, such as when re-using a publish root
```
tst-lsavoie, line 284: I think in this case the relevant user is the "default Apache user", not the "default fusion user".

earth_enterprise/rpms/opengee-server/snippets/post-install.sh
```diff
@@ -125,14 +125,11 @@ fix_postinstall_filepermissions()
     chown $GEAPACHEUSER:$GEGROUP $BASEINSTALLDIR_OPT/gehttpd/htdocs/.htaccess
     chown -R $GEAPACHEUSER:$GEGROUP $BASEINSTALLDIR_OPT/gehttpd/logs
 
-    # Publish Root
+    # Publish Root - note these are not recursive
```
tst-lsavoie, line 128: Technically this file works differently than opengee-fusion/snippets/post-install.sh. For fusion, we check the permissions of the top level asset root directory and only run chown if those permissions are wrong. For server, we never run chown. I'm not sure that it matters a lot but it is somewhat inconsistent. Personally I prefer the server method - don't run a chown at all and let the user make the decision to do that. We'll want to make sure we clearly state somewhere that if your asset root or publish root permissions are wrong you'll need to run the appropriate command with the chown option.

I'll leave it up to you whether you want to address this or not. It would be nice to have them consistent but I can live without it.

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -822,10 +822,10 @@ install_or_upgrade_asset_root()
         if [ $IS_SLAVE == false ]; then
             # TODO: Verify this logic -- this is what is defined in the installer documentation, but need confirmation
             if [ $NEW_GEGROUP == true ] || [ $NEW_GEFUSIONUSER == true ]; then
-                NOCHOWN=""
+                DOCHOWN="--chown"
```
tst-lsavoie, line 825: Should we add similar checks in install_server.sh for consistency sake?

docs/geedocs/docsrc/answer/fusionAdministration/relocateAssetRoot.rst
tst-jlarocco: I left it out because step says, "Create your new publish root", and if it's being created then the directories will be created with the correct owner and permissions without needing to chown.

earth_enterprise/src/server/geconfigurepublishroot
```diff
@@ -56,9 +58,11 @@ EOF
 my $allow_symlinks = 0;
 my $show_usage = 0;
 my $noprompt = 0;
+my $chown = 0;
```
tst-lsavoie, line 61: There is a "usage" print out right above this. Chown should be added there.

earth_enterprise/src/server/geconfigurepublishroot
```diff
@@ -90,30 +94,41 @@ if ($noprompt eq 0) {
...
+            return;
+        }
+        chown $uid, $gid, $File::Find::name;
+        # root's umask could leave the dirs with less permissions than we want
+        # geapacheuser:gegroup will need to write to these dirs which is why 775.
+        chmod 0775, $File::Find::name if -d;
+        # Don't set execute bit on files
+        chmod 0664, $File::Find::name if -f;
+    }
+    File::Find::find({wanted=>\&wanted, follow=>0}, ($stream_dir, $search_dir));
```
tst-lsavoie, line 130: Minor: this only changes the permissions for the directories under /gevol/published_dbs, not the permissions on /gevol/published_dbs itself. I guess that's the way it's always been, so that's probably fine.

earth_enterprise/rpms/opengee-fusion/snippets/post-install.sh
tst-jlarocco: Yeah, best to be consistent.  Since we decided not to chown the publish root, I'll remove the chown here as well.

docs/geedocs/docsrc/answer/fusionAdministration/relocateAssetRoot.rst
tst-jlarocco: I left it out on line 69 because it says "Create your new publish root", and it creates it with the correct owner and permissions without needing to `chown`.

tst-lsavoie (changes requested):
In summary, here are the changes I'd like to see:

- commandReference.rst - spelling of "defualt"
- commandReference.rst - should mention the apache user instead of the fusion user for geconfigurepublishroot
- usage message in geconfigurepublishroot
- don't forget to rebuild the docs

In addition, the following changes would be nice but not necessary:

- don't do chown in fusion installer (except when creating /gevol/assets or gefusionuser)

earth_enterprise/src/installer/install_fusion.sh
tst-jlarocco: I hadn't made any logic changes to the regular installers, but it would be nice to have consistency between the RPM and non-RPM installers.

Since we decided to remove the privilege changes for the publish root from the RPM installer, I'll remove the `--chown` here and update message, and also remove the `chown -R` calls in `earth_enterprise/src/installer/install_server.sh`

That way all of the installers should be consistent about not changing permissions and slowing down installation.

earth_enterprise/rpms/opengee-server/snippets/post-install.sh
tst-jlarocco: Yeah, I agree.  I'll remove $DOCHOWN and update the message here as well.

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -822,10 +822,10 @@ install_or_upgrade_asset_root()
         if [ $IS_SLAVE == false ]; then
             # TODO: Verify this logic -- this is what is defined in the installer documentation, but need confirmation
             if [ $NEW_GEGROUP == true ] || [ $NEW_GEFUSIONUSER == true ]; then
-                NOCHOWN=""
-                UPGRADE_MESSAGE="\nThe upgrade will fix permissions for the asset root and source volume. This may take a while.\n"
+                UPGRADE_MESSAGE="WARNING: The installer detected the asset root may have \
+                    incorrect permissions! After installation you may need to run \n\
+                    'geconfigureassetroot --noprompt --chown --assetroot $ASSET_ROOT'"
```
tst-lsavoie, line 827: Minor: should we say `/opt/google/bin/geconfigureassetroot` here?

earth_enterprise/src/installer/install_fusion.sh
```diff
@@ -822,10 +822,10 @@ install_or_upgrade_asset_root()
         if [ $IS_SLAVE == false ]; then
             # TODO: Verify this logic -- this is what is defined in the installer documentation, but need confirmation
             if [ $NEW_GEGROUP == true ] || [ $NEW_GEFUSIONUSER == true ]; then
-                NOCHOWN=""
-                UPGRADE_MESSAGE="\nThe upgrade will fix permissions for the asset root and source volume. This may take a while.\n"
+                UPGRADE_MESSAGE="WARNING: The installer detected the asset root may have \
```
tst-lsavoie, line 825: There's a potential issue on line 818 above (I can't put the comment there so I'm putting it here). If you do the following:

1. Install Fusion
2. Put a bunch of files with the wrong ownership under /gevol/fusion
3. Uninstall Fusion
4. Install Fusion again

then this script will run a potentially costly chown operation. The key is that you uninstall Fusion and install it again rather than upgrade in-place. In that case it will be considered a "new" installation even though /gevol/assets already exists and it will run the chmod.

I'm not sure how much this matters. It's likely to be a rare occurrence, so I could be convinced not to worry about it.

earth_enterprise/src/installer/install_server.sh
```diff
@@ -649,8 +649,13 @@ fix_postinstall_filepermissions()
...
   chmod 644 "$PUBLISHER_ROOT/.config"
   chmod 755 "$PUBLISHER_ROOT"
-  chown -R "$GEAPACHEUSER_NAME:$GRPNAME" "$PUBLISHER_ROOT/stream_space"
-  chown -R "$GEAPACHEUSER_NAME:$GRPNAME" "$PUBLISHER_ROOT/search_space"
+
+  SEARCH_OWNER=`find "$PUBLISHER_ROOT/search_space" -maxdepth 0 -printf "%g:%u"`
+  STREAM_OWNER=`find "$PUBLISHER_ROOT/stream_space" -maxdepth 0 -printf "%g:%u"`
+  if [ "$SEARCH_OWNER" != "$GEGROUP:$GEAPACHEUSER" -o "$STREAM_OWNER" != "$GEGROUP:$GEAPACHEUSER" ] ; then
+        printf "WARNING: The installer detected the publish root may have incorrect permissions! After installation you may need to run \n\
+sudo /opt/google/bin/geconfigurepublishroot --noprompt --chown --path=$PUBLISHER_ROOT"
```
tst-lsavoie, line 657: I think you need a `\n` on the end of this. This is the output I'm getting:

```
WARNING: The installer detected the publish root may have incorrect permissions! After installation you may need to run 
sudo /opt/google/bin/geconfigurepublishroot --noprompt --chown --path=/gevol/published_dbsStarting server
waiting for server to start.... done
```

There may be similar issues in the install_fusion script.

earth_enterprise/src/installer/install_fusion.sh
tst-lsavoie: Another point here - this message only prints out if the script has to create gefusionuser or gegroup. If the user and group already exist we'll never see this message. Similarly, if the user and group don't exist but the permissions are correct on /gevol/assets, this message will print out anyway.

The logic in this area is kind of convoluted and could probably use a good refactor.

earth_enterprise/src/installer/install_server.sh
tst-lsavoie: I'm seeing this message print out on a fresh install where no /gevol exists before the install for some reason. I'm not sure why. The permissions are correct when it's created.

tst-lsavoie:
I think the docs need to be rebuilt again.

As a side note, when I install Fusion or Server I get the 5.3.7 docs, so I don't see your changes. I don't think that has anything to do with this PR, though.

tst-lsavoie (changes requested):
This is what I see:

- Rebuild the documentation
- Add a newline at the end of the new warning messages where needed
- Suggest refactoring the permissions check - move the check to somewhere near the end of the respective install scripts and always check permissions, regardless of the type of install/upgrade.
- Maybe do something about the case where you're installing Fusion for the first time but /gevol already exists and you run a potentially costly chmod.

earth_enterprise/rpms/opengee-fusion/snippets/post-install.sh
```diff
@@ -150,10 +150,10 @@ install_or_upgrade_asset_root()
         if [ "$IS_SLAVE" = "false" ]; then
             OWNERSHIP=`find "$ASSET_ROOT" -maxdepth 0 -printf "%g:%u"`
             if [ "$OWNERSHIP" != "$GEGROUP:$GEFUSIONUSER" ] ; then
-                NOCHOWN=""
-                UPGRADE_MESSAGE="The upgrade will fix permissions for the asset root and source volume. This may take a while."
+                UPGRADE_MESSAGE="WARNING: The installer detected the asset root may have  incorrect permissions! \
+After installation you may need to run \n\n\
+$BASEINSTALLDIR_OPT/bin/geconfigureassetroot --noprompt --chown --assetroot $ASSET_ROOT\n\n"
```
tst-lsavoie, line 155: Turns out this command doesn't work. You have to include the `--repair` flag. The same likely applies to install_fusion.sh.

Strangely enough, the command for geconfigurepublishroot works fine.

earth_enterprise/rpms/opengee-fusion/snippets/post-install.sh
```diff
@@ -150,10 +150,10 @@ install_or_upgrade_asset_root()
         if [ "$IS_SLAVE" = "false" ]; then
             OWNERSHIP=`find "$ASSET_ROOT" -maxdepth 0 -printf "%g:%u"`
             if [ "$OWNERSHIP" != "$GEGROUP:$GEFUSIONUSER" ] ; then
-                NOCHOWN=""
-                UPGRADE_MESSAGE="The upgrade will fix permissions for the asset root and source volume. This may take a while."
+                UPGRADE_MESSAGE="WARNING: The installer detected the asset root may have  incorrect permissions! \
```
tst-lsavoie, line 153: Minor: double space in "have  incorrect".

earth_enterprise/rpms/opengee-fusion/snippets/post-install.sh
tst-lsavoie: Minor: while you're fixing this, you might as well add `sudo` at the beginning of this command (and make the same change in install_fusion.sh).

tst-lsavoie (changes requested):
Looks like the last fix is adding `--repair` to the recommended `geconfigureassetroot` calls.

tst-lsavoie:
I noticed that if the permissions on /gevol/assets are wrong we get duplicate warnings when installing Fusion:
```
WARNING: The installer detected the asset root may have incorrect permissions! After installation you may need to run 

sudo /opt/google/bin/geconfigureassetroot --noprompt --chown --repair --assetroot /gevol/assets

Switching to (gefusionuser, gegroup)
Validating assetroot
No volume changes to save.
Configured /gevol/assets.
No upgrade needed.
Fusion Fatal:	
/gevol/assets needs to be repaired.
You must run the following command:
  geconfigureassetroot --repair --assetroot /gevol/assets
```
We could add `--chown` to the second command or just live with it.

tst-lsavoie (approved):
I reviewed the code and tested the RPM and script installs in several scenarios with good and bad permissions. Everything works as expected.

tst-hrajamoney (approved):

PR #1918, opened 2021-03-10
Added the full path to cmdline

### LINKED ISSUE #1674: Error importing vector resources with full path right after install
**Describe the bug**

It is known that if you do not reboot/logout/something similar after installing Open GEE that you have to run commands using the full path (see issue #110). However, even if you use a full path, a vector import still fails.

**To Reproduce**

1. Install Open GEE. Do not reboot, log out, or take any other action that will update your path.
1. Run the download_tutorial.sh script to get the tutorial files.
1. Run `genewvectorresource -o Tutorial/Resources/Vector/CAHighways /opt/google/share/tutorials/fusion/Vector/california_roads_line.shp`. This command fails as expected because `genewvectorresource` is not in your path.
1. Run `/opt/google/bin/genewvectorresource -o Tutorial/Resources/Vector/CAHighways /opt/google/share/tutorials/fusion/Vector/california_roads_line.shp`. This will still fail even though you specified the full path with the following error message:

```
Fusion Warning:	Unable to exec gevectorimport --validate --layer 0 -o foo.kvp --filelist /tmp/khtmphv1BZz
Fusion Fatal:	Unable to validate vector import
```

You can work around the issue by resetting your path (logging out and back in, for example), after which the command will succeed.

Most likely this command calls some other Open GEE executable without specifying the full path.

**Expected behavior**

The last command succeeds because you specified the full path.

tst-hrajamoney:
Fixes #1674

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/tools/genewimageryresource.cpp
+++ b/earth_enterprise/src/fusion/autoingest/tools/genewimageryresource.cpp
@@ -290,7 +290,7 @@ main(int argc, char *argv[]) {
 
       // validate mosaic inputs
       CmdLine cmdline;
-      cmdline << "gevirtualraster" << "--validate" << "-o" << "foo.khvr";
+      cmdline << "/opt/google/bin/gevirtualraster" << "--validate" << "-o" << "foo.khvr";
       if (req.config.fill.size()) {
         cmdline << "--fill" << req.config.fill;
       }
--- a/earth_enterprise/src/fusion/autoingest/tools/genewvectorresource.cpp
+++ b/earth_enterprise/src/fusion/autoingest/tools/genewvectorresource.cpp
@@ -250,7 +250,7 @@ main(int argc, char *argv[]) {
     // validate inputs files
     {
       CmdLine cmdline;
-      cmdline << "gevectorimport" << "--validate"
+      cmdline << "/opt/google/bin/gevectorimport" << "--validate"
               << "--layer" << req.config.layer
               << "-o" << "foo.kvp";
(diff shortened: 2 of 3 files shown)


tst-nrichman (approved):

earth_enterprise/src/fusion/autoingest/tools/genewimageryresource.cpp
```diff
@@ -290,7 +290,7 @@ main(int argc, char *argv[]) {
 
       // validate mosaic inputs
       CmdLine cmdline;
-      cmdline << "gevirtualraster" << "--validate" << "-o" << "foo.khvr";
+      cmdline << "/opt/google/bin/gevirtualraster" << "--validate" << "-o" << "foo.khvr";
```
tst-lsavoie, line 293: Is there a way to do this that wouldn't cause problems if we start installing somewhere other than `/opt/google/bin`? For example, could we just add `/opt/google/bin` to the path when we run the command, or run it from the same directory as the calling binary, or is there some configuration file we can get /opt/google/bin from, or something like that? I can live with it this way but it would be nice to have it a bit more flexible.

tst-hrajamoney: I decided this way, as there are lot of places in code, which uses this absolutepath.

tst-lsavoie:
I found one more place with a similar error: gemunge.cpp line 340. I have no idea how it's used, but I guess we should fix it too.

tst-lsavoie (approved):
Tested and everything looks good.

PR #1920, opened 2021-03-12
Background color fix for jpgs in wms

tst-mswartz:
Verification steps:
1. Publish a globe with "wms" enabled in the admin panel
2. Using qgis or a script, view the wms content..  In qgis try connecting using jpg
3. Background is no longer black by default

### DIFF
--- a/earth_enterprise/src/server/wsgi/wms/ogc/common/tiles.py
+++ b/earth_enterprise/src/server/wsgi/wms/ogc/common/tiles.py
@@ -70,9 +70,15 @@ def StitchTiles(tiles_array, layer_properties, tilepixel_rect,
   set_pixel_to_bgcolor = (layer_properties.image_format == "image/png" and
                           layer_properties.is_transparent == "FALSE")
 
-  bgcolor = (layer_properties.bgcolor if set_pixel_to_bgcolor
-             else _NO_DATA_PIXELS)
-  alpha = _OPAQUE_ALPHA if set_pixel_to_bgcolor else _TRANSPARENT_ALPHA
+  bgcolor = ALL_WHITE_PIXELS
+  if layer_properties.bgcolor:
+      bgcolor = layer_properties.bgcolor
+
+  alpha = _TRANSPARENT_ALPHA
+
+  if set_pixel_to_bgcolor:
+      #bgcolor = layer_properties.bgcolor
+      alpha = _OPAQUE_ALPHA
 
   # Alpha channel is not required for jpeg formats.
   if layer_properties.image_format == "image/jpeg":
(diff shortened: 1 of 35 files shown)


earth_enterprise/src/server/wsgi/wms/ogc/implementation/common.py
```diff
@@ -90,9 +90,12 @@ def GenerateOutputCommon(self):
...
       # Otherwise str.decode("hex") will throw
       # "TypeError: Non-hexadecimal digit found" error.
-      if bgcolor[:2] == "0x":
-        bgcolor = bgcolor[2:]
-      bgcolor = tuple(ord(c) for c in bgcolor.decode("hex"))
+      try:
+        if bgcolor[:2] == "0x":
+          bgcolor = bgcolor[2:]
+        bgcolor = tuple(ord(c) for c in bgcolor.decode("hex"))
+      except:
```
tst-mbiancalana, line 97: This check on the BGCOLOR value is a nice addition. The BGCOLOR field is optional in a GetMap request, and I didn't see anything in the WMS implementation spec about returning an exception on bad BGCOLOR values. So it's better to set the value to the default and carry on with the request, as you've done here. I like it much better this way.

The bare except is really broad, though. We should specifically catch the TypeError mentioned above here.
- [ ] Narrow the except to catch only `TypeError`s 

It might also be worth noting in a comment for posterity that this whole conversion from a hex string looks different in Python 3 and raises a ValueError when it encounters problems.

earth_enterprise/src/server/wsgi/wms/ogc/common/tiles.py
```diff
@@ -70,9 +70,15 @@ def StitchTiles(tiles_array, layer_properties, tilepixel_rect,
...
-             else _NO_DATA_PIXELS)
-  alpha = _OPAQUE_ALPHA if set_pixel_to_bgcolor else _TRANSPARENT_ALPHA
+  bgcolor = ALL_WHITE_PIXELS
+  if layer_properties.bgcolor:
+      bgcolor = layer_properties.bgcolor
+
+  alpha = _TRANSPARENT_ALPHA
+
+  if set_pixel_to_bgcolor:
+      #bgcolor = layer_properties.bgcolor
```
tst-mbiancalana, line 80: This comment line should go as well.
- [ ] Remove unneeded comment

tst-mbiancalana (changes requested):
I tested the changes and they work!
Just a couple of stylistic changes requested.

tst-mbiancalana:
The merge of `release_5.3.7.1` into this PR should be backed out (either by rebase or other means). That branch hasn't been merged into master yet (though it could be at any time), and it will be confusing to do that merge as part of this PR.
- [ ] Back out merge of `release_5.3.7.1` branch into this PR

tst-tcalder (commented):
This looks good, I agree with @tst-mbiancalana , once those changes are made, this should be set to go.

tst-mbiancalana:
@tst-mswartz I've looked at this again, and I'm not sure what I saw that made me think this needed the commits from `release_5.3.7.1` removed. As you pointed out when we talked, the diff is just your changes. I'm marking that task as done (obsolete) and approving the PR. Sorry for the confusion.

tst-mbiancalana (approved):

tst-tcalder (approved):

PR #1926, opened 2021-03-26
Handle very high resolution imagery

### LINKED ISSUE #1925: Fusion cannot process very high level images
**Describe the bug**

Fusion will refuse to process images that are higher than a certain resolution (i.e., zoom level 24). The reason for this is because the number of pixels and tiles at zoom levels higher than 24 starts to be too large to store in a 32-bit integer. However, instead of rejecting such imagery, Fusion could simply downsample it to level 24 as a part of the reprojection that it's already doing.

**To Reproduce**

Create a resource from the attached image and try to build it in Fusion. The build will fail and the logs will show an error similar to the one below:
```
Fusion Fatal:	input pixel size with reprojection: (xy) 0.0185 m, 0.0185 m.
This tool supports pixel size upto: (xy) 0.0186 m, 0.0186 m.
You may down-sample the imagery and retry.
```

[toohighres.tiff.zip](https://github.com/google/earthenterprise/files/6214562/toohighres.tiff.zip)

**Expected behavior**

The build succeeds.

tst-lsavoie:
This change allows Fusion to process images that are higher resolution than the native resolution of level 24. It sets the target level for imagery to level 24 (at the maximum), which means gerasterimport will reproject the image to level 24 resolution during its normal processing. No additional processing steps are added; this just changes the parameters that are used in the existing reprojection.

This PR includes new unit tests, and I've run some additional integration tests with images that exercise various code paths. I manually verified that each image is rendered correctly.

Fixes #1925


tst-mreinert (approved):

tst-jlarocco (approved):
Code looks good and unit tests all passed.

PR #1935, opened 2021-04-06
Fix vector overflows

### LINKED ISSUE #1934: Possible overflow in vector handling
**Describe the bug**

There are potential overflows in our vector handling where code accesses the 4th element of a 3-element vector. The lines in question are in `src/fusion/rasterfuse/vipm/vector.h`, lines 195, 200, and 201.

It's possible that this code is unused.

**To Reproduce**

There is no known way to trigger this bug.

**Expected behavior**

Open GEE does not read outside the bounds of vectors.

tst-lsavoie:
This fixes some vector overflows where the 4th element of a 3-element vector was accessed. The problem code wasn't used, so I simply deleted it. I also removed some other unused functions and added unit tests. I tried to only remove functions such that the compiler would catch any errors except that I removed `operator=` from `etMat3` because it's only used in one place so it was easy to determine that it wasn't needed.

Fixes #1934

### DIFF
--- a/earth_enterprise/src/fusion/rasterfuse/SConscript
+++ b/earth_enterprise/src/fusion/rasterfuse/SConscript
@@ -65,3 +65,6 @@ geraster2gigapxl = env.executable('geraster2gigapxl',
 env.install('fusion_bin', [gerasterimport, gepackgen, geraster2kml])
 env.install('tools_bin', [geraster2gigapxl])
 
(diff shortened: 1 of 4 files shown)


tst-hrajamoney (approved):

tst-jlarocco (approved):
Code looks good and the new unit test passed.
That was a lot of unused code.

PR #1940, opened 2021-04-12
Update docs for python 3 and remove python 2

tst-cjeffries:
Updates documentation to reflect python 3 migration. Removes python 2 and references to it.


tst-basilhuffman:
might want to give instructions on building python 3.8 on the other build .md files

earth_enterprise/BUILD_RHEL_CentOS.md
```diff
@@ -223,19 +223,12 @@ shunit2 was installed in a previous step.
...
-```bash
-sudo yum install python python-pip python-devel
-```
+In order to build OpenGEE, Python 3.8 is needed. If you have access to it through public repos, we suggest installing it that way.
+If you do not have access, Python 3.8 can be built and installed by running the following from `earthenterprise`:
 
 ### All distros
 
-The following script will build and install Python 2.7 and Python 3.8 from source if they are needed. From `earthenterprise` run:
+The following script will build and install Python 3.8 from source if it are needed. From `earthenterprise` run:
```
tst-dsugar, line 231: if it **is** needed

PR #1944, opened 2021-04-16
added vector size check after calling TokenizeString

### LINKED ISSUE #1943: Fusion Code does not check c++ vector size before accessing indexes
**Description**
Fusion includes a function called "TokenizeString" that splits a string into a vector of tokens. There are two places in the code where this function is called and the resulting vector's size is not checked before it is accessed.

The two places where the bounds are not checked are:

JsonUtils.cpp, in the function "LookAtJson". It is assumed that "tokens" will have at least 3 elements, but this is not checked.

geaddtomapproject.cpp, when processing command line arguments in the main function. It is assumed that "tokens" will have at least 1 element, but this is not checked. 

**Steps To Reproduce**

These bugs were found through code inspection.

geaddtomapproject.cpp
1. genewmapproject -o testproject
2. geaddtomapproject -o testproject ""  -> segfault

JsonUtils.cpp
1. Drop a pin in google earth and save the point to create a kml/kmz with a single point in it
2. Create a vector resources.  CAHighways from the tutorial set is fine
3. Create a vector project
4. Add the resource to the project.  Modify the layer properties and set the look at field using the point you created in step 1
5. Build the vector project
6. Edit the  dbroot.kva/ver001/dbrootconfig.xml file for the vector project you just built.  Find the xml element look_at.  Change the value from the pipe delimited value to a single integer, something like 123
7. Restart gefusion(this step may not be necessary)
8. Build a globe using the now corrupted vector project
9. Run geserveradmin --adddb and --pushdb for the globe you just built
10. Run geserveradmin --publishdb for the globe you just built.  This step will fail with a no status returned message
11. Inspect the log file at /opt/google/gehttpd/logs/error_log to see the seg fault message

tst-mreinert:
fixes #1943 

Add calls to std::vector::size prior to accessing vector elements in JsonUtils.cpp and geaddtomapproject.cpp

### DIFF
--- a/earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
+++ b/earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
@@ -297,8 +297,7 @@ std::string JsonUtils::LookAtJson(const std::string& lookAtSpec) {
 
   std::map<std::string, std::string> field_map;
 
-  if (tokens.size() >= 3)
-  {
+  if (tokens.size() >= 3) {
     double altitude = strtod(tokens[2].c_str(), NULL);
     int zoomLevel = AltitudeToZoomLevel(altitude);
 
@@ -307,9 +306,7 @@ std::string JsonUtils::LookAtJson(const std::string& lookAtSpec) {
     field_map["lng"] = tokens[0];
     field_map["zoom"] = Itoa(zoomLevel);
     field_map["altitude"] = DoubleToString(altitude);
-  }
-  else
-  {
+  } else {
     int zoomLevel = AltitudeToZoomLevel(0);
     field_map["lat"] = "0";
     field_map["lng"] = "0";
--- a/earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
+++ b/earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
@@ -295,15 +295,27 @@ std::string JsonUtils::LookAtJson(const std::string& lookAtSpec) {
   std::vector<std::string> tokens;
   TokenizeString(lookAtSpec, tokens, "|");
 
-  double altitude = strtod(tokens[2].c_str(), NULL);
-  int zoomLevel = AltitudeToZoomLevel(altitude);
-
-  // Create the map of field name-value pairs for the output JSON.
   std::map<std::string, std::string> field_map;
-  field_map["lat"] = tokens[1];
-  field_map["lng"] = tokens[0];
-  field_map["zoom"] = Itoa(zoomLevel);
-  field_map["altitude"] = DoubleToString(altitude);
+
+  if (tokens.size() >= 3)
+  {
+    double altitude = strtod(tokens[2].c_str(), NULL);
+    int zoomLevel = AltitudeToZoomLevel(altitude);
+
+    // Create the map of field name-value pairs for the output JSON.
+    field_map["lat"] = tokens[1];
+    field_map["lng"] = tokens[0];
+    field_map["zoom"] = Itoa(zoomLevel);
+    field_map["altitude"] = DoubleToString(altitude);
+  }
+  else
+  {
+    int zoomLevel = AltitudeToZoomLevel(0);
+    field_map["lat"] = "0";
+    field_map["lng"] = "0";
+    field_map["zoom"] = Itoa(zoomLevel);
+    field_map["altitude"] = "0";
+  }
   return JsonObject(field_map);
 }
 
--- a/earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
+++ b/earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
@@ -1,4 +1,4 @@
-// Copyright 2018 the Open GEE Contributors
+// Copyright 2021 the Open GEE Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
```diff
@@ -295,15 +295,27 @@ std::string JsonUtils::LookAtJson(const std::string& lookAtSpec) {
...
-  int zoomLevel = AltitudeToZoomLevel(altitude);
-
-  // Create the map of field name-value pairs for the output JSON.
   std::map<std::string, std::string> field_map;
-  field_map["lat"] = tokens[1];
-  field_map["lng"] = tokens[0];
-  field_map["zoom"] = Itoa(zoomLevel);
-  field_map["altitude"] = DoubleToString(altitude);
+
+  if (tokens.size() >= 3)
```
tst-lsavoie, line 300: Very minor: the style tends to be:
```
if (tokens.size() >= 3) {
```
rather than
```
if (tokens.size() >= 3)
{
```
That applies to both of the files you've changed.

tst-lsavoie:
How difficult would it be to add unit tests for JsonUtils.cpp, or at least the part that you modified? I get that adding unit tests to geaddtomapproject.cpp is going to be a problem since it already has a `main` function, but it would be nice to add some tests where we can.

tst-lsavoie (approved):
Code changes look good. I'd like to see some unit tests if possible, but there's nothing wrong with the code as-is.

tst-jlarocco (approved):
I was able to run:
`
genewmapproject -o testproject
geaddtomapproject -o testproject "" 
`
And also build a project with corrupted data in "dbroot.kva/ver001/dbrootconfig.xml", which previously triggered this bug.

Both tests passed, and the code looks good.

It would be nice to get unit tests for these if we could but not a show stopper IMO.

earth_enterprise/src/fusion/autoingest/JsonUtils.cpp
```diff
@@ -295,15 +295,24 @@ std::string JsonUtils::LookAtJson(const std::string& lookAtSpec) {
...
+    double altitude = strtod(tokens[2].c_str(), NULL);
+    int zoomLevel = AltitudeToZoomLevel(altitude);
+
+    // Create the map of field name-value pairs for the output JSON.
+    field_map["lat"] = tokens[1];
+    field_map["lng"] = tokens[0];
+    field_map["zoom"] = Itoa(zoomLevel);
+    field_map["altitude"] = DoubleToString(altitude);
+  } else {
+    int zoomLevel = AltitudeToZoomLevel(0);
```
tst-lsavoie, line 310: Very minor: it just occurred to me that altitude 0 might not be the best option. That's likely zoomed pretty far in. A better default altitude might be closer to 1,000,000 meters (about 1,000 miles). That gives you a fairly wide view.

Of course, we'll hit this so rarely that it probably doesn't matter much.

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -0,0 +1,110 @@
...
+class JsonUnitTest : public testing::Test, public JsonUtils {
+  public:
+    JsonUtilsAccessor jsonUtils;
+};
+
+// Testing the JsonBool function
+TEST_F(JsonUnitTest, JsonBoolTrue) {
+    const bool trueBool = true;
+    const std::string expectedTrue= "true";
+    std::string result = jsonUtils.RunJsonBool(trueBool);
```
tst-lsavoie, line 53: Minor: I think that since this is a fixture and the test fixture inherits from `JsonUtils`, you can just call `JsonBool` directly. Then you don't need the Accessor class.

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -0,0 +1,110 @@
+// Copyright 2018 the Open GEE Contributors
```
tst-lsavoie, line 1: Copyright year should be 2021

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
tst-lsavoie: Or, on the flip side, if you leave it like it is I don't think `JsonUnitTest` needs to inherit from `JsonUtils`.

earth_enterprise/src/fusion/autoingest/JsonUtils_unittest.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+TEST_F(JsonUnitTest, LookAtValid) {
+    const std::string lookAtString = "-93.02119299429238|38.88758732203867|4963567.527932385|0.4781091614682265|6.862074527538373";
+    const std::string expectedJson= "{\naltitude : 4.96357e+06,\nlat : 38.88758732203867,\nlng : -93.02119299429238,\nzoom : 3\n}\n";
+    std::string result = jsonUtils.RunLookAtJson(lookAtString);
+
+    EXPECT_EQ(expectedJson, result);
+}
+
+TEST_F(JsonUnitTest, LookAtInvalid) {
```
tst-lsavoie, line 74: Minor: you could also add a test for LookAt with an empty string. That will increase coverage.

tst-lsavoie (changes requested):
I left a few suggestions. The only thing that really should be addressed is the copyright date.

tst-lsavoie (approved):
Everything looks good and tests pass.

PR #1945, opened 2021-04-16
WIP: fix disappearing vector points

tst-basilhuffman:
possible solution is to auto-draw a selection covering the entire globe


tst-basilhuffman:
problem has existed since before switching to qt4, closing

PR #1957, opened 2021-05-04
Python 3 Migration WMS

tst-cjeffries:
Rectifies issues with WMS and Python 3.

To test:
1. Build and Install (Build and install according to the build docs (https://github.com/tst-cjeffries/earthenterprise/blob/python_3_wms_bug/earth_enterprise/BUILD_RHEL_CentOS.md)
2. Run the tutorials.
`sudo /opt/google/share/tutorials/fusion/download_tutorial.sh`
`sudo cp earthenterprise/earth_enterprise/src/fusion/tools/gee_test/tutorial_files/providers.xml /gevol/assets/.userdata/`
`sudo env "PATH=$PATH:/opt/google/bin" earthenterprise/earth_enterprise/src/fusion/tools/gee_test/run_tutorial.sh`
Then wait until the tutorials are finished building. This can be monitored in with the status scripts in that same directory.
`sudo env "PATH=$PATH:/opt/google/bin" earthenterprise/earth_enterprise/src/fusion/tools/gee_test/publish_tutorial.sh`
3. Go to the admin page and republish `SF_2d_Merc` with WMS enabled.
4. Try the GetCapabilties Request:
`curl "http://tstserver.vm/SFDb_3d_WMS/wms?version=1.3.0&request=GetCapabilities&service=WMS"`
5. Verify that the returned XML looks correct.
6. Open Earth Client and connect to http://kh.google.com
7. Go to the “Add” menu at the top and click “Image Overlay”
8. Give your overlay a name
9. Click the “Refresh” tab, then “WMS Paramters”
10. Click the “Add…” button at the top, paste in the URL of the WMS database you just published, and click OK.
11. Click “Proceed” if it asks about an untrusted certificate (possibly multiple times)
12. Click “[O] Imagery”, then “Add->”.
13. Click OK twice.
14. The WMS overlay should appear on top of the basemap. Verify that it has the expected content, and remember that it is very slow. If a greyed out area with a giant arrow appears instead, right click on the new layer you added and click “Refresh”. This should force the imagery to appear.
15. Open QGIS
16. Go to Layer > Add Layer > Add WMS ... Layer
17. Create the Layer with the correct url.
18. Add imagery as a JPEG layer
19. Add vector info as a PNG layer.
20. Verify that both layers can be viewed.


earth_enterprise/src/server/wsgi/wms/ogc/common/geom.py
```diff
@@ -44,6 +44,12 @@ def __div__(self, b):
       return Pair(self.x / b.x, self.y / b.y)
     else:
       return Pair(self.x / b, self.y / b)
+  
+  def __truediv__(self, b):
+    if isinstance(b, Pair):
+      return Pair(self.x // b.x, self.y // b.y)
```
tst-ccamp, line 50: Any reason why this uses // instead of the / that the other functions use?

tst-ccamp: I think this could use some comments

earth_enterprise/src/server/wsgi/wms/ogc/service/wms_request_app.py
```diff
@@ -57,6 +57,11 @@ def __call__(self, environ, start_response):
       response_headers.append(content_type)
 
     response_body = output
+    try:
+      response_body = response_body.encode('ascii')
+    except (UnicodeDecodeError, AttributeError):
+      pass
```
tst-ccamp, line 63: Shouldn't we at least log something?

earth_enterprise/src/server/wsgi/wms/ogc/common/geom.py
tst-cjeffries: It needs fixed. Its supposed to just be standard division. I saw that earlier today thinking I forgot to fix it, but it was actually correct  in the first place.

earth_enterprise/src/server/wsgi/wms/ogc/service/wms_request_app.py
tst-cjeffries: This isnt an error so to speak. Essentially, we previously were returning images as strings, and now we return them as bytes, which we dont want to encode. This allows us to handle both scenarios. Logging here would fill the logs. I could reformat this as if statement, if that would be easier to read.

tst-ccamp: I think the best solution here is <drumroll please> ... a code comment explaining what you just said there :)

tst-cjeffries: Will do.

tst-ccamp (approved):
Finished testing and approved

tst-basilhuffman (approved):

earth_enterprise/src/server/wsgi/wms/ogc/common/geom.py
```diff
@@ -44,6 +44,12 @@ def __div__(self, b):
       return Pair(self.x / b.x, self.y / b.y)
     else:
       return Pair(self.x / b, self.y / b)
+  
+  def __truediv__(self, b):
```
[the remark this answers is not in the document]
tst-cjeffries: Comment to address.

PR #1964, opened 2021-05-17
Enable setgid support for fusion ui

tst-dsugar:
Makes fusionui work when chmod g+s and setcap are applied to binary.


tst-basilhuffman (approved):
tested. approve

tst-cjeffries (approved):
Lgtm

PR #1965, opened 2021-05-19
Python 3 Pip Install Cleanup

tst-cjeffries:
Removes psycopg2 and defusedxml from the pip install instructions as we build them ourselves.


tst-basilhuffman (approved):

tst-basilhuffman (approved):
lgtm

PR #1973, opened 2021-06-02
Unbundle openssl

tst-basilhuffman:
Bundling OpenSSL is no longer required and the system packages should be used instead.


tst-ccamp (approved):
LGTM

tst-dsugar (approved):

